PR #236, opened 2022-04-01
[datasource] Google Sheets enhancements

### LINKED ISSUE #239: [sheets] Rewrite the `types` to create separate ones for `exec` and `execPrivate` queries

### LINKED ISSUE #265: [sheets] Handle file search during spreadsheet selection
- The Drive API `files.list` method [returns](https://developers.google.com/drive/api/v3/reference/files/list) a `nextPageToken` in case the list of files returned is incomplete. 

- The initial idea was to set the search value (`inputValue`) to a state variable (`spreadsheetQuery`) and make a debounced version of it (`debouncedSpreadsheetQuery`) the query key, and only call `setSpreadsheetQuery` IF there is a `nextPageToken` set in the previous response.

- However, this breaks in the scenario where the user returns to a wider search from a narrower one, i.e. reduces characters in the search value.

- The option that appears to be the best seems to be calling this debounced search every time the input changes, regardless of a `nextPageToken` being set in the previous response or not.

@Janpot Does this make sense?

### LINKED ISSUE #311: [sheets] Fix expiration access with `refresh_token`

bharatkashyap:
- Fixes #239
- Fixes #265 
- Fixes #311 


bharatkashyap:
Test app that I created 12 hours ago: 
https://toolpad-pr-236.onrender.com/_toolpad/app/cl280m3j30003a1kpalmidzdy/editor/pages/cl281we3j00013f675rer18yv

I think the refresh token issue is resolved

newguy-123:
So basically now, any new Google sheets App that a user will create will continue to show the data forever or there are some limits? Of course assuming user doesn't withdraw the granted permissions.

bharatkashyap:
> So basically now, any new Google sheets App that a user will create will continue to show the data forever or there are some limits? Of course assuming user doesn't withdraw the granted permissions.

Yes, data will persist forever (essentially as long as the Refresh Token stays valid, see [this](https://developers.google.com/identity/protocols/oauth2#expiration))

packages/toolpad-app/src/toolpadDataSources/googleSheets/types.ts
```diff
@@ -55,7 +55,43 @@ export type GoogleSpreadsheet = {
   values: any[][];
 };
 
-export type GoogleSheet = {
+export type GoogleDriveFiles = {
```
Janpot, line 58: Does the Google Drive API export a type for this that we can use? Because if they ever change their response in some version, it could break undetected.

packages/toolpad-app/src/toolpadDataSources/googleSheets/server.ts
```diff
@@ -43,45 +54,64 @@ function createOAuthClient() {
...
+    const response = await sheetsClient.spreadsheets.get({
+      spreadsheetId: spreadsheetId ?? undefined,
+      includeGridData: false,
+    });
+    if (response.statusText === 'OK') {
       return response.data;
-    } catch (error) {
-      throw new Error(`Unable to fetch spreadsheets`);
     }
+    throw new Error(`Failed to fetch query: "${JSON.stringify(query)}"`);
```
Janpot, line 112: Is there any information in the response that could help explain the error, that we can add to the error message? Like an error code or something?

packages/toolpad-app/src/toolpadDataSources/googleSheets/server.ts
```diff
@@ -103,30 +133,34 @@ async function exec(
...
-    const response = await sheets.spreadsheets.values.get({
-      spreadsheetId: query.spreadsheet.id,
-      range: query.ranges,
-    });
-    if (response.statusText === 'OK') {
-      const { values } = response.data;
-      if (values && values.length > 0) {
-        const headerRow = values.shift() ?? [];
-        const fields = headerRow.reduce((acc, currValue) => ({ ...acc, [currValue]: '' }), {});
+  if (!query) {
```
Janpot, line 136: Not sure how we can enter this conditional, the types seem to suggest this is impossible

packages/toolpad-app/src/toolpadDataSources/googleSheets/client.tsx
```diff
@@ -1,149 +1,176 @@
...
-      enabled: Boolean(query.spreadsheet) && query.type === GoogleSheetsActionType.FETCH_SHEET,
+      keepPreviousData: true,
+      enabled: Boolean(value?.spreadsheetId),
     },
   );
 
-  const { isLoading: isApiDataLoading } = useQuery(
-    ['execApi', query.ranges],
+  const fetchedSpreadsheet: PrivateApiResult<GoogleSpreadsheet> = useQuery(
+    ['sheetGet', value?.spreadsheetId],
```
Janpot, line 65: Would it make sense to do

```suggestion
    ['sheets', GoogleSheetsPrivateQueryType.FETCH_SPREADSHEET, value?.spreadsheetId],
```

for these query keys?

packages/toolpad-app/src/toolpadDataSources/googleSheets/types.ts
bharatkashyap: Yes, I can potentially import the types exported by `@googleapis`. Would it make sense to do that for all the types then?

packages/toolpad-app/src/toolpadDataSources/googleSheets/server.ts
bharatkashyap: `response.status` contains a `number` with the HTTP Error code. We could do:

```js
throw new Error(`Error ${response.status}: Failed to fetch "${JSON.stringify(query)}"`
```

packages/toolpad-app/src/toolpadDataSources/googleSheets/client.tsx
bharatkashyap: Yes, would it make sense then that the first element in the array is `sheets` for all of them?

packages/toolpad-app/src/toolpadDataSources/googleSheets/types.ts
Janpot: I don't know, which other types? It could also make sense to just pick out of it only the things we are interested in. e.g. `Pick<GoogleDriveApi.Files, 'id' | 'name' | '...'>`

packages/toolpad-app/src/toolpadDataSources/googleSheets/client.tsx
Janpot: Not strictly necessary, but as Toolpad grows, we should take care that queries can't clash across the application. e.g. if tomorrow we support office 360 excel files or something there may be another `ExcelPrivateQueryType.FETCH_SPREADSHEET` constant with the same name. Prefixing the query key with something like the datasource name prevents these collisions from ever happening

Janpot (approved):

oliviertassinari:
I'm deleting the token since this PR is merged

Screenshot 2022-04-28 at 01 42 27

Google didn't like it:

Screenshot 2022-04-28 at 01 42 41

to get our production project approved. So I have created a dev project in case we need it:

Screenshot 2022-04-28 at 01 46 06

It seems to match the best practices:

> A general recommendation is to have one project per application per environment. For example, if you have two applications, "app1" and "app2", each with a development and production environment, you would have four projects: app1-dev, app1-prod, app2-dev, app2-prod

https://cloud.google.com/docs/enterprise/best-practices-for-enterprise-organizations

PR #285, opened 2022-04-15
Differentiate the staging and PR preview plan

oliviertassinari:
I assume that overtime the two plan will fork


Janpot:
Yes, currently this is not enabled yet, we can try it out with changing
```yml
previewsEnabled: true
```
at the top

Janpot (approved):

PR #318, opened 2022-04-26
Remove jsdocs rules

Janpot:
We're not relying on them right now and they're not working at the moment anyway.
They should use `FunctionDeclaration` in contexts instead of `TSFunctionType`. Not sure why these are configured for `TSFunctionType`, perhaps an older AST format that's not used anymore by `@typescript-eslint/parser`?


Janpot:
ok, understood what this is for. We're still not using it, but we may as well keep it. Working as intended

PR #340, opened 2022-04-29
[sheets] Add `rowFieldId` to accept user defined Id fields

bharatkashyap:


packages/toolpad-components/src/DataGrid.tsx
```diff
@@ -134,6 +146,33 @@ const DataGridComponent = React.forwardRef(function DataGridComponent(
...
+      return;
+    }
+
+    if (rowIdFieldProp || Boolean(rows?.[0]?.id)) {
+      idFieldInitRef.current = true;
+      return;
+    }
+
+    const mappedRows = addIdField(rows);
+    nodeRuntime.setProp('rows', mappedRows);
```
Janpot, line 160: There seems to be a misunderstanding on what this method is for. I'm updating the name https://github.com/mui/mui-toolpad/pull/342. Hopefully that makes it more clear

packages/toolpad-components/src/DataGrid.tsx
```diff
@@ -122,6 +131,9 @@ const DataGridComponent = React.forwardRef(function DataGridComponent(
 
   const columnsInitRef = React.useRef(false);
   const hasColumnsDefined = columnsProp && columnsProp.length > 0;
+
+  const idFieldInitRef = React.useRef(false);
```
Janpot, line 135: I don't think there's a need for refs, effects, debouncing... 

If my understanding is correct, then all this needs to do is derive some state for the rows from the props. Basically it's the following line that needs to be updated:

```tsx
const rows: GridRowsProp = rowsProp || dataQueryRows || EMPTY_ROWS;
```

It should become something along the lines of:

```tsx
  const rows: GridRowsProp = React.useMemo(() => {
    const parsedRows = rowsProp || dataQueryRows || EMPTY_ROWS;
    if (parsedRows.length <= 0 || rowIdField || parsedRows[0].id) {
      return parsedRows;
    }
    return parsedRows.map((row, id) => ({ ...rows, id }));
  }, [dataQueryRows, rowsProp, rowIdField]);
```

packages/toolpad-components/src/DataGrid.tsx
```diff
@@ -134,6 +146,33 @@ const DataGridComponent = React.forwardRef(function DataGridComponent(
...
+      return;
+    }
+
+    const mappedRows = addIdField(rows);
+    nodeRuntime.setProp('rows', mappedRows);
+
+    idFieldInitRef.current = true;
+  }, [rowIdFieldProp, rows, nodeRuntime]);
+
+  const debouncedRowIdFieldProp = useDebounced(rowIdFieldProp, 300);
```
Janpot, line 165: Let's debounce this in the editor instead, before it updates the dom. No need to make the app rerender on every keystroke if we're going to debounce it in the end anyway

bharatkashyap: Understood - I've added debouncing logic in the `StringPropEditor` so that the dom updates are debounced themselves. Reimplemented the logic since I couldn't find the scope to re-use `useDebounced` there.

Janpot: I've been thinking about this and we can probably get rid of debouncing altogether if we merge https://github.com/mui/mui-toolpad/pull/355

packages/toolpad-components/src/utils/useDebounced.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react';
```
Janpot, line 1: This can be removed

Janpot (approved):

PR #341, opened 2022-04-30
Use quickjs runtime everywhere

Janpot:
Replace the iframe for evaluating js expressions with the quickjs runtime, both server and client. We're using the same method again everywhere to evaluate expressions.


Janpot:
Converting back to draft. Rethinking bindings on the page first to include more use-cases

PR #369, opened 2022-05-05
Change the binding button's icons to indicate what clicking the button does

mbrookes:
> If you decide to use a single control to indicate both state and what will happen next, how should that control be labeled?
> [...]
> If there is no text label (presumably because the icon is clear enough), then the _icon should change depending on the state it moves the system to_.
> A classic example is the interplay between the Play and Pause icons seen in the YouTube example above and present in most video players.

https://www.nngroup.com/articles/state-switch-buttons/

### DIFF
--- a/packages/toolpad-app/src/components/AppEditor/BindingEditor.tsx
+++ b/packages/toolpad-app/src/components/AppEditor/BindingEditor.tsx
@@ -178,7 +178,7 @@ export function BindingEditor<V>({
       onClick={handleOpen}
       color={hasBinding ? 'primary' : 'inherit'}
     >
-      {hasBinding ? <LinkIcon fontSize="inherit" /> : <LinkOffIcon fontSize="inherit" />}
+      {hasBinding ? <LinkOffIcon fontSize="inherit" /> : <LinkIcon fontSize="inherit" />}
     </IconButton>
   );
 


Janpot:
No objection to this. Just one nuance compared to the article: the states are "bound"/"not bound", but the action is not a toggle. It's "open dialog to bind" or "open dialog to unbind/update binding". It could also make sense to to have the `AddLink` icon for the "unbound" state and the `Link` for the "bound" state.

Screenshot 2022-05-05 at 20 40 19

oliviertassinari:
I agree with Jan on this one. This icon communicates the state, it doesn't perform an action on the state. I think that the UX is better on HEAD compared to this PR.

mbrookes:
> Just one nuance compared to the article: the states are "bound"/"not bound", but the action is not a toggle.

Yes, absolutely. If it was a menu item would have an ellipsis at the end of the description to indicate that it opens a dialog rather than triggering an immediate change of state, but that isn't possible here. That does make me wonder if tooltips for buttons that open a dialog or similar should offer the same affordance, but it isn't something I've come across (or at least noticed).

mbrookes:
> Since we have a new database in PR preview, maybe we don't need to set a basic auth password?

Makes sense. For now, what's the password please?

PR #371, opened 2022-05-06
Implement suggestions around the binding editor

Janpot:
* Use different icons to avoid confusion with toggle switches
* Expand tooltip to reflect the action that will be taken

Closes https://github.com/mui/mui-toolpad/pull/369


mbrookes:
Looks good. I still think it would be good to use an ellipsis on the tooltip for a button that opens a dialog rather than being a toggle button.

Off topic to this PR, but I noticed that it's possible to submit a garbage binding, which results in:

image

Janpot:
> Looks good, although I still think it would be good to use an ellipsis on the tooltip to indicate that it opens a dialog, and is not a toggle button.

Just to confirm I understand this correctly, how would that look exactly? Something like `bind the "rows" property...` instead of `bind the "rows" property`?

> If we're strictly following Material Design, it doesn't use the tooltip arrow. And I'm not sure about having the primary color on the button icon for the configured binding, but happy to be schooled on that!

I think my initial thinking was to treat it similar to a checkbox, which also uses primary color for "active" state. But I'm certainly not the person to do schooling on material design 🙂 . Maybe Gerda will give us another perspective on this?

> Off topic to this PR, but I noticed that it's possible to submit a garbage binding, which results in:

Yes, this is currently as expected. You can't have JS bindings without error handling. This is simply the catch-all for a single element on the page displaying the first error that appeared in a binding. We will improve the UX by making individual error states for each property so that it's clearer what exactly is erroring.

mbrookes:
> Something like `bind the "rows" property...`

Yes, although I'm not aware of a precedent for that outside of menus.

Janpot:
Adding this in https://github.com/mui/mui-toolpad/pull/487

PR #372, opened 2022-05-06
[sheets] Change the row id select control to a custom component

bharatkashyap:
https://user-images.githubusercontent.com/19550456/167110237-33939ed3-821c-45a7-b7ab-4497521e6c11.mov


bharatkashyap:
@Janpot 

Two issues with providing the list of all columns in the dropdown:

1. The "id" field gets added to the dropdown even thought it does not actually exist in the data
2. Selecting a column for which there are missing values in any row stops the rendering of the DataGrid and replaces it with an error 

For the first, we can limit the list to only the columns present in `props.rows[0]`, but how does the user "go back" to a system-generated "id" column if they wish to?

for the second, perhaps it's okay to show an error and nothing else is required on our end?

Janpot:
> Selecting a column for which there are missing values in any row stops the rendering of the DataGrid and replaces it with an error

We should think about how we can use the DataGrid error UI to display this. But let's concentrate on that separately.

Janpot (approved):

PR #377, opened 2022-05-09
Default REST datasource

### LINKED ISSUE #364: Add default Rest datasource
Add a default Rest datasource when a project is created so that a user can start creating APIs right away

bharatkashyap:
- Fixes #364

Earliest implementation of adding a default `rest` connection inside `createApp`.  

Perhaps it is better to add this logic inside the client after `createAppMutation.mutateAsync([name]);`,the issue there is that the `dom` doesn't exist yet and the `useDom()` hook isn't callable at that point.

### DIFF
--- a/packages/toolpad-app/src/server/data.ts
+++ b/packages/toolpad-app/src/server/data.ts
@@ -159,7 +159,6 @@ export async function createApp(name: string): Promise<App> {
         status: appDom.createConst(null),
       },
     });
-    console.log(newNode);
     const newDom = await appDom.addNode(dom, newNode, appNode, 'connections');
     await saveDom(app.id, newDom);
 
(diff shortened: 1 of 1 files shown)


packages/toolpad-app/src/server/data.ts
```diff
@@ -151,7 +151,17 @@ export async function createApp(name: string): Promise<App> {
...
-    await saveDom(app.id, dom);
+    const appNode = appDom.getApp(dom);
+    const newNode = appDom.createNode(dom, 'connection', {
+      attributes: {
+        dataSource: appDom.createConst('rest'),
+        params: appDom.createSecret({ name: 'rest' }),
+        status: appDom.createConst(null),
+      },
+    });
+    console.log(newNode);
```
Janpot, line 162:
```suggestion
```

Janpot (approved):
Can be merged after addressing the `console.log`

PR #396, opened 2022-05-10
First pass at creating queries+APIs [WIP]

Janpot:
Initial pass


prakhargupta1:
A concern that I felt in this version: I can't see add query button option until I create a page. It will be a problem for those who like to establish DB connection and fetch data first.

Like we are showing a Connection by default. I suggest we also show the first blank page by default. This will also make discovery of Components slider easier. Currently, it appears only after the page is created.

PR #400, opened 2022-05-10
Post data-fetch transformation

### LINKED ISSUE #366: Datafetching, serverside post-transformation
## Problem

My API data needs shaping before it's usable on the client. I'd like to be able to have it pretransformed on the server so that I can use it in my preferred shape on the browser

## Solution

In the API editor, at the bottom, add a code editor that allows writing a transformation to be applied to the API result. This should contain a javascript function where the first parameter is the data to be transformed and it returns the transformed object. e.g.

```ts
(data) => Object.entries(data.myProperty).map((data, index) => ({ ...data, id: index }))
```

Apply this transformation serverside before sending the API response

## Note

Make sure to add this to https://github.com/mui/mui-toolpad/issues/365  if that gets implemented first

bharatkashyap:
- Fixes #366 

### DIFF
--- a/packages/toolpad-app/src/server/data.ts
+++ b/packages/toolpad-app/src/server/data.ts
@@ -362,17 +362,6 @@ async function applyTransform<Q>(
   };
 }
 
-async function applyTransform<Q>(
-  api: appDom.ApiNode<Q>,
-  result: ApiResult<{}>,
-): Promise<ApiResult<{}>> {
-  return {
-    data: await evalExpression(
-      `${api.attributes.transform?.value}(${JSON.stringify(result.data)})`,
-    ),
-  };
-}
-
 export async function execApi<Q>(
   appId: string,
   api: appDom.ApiNode<Q>,
--- a/packages/toolpad-app/src/server/data.ts
+++ b/packages/toolpad-app/src/server/data.ts
@@ -372,7 +372,7 @@ export async function execApi<Q>(
     const apiResult = await dataSource.exec(connection, api.attributes.query.value, params);
     return {
       data: await evalExpression(
-        `${api.attributes.transform.value}(${JSON.stringify(apiResult.data)})`,
+        `${api.attributes.transform?.value}(${JSON.stringify(apiResult.data)})`,
       ),
     };
   }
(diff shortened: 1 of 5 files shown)


packages/toolpad-app/src/server/handleDataRequest.ts
```diff
@@ -29,6 +29,18 @@ export default async (
   const apiNodeId = req.query.queryId as NodeId;
   const dom = await loadVersionedDom(appId, version);
   const api = appDom.getNode(dom, apiNodeId, 'api');
+  if (api.attributes.transformEnabled.value) {
```
Janpot, line 32: Won't this fail for existing apis in toolpad? Their `transformEnabled` property and `transform` property will be `undefined`

bharatkashyap: Yes, correct - should use optional chaining for the `transform` and `transformEnabled` accesses everywhere

packages/toolpad-app/src/server/data.ts
```diff
@@ -376,7 +377,14 @@ export async function execApi<Q>(
       `Unknown connection "${api.attributes.connectionId.value}" for api "${api.id}"`,
     );
   }
-
+  if (api.attributes.transformEnabled?.value) {
+    const apiResult = await dataSource.exec(connection, api.attributes.query.value, params);
```
Janpot, line 381: To express the logic more clearly, perhaps extract it out of the `if` statement like:

```tsx
let result  = await exec()
if (transform) {
  result = transform(result)
}
return result;
```

packages/toolpad-app/src/server/handleDataRequest.ts
```diff
@@ -29,6 +29,18 @@ export default async (
...
   const api = appDom.getNode(dom, apiNodeId, 'api');
+  if (api.attributes.transformEnabled?.value) {
+    const apiResult = await execApi(
+      appId,
+      api,
+      req.query.params ? JSON.parse(req.query.params as string) : {},
+    );
+    res.json({
+      data: await evalExpression(
+        `${api.attributes.transform?.value}(${JSON.stringify(apiResult.data)})`,
```
Janpot, line 40: Isn't it already transformed inside of `execApi`?

bharatkashyap: Yes this needn't be there - for some reason I think I found a case in the early exploration when both functions need independent transformation, but testing now and can't find such a case anymore

bharatkashyap:
Two-dot diff for this branch: https://github.com/mui/mui-toolpad/compare/4fa3752..f1c195e

Janpot (approved):

PR #412, opened 2022-05-13
Export prop controls as FCs

bharatkashyap:
- Refactor the property control components to export React Functional Components directly


Janpot (approved):

packages/toolpad-app/src/components/propertyControls/index.tsx
```diff
@@ -10,9 +9,10 @@ import GridColumns from './GridColumns';
 import HorizontalAlign from './HorizontalAlign';
 import VerticalAlign from './VerticalAlign';
 import RowIdFieldSelect from './RowIdFieldSelect';
+import { EditorProps } from '../../types';
 
 const propTypeControls: {
-  [key in ArgControlSpec['type']]?: PropControlDefinition;
+  [key in ArgControlSpec['type']]?: React.FC<EditorProps<any>>;
```
oliviertassinari, line 15: React.RC adds the `children` prop, which doesn't seem to be accepted. Maybe?
```suggestion
  [key in ArgControlSpec['type']]?: (props: EditorProps<any>) => React.ReactNode;
```

Janpot: Upcoming changes to the `@types/react` will remove the `children` prop from `React.FC`. So we should be good once we upgrade I think. (waiting for `react-query` to hit v4 stable)

Issue #420, opened 2022-05-15
Rename page title based on context


oliviertassinari:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

I'm on this page:

<img width="258" alt="Screenshot 2022-05-16 at 01 22 22" src="https://user-images.githubusercontent.com/3165635/168498463-3a5d2110-f3a8-4331-ac16-a38874b47869.png">
 
but the tabs tell me nothing about it:

<img width="216" alt="Screenshot 2022-05-16 at 01 22 59" src="https://user-images.githubusercontent.com/3165635/168498489-1b86f56d-acef-4420-a33f-d3caa001c43f.png">

### Examples 🌈

I would expect: "Name of the page | Toolpad editor"

### Motivation 🔦

Make it easier to find previous active work. I have struggled many times to find my editor back among my long list of open tabs 😁. It's a bit related to #234 for the favicon.

PR #421, opened 2022-05-16
Add some page titles

Janpot:
Add a title for page and component editors, and also for the page runtime

closes https://github.com/mui/mui-toolpad/issues/420


packages/toolpad-app/src/utils/usePageTitle.ts
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+
+export default function usePageTitle(title: string) {
```
oliviertassinari, line 3: So no Next.js https://nextjs.org/docs/api-reference/next/head 😁.

Janpot: We can use it. At the moment Next.js is mostly used for top-level routing and API routes. Almost all other parts of Toolpad are Next.js agnostic (The editor and the applications use React router, lazy loading uses `React.lazy`,...). It depends on how much we want to lock into their APIs. 
In my opinion, their routing doesn't always work all that well for SPAs.

PR #426, opened 2022-05-16
Use docker in render.com

Janpot:
Try out using the Dockerfile for starting the render web service, to avoid divergence and breakage between docker and render.com


Janpot:
Works but about half as slow. We added a smoke test in the meantime

PR #435, opened 2022-05-17
Make the editor deal with pages that have deleted components on them

Janpot:
Instead of crashing the whole app


packages/toolpad-app/src/components/AppEditor/PageEditor/ComponentEditor.tsx
```diff
@@ -67,7 +67,7 @@ function SelectedNodeEditor({ node }: SelectedNodeEditorProps) {
 
   return (
     <Stack direction="column" gap={1}>
-      <Typography variant="subtitle1">Component: {component.displayName}</Typography>
+      <Typography variant="subtitle1">Component: {component?.displayName || '<unkown>'}</Typography>
```
apedroferreira, line 70: `<unknown>` instead of `unkown`? :)

PR #439, opened 2022-05-18
Hide APIs

Janpot:
Pulled from https://github.com/mui/mui-toolpad/pull/396 So we can deploy it separately


apedroferreira (approved):
code looks good to me!

i noticed you tend to use the ternary operator for rendering things conditionally, while i usually use something like `apis.length > 0 && <ContentToRender>` - let me know if it's preferable to use the ternary operator in this project,

Janpot:
I always prefer ternary as I once was bitten by the same bug as in https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

Janpot:
@apedroferreira great find, but the version of `eslint-plugin-react` containing the rule isn't released yet. We can enable it once it gets released.

PR #441, opened 2022-05-19
Start on CSP

Janpot:
Continuation of https://github.com/mui/mui-toolpad/pull/269 (closing that one since it's acting strange on render.com)


packages/toolpad-app/src/csp.ts
```diff
@@ -0,0 +1,15 @@
+import { randomBytes } from 'crypto';
+
+export function generateNonce(): string {
+  return randomBytes(16).toString('base64');
+}
+
+export function getCsp(nonce: string): string {
```
apedroferreira, line 7: why don't we set the CSP in `next.config.js`? is there an advantage to doing it this way?

Janpot: We need a nonce for an inline script that needs to execute before react is loaded. The script is used to load the react devtools agent when the page is opened in the editor iframe.
https://github.com/mui/mui-toolpad/blob/26305678c1e49532eaa55375a9ad01ff9fea5c49/packages/toolpad-app/pages/_document.tsx#L18-L31

PR #448, opened 2022-05-20
Remove default props in favour for defaultValue in argTypes

Janpot:
@apedroferreira


packages/toolpad-app/src/components/AppEditor/CodeComponentEditor/index.tsx
```diff
@@ -176,7 +181,15 @@ function CodeComponentEditorContent({ theme, codeComponentNode }: CodeComponentE
 
   const { Component: GeneratedComponent, error: compileError } = useCodeComponent(input);
 
-  const CodeComponent: React.ComponentType<any> = useLatest(GeneratedComponent) || Noop;
+  const CodeComponent: ToolpadComponent = useLatest(GeneratedComponent) || Noop;
+
+  const defaultProps = React.useMemo(() => {
```
apedroferreira, line 186: My preferred way of doing something like this would be `Object.keys(...).reduce(...)` or using a mapping utility from Lodash, maybe it's (arguably) more readable and less verbose?
Anyway this is fine too, just leaving other some possible ideas.

apedroferreira (approved):

packages/toolpad-app/src/components/AppEditor/CodeComponentEditor/index.tsx
Janpot: I don't thing `reduce` is necessarily a more readable method (It more often [is not](https://twitter.com/jaffathecake/status/1213077702300852224) in most cases, but this is entering personal preference territory). But I understand the concern, I've added a `mapValues` utility.

apedroferreira: Yeah, they're just different ways of writing and down to personal taste, I was just trying to get your input on how I would probably have done it.
Those utilities are great, thanks for that!
And I agree on the comment about lodash, especially for simple stuff like this!

PR #449, opened 2022-05-20
Default page on new apps

### LINKED ISSUE #438: Add default page on new apps
* Initialize a new empty page for new applications
* When accessing the root app url in the editor
  * there are pages => redirect to the first page
  * there are no pages => show empty state (button "create a page")

bharatkashyap:

Fixes #438

### DIFF
--- a/packages/toolpad-app/src/components/AppEditor/AppEditorShell.tsx
+++ b/packages/toolpad-app/src/components/AppEditor/AppEditorShell.tsx
@@ -15,12 +15,10 @@ import {
 } from '@mui/material';
 import * as React from 'react';
 import { useForm } from 'react-hook-form';
-import { Outlet, useNavigate, useLocation } from 'react-router-dom';
+import { Outlet, useNavigate } from 'react-router-dom';
 import RocketLaunchIcon from '@mui/icons-material/RocketLaunch';
 import DialogForm from '../DialogForm';
-import { useDomLoader, useDom } from '../DomLoader';
-import { LocationState } from '../../types';
-import * as appDom from '../../appDom';
+import { useDomLoader } from '../DomLoader';
 import ToolpadAppShell from '../ToolpadAppShell';
 import PagePanel from './PagePanel';
 import client from '../../api';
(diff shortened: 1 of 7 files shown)


packages/toolpad-app/src/components/AppEditor/AppEditorShell.tsx
```diff
@@ -109,6 +111,28 @@ export interface ToolpadAppShellProps {
 
 export default function AppEditorShell({ appId, ...props }: ToolpadAppShellProps) {
   const domLoader = useDomLoader();
+  const dom = useDom();
```
Janpot, line 114: I don't think that reasoning is correct. It's not "first visit to the Editor", it's "visit the root of the editor while tehre exists pages". I'm not sure this logic belongs in this place. I think you'll need to use a [`Navigate`](https://reactrouter.com/docs/en/v6/components/navigate) component under the `/` route, here https://github.com/mui/mui-toolpad/blob/0f703b1252920837c3f8843053e19e6840d27f58/packages/toolpad-app/src/components/Studio.tsx#L18

Something like

```tsx
<Route path="/" element={hasPages ? <Navigate to={firstPageUrl} /> : <Home />} />
```

Janpot (changes requested):

packages/toolpad-app/src/components/AppEditor/AppEditorShell.tsx
```diff
@@ -109,6 +111,28 @@ export interface ToolpadAppShellProps {
...
+
+  /* Redirect to first page (if it exists) if (any of the following):
+   * 1. first visit to the Editor
+   * 2. if being redirected from a delete operation
+   */
+  React.useEffect(() => {
+    const app = appDom.getApp(dom);
+    const { pages = [] } = appDom.getChildNodes(dom, app);
+    if (
+      pages?.length &&
```
Janpot, line 129: Two Side note: 
- no need to use optional chaining `?` if it can't be `null`/`undefined`
- Always use explicit boolean check for `[].length`: `pages.length > 0`

packages/toolpad-app/src/components/AppEditor/AppEditorShell.tsx
bharatkashyap: Using the `useDom` and `useLocation` hooks inside the `<Studio>` component may not be feasible, since the dom isn't completely loaded by the time the component is rendered, and `useLocation` is only allowed within the context of a `<Router>` component

packages/toolpad-app/src/components/AppEditor/index.tsx
```diff
@@ -42,14 +44,39 @@ interface FileEditorProps {
...
         <Route path="codeComponents/:nodeId" element={<CodeComponentEditor appId={appId} />} />
-        <Route path="*" element={null} />
+        <Route path="codeComponents/:nodeId" element={<CodeComponentEditor appId={appId} />} />
+        <Route
+          index
+          element={
+            hasPages ? (
+              <Navigate to={`pages/${firstPage.id}`} />
+            ) : (
+              <NotFoundEditor
```
Janpot, line 73: Perhaps we can make a page that has the text "This app has no pages." and a button underneath that says "create page" that opens the page creation dialog?

bharatkashyap: Does something like this make sense? 
Screenshot 2022-05-24 at 6 37 17 PM

Intentionally trying to avoid the pattern of one line text and a button centred on the page

Janpot: > Intentionally trying to avoid the pattern of one line text and a button centred on the page

I don't see a problem with that. I don't think `Alert` is the right paradigm for an empty state.

bharatkashyap: Screenshot 2022-05-25 at 5 13 55 PM

bharatkashyap:
A side note: deleting "any" type of node (CodeComponents, Connections) will cause the redirect to happen now. Seems okay to me for now

Janpot (approved):
👌

PR #450, opened 2022-05-20
App Editor enhancements: column components

apedroferreira:
- Enhance app editor with new column components that take a span from 1 to 12.
- Disable placing rows explicitly.
- Automatic removal or addition of rows according to https://www.notion.so/mui-org/App-Editor-Enhancements-070bf8ba0e2145a2aaaf50d67a24983e.

Screen Shot 2022-05-24 at 19 58 11

@Janpot Please review and let me know if it's all good, and if we should merge this already as a first step.
I've also created a new utils file for some constants/operations that were being used a lot, let me know if that approach is good or if you prefer some other way, I'm totally fine with changing it.

Will try to enable drag & drop with quadrants next up, so that columns can be created automatically when an element is added under another, among other features that will should improve the UX a lot.
Also we can eventually remove the spacing in rows and columns, or being able to edit them or see them, if we later decide that's the way to go.

### DIFF
--- a/packages/toolpad-app/src/components/AppEditor/PageEditor/ComponentCatalog.tsx
+++ b/packages/toolpad-app/src/components/AppEditor/PageEditor/ComponentCatalog.tsx
@@ -7,7 +7,7 @@ import * as appDom from '../../../appDom';
 import { useDom } from '../../DomLoader';
 import { usePageEditorApi } from './PageEditorProvider';
 import { useToolpadComponents } from '../toolpadComponents';
-import { PAGE_ROW_COMPONENT_ID } from '../../../utils/components';
+import { PAGE_ROW_COMPONENT_ID } from '../../../toolpadComponents';
 
 const WIDTH_COLLAPSED = 50;
 
--- a/packages/toolpad-app/src/components/AppEditor/PageEditor/ComponentCatalog.tsx
+++ b/packages/toolpad-app/src/components/AppEditor/PageEditor/ComponentCatalog.tsx
@@ -92,23 +92,25 @@ export default function ComponentCatalog({ className }: ComponentCatalogProps) {
         <Collapse in={!!openStart} orientation="horizontal" timeout={200} sx={{ height: '100%' }}>
           <Box sx={{ width: 300, height: '100%', overflow: 'auto' }}>
             <Box display="grid" gridTemplateColumns="1fr" gap={1} padding={1}>
-              {Object.entries(toolpadComponents).map(([componentId, componentType]) => {
-                if (!componentType) {
-                  throw new Error(
-                    `Invariant: Component definition for "${componentId}" is undefined`,
+              {Object.entries(toolpadComponents)
+                .filter(([componentId]) => componentId !== 'PageRow')
+                .map(([componentId, componentType]) => {
+                  if (!componentType) {
+                    throw new Error(
+                      `Invariant: Component definition for "${componentId}" is undefined`,
+                    );
+                  }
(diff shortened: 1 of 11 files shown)


packages/toolpad-app/src/components/AppEditor/PageEditor/RenderPanel.tsx
```diff
@@ -688,7 +688,8 @@ export default function RenderPanel({ className }: RenderPanelProps) {
 
           const isOnlyRowElement =
             dragParent &&
-            getElementNodeComponentId(dragParent as appDom.ElementNode) === PAGE_ROW_COMPONENT_ID &&
+            (dragParent as appDom.ElementNode).attributes.component?.value ===
```
apedroferreira, line 691: is there a better way to get around the types here?
the default type doesn't let me try to use `component`

packages/toolpad-app/src/components/AppEditor/PageEditor/ComponentCatalog.tsx
```diff
@@ -92,23 +93,25 @@ export default function ComponentCatalog({ className }: ComponentCatalogProps) {
         <Collapse in={!!openStart} orientation="horizontal" timeout={200} sx={{ height: '100%' }}>
           <Box sx={{ width: 300, height: '100%', overflow: 'auto' }}>
             <Box display="grid" gridTemplateColumns="1fr" gap={1} padding={1}>
-              {Object.entries(toolpadComponents).map(([componentId, componentType]) => {
-                if (!componentType) {
-                  throw new Error(
-                    `Invariant: Component definition for "${componentId}" is undefined`,
+              {Object.entries(toolpadComponents)
+                .filter(([componentId]) => componentId !== PAGE_ROW_COMPONENT_ID)
```
apedroferreira, line 97: just filtering out rows here

apedroferreira:
also can we run prettier (and lint) on pre-commit if they're not running there? just so that the CI won't fail for stuff like this.
i can add it if you think it's cool.

packages/toolpad-app/src/components/AppEditor/PageEditor/RenderPanel.tsx
Janpot: The parent of an element can either be a `PageNode` or another `ElementNode`. A `PageNode` doesn't have a `component` attribute. That's why typescript is complaining here.

There are methods like `appDom.isPage`, or `appDom.isElement` that can act as type guards. You can use it in a  conditional to get rid of the type warnings without resorting to an `as` type assertion.

apedroferreira: ok, i'll keep the `as` if it's ok, so i don't to have to nest part of the logic inside an `if` - i feel like the way i wrote it is pretty explicit but lmk if you disagree

apedroferreira: ok i'll add the checks then, i get your point too

packages/toolpad-app/src/utils/components.ts
```diff
@@ -0,0 +1,8 @@
+import type * as appDom from '../appDom';
```
Janpot, line 1: Kinda feels like this logic belongs under `./src/toolpadComponents` so that we can reserve `utils` for things that are toolpad agnostic

Janpot (approved):

Janpot:
👌  Looking great!

PR #452, opened 2022-05-20
Get rid of next.js custom server

Janpot:
It was only there to support basic auth.

_edit_

Leaving this in draft for a while. _middlewares is a beta feature and they just announced there will be a couple of changes to it. This PR should be compatible with the new changes.


apedroferreira (approved):
nice!

Issue #432, opened 2022-05-17
Simplify dev workflow


Janpot:
Currently one has to run `yarn dev` some where to build all the projects, and then `yarn cli` to run. We should consider combining the commands and have a single `yarn dev` command start the cli and watch/build all the packages

apedroferreira:
Yes, a single `yarn dev` script to run the app locally should cover all my (very minor) pain points when trying to run the project locally:
- `yarn dev` being a command that builds instead of running the app as is typical/expected in many projects
- `yarn dev` not exiting because it watches for changes in the code, which would not be the intended usage for someone who is just setting up the project to run locally for the first time
- `yarn cli` not being a very explicit name for the command that runs the app

PR #460, opened 2022-05-24
Remove cli package

Janpot:
It wasn't doing anything significant anymore. Removing it to make it easier to split up to ee and cloud.

This also closes https://github.com/mui/mui-toolpad/issues/432


CONTRIBUTING.md
```diff
@@ -29,7 +29,6 @@ If you would like to hack on MUI Toolpad or want to run the latest version, you
 
    ```sh
    yarn install
```
apedroferreira, line 31: there's also some text here about "building in watch mode", we might want to move that text to another step too?

PR #463, opened 2022-05-25
Run prettier in pre-commit hook

apedroferreira:
On every commit, run prettier for staged files inside the `toolpad-app` package.


apedroferreira:
> I don't have the best experiences with pre-commit hooks. You don't want to install the vscode extension?

i have the extension but on saving some files they don't get fixed the exact same way as running `yarn prettier`.
i never had any issues with pre-commit hooks... at least when they're simple like this.
but i don't mind us keeping things the same, this PR was just a suggestion

> We're also already using the prettier cli from the @mui/material monorepo which checks staged files
i guess we don't need lint-staged then

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
```
Janpot, line 1: Is this supposed to be committed? How does husky make this cross platform?

apedroferreira: but yeah that's how it works, i think the precommit script is supposed to be commited

apedroferreira:
Let's not merge this for now, doesn't seem so simple to make it work with our own `yarn prettier` script after all

PR #465, opened 2022-05-25
lint staged

Janpot:
@apedroferreira How about this?

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+yarn prettier
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-yarn prettier
+yarn prettier check-changed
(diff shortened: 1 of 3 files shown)


apedroferreira:
> @apedroferreira How about this?

it throws an error when there are prettier issues?
not sure, ideally it would fix them automatically :/ 
also the error it shows when the checks fail is a bit ugly

but like i said i was not being able to make prettier include the formatting fixes in the commit changes when i tried it...

apedroferreira:
something like `yarn prettier && git add .` works and includes the formatting changes in the commit, but of course it's bad because it adds every file even if it's not staged...
so maybe my first PR with `lint-staged` wasn't so bad as it only applies to staged files -  it was the first commit here https://github.com/mui/mui-toolpad/pull/463/commits/4243e7a0c7f21597d3035eb788a4015309a979d4

apedroferreira:
> > ideally it would fix them automatically
> 
> I'm not sure I feel comfortable with automations committing changes that I didn't explicitly add myself.

i can see that, i'm fine with keeping things as they are then, no problem with it! :)

Janpot:
This runs it on changed files, not on staged files, so not very useful after all. `lint-staged` would be better in that sense, but still I sometimes have unstaged chunks that may contain lint errors.

Is there a way to make the hook optional, so that you may enable it just for you?

apedroferreira:
> This runs it on changed files, not on staged files, so not very useful after all. `lint-staged` would be better in that sense, but still I sometimes have unstaged chunks that may contain lint errors.
> 
> Is there a way to make the hook optional, so that you may enable it just for you?

don't worry about it, i don't think it's worth adding something just so i can use it
running `yarn prettier` manually is still fine

Issue #101, opened 2022-02-01
drag & drop next steps


Janpot:
# Visual editor

This RFC tries to specify a paradigm for our visual editor. For internal tooling we want something that allows for flexible layout, yet is constrained enough to require minimal manual tweaking.

## Regular components

- All components conform to the width given to them by the layout algorithm. For some components this means they will have to centre themselves in a box (e.g. a button my take width from its content, which may be more narrow than the available width) some components will have an option to expand to the full width of their box (e.g. `fullWidth` property on `TextField`). Other components will horizontally adapt to any width imposed on them (e.g. image, datagrid)
- All components define the height they want to be rendered at. For some components this will be always the same, e.g. textinputs have a fixed height. And for some it will be configurable. e.g. data grid or image

## Layout components

In terms of layout we identify two types of components, columns and rows.

### column

a column works vertically. It acts as a component on the horizontal axis (adapt to the width imposed to it by the layout algorithm). It serves as the container for rows. Its only possible children are rows and its only possible parent is a row. A only needs to be manually created when the user wants to build something like a sidebar.

- A column only has rows as its children. Dropping a component on a column where there is no row, implicitly creates a row. Removing the last child from a row, removes that row from the page.

### rows

Rows are never created explicitly, they are implicitly created depending on the drop position of certain elements

- A row has the same width as the column it’s under.
- A row is a 12 column grid layout into which its children take their place.
- A child of a row takes up a certain amount of columns (span). The user can adjust the span of an element with drag handles on the left and right side of the element.

### page

On the top level, a page acts as a column

## Drop algorithm

We must avoid layout shifting as much as possible, therefor we will not live preview elements as they are dragged on the page. Dragging a new element only highlights cursors as available drop targets. Dragging an existing element, to move it around, stays in place until the drop is performed.

The biggest challenge of drag and drop visual editor is to calculate a predictable insertion position for the drop target. I propose the following algorithm:

- around top level rows there is a 5px (?) band which always takes precedence as a drop target to create a new row that contains the element being dragged. We may shift the page rows apart by the same gap during dragging if this can be done with minimal layout shift.

For any other [X,Y] mouse position during the drag 

- find the most deeply nested component under the mouse
- divide the component into 4 quadrants by connecting its corners diagonally, determine which quadrant the mouse is over to determine the side of the box that will be the drop target.
[...]

PR #466, opened 2022-05-26
Improve visual editor UX

apedroferreira:
https://www.notion.so/mui-org/App-Editor-Enhancements-070bf8ba0e2145a2aaaf50d67a24983e

Related to #101

This is an improvement proposal for our drag & drop. It may not be perfect yet but I feel like it's an improvement and a step in the right direction.
There might still be uncaught bugs or small UX issues, but overall seems fine from what has been tested. The rest can be fixed and improved after receiving feedback.

Refactored drag & drop with several features:
- replace old slots system with "quadrants" over every element to determine where the new element will be placed. A center area in container elements is also accounted for besides top, right, bottom and left.
- remove ability to add row and columns - those are automatically created and can only be placed at the top level (can be changed, it's just a proposal to experiment with)
- restored some of the previous defaults for page rows and columns, since I felt like I should not have changed them in previous PRs

Video example:

https://user-images.githubusercontent.com/10789765/172681252-8299025a-7e60-4c03-a308-8a1cac17361c.mov

@Janpot I tried to be careful but let me know if I deleted something I shouldn't, or if some of the new utility methods I created would be better in some other place.


packages/toolpad-core/src/types.ts
```diff
@@ -84,25 +84,25 @@ export interface FunctionValueType extends ValueTypeBase {
...
-  | 'slots' // slots in canvas
-  | 'multiSelect' // multi select ({ type: 'array', items: { type: 'enum', values: ['1', '2', '3'] } })
-  | 'date' // date picker
-  | 'json' // JSON editor
-  | 'GridColumns' // GridColumns specialized editor
-  | 'HorizontalAlign'
-  | 'VerticalAlign'
-  | 'function'
-  | 'RowIdFieldSelect'; // Row id field specialized select
+    | 'boolean' // checkbox
```
Janpot, line 87: Is it your editor prettier extension doing that? If so, let's debug that next week

apedroferreira: maybe it's something in my global vscode settings, i'll take a look

apedroferreira:
@Janpot I've made all the changes and added all the fixes we talked about yesterday, hopefully it's better now.
The next step would be implicit element containers that take up all remaining space in a row, but that will be a separate PR.

Janpot:
The space between two elements is not behaving the best

https://user-images.githubusercontent.com/2109932/172687667-090c8ba8-564f-46b3-bd57-1ccc668db295.mov

apedroferreira:
> The space between two elements is not behaving the best
> 
>  Screen.Recording.2022-06-08.at.20.12.51.mov

apedroferreira:
> The space between two elements is not behaving the best
> 
>  Screen.Recording.2022-06-08.at.20.12.51.mov

i see what happened there, you dragged to the space between two elements which means you were hovering over the parent, and you placed it in the center of the parent, which adds an element as the last element in the parent...

i can try to handle situations like this but will have to change to logic to somehow not consider exactly the element you're hovering over - and be able to see which elements are close to the cursor in some way

Janpot:
I think a misconception that this implementation makes is that slots go on the edges of the components. Components are laid out in a structure that can have a configurable gap. If the slot goes on the edge, that makes it so there are two slots in between two consecutive children.

I think a better mental model would be that the layout components (rows, columns) divide the page in rectangles that the user components (button, input, table,...) occupy.  Insert slots appear on the edges of these rectangles. For the drag&drop algorithm, these rectangles don't have spacing, the spacing only applies to the components that occupy the grid.

apedroferreira:
> I think a misconception that this implementation makes is that slots go on the edges of the components. Components are laid out in a structure that can have a configurable gap. If the slot goes on the edge, that makes it so there are two slots in between two consecutive children.
> 
> I think a better mental model would be that the layout components (rows, columns) divide the page in rectangles that the user components (button, input, table,...) occupy. Insert slots appear on the edges of these rectangles. For the drag&drop algorithm, these rectangles don't have spacing, the spacing only applies to the components that occupy the grid.

yeah, i wonder if i should have kept the previous slots and added the new behavior on top of it somehow... but it seemed like a good idea to do a complete refactor before, as i had discussed with you.
i'll try to change things to work like that under the new system, i agree that that approach feels more intuitive in that way.

packages/toolpad-app/src/appDom.ts
```diff
@@ -701,6 +700,58 @@ export function getNodeIdByName(dom: AppDom, name: string): NodeId | null {
...
+}
+
+export function getNewParentIndexBeforeNode(dom: AppDom, node: AppDomNode, parentProp: string) {
+  const parent = getParent(dom, node);
+
+  if (!parent) {
+    throw new Error(`Invariant: Node: "${node.id}" has no parent`);
+  }
+
+  const siblings: readonly AppDomNode[] = (getChildNodes(dom, parent) as any)[parentProp] ?? [];
```
Janpot, line 732: Wouldn't this work as well?
```suggestion
  const siblings = getSiblings(dom, node)
```

packages/toolpad-app/src/components/AppEditor/PageEditor/RenderPanel.tsx
```diff
@@ -619,106 +529,334 @@ export default function RenderPanel({ className }: RenderPanelProps) {
...
+
+        const isDraggingOverPage = appDom.isPage(activeDropNode);
+        const isDraggingOverRow = appDom.isElement(activeDropNode) && isPageRow(activeDropNode);
+        const isDraggingOverContainer = activeNodeDropInfo
+          ? hasContainerComponent(activeNodeDropInfo)
+          : false;
+
+        let centerAreaFraction = 0;
+        if (isDraggingOverContainer) {
+          centerAreaFraction = 0.625;
```
Janpot, line 556: I believe we can make this binary. no need for `centerAreaFraction`. Either it's a container and we detect overlap with `rectContainsPoint` or it's not and we detect the edge using the quadrants algorithm.
(I'd also name it `getDiagonalQuadrant(rectangle, x, y)` or so to be a bit more descriptive)

Janpot:
The slot for the container element seems to have lost its margins left and right

Screenshot 2022-06-13 at 10 24 07

compared to:

Screenshot 2022-06-13 at 10 25 52

packages/toolpad-app/src/appDom.ts
```diff
@@ -701,6 +700,58 @@ export function getNodeIdByName(dom: AppDom, name: string): NodeId | null {
   return index.get(name) ?? null;
 }
 
+export function getNewFirstParentIndexInParentNode(
```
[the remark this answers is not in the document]
apedroferreira: i guess i deleted whatever was using them at some point, yeah good catch, will delete then

packages/toolpad-app/src/components/AppEditor/PageEditor/RenderPanel.tsx
apedroferreira: i see, this answers my previous comment, ok let's try that

apedroferreira:
Yeah I noticed that changed, as it seems not be what we want will try to change it back.
And also will look into the changes that might've broken `CustomLayout`.

packages/toolpad-app/src/appDom.ts
apedroferreira: actually the variable name should be `children` or something because i still want the node itself to be in the array - will change it to be more clear

Janpot: Is there a way we can get rid of the `as any`? Maybe witha  runtime check?

apedroferreira: i must've added it temporarily but then forgot to fix the types, thanks for catching that

apedroferreira:
@Janpot this can be tested/reviewed now, hopefully not much more is needed - i tried to find the best solutions/tradeoffs but this logic can get pretty complex.
took a lot of work to refactor for custom components / multiple slots... those components still don't show their content for some reason but other than that things should be alright.

Janpot:
Few things I'm noticing:

1. https://user-images.githubusercontent.com/2109932/175768055-7f580fdf-9976-48d6-b9ff-02940c8af139.mov

   Pretty strange that the whole thing suddenly shrinks after I add a button on top. Been tweaking a bit on the layout components in https://github.com/mui/mui-toolpad/pull/602 to see if I can make them behave more like I'd expect. It does break the slots algorithm, I didn't look into that.

2. None of the components are interactive anymore in the canvas. You used to be able to click the selected button, or fill the selected TextField

3. Can't insert at the top of the page:

    https://user-images.githubusercontent.com/2109932/175768158-9d4abe8e-f385-4f40-8e62-6a60090184fe.mov

apedroferreira:
1 and 3 seem like similar issues, let's see if increasing the drop area of the top row to include the top padding of the page gets rid of the issue.
number 2 must've been a regression i've missed, will try to restore things to how they were before.

apedroferreira:
@Janpot this might be mergeable now, i thing the UX is feeling alright in general, lmk if you find any more issues!
i've seen just a few minor issues in some rare cases, i can handle those later along with everyone's feedback

prakhargupta1:
When dropping a component, the default alignment on the left felt better than the central alignment.
It could be a personal choice also I don't know how much rework would it require, so just sharing my view.

Janpot (approved):
👍  Great work! Let's merge this and continue polishing in follow-up PRs

PR #467, opened 2022-05-26
Add Menu to App card

### LINKED ISSUE #292: App naming and menu item
Show a menu item per app that allows:

- [ ] 1. Duplicating an app under a new name.

- [x] 2. Rename an existing App.

- [x] 3. Deleting App should also happen from the menu item and not from the App card.

- [ ] Currently, Apps with the same name can be created. Let's add a database constraint and show an error when trying to name/rename an app to an app with existing name

bharatkashyap:
- Fixes #292 

https://user-images.githubusercontent.com/19550456/170538781-70b0743d-9e62-4d97-9fe7-cfecb0872fc5.mov

^ The version with "try" is an example of hitting "Escape" when renaming

### DIFF
--- a/packages/toolpad-app/src/components/Home.tsx
+++ b/packages/toolpad-app/src/components/Home.tsx
@@ -156,7 +156,6 @@ function AppCard({ app, onDelete, onUpdate }: AppCardProps) {
     async (name: string) => {
       if (app?.id && onUpdate) {
         onUpdate(await client.mutation.updateApp(app.id, name));
-        // await client.refetchQueries('getApps');
       }
     },
     [app?.id, onUpdate],
@@ -277,20 +276,21 @@ function AppCard({ app, onDelete, onUpdate }: AppCardProps) {
 }
 
 export default function Home() {
-  const { data: apps = [], status, error } = client.useQuery('getApps', []);
+  const [apps, setApps] = React.useState<App[]>([]);
+
+  const { status, error } = client.useQuery('getApps', [], {
+    onSuccess: (data: App[]) => setApps(data),
+  });
 
   const [createDialogOpen, setCreateDialogOpen] = React.useState(false);
 
   const [deletedApp, setDeletedApp] = React.useState<null | App>(null);
 
-  const [updatedApp, setUpdatedApp] = React.useState<null | App>(null);
-
-  const fetchedApps = React.useMemo(() => {
-    if (updatedApp) {
-      return apps.map((app) => (app.id === updatedApp.id ? updatedApp : app));
-    }
-    return apps;
-  }, [updatedApp, apps]);
+  const onUpdate = React.useCallback((updatedApp: App) => {
+    setApps((previousApps) => {
+      return previousApps.map((prevApp) => (prevApp.id === updatedApp.id ? updatedApp : prevApp));
+    });
+  }, []);
 
   return (
     <ToolpadShell>
@@ -322,12 +322,12 @@ export default function Home() {
               case 'error':
                 return <Alert severity="error">{(error as Error)?.message}</Alert>;
               case 'success':
-                return fetchedApps.length > 0
-                  ? fetchedApps.map((app) => (
+                return apps.length > 0
+                  ? apps.map((app) => (
                       <AppCard
                         key={app.id}
                         app={app}
-                        onUpdate={setUpdatedApp}
+                        onUpdate={onUpdate}
                         onDelete={() => setDeletedApp(app)}
                       />
                     ))
--- a/packages/toolpad-app/src/components/Home.tsx
+++ b/packages/toolpad-app/src/components/Home.tsx
@@ -2,14 +2,15 @@ import {
   Alert,
   Button,
   Card,
-  CardActionArea,
+  CardHeader,
   CardActions,
-  CardContent,
   Container,
   Dialog,
   DialogActions,
   DialogContent,
   DialogTitle,
+  Menu,
+  MenuItem,
   Skeleton,
   TextField,
   Toolbar,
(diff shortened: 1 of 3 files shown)


bharatkashyap:
@Janpot For duplicating apps, should there be an additional rpc which creates a new app, fetches all domNodes of the original app and sequentially adds them to the new one?

packages/toolpad-app/src/components/Home.tsx
```diff
@@ -112,45 +120,178 @@ function AppDeleteDialog({ app, onClose }: AppDeleteDialogProps) {
...
+        anchorEl={menuAnchorEl}
+        open={menuOpen}
+        onClose={handleMenuClose}
+        MenuListProps={{
+          'aria-labelledby': 'basic-button',
+          dense: true,
+        }}
+      >
+        <MenuItem onClick={handleRenameClick}>
+          <DriveFileRenameOutlineIcon sx={AppCardMenuIconStyles} />
```
Janpot, line 263: Perhaps we can do it like in the demo? https://mui.com/material-ui/react-menu/#icon-menu

packages/toolpad-app/src/server/data.ts
```diff
@@ -166,6 +166,17 @@ export async function createApp(name: string): Promise<App> {
   });
 }
 
+export async function updateApp(appId: string, name: string): Promise<App> {
+  return prisma.$transaction(async () => {
```
Janpot, line 170: I don't believe this needs to be wrapped in a transaction?

packages/toolpad-app/src/components/Home.tsx
```diff
@@ -112,45 +120,178 @@ function AppDeleteDialog({ app, onClose }: AppDeleteDialogProps) {
...
+  const { status, error } = client.useQuery('getApps', [], {
+    onSuccess: (data: App[]) => setApps(data),
+  });
 
   const [createDialogOpen, setCreateDialogOpen] = React.useState(false);
 
   const [deletedApp, setDeletedApp] = React.useState<null | App>(null);
 
+  const onUpdate = React.useCallback((updatedApp: App) => {
+    setApps((previousApps) => {
```
Janpot, line 290: You should refetch the app list instead, so that the cache gets renewed. Like in https://github.com/mui/mui-toolpad/blob/3ebb35665c46646ace00b2c972cd1f0d637c4f6b/packages/toolpad-app/src/components/Home.tsx#L91

You can even directly use the `refetch` method returned by `client.useQuery('getApps'` in this case.

packages/toolpad-app/src/components/Home.tsx
```diff
@@ -112,45 +120,178 @@ function AppDeleteDialog({ app, onClose }: AppDeleteDialogProps) {
...
+    </React.Fragment>
   );
 }
 
 export default function Home() {
-  const { data: apps = [], status, error } = client.useQuery('getApps', []);
+  const [apps, setApps] = React.useState<App[]>([]);
+
+  const { status, error } = client.useQuery('getApps', [], {
+    onSuccess: (data: App[]) => setApps(data),
```
Janpot, line 282: not necessary, use `refetch` to refresh the apps

Janpot:
> For duplicating apps, should there be an additional rpc which creates a new app, fetches all domNodes of the original app and sequentially adds them to the new one?

This should fully happen on the server. But we first need to decide whether toolpad app dom nodes should have globally unique ids. Let's do it in a separate PR

packages/toolpad-app/src/components/Home.tsx
bharatkashyap: The issue with performing a refetch was that it modified the order of the apps returned, so every rename operation caused the cards to switch around

Janpot: We shouldn't order them alphabetically then, we should order by "last edited".

Janpot: By "last edited" I mean the app dom, not the database record for the app. We'll need to intoruduce another field in the db schema to express this.

bharatkashyap: Should I use `refetch` for the scope of the current PR and add this in a separate PR?

Janpot: Yep, that's good 👍

apedroferreira: not sure how we're doing it, but ideally with GraphQL caching if we return `id` and `name` from the mutation it would just update the name of the app in the cache, and we wouldn't need to refetch?

Janpot: `refetch` in the context of `react-query` is essentially just invalidating its cache for that list query. You don't need to do anything else in terms of updating UI.

bharatkashyap:
- Also adding the error on rename clash and order by last edited time to this PR, and moving "Duplicate" to a separate PR.

https://user-images.githubusercontent.com/19550456/172612608-3cdda05f-640c-46ba-bff4-87bfd96ed6d2.mov

Janpot:
> Also adding the error on rename clash

The way I would prefer this flow to work would be is that if while saving the name a conflict is detected, show a popup that explains that and asks whether you want to discard the new name or keep editing.

We should probably also think about a reusable `EditableText` component as we'll need something similar to replace the current `NodeNameEditor`. But we can extract something in a folow-up.

bharatkashyap:
https://user-images.githubusercontent.com/19550456/172625836-2939d557-8a40-403b-bdcf-806d7c17fe4a.mov

bharatkashyap:
> 👍 Looks great. To be consistent with the rest of the app I would put the positive action (keep editing) at the right and the negative action (discard) at the left

Screenshot 2022-06-08 at 3 42 25 PM

Janpot (approved):

PR #474, opened 2022-05-27
bring back some spacing

Janpot:
@apedroferreira It doesn't have to necessarily happen through this prop, but I believe we should aim at making it easy to make good looking layout by default. I propose we bring this back temporarily so that we can use Toolpad master for the demos. wdyt?


apedroferreira:
> @apedroferreira It doesn't have to necessarily happen through this prop, but I believe we should aim at making it easy to make good looking layout by default. I propose we bring this back temporarily so that we can use Toolpad master for the demos. wdyt?

sounds good! i was just playing around with the values, i think any of these things is still up to discussion/changes at this point.

apedroferreira (approved):

packages/toolpad-components/src/PageRow.tsx
```diff
@@ -30,7 +30,7 @@ export default createComponent(PageRow, {
   argTypes: {
     spacing: {
       typeDef: { type: 'number' },
-      defaultValue: 0,
+      defaultValue: 2,
```
oliviertassinari, line 33: I think that we should aim for dense UIs, I would propose instead:

```suggestion
      defaultValue: 1,
```

Janpot: Agreed. Also, since at the moment this adds both margin and gap to the rows, it creates a gap of `2` when two rows stack. We should probably make it so that this only controls `gap`, and provide a similar thing for columns so that every gap between elements has the same size throughout by default.

oliviertassinari:
A couple of things that got me confused. I mean differences between what happened and what I intuitively expected. Some of these expected behaviors might actually help the many and not just me (as sometimes it can harm too to overoptimize for one's personal expectations):

1. When we create a new column, we might want to set its span to 6. It took me some time to understand why the button was wrapping:

Screenshot 2022-05-29 at 18 39 23

when adding a column and a button inside it: https://toolpad-pr-474.onrender.com/_toolpad/app/cl3ritkw60003adpgc1oznloh/editor/pages/cl3riv3gu00033f6b37iul8r0

2. If I add a column and then try to delete it, it crashes.

Screenshot 2022-05-29 at 18 40 29

https://toolpad-pr-474.onrender.com/_toolpad/app/cl3ritkw60003adpgc1oznloh/editor/pages/cl3rj1xtu00043f6b4l5waeqi

3. Should we expect to be able to create a Row manually? I have caught myself trying to add one but couldn't find it.
4. Why is the top-level row in https://toolpad-pr-474.onrender.com/_toolpad/app/cl3ritkw60003adpgc1oznloh/editor/pages/cl3rj1xtu00043f6b4l5waeqi called `pageRow3` and not `pageRow1`? In https://toolpad-pr-474.onrender.com/_toolpad/app/cl3ritkw60003adpgc1oznloh/editor/pages/cl3riv3gu00033f6b37iul8r0, the first row is called `pageRow` and the second `pageRow1` it's also confusing.
5. In this configuration, with the Row component selected:

Screenshot 2022-05-29 at 18 44 15

I expected that clicking on the left panel "point1" page would give me focus on the whole page UI on the right panel. It doesn't. Instead, I have to switch to another page and back to it, to be able to modify the params..

6. When the spacing on a Row is 0, it's impossible to configure it anymore. e.g. https://toolpad-pr-474.onrender.com/_toolpad/app/cl3ritkw60003adpgc1oznloh/editor/pages/cl3rj3ofo00023f6bd7gck5v6

Screenshot 2022-05-29 at 18 53 04

Maybe it's a use case for a breadcrumb, I don't know:

Screenshot 2022-05-29 at 18 56 06

(Webflow)

7. When I add a column, I end up with too much spacing, but more important, too much complexity. Do we really need 4 boxes here?

Screenshot 2022-05-29 at 18 59 46

https://toolpad-pr-474.onrender.com/_toolpad/app/cl3ritkw60003adpgc1oznloh/editor/pages/cl3rjoqcq000b3f6bsb9khkru

8. Starting from step 7. How do we add a header on top of each table? I can't figure it out.

Janpot:
1. I feel like it should take up all the space available, and if something gets inserted in between two items in a row, it should redestribute the spans to give each item equal span. Then the spans should be adjustable with drag hanldes.

2. I can't reproduce

3. I think rows and columns shouldn't be created manually, they should appear/disapear depending on where you insert an element.

4. There's a renaming algorithm that prevents name clashes. As it has no relation to the layout algorithm, this an probably be moved to another discussion.

5. We should make the page panel more accessible, yes. Or even make it always visible like in https://github.com/mui/mui-toolpad/pull/289

6. We will need a breadcrump of some sort to traverse the upward hierarchy of an element (makeswift also has an interesting slution for this), as well as a hierarchy explorer as proposed by https://github.com/mui/mui-toolpad/issues/490

7. This is the problem that I describe in https://github.com/mui/mui-toolpad/pull/474#discussion_r884297867. Spacing is adding margin+gap. It shouldn't add margin since they compound.

8. In the proposed algorithm as I described it in the RFC, you would drag an element in the top quadrant of the data grid. a cursor would appear on top of the datagrid, same width. when you drop the element there, the following should happen:
    - a new column is created in place of the datagrid
    - your new element is placed in the column
    - the data grid is moved inside of that column under the element you just added
  
     in that system rows and columns get created/removed on the fly, as needed. You would never create them manually. a large portion of that RFC hasn't been implemented yet.

    > it makes me wonder if we shouldn't start from a CSS grid model

     Yes, good question, I've been wondering this too. So far it has been very challenging to get opinions on this during the team meetings.

mbrookes:
Regarding CSS grid, that seems like a given. Flexbox has been coerced (abused?) for this use case because there wasn't a better alternative. With IE now finally dead, and evergreen browsers supporting CSS grid, it makes sense to take advantage of it.

Janpot:
Adjusted the spacing according to my comment https://github.com/mui/mui-toolpad/pull/474#discussion_r884297867

Also threw in a `Container` for good measure. This is how those previous screenshot look now:

Screenshot 2022-05-30 at 09 32 34

Screenshot 2022-05-30 at 09 32 27

apedroferreira:
Hey, Jan's answers were pretty good, just answering some of the points myself:
1. Nothing decided here yet, but the idea right now is to make new columns take up the whole remaining horizontal space when added.
2. Lmk if you got a repro, I'm completely rewriting this logic though right now.
6. Drag handles should solve this and then allow us to see which spacing works best.
7. The boxes were just so we can see where elements are and test out these changes - i don't think the final solution should have so many rectangles :D It should be much simpler.

CSS grid sounds good, but we would need some kind of proposal for how it should work!

Issue #486, opened 2022-05-28
"Remove binding" button doesn't close the binding dialog


mbrookes:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

The binding is cleared, but the dialog stays open.

### Expected behavior 🤔

The dialog should close.

### Steps to reproduce 🕹

Steps:

1. Open an existing binding
2. Click "Remove binding"

Janpot:
👍  https://github.com/mui/mui-toolpad/pull/487

PR #487, opened 2022-05-28
Close binding editor on "remove binding"

Janpot:
also tweak tooltip with `...` and make sure it doesn't interfere with overlapping controls

closes https://github.com/mui/mui-toolpad/issues/486

### DIFF
--- a/packages/toolpad-app/src/components/AppEditor/BindingEditor.tsx
+++ b/packages/toolpad-app/src/components/AppEditor/BindingEditor.tsx
@@ -138,6 +138,11 @@ export function BindingEditor<V>({
     handleClose();
   }, [onChange, input, handleClose]);
 
+  const handleRemove = React.useCallback(() => {
+    onChange(null);
+    handleClose();
+  }, [onChange, handleClose]);
+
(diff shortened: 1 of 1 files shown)


packages/toolpad-app/src/components/AppEditor/BindingEditor.tsx
```diff
@@ -155,11 +160,11 @@ export function BindingEditor<V>({
 
   const TooltipComponent = error ? ErrorTooltip : Tooltip;
   const tooltipTitle: string =
-    error ?? (hasBinding ? `Update "${label}" binding` : `Bind "${label}"`);
+    error ?? (hasBinding ? `Update "${label}" binding...` : `Bind "${label}"...`);
```
mbrookes, line 163: Thanks for the fix. If you're onboard with using an ellipsis in tooltips to indicate that a button opens a dialog, then perhaps the unicode "…" is preferable to  three full-stops/periods "..."?

Janpot: https://github.com/mui/mui-toolpad/pull/519#event-6727268978

PR #501, opened 2022-05-30
Improve validity check for Google Sheets connection

### LINKED ISSUE #428: [sheets] Validate connections using API
- Instead of using the presence of an unexpired `access_token` on the client, use the token info OAuth APIs to check if the connection has not been manually revoked by the user

bharatkashyap:
- Fixes #428  


packages/toolpad-app/src/components/AppEditor/ConnectionEditor/index.tsx
```diff
@@ -64,20 +65,26 @@ function ConnectionEditorContent<P>({
...
         {dataSource ? (
-          <ConnectionParamsEditor
-            dataSource={dataSource}
-            value={connectionNode.attributes.params.value}
-            onChange={handleConnectionChange}
-            handlerBasePath={`/api/dataSources/${dataSourceType}`}
-            appId={appId}
-            connectionId={connectionNode.id}
-          />
+          <QueryEditorContextProvider value={connectionEditorContext}>
```
Janpot, line 78: Wouldn't it be confusing to use `QueryEditorContext` in the `ConnectionEditor`? Do we need a `ConnectionEditorContext`?

bharatkashyap: I thought of adding that but it felt very redundant because we need the exact same attributes; in that case we would also need to create a separate version of the `usePrivateQuery` hook

Janpot: do we need to rename it to `DataSourceConnectionContext`?

bharatkashyap: The `DataSource` part seems redundant to me - `ConnectionContext` feels succinct enough. What do you think?

bharatkashyap:
Quick tweak to display connected user details:
Screenshot 2022-06-15 at 6 47 47 PM

Janpot (approved):

Issue #516, opened 2022-06-01
Convert UI to MUI dark theme


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

This in an exploratory discussion about Toolpad supporting a dark mode. There can be 2 scenarios:

1. Dark mode by default.

- Would all devs prefer it? Though many will as it reduces eye strain in low light conditions. But for a dev working during the daytime it might not be preferred.
- Landing page, docs will have to be dark too

2. Option to switch between dark and light or maybe match system settings. It makes more sense but comes with maintenance overhead.

- Docs will also have to support light and dark mode. It would mess with the screenshots/videos?

From a [survey](https://medium.com/dev-channel/let-there-be-darkness-maybe-9facd9c3023d) conducted for Chrome, of all (201) the users who have tried Dark mode. 75% prefer to keep it always on.

### Examples 🌈

https://webflow.com/ (White canvas, gray everywhere else)
https://www.bildr.com/ (White canvas, black everywhere else)
Tooljet: Supports switching

### Motivation 🔦

https://github.community/t/how-popular-is-dark-mode/226624/2
https://community.retool.com/t/developer-dark-mode/6935

https://blog.superhuman.com/why-do-people-use-dark-mode/

oliviertassinari:
Some data: https://twitter.com/MUI_hq/status/1528403016587657220

oliviertassinari:
@mbrookes so you are proposing what https://framer.com / https://figma.com is doing. What they have implemented feels great.

I agree that we should eventually get there. VS Code supports light mode too. For the first 4 years I was programing, I was using a light theme in my IDE. 

Regarding forcing dark mode, I'm personally not in the targeted audience of who this would benefit to make it flexible. I could code on VS Code in the beach this summer almost as if I was indoor 😁.

Janpot:
> I don't recall implementing it for the docs being particularly challenging

I agree with that, especially since atm, it seems we're starting out by just copying the theme from the docs and leaving all the light mode parts in.

bharatkashyap:
Agreed - I've added a theme mode toggle here https://github.com/mui/mui-toolpad/pull/988

PR #517, opened 2022-06-01
Dark mode exploration

Janpot:
Quick look into what would obviously break if we switch on dark mode.
So far just needs a tweak to the component catalog, and the runtime `CssBaseline` must be nested under the runtime theme

To Do:

- [ ] code editor needs themeing
- [ ] mode toggle

related https://github.com/mui/mui-toolpad/issues/516


Janpot:
Might as well merge this already

Janpot:
> Can't open the Render URL. It says deploy failed: Out of memory.

I'll look into that. 

To be clear, this doesn't add a dark mode to Toolpad editor. Just tries to get a sense of the current compatibility with a dark theme.

It does add a dark mode option to toolpad apps theme.

Janpot:
Looks like we're hitting https://github.com/prisma/prisma/issues/11413 
See the issue that has been linked to: https://github.com/redwoodjs/redwood/issues/4271
Describes the exact same problem happening on the exact same command that we run `prisma migrate deploy`

PR #540, opened 2022-06-09
Some tweaks to connection editor

Janpot:
* Move save button to the bottom right
* Make save button contained
* wrap editor in a responsive container


joserodolfofreitas:
A bit late on the comment, but that's on target

PR #543, opened 2022-06-09
Fix Overlay sizing

Janpot:
We need to make the overlay a child of the app root, so that it can size relative to it. Took the opportunity to remove a few hacks from the `RenderPanel` related to its communication with the canvas

https://user-images.githubusercontent.com/2109932/172901286-443a1bcb-283b-4857-a8e5-38b04988e35f.mov


apedroferreira (approved):

apedroferreira (approved):
great job, thanks for this, we definitely should make the `RenderPanel` file much lighter and this helps a lot :D
doesn't seem like anything should conflict with my work!

PR #546, opened 2022-06-10
Add autocomplete to post-request transform editor

### LINKED ISSUE #503: Query post-transformation, autocompletion support
Currently it's unintuitive to build a post-transformation. The code editor doesn't help the user at all with autocomplete. One possible solution could be to, instead of expecting a function expression, we expect a function body, we can the supply `data` in the `globalScope` property of the editor to get autocomplete for free.

```tsx
return data;
```
instead of
```tsx
(data) => {
  return data;
}
```

@bharatkashyap what do you think?

### LINKED ISSUE #507: Transform function undefined on creation
### Duplicates

### Latest version

### Current behavior 😯

Screenshot 2022-05-31 at 1 28 13 PM

### Expected behavior 🤔

It should return the existing data

### Steps to reproduce 🕹

Steps:

1. Add a Google Sheets query
2. Enable the "Transform API response"
3. Re-run the Preview

### Context 🔦

_No response_

### Your environment 🌎

_No response_

bharatkashyap:
- Fixes #503
- Fixes #507 


packages/toolpad-app/src/components/AppEditor/PageEditor/QueryEditor.tsx
```diff
@@ -283,11 +284,36 @@ function QueryNodeEditorDialog<Q, P>({
...
+  const untransformedQueryPreview = client.useQuery(
+    'execQuery',
+    input
+      ? [
+          appId,
+          {
+            ...input,
+            attributes: {
+              ...input.attributes,
+              transform: { value: '', type: 'const' },
```
Janpot, line 296: Shouldn't just overwriting `transformEnabled` be enough?

```suggestion
```

Janpot: Maybe we should abstract this in a `withTransformDisabled(input)` function?

packages/toolpad-app/src/components/AppEditor/PageEditor/QueryEditor.tsx
```diff
@@ -283,11 +284,36 @@ function QueryNodeEditorDialog<Q, P>({
...
+    { retry: false },
+  );
+
   const handleUpdatePreview = React.useCallback(() => {
     setPreviewQuery(input);
     setPreviewParams(paramsObject);
   }, [input, paramsObject]);
 
+  const handleTransformEditorFocus = React.useCallback(() => {
+    if (input) {
```
Janpot, line 312: Interesting, does it try to refetch even though the query is disabled (because we provided `null` as the key)?

packages/toolpad-app/src/components/AppEditor/PageEditor/QueryEditor.tsx
```diff
@@ -384,13 +410,26 @@ function QueryNodeEditorDialog<Q, P>({
...
                   }
                 />
-
-                <JsExpressionEditor
-                  globalScope={{}}
-                  value={input.attributes.transform?.value ?? '(data) => {\n  return data;\n}'}
-                  onChange={handleTransformFnChange}
-                  disabled={!input.attributes.transformEnabled?.value}
-                />
+                <Stack direction={'row'} spacing={2}>
```
Janpot, line 413: Wondering whether we shouldn't make this all just part of the `JsExpressionEditor`. Since it's very similar to what's being used in the binding editor?

packages/toolpad-app/src/components/AppEditor/PageEditor/QueryEditor.tsx
bharatkashyap: Yes, that's exactly what is happening

packages/toolpad-app/src/components/AppEditor/PageEditor/QueryEditor.tsx
bharatkashyap: No, since then any change to the `transform` causes a refetch to get triggered

packages/toolpad-app/src/components/AppEditor/PageEditor/QueryEditor.tsx
bharatkashyap: I think it's okay to keep them separate because that keeps it flexible for the user to configure them as needed in the area of the codebase where they are getting used

Janpot:
a few suggestions after trying it out:

* Let's position the refresh button absolutely in the top-right corner as an icon button?
* Let's hide/disable the refresh button when the data is up to date?
* Let's automatically refresh the data when the code editor is focused?
* There is no overflow on the object explorer, it just grows indefinitely when expanding large objects.

bharatkashyap:
> * Let's hide/disable the refresh button when the data is up to date?

Is a deep comparison of the returned data from the two queries (untransformed vs transformed) the correct way to do this?

Janpot (approved):
Let's merge 👍

Issue #76, opened 2022-01-25
action: navigate


Janpot:
button click:
- navigate to other page
- bind page state to navigation query parameters

newguy-123:
https://docs.appsmith.com/v/v1.2/core-concepts/connecting-ui-and-logic/sharing-data-across-pages

PR #565, opened 2022-06-16
Introduce running actions on events

Janpot:
* Support running scripts on events
* Support navigations on events

closes https://github.com/mui/mui-toolpad/issues/76


oliviertassinari:
I have disabled these noisy notifications:

Screenshot 2022-06-17 at 12 19 03

Issue #496, opened 2022-05-29
Make properties panel resizable


Janpot:
From https://github.com/mui/mui-toolpad/issues/309

> UX: Sometimes, I wish I could resize the right-side panel.

PR #570, opened 2022-06-17
Make component panel resizable

Janpot:
Pull the SplitPane implementation out of https://github.com/mui/mui-toolpad/pull/289 and instead use it to make the component panel resizable.

Closes https://github.com/mui/mui-toolpad/issues/496


oliviertassinari:
At some point in the future, we could persist the panel sizes between sessions 😁. I have taken a note that Toolpad could test MUI X window splitter in the future in the benchmark list of https://github.com/mui/mui-x/issues/512.

Janpot:
> * At some point in the future, we could persist the panel sizes between sessions 😁.

We already have the primitives available so I just created https://github.com/mui/mui-toolpad/pull/604

PR #605, opened 2022-06-24
Remove some obsolete components

Janpot:
* all pages are now wrapped in a `Container` by default
* `CustomLayout` was just around to test slot properties on custom components


Janpot:
It'd probably makes sense to just create a code component to test that.

apedroferreira:
> It'd probably makes sense to just create a code component to test that.

yeah that works too

PR #542, opened 2022-06-09
[docs] Bootstrap docs/landing page site

bharatkashyap:
Deploy preview:

- https://deploy-preview-542--mui-toolpad-docs.netlify.app/toolpad/
- https://deploy-preview-542--mui-toolpad-docs.netlify.app/toolpad/getting-started/setup/

Partially fixes #477 
Fixes #488

PR #606, opened 2022-06-24
[docs] Fix Netlify deploy with a dummy index.html

oliviertassinari:
https://deploy-preview-606--mui-toolpad-docs.netlify.app/.

Until #542 is merged, I think that it would be better to keep the CI to default to green, e.g. #605. It will also help test that https://mui.com/toolpad/ proxy works as expected.


oliviertassinari:
A side observation: for community contributions, it won't fly very well:

Screenshot 2022-06-25 at 00 58 11

Janpot:
👍 agreed, IAC all things. Thanks!

Issue #615, opened 2022-06-29
Make monaco work offline


Janpot:
`@monaco-editor/react` by default doesn't work offline. Probably want a more direct integration in the flavour of https://www.swyx.io/how-to-add-monaco-editor-to-a-next-js-app-ha3

PR #619, opened 2022-07-01
Make monaco editor work offline

Janpot:
Get rid of the `@monaco-editor/react` wrapper and write a custom React/Monaco integration.

Fixes https://github.com/mui/mui-toolpad/issues/615


mbrookes:
> write a custom React/Monaco integration

🤯

Janpot:
Having more issues in CI than expected, converting to draft again

challenges faced in this PR:

- monaco requires being able to load workers from url. It also imports CSS in its source files.
- There is no official microsoft react wrapper for monaco, despite its name `@monaco-editor/react` is a 3rd party wrapper.
- I'm using `monaco-editor-webpack-plugin` to build the workers.
- I inlined the css parts from `next-transpile-modules`. Using `next-transpile-modules` directly has too many issues. It makes it so the javascript is compiled as well and the the SWC compiler in next.js seems to compile them incorrectly. babel is too slow. We only need the css extracted.
  - yet it's still causing memory issues in CI. upgrading to a `medium+` resource class helped. Trying to optimize the setup.
- The react wrapper interface itself is similar to the `@monaco-editor/react`. We're going to use suspense to load the component instead.
- Works well locally, in dev mode. But the docker smoke test is still failing and I'm seeing weird artificts in render.com.

PR #644, opened 2022-07-08
Extract canvas logic in top level layout

Janpot:
The components folder is starting to become unwieldy, so far we've treated the "application editor" as just another component. It's not meant for reuse. I'm proposing to extract this into its own top level folder, which can be more seen as entry-points into the application
Aiming towards a better top-level project layout. I propose:

* `./server`: all server only code (OK)
* `./runtime`: toolpad application runtime (OK)
* `./canvas`: application runtime with overrides, used inside the editor (this PR)
* `./toolpad`: Editor application (currently interleaved with ./components)
* `./components`: shared components (currently contains Toolpad editor as well)
* `./utils`: shared generic utility functions, do not put toolpad specific functionality in here (OK)
* `./toolpadDatasources`: available datasources, will rename to `./toolpad`
* `./`: toolpad specific libraries, we will let this folder grow until it becomes unwieldy and needs organization

### DIFF
--- a/packages/toolpad-app/pages/app-canvas/[[...path]].tsx
+++ b/packages/toolpad-app/pages/app-canvas/[[...path]].tsx
@@ -1,6 +1,6 @@
 import type { NextPage } from 'next';
 import * as React from 'react';
-import AppCanvas, { AppCanvasProps } from '../../src/runtime/AppCanvas';
+import AppCanvas, { AppCanvasProps } from '../../src/canvas';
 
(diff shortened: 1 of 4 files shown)


Janpot:
> sounds good, maybe the `toolpad` folder could be `editor`

My only objection would be that it is more than just the application editor, it also houses the application overview, application management, (future) user management, the release process

> and toolpadDatasources could be just datasources?

👍

Issue #359, opened 2022-05-02
Support resizing of components


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Improve the visual editor by adding the feature to resize components through a resize handle (inputs horizontal only, data grids vertical). It should be extensible enough so that we can share the logic with other components. Bonus points if there’s a way for custom components to integrate resize handles as well.

### Motivation 🔦

Retool, Appsmith both have this option

apedroferreira:
Closing, as basic resizing functionality has been merged.
However, even though the current resizing features are sufficient I don't think they're complete enough. All components with variable height or width should allow those properties to be resizable with handles, as well as their spacing.
I've created an issue for a refactor (https://github.com/mui/mui-toolpad/issues/712) that should make it easier to improve the resizing behavior. Will create one or 2 issues for improving the resizing of components and their spacing too.

PR #645, opened 2022-07-08
Editor - Resizing elements inside page rows

apedroferreira:
Related issues: https://github.com/mui/mui-toolpad/issues/101 and https://github.com/mui/mui-toolpad/issues/359

Add ability to resize node elements inside page rows, in the editor:

https://user-images.githubusercontent.com/10789765/179825987-01df4485-717d-4085-9742-ac2580d8a592.mov

- allow dragging sides of elements to resize them
- show grid overlay while dragging - snap to grid columns to make it easier for users to build a good-looking layout
- show resize preview while dragging - changing the styles in the DOM directly, using a CSS transform scale
- control grid layout inside rows through a prop set directly in the runtime - use the `layout` namespace in page row children to control the value of that prop
- includes some necessary small adjustments to some editor components too

closes https://github.com/mui/mui-toolpad/issues/681


apedroferreira:
@Janpot Resizing to the outside kind of works (need to fix resizing to the inside), but performance is pretty bad at the moment... (will refactor things a bit later too, I'm mostly trying to make it work now)
Are we doing any kind of debouncing/delay when updating the DOM os something? Just need to know, otherwise I can try to find out how to make the performance smoother.

packages/toolpad-app/src/components/AppEditor/PageEditor/RenderPanel.tsx
```diff
@@ -757,11 +813,119 @@ export default function RenderPanel({ className }: RenderPanelProps) {
...
   }, [nodesInfo, pageNodes]);
 
+  const updatePageRowLayout = React.useCallback(
+    (pageRowNode: appDom.ElementNode) => {
+      const children = appDom.getChildNodes(dom, pageRowNode).children;
+      const layoutColumnSizes = children
+        .map((child) => child.layout?.columnSize || appDom.createConst(0))
+        .map((constAttribute) => constAttribute.value);
+
+      const totalLayoutColumnsSize = layoutColumnSizes.reduce((acc, size) => acc + size, 0);
```
apedroferreira, line 823: will extract this column size normalization logic later

apedroferreira:
i'll also probably remove the edge drag areas outside of the `NodeHud`, doesn't make sense for them to be there.

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/RenderPanel.tsx
```diff
@@ -1058,6 +1142,16 @@ export default function RenderPanel({ className }: RenderPanelProps) {
                 lastContainerChild.parentProp,
                 parent.parentIndex,
               );
+
+              if (isPageColumn(parent)) {
```
apedroferreira, line 1146: i'll probably improve this function later to not have to repeat as much logic- should be done separately though as it relates to the whole editor and not just resizing

Janpot (approved):
👍  lgtm

PR #657, opened 2022-07-14
Move the toolpad app out of the components folder

Janpot:
Follow up on https://github.com/mui/mui-toolpad/pull/644 
Moves toolpad application components it into a top-level `./toolpad` folder so that we reserve the `./components` folder for reusable react components.

For now kept the name `./toolpad` since I feel like `./editor` doesn't quite capture the breadth of what's in this folder (also has releases/app overview/...). But open for discussion


bharatkashyap (commented):
Screenshot 2022-07-14 at 8 41 45 PM

I would think that a name that is more semantically precise - like the other folders - might be better; for instance, I wouldn't be able to understand why the files in the `toolpad` folder are grouped together whereas it is evident for all other folders.

Perhaps `home` or `appShell`?

apedroferreira:
> Screenshot 2022-07-14 at 8 41 45 PM
> 
> I would think that a name that is more semantically precise - like the other folders - might be better; for instance, I wouldn't be able to understand why the files in the `toolpad` folder are grouped together whereas it is evident for all other folders.
> 
> Perhaps `home` or `appShell`?

or maybe just `app`? not sure if that's an improvement, just a possible suggestion.
anyway i think these folder changes are already an improvement.

apedroferreira:
> The name "app" is too ambiguous, it could refer to the apps you create inside of toolpad. I'm thinking "designer", "workspace", "studio",...

I usually put the non-reusable components inside folders for each page of the website, in a `pages` folder - not sure how much that could apply to this app though since pages can have so many specific sections.
But anyway any of those names seem fine to me, or we can keep `toolpad` too for now :)

PR #658, opened 2022-07-14
Allow duplicating apps

### LINKED ISSUE #843: Duplicate apps
- Subset of #437

bharatkashyap:
- Add option to "Duplicate" app in the Home screen app card
- Fixes #843  
- Useful for, among other things, allowing users to duplicate template apps present by default in the instance

- [x] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/material-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request)

https://user-images.githubusercontent.com/19550456/190364240-0ab78713-430e-447a-9d1e-c1cd942cb8b3.mov


packages/toolpad-app/src/server/data.ts
```diff
@@ -233,6 +233,21 @@ export async function updateApp(appId: string, name: string): Promise<App> {
   });
 }
 
+export async function duplicateApp(appId: string): Promise<App> {
+  const app = await prisma.app.findUnique({ where: { id: appId } });
+  const dom = await loadPreviewDom(appId);
+  const duplicateDom = appDom.duplicateDom(dom);
+  if (!app) {
```
apedroferreira, line 240: not a big deal, but this check could be right after `findUnique` - just makes it a bit easier to follow and avoids doing unnecessary work.

packages/toolpad-app/src/components/Home.tsx
```diff
@@ -146,6 +147,11 @@ function AppCard({ app, activeDeployment, onDelete }: AppCardProps) {
   const [editingTitle, setEditingTitle] = React.useState<boolean>(false);
   const [appTitle, setAppTitle] = React.useState<string | undefined>(app?.name);
   const appTitleInput = React.useRef<HTMLInputElement | null>(null);
+  const duplicateAppMutation = client.useMutation('duplicateApp', {
```
apedroferreira, line 150: are we handling errors if the mutation fails?

apedroferreira (approved):
nice job, just left a couple of comments but no big deal, seemed good to me but Jan might know a few things more

also not sure if we should open the duplicated project right after we create it, maybe we should just show it in the project list

Janpot (changes requested):
Some properties refer to other nodes. e.g. the page id in a navigate action, or the connection id in a query. Those links will be broken when you give nodes new ids, as they will refer to nodes in a different dom.

We need a structural way of dealing with these foreign key style references

bharatkashyap:
> also not sure if we should open the duplicated project right after we create it, maybe we should just show it in the project list

I agree with showing the list of apps; saving one click might not be worth sacrificing the instant feedback of the operation having completed successfully

Janpot:
@bharatkashyap I also needed a dom duplication method for https://github.com/mui/mui-toolpad/pull/786 so [quickly wrote one](https://github.com/mui/mui-toolpad/pull/786/files#diff-299a4d9f06f69284f5c11960ef97e6a3723fbcefe950857a324dffd50c58e876R880). We can consolidate when you pick this PR back up.

Janpot:
Just realized that after https://github.com/mui/mui-toolpad/pull/778, rewriting the indices shouldn't be necessary anymore

packages/toolpad-app/src/server/data.ts
```diff
@@ -455,3 +455,13 @@ export async function loadDom(appId: string, version: VersionOrPreview = 'previe
 export async function loadRenderTree(appId: string, version: VersionOrPreview = 'preview') {
   return appDom.createRenderTree(await loadDom(appId, version));
 }
+
+export async function duplicateApp(id: string): Promise<AppMeta> {
+  const app = await prismaClient.app.findUnique({ where: { id } });
```
Janpot, line 460: Isn't this check already happening in `loadPreviewDom`?

bharatkashyap: I feel like that adds an unnecessary step and is a UX downgrade. The standard expectation with a duplicate/"make a copy" interaction is not a dialog box from what I can see across most benchmarks. 

I propose we could add an assertion to satisfy the type system: `app!.name ...`

bharatkashyap: I think being taken to the duplicated app directly from the list page might be an overreach:

On the list page, on duplication, I would expect that the overall view remains consistent and another app appears in the list with an indication to it being the duplicated app.  In this case it is the "(copy)" added to the name. We could potentially even add a selected outline/focus indicator to it.

Bechmarking with the Google Drive experience:

https://user-images.githubusercontent.com/19550456/189125012-62122cc6-7584-4ada-921b-ca2251783417.mov

"Dialog box -> Redirected to the new app" is an okay expectation for me when I am inside the "editor" itself; in the future, if we allow duplication while someone is inside the editor, I think we should follow the dialog box UX (We could potentially include other options along with the title in that case.)

Again, benchmarking with Google Drive:

https://user-images.githubusercontent.com/19550456/189125657-78e91f6c-f10e-4621-b0ca-98c3fc21abd4.mov

I don't think the rename action is that immediately important; and the pain of having to go back to the list view to rename should be mitigated by adding this inline renaming functionality in the future:

https://user-images.githubusercontent.com/19550456/189129657-10ec5bbe-7679-4b6b-8eb8-acac00052ad7.mov

Janpot: In any case, both implementations can be built by passing the name to `duplicateApp`

test/integration/appDuplicate.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+  await page.locator('button:has-text("create new")').click();
+
+  await page.fill('[role="dialog"] label:has-text("name")', name);
+
+  await page.click('[role="dialog"] button:has-text("create")');
+
+  await page.waitForNavigation({ url: /\/_toolpad\/app\/[^/]+\/editor\/pages\/[^/]+/ });
+}
+
+test('duplicate app from home flow', async ({ page }) => {
```
apedroferreira, line 15: we've created these page object models (https://playwright.dev/docs/test-pom) to try to encapsulate some common behaviors in different pages.
you can probably use the `ToolpadHome` model for some things in these tests?

bharatkashyap: This looks very useful! For the app itself, do you see us making an AppCard/AppRow model as well to be able to capture the app-specific menu and buttons?

Janpot: > do you see us making an AppCard/AppRow model as well to be able to capture the app-specific menu and buttons?

Probably not. Why wouldn't `openApp(name)`, `editApp(name)`, `renameApp(name, newName)`,... work?

bharatkashyap: Got it! I struggled a bit with making the test to check actual duplication work but can't seem to make it pass. I'll move it to a [separate PR](https://github.com/mui/mui-toolpad/tree/duplicate-app-test) and keep the minimal test which only checks whether the duplicated apps have been created.

test/integration/appDuplicate.spec.ts
```diff
@@ -0,0 +1,19 @@
+import * as path from 'path';
+import { test, expect } from '../playwright/test';
+import { ToolpadHome } from '../models/ToolpadHome';
+import { readJsonFile } from '../utils/fs';
+import * as locators from '../utils/locators';
+
+test('duplicate app from home flow', async ({ page }) => {
+  const dom = await readJsonFile(path.resolve(__dirname, './duplicateNavigationDom.json'));
```
apedroferreira, line 8: if we're not checking anything about this DOM maybe we don't need it?
(but if you're gonna check the app itself in the runtime could be useful of course)

apedroferreira: on that note, we should probably make a simple check that the duplicated dom is the same?

bharatkashyap: That's what the effort here https://github.com/mui/mui-toolpad/pull/1007 is about, removed that from this for now

apedroferreira: just saw the test changes, looks good!

test/models/ToolpadHome.ts
```diff
@@ -89,4 +89,23 @@ export class ToolpadHome {
 
     return { id: idMatch[1], name };
   }
+
+  async duplicateApplication(appName: string) {
+    await this.page.click(
+      `[role="row"] >> has="input[value='${appName}']" >> [aria-label="Application menu"]`,
```
Janpot, line 95: I'd suggest to create a `getAppRow` function that returns a locator for the row, from which you can create a locator for the menu button.

test/models/ToolpadHome.ts
```diff
@@ -89,4 +89,23 @@ export class ToolpadHome {
...
+    );
+    await this.page.click('[role="menuitem"]:has-text("Duplicate"):visible');
+
+    // Navigate to the new app
+    await this.page.click(`[role="row"] >> a:has-text("Edit")`);
```
Janpot, line 100: reuse that `getAppRow` locator function here

test/models/ToolpadHome.ts
apedroferreira: we already have this function, could probably use it

Janpot (approved):

Issue #666, opened 2022-07-19
Applications cards overflow badly


Janpot:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

<img width="499" alt="Screenshot 2022-07-19 at 17 25 07" src="https://user-images.githubusercontent.com/2109932/179788875-811f32b9-bcd0-4a53-978f-c0f260075ec5.png">

### Expected behavior 🤔

Proper overflow behavior. Ellipses maybe? Or should this UI be a table rather than cards?

### Steps to reproduce 🕹

Steps:

1.
2.
3.
4.

apedroferreira:
i think ellipses would solve it, but the title text should be small enough so that titles that aren't too long are always fully readable

PR #672, opened 2022-07-20
Fix app name overflow

bytasv:

1. Show ellipses for overflown text (I'm just not sure if it's fine that we now also truncate multi word names? 🤔 )
2. In addition I've added tooltip with a longer enter delay so that users can inspect actual value even if it's truncated.
3. Reduced size of name so that we have more space before truncating

image

image

4. I also noticed that it gets broken when editing app, so I fixed it there as well:

image

@Janpot regarding table vs card view - lemme know if we should proceed that way, I think this could be a quick tweak - we could just add toolbar with icons and depending on local view state render either card or table

Off topic: I remember in previous team we implemented `TypographyOverflow` component - https://picasso.toptal.net/?path=/story/components-typographyoverflow--typographyoverflow - maybe it would be worth considering pitching similar functionality for the core component? 🤔 

Closes https://github.com/mui/mui-toolpad/issues/666


Janpot:
Original PR: https://github.com/mui/mui-toolpad/pull/670

Issue #613, opened 2022-06-27
In custom component interface, move 'Update' button to the bottom right.


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

<img width="1089" alt="Screenshot 2022-06-27 at 11 49 42 AM" src="https://user-images.githubusercontent.com/92228082/175925241-39d439d5-5e66-4b84-a48b-777ca6068f7f.png">

In the current placement, it gets easily missed.

prakhargupta1:
Closed via #671

Janpot:
That PR was closed without merging. Merging the [new PR](https://github.com/mui/mui-toolpad/pull/673) will automatically close this ticket

PR #673, opened 2022-07-20
Tweak update button position and visibility in component editor

bytasv:

image

- Moved toolbar below the editor
- Also changed variant for button so it's more visible. IMO since its the main CTA it could use that extra visibility
- Fix split-pane height issue (seems that because of its using absolute positioning (weird) it breaks when immediate parent doesn't have relative or any other position set). FYI I've addded fix to `packages/toolpad-app/src/components/SplitPane.tsx` as to me it seems more intuitive to work like that everywhere. I've also checked other places where `SplitPane` is used and it seems the rest is still working as expected

Closes https://github.com/mui/mui-toolpad/issues/613


bytasv:
@Janpot when you get a chance please re-review, I've changed approach and addressed overscroll issue

packages/toolpad-app/src/components/SplitPane.tsx
```diff
@@ -38,6 +38,9 @@ const WrappedSplitPane = React.forwardRef<
       // Some sensible defaults
       minSize={20}
       maxSize={-20}
+      style={{
+        position: 'relative',
```
Janpot, line 42: Yep, makes sense. Perhaps it could make a bit more sense on the `paneStyle` instead, since that's the exact rectangle we want that editor to size to?

Janpot:
Screenshot 2022-07-21 at 20 14 00

👌 horizontal scrollbars

packages/toolpad-app/src/components/SplitPane.tsx
bytasv: That was my initial though too, I tried it but it seems that `paneStyle` is responsible for other components so I had to use `style`

Janpot: Ok, so it's the whole splitpane sizing badly. I'm a bit concerned that this will mess up existing splitpane usage. Perhaps it's safer to [keep it absolute](https://github.com/mui/mui-toolpad/pull/678) and have it size against its parent?

Janpot (approved):

Issue #662, opened 2022-07-19
Query editor: make panels resizable


Janpot:
Rest/google sheets queries have a split panel between editor and preview, we can make this resizable, just like the function query editor.

PR #676, opened 2022-07-21
Add split panes to connections dialog

bytasv:

image

I've added `SplitPane` however I'm not a big fan of implementation right now. I had to add `ConnectionWrapper` which is a pretty dumb component which allows either to use split pane or not. I think to do it more cleaning we would need to refactor layout and maybe unify each connection type 🤔


Janpot (approved):

bytasv:
Closes https://github.com/mui/mui-toolpad/issues/662

PR #677, opened 2022-07-21
Built-in Map Component

bharatkashyap:

Quick experiment with adding a Map Component as a built-in:

https://user-images.githubusercontent.com/19550456/180256890-ac487c79-a23e-4db2-8ab0-e99ec92bd088.mov

Compared with creating a Map component as a[ Custom Code component](https://master--toolpad.mui.com/_toolpad/app/cl5s7o6et3701apke1fb0ismp/editor/codeComponents/cl5s7oigy00003b6gjil2l5yx) which can offer the same functionality in a pro-code way:

Screenshot 2022-07-21 at 9 17 08 PM


packages/toolpad-components/src/Map.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+const MapComponent = ({ latitude, longitude, zoom, sx, style }: MapComponentProps) => {
+  return (
+    <Map
+      key={`${latitude}-${longitude}-${zoom}`}
+      initialViewState={{ latitude, longitude, zoom }}
+      style={sx as React.CSSProperties}
+      mapStyle={style}
+      mapboxAccessToken={
+        'pk.eyJ1IjoibWFib25nIiwiYSI6ImNrMm9qN2tiYTEwc3ozZG41emx6bW9uZnQifQ.PhojWq3UwsAlPB7LBvJiTw'
```
bytasv, line 22: would it make sense to place it in some config file? 🤔

bharatkashyap: Yes, for now I didn't add it as an environment variable for the ease of being able to view it in the render.com preview; I wasn't sure how we would add environment variables to code in `toolpad-components` folder on render.com

Janpot: The easiest way for now would probably be by using a build time variable.
- add a key [here](https://github.com/mui/mui-toolpad/blob/299a8ccf1af5e09f44c715f41579057470b7d1a5/packages/toolpad-app/src/config.ts#L25)
- pass the var [here](https://github.com/mui/mui-toolpad/blob/299a8ccf1af5e09f44c715f41579057470b7d1a5/packages/toolpad-app/next.config.mjs#L25)
- Use it in your code as `process.env.MAPBOX_ACCESS_TOKEN` (exactly like that, so that webpack can find/replace it)

[Here's](https://github.com/mui/mui-toolpad/blob/299a8ccf1af5e09f44c715f41579057470b7d1a5/packages/toolpad-app/src/config.ts#L4) some more info about the different ways configuration can be passed to Toolpad. (For this case, only 1. will work. 3. doesn't apply as these are browserside components. If we want 2., then we're going to need to add a context or something for it in `toolpad-core`)

packages/toolpad-app/next.config.mjs
```diff
@@ -26,6 +26,7 @@ function parseBuidEnvVars(env) {
     TOOLPAD_TARGET: target,
     TOOLPAD_DEMO: env.TOOLPAD_DEMO || '',
     TOOLPAD_VERSION: pkgJson.version,
+    MAPBOX_ACCESS_TOKEN: env.MAPBOX_ACCESS_TOKEN || '',
```
Janpot, line 29: Should we maintain the `TOOLPAD_` prefix for all our variables? `TOOLPAD_MAPBOX_ACCESS_TOKEN`

PR #680, opened 2022-07-22
Fix hidden tooltips issue in code editor

bytasv:

image

I fixed the issue of tooltip being hidden. Tested other usages of split pane all seems still fine, however I'm not really sure what was meant by split pane overflow being hidden 🤔 I'm trying to think of a reasons but can't come up with anything, for me it seems logical that the consumer would be controlling overflow 🤔

### DIFF
--- a/packages/toolpad-app/src/components/SplitPane.tsx
+++ b/packages/toolpad-app/src/components/SplitPane.tsx
@@ -43,7 +43,11 @@ const WrappedSplitPane = React.forwardRef<
         // Prevent the content from stretching the Panel out
         minWidth: 0,
         minHeight: 0,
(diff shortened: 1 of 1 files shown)


Janpot:
This doesn't fix the problem at the root, we'll still see the issue in dialogs:

Screenshot 2022-07-22 at 12 40 08

There may be legitimate reasons to have the editor nested in a container with hidden overflow. Isn't there a way we can make `monaco-editor` attach these tooltips to the body or something?

bytasv:
Oh, good catch, I will take a look 🙇

bytasv:
Closing as per https://github.com/mui/mui-toolpad/pull/682

Issue #667, opened 2022-07-19
Code Editor tooltips get clipped by the Code Editor bounds


Janpot:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

This happens in several of the code editors when hovering over a variable close to the border:

e.g. in the custom components editor:

<img width="586" alt="Screenshot 2022-07-19 at 17 28 15" src="https://user-images.githubusercontent.com/2109932/179789649-d4c868d2-8d45-429e-866a-9dccdbc2eb8b.png">

### Expected behavior 🤔

The whole tooltip gets shown

### Steps to reproduce 🕹

Steps:

1.
2.
3.
4.

### Context 🔦

Related to https://github.com/microsoft/monaco-editor/issues/181 ?

PR #682, opened 2022-07-22
Fix overflow widgets for monaco

Janpot:
Fixes https://github.com/mui/mui-toolpad/issues/667

Screenshot 2022-07-22 at 15 46 48

Issues compared to the previous situation:
- [ ] Hovering over the tooltip doesn't keep it open https://github.com/microsoft/monaco-editor/issues/2156
  Fix should come in 0.34 according to https://github.com/microsoft/vscode/pull/142160#issuecomment-1055738860
- [ ] Doesn't really work when hovering on `createComponent` on the third line

Both of these issues are resolved on `monaco-editor@0.34.0-dev.20220722`

Closes https://github.com/mui/mui-toolpad/pull/680


packages/toolpad-app/src/components/MonacoEditor.tsx
```diff
@@ -183,6 +187,15 @@ export default React.forwardRef<MonacoEditorHandle, MonacoEditorProps>(function
       const pathUri = monaco.Uri.parse(`./scripts/${cuid()}${getExtension(language)}`);
       const model = monaco.editor.createModel(value || '', language, pathUri);
 
+      if (!overflowWidgetsDomNodeRef.current) {
+        const overflowWidgetsDomNode = document.createElement('div');
```
bytasv, line 191: NIT: just wondering - would it be possible to use `Portal` and just pass a ref? 🤔 
maybe something like this:
```jsx
<Portal>
  <div
    className='monaco-editor'
    ref={overflowWidgetsDomNodeRef}
    style={{ zIndex: overflowWidgetsDomNodeZindex }}
  />
</Portal>
```

Janpot: You mean using [`react-portal`](https://www.npmjs.com/package/react-portal)? I think that should be possible yes

bytasv: Hmm, I was thinking about one that MUI provides: https://mui.com/material-ui/react-portal/ but not sure if that's possible 🤔 anyways, that would apply if it doesn't complicate the matters, but seems that with MUI Portal we would need a way to pass container element all the way to body, unless we could do:
```jsx
<Portal container={document.body} />
```

Janpot: Just tried it out and it looks like the ref gets assigned later than the initial `useEffect` call, which means it's not available yet when the editor is created.

Looking at [the implementation](https://github.com/mui/material-ui/blob/1490632fc1d91a7d417e33f333db080293590301/packages/mui-base/src/Portal/Portal.js#L20) it makes sense, an effect has to run once before the portal gets rendered. I think this is done to avoid mismatches during hydration.

bytasv (approved):

PR #690, opened 2022-07-25
Add List view for apps as default

### LINKED ISSUE #689: Renaming apps breaks styling
### Duplicates

### Latest version

### Current behavior 😯

https://user-images.githubusercontent.com/19550456/180843860-c5768681-2d25-4154-b2b4-38eb80854d57.mov

### Expected behavior 🤔

The styling of the text field and the label of the app should be consistent when renaming.

### Steps to reproduce 🕹

Steps:

1. Navigate to the home page
2. On any app card, click on the extended options menu
3. Select "Rename"

### Context 🔦

_No response_

### Your environment 🌎

_No response_

bharatkashyap:

- Added a list view for apps with a toggle button to switch to grid,
  -  Possible extension to allow collapsible rows for folders

- Removed the "Apps" heading in favour of a Tabs list in the AppBar,
   - Possible extension to allow a global "Data" or "Connections" tab

- Rewrote the `EditableText` component to use only `TextFields` and leak minimal implementation details

- Added a `AppNameEditable` component (along with some other to allow reuse between `Card` and `Row`) to store app name specific renaming logic

- Remove tooltips from app names

- Fixed a typo in getting readable duration

- Fixes #689 

https://user-images.githubusercontent.com/19550456/182039663-cd939cbf-7a91-4cef-b12b-8eea1277400b.mov


packages/toolpad-app/src/components/EditableText.tsx
```diff
@@ -11,25 +18,42 @@ interface EditableTextProps {
   onBlur: (event: React.FocusEvent<HTMLInputElement>) => void;
   variant?: TypographyVariant;
   size?: 'small' | 'medium';
+  inputProps?: SxProps;
```
bytasv, line 21: I find this slightly misleading - it's named as `inputProps` but it can only control `sx` - would it make sense to name it `inputStyle` or `inputSx` instead? 🤔

Janpot:
I believe one of the goals of this component should be to hide away the implementation details of how we switch between editable/non-editable state. This PR seems to leak more implementation details than before. In the ideal scenario, consumers of this component don't know whether it's built by switching between typography and textfield, or by styling a single input. I suggest we base the styling purely on `variant` and `sx` props for both non-editable and editable state. If that's too hard we can perhaps compromise on `sx` only as a start.

Keep in mind that the typography styles are available in the `theme` object under `theme.typography[variant]`

packages/toolpad-app/src/components/EditableText.tsx
```diff
@@ -1,39 +1,94 @@
...
-  onKeyUp: (event: React.KeyboardEvent<HTMLInputElement>) => void;
-  onBlur: (event: React.FocusEvent<HTMLInputElement>) => void;
+  onKeyUp?: (event: React.KeyboardEvent<HTMLInputElement>) => void;
+  onBlur?: (event: React.FocusEvent<HTMLInputElement>) => void;
   variant?: TypographyVariant;
   size?: 'small' | 'medium';
+  sx?: SxProps;
+  helperText?: React.ReactNode;
+  disabled?: boolean;
+  inputProps?: InputBaseComponentProps;
```
Janpot, line 22: I don't see this prop used anywhere. If there's no use for it, I would remove it. No need to leak abstraction if we don't have to.

packages/toolpad-app/src/components/EditableText.tsx
```diff
@@ -1,39 +1,94 @@
...
   defaultValue?: string;
-  loading: boolean;
-  editing: boolean;
-  isError: boolean;
+  editing?: boolean;
+  isError?: boolean;
   errorText?: string;
-  onKeyUp: (event: React.KeyboardEvent<HTMLInputElement>) => void;
-  onBlur: (event: React.FocusEvent<HTMLInputElement>) => void;
+  onKeyUp?: (event: React.KeyboardEvent<HTMLInputElement>) => void;
```
Janpot, line 15: I'm not super fond of leaking these key handlers. Would it be make sense to build this using controlled properties only? i.e. `value` and `onChange` that act like a regular `TextField`. And to change editable state, `open` and `onClose`, analog to e.g. the MUI `Dialog`, `Menu`, `Drawer`,...

test/e2e-smoke/basic.spec.ts
```diff
@@ -19,9 +19,13 @@ test('basic app creation flow', async ({ page }) => {
 
   await page.click('[aria-label="Home"]');
 
-  const appCardSelector = `[role="article"]:has-text("${appName}")`;
+  const appRow = await page.locator(`[role="row"]`, {
+    has: await page.locator(`input[value="${appName}"]`),
```
Janpot, line 23: I'd prefer if we could keep the old locator. I have a hunch that there's an accessibility issue here. What happens if we add the `aria-readonly` to the input when it's not editable?

Janpot:
@bharatkashyap It seems like the title and description of this PR are a bit off?

test/e2e-smoke/basic.spec.ts
bharatkashyap: Adding the `aria-readonly` still doesn't help use the original selector, which relied on a `has-text`; but now since the name resides in the `value` of an `input`, `has-text` fails to find the element.

I've modified it a bit to use the selector format instead of locators, and now seems like the tests pass for Firefox as well

packages/toolpad-app/src/components/EditableText.tsx
```diff
@@ -1,42 +1,38 @@
...
   defaultValue?: string;
+  disabled?: boolean;
   editing?: boolean;
-  isError?: boolean;
   errorText?: string;
-  onKeyUp?: (event: React.KeyboardEvent<HTMLInputElement>) => void;
-  onBlur?: (event: React.FocusEvent<HTMLInputElement>) => void;
-  variant?: TypographyVariant;
+  helperText?: React.ReactNode;
+  isError?: boolean;
```
Janpot, line 10: Shall we keep the same interface as MUI the `TextField`? just a `error` and `helperText`

packages/toolpad-app/src/components/EditableText.tsx
```diff
@@ -1,42 +1,38 @@
...
-  TypographyVariant,
-  SxProps,
-  InputBaseComponentProps,
-} from '@mui/material';
+import { TextField, Theme, TypographyVariant, SxProps } from '@mui/material';
 
 interface EditableTextProps {
   defaultValue?: string;
+  disabled?: boolean;
   editing?: boolean;
```
Janpot, line 7: In terms of naming maybe it makes more sense to go use the adjective `editable` instead of the verb?

bharatkashyap: Do you mean `editable` instead of `editing`? I feel like it conveys different things: `editable` would mean whether the component can ever be edited; whereas `editing` to me conveys the idea that this prop is tied to its "state"

bytasv: My initial thoughts were the same, however if we looked at https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Editable_content - it kinda works and does what we are trying to do here as well, no?
(it's not without a reason why naming in programming is one of the most difficult things :D )

Janpot: Do you mean using a `Typography` and toggle `contentEditable` instead of using a `TextField`? It's an interesting idea, but we'd have to disable pasting formatted text and deal with accessibility and so on. I don't think it will necessarily make things simpler.

Or do you just propose to name the prop `contentEditable`? I think the existence of the prop illustrates that there's a preference of "editable" over "editing". i don't necessarily think we should use `contentEditable`, we're not really making the "content" editable the same way `contentEditable` does.

apedroferreira: Just adding my opinion - I agree with Bharat in the sense that something like `editable: false` to me would mean the component would be read-only or something...

Then again if the `editing` prop is not internal state and comes from outside the component, maybe the prop should express something like an "edit mode"?

bytasv: > Do you mean using a Typography and toggle contentEditable instead of using a TextField? It's an interesting idea

Oh sorry, my bad, I just added more confusion 😅 

No, I just wanted to refer to the prop name that is used `contentEditable` in our case that could be `editable` just as you proposed. I was also on the verge of whether it should `editing` but my choice went with using same format as `contentEditable` uses 🙂

bharatkashyap: I think `contentEditable` expresses a "permanent" nature of the element as being able to be edited; whereas what I was trying to express with the prop name is the "temporary" state of the component as either being able to be edited or not.

bharatkashyap: I think `contentEditable` expresses a "permanent" nature of the element as being able to be edited, whereas I was trying to express that the prop controls only the "temporary" state of the component, which is why "editing".  

Perhaps "focussed" could be clearer?

bytasv: Overall I think that neither name would is wrong so choosing either should be fine. I'd say we should pick one which is preferred by majority (don't count me). After all, after such a long debate I don't think either of us will be confused by what `editing` or `editable` for this component means 😅 

Just my 2 last cents which might be inline to what Pedro wrote - to me `editing` may sound a little bit more like a state defining value that we pass around, where `editable` defines kinda of interface so in the end it could be even something like this:
```jsx
const [editing, setEditing] = useState()
...
return <Field editable={editing} />
```

bharatkashyap: > ```js
> const [editing, setEditing] = useState()
> ...
> return <Field editable={editing} />
> ```

That makes sense to me. We can go ahead with `editable`

packages/toolpad-app/src/toolpad/Home.tsx
```diff
@@ -271,51 +368,131 @@ function AppCard({ app, activeDeployment, onDelete }: AppCardProps) {
...
     </React.Fragment>
   );
 }
+interface TabPanelProps {
+  children?: React.ReactNode;
+  index: number;
+  value: number;
+}
+
+function TabPanel(props: TabPanelProps) {
```
Janpot, line 463: In [other places](https://github.com/mui/mui-toolpad/blob/54095761505e65617bb043f1b499c6095d380858/packages/toolpad-app/src/toolpad/AppEditor/BindingEditor.tsx#L245) we use the MUI `TabContext` and `TabPanel`.

packages/toolpad-app/src/toolpad/Home.tsx
```diff
@@ -334,55 +511,119 @@ export default function Home() {
...
-                      );
-                    })
-                  : 'No apps yet';
-              default:
-                return '';
-            }
-          })()}
-        </Box>
-      </Container>
+      <TabPanel index={0} value={tabIndex}>
```
Janpot, line 529: Does this have to be part of this PR?

packages/toolpad-app/src/toolpad/Home.tsx
```diff
@@ -334,55 +511,119 @@ export default function Home() {
...
+                          );
+                        })
+                      : 'No apps yet';
+                  default:
+                    return <AppCard />;
+                }
+              })()}
+            </Box>
+          ) : null}
+          <Table aria-label="apps list" size="medium">
```
Janpot, line 595: Should the table be there even when in viewMode "grid"?

packages/toolpad-app/src/toolpad/Home.tsx
```diff
@@ -334,55 +511,119 @@ export default function Home() {
...
+                value="grid"
+                aria-label="grid view"
+                color={viewMode === 'grid' ? 'primary' : undefined}
+              >
+                <GridViewIcon />
+              </ToggleButton>
+            </ToggleButtonGroup>
+          </Toolbar>
+          {viewMode === 'grid' ? (
+            <Box
```
Janpot, line 557: Would it make sense to extract a `AppsGridView` and a `AppsTableView` component, with the exact same interface and switch between them bases on `viewMode`? e.g.

```tsx
const AppsView = viewMode === 'grid' ? AppsGridView : AppsTableView
// ...

<AppsView apps={...} />
```

packages/toolpad-app/src/toolpad/Home.tsx
bharatkashyap: This is to move the "Apps" heading into an "Apps" tab

Janpot: Is there a need for this right now? It seems a bit redundant. Also, it looks (and works) completely different from how this menu is in the editor. Let's leave it out of the scope of this PR.

packages/toolpad-app/src/toolpad/Home.tsx
bharatkashyap: Nope, fixed this in the refactor to `AppView`

Janpot:
Resolved the conflicts that emerged after merging https://github.com/mui/mui-toolpad/pull/723

Janpot (approved):
👌

Issue #77, opened 2022-01-25
Support mutations


Janpot:
introduce concept of mutations:
- model after and build on top of `react-query` mutations? 
- Create a "mutation" abstraction that is analog to the "query" abstraction, but instead of providing a `.data` property in the bindings, it provides `.mutate` (or so) function that can be called in onClick handlers (It's already possible to call arbitrary javascript in these handlers)
- currently the queries make available a `.refetch` function to the bindings, this can be used to refresh any query you want after a mutation.

PR #698, opened 2022-07-27
Support mutations as non-idempotent version of queries

Janpot:
* serverless function only for now, call from js action only for now. Let's iterate
* Create a mutation, with ability to bind parameters to values available in the page
* Call mutations through `mutation.call` in event handlers. Using javascript biding allows you to perform other actions in parallel, or run refetches after the mutation is done.
* Track mutation state through `isLoading` and `error`

https://user-images.githubusercontent.com/2109932/184329010-9932d3ca-1937-4d4d-ba00-3c3e74b7c8f0.mov

Closes https://github.com/mui/mui-toolpad/issues/77


Janpot:
Ok, let's see if we can move this forward

apedroferreira:
looks great too! UX feedback:
- ideally the default experience should probably be a simple interface where you just need to define a method (POST as default), the URL and the body, or something close to that? and then being able to edit the whole implementation as a more advanced option.
- it probably wouldn't be immediately obvious to me than we can use top level `await` in the `onClick` JS expression, not sure if that's a relevant issue
- later on maybe we could be able to somehow make it possible for mutations and queries to work over the same dataset? but i guess that's something for later

PR #721, opened 2022-08-02
Rest datasource configure bodies and headers

Janpot:
- [x] configure body
  - [x] raw body, with selectable content type
  - [x] x-www-form-urlencoded
- [x] configure url query parameters
- [x] configure headers
- [x] move everything but the method+url under tabs
- [x] move parameters panel to the same place as postgres and function datasources

https://user-images.githubusercontent.com/2109932/188429030-b70fe2cb-d81f-479a-89fb-265c8c32e8e4.mov

Fixes https://github.com/mui/mui-toolpad/issues/850


apedroferreira:
cool!
a couple questions/thoughts:
- i don't know what the "Parameters" below are. could they be included with those tab options if it makes sense?
- the "transform" functionality is logically a bit different to the other tabs, as it's not something as inherent to the request itself. i wonder if we should separate it somehow? but i also don't dislike it being there in a tab, seems easy to use

packages/toolpad-app/src/toolpadDataSources/rest/client.tsx
```diff
@@ -201,6 +211,27 @@ function QueryEditor({
     }));
   }, []);
 
+  const handleBodyChange = React.useCallback((newBody: Maybe<Body>) => {
```
apedroferreira, line 214: should we have a reusable function for these? the callbacks seem pretty similar.

apedroferreira: but if we predict these might changeover time it's best to keep them separate though, just a thought

Janpot:
> i don't know what the "Parameters" below are. could they be included with those tab options if it makes sense?

It's parameters to the datasource query itself. this is where you can bind state of the page, which is usable inside of the query bindings. I moved it to this place because it's the same location as in the function and postgresql datasources.

packages/toolpad-app/src/toolpadDataSources/rest/client.tsx
Janpot: One could invent a hook like:

```tsx
function useQueryChangeHandler<F extends Function>(
  setInput: React.Dispatch<React.SetStateAction<QueryEditorModel<FetchQuery>>>,
  property: string,
  producer: F,
): F {
  return React.useCallback<F>(
    (...args: any[]) => {
      setInput((existing) => ({
        ...existing,
        query: { ...existing.query, [property]: producer(...args) },
      }));
    },
    [producer, property, setInput],
  );
}
```

and use as 

```tsx
const handleMethodChange = useQueryChangeHandler('method', (event) => event.target.value);
const handleMethodChange = useQueryChangeHandler('url', (newUrl) => newUrl);
// ...
```

But that `produce` parameter would invalidate the callback on each render. So we could declare those functions in module scope, or wrap them with `useCallback` as well, but at which point do all the abstractions become less readable again? The way it's done now is idiomatic and explicit and easy to understand. I kind of prefer it, even though it could probably be expressed in less lines of code. I was thinking for the future [we could use `immer`](https://github.com/mui/mui-toolpad/issues/763) to remove some of the duplication inside the `setInput`.

e.g. from
```tsx
  const handleMethodChange = React.useCallback((event: React.ChangeEvent<HTMLInputElement>) => {
    setInput((existing) => ({
      ...existing,
      query: { ...existing.query, method: event.target.value },
    }));
  }, []);
```
to
```tsx
  const handleMethodChange = React.useCallback(
    (event: React.ChangeEvent<HTMLInputElement>) =>
      setInput(
        produce((draft) => {
          draft.query.method = event.target.value;
        }),
      ),
    [],
  );
```

apedroferreira: yeah, it's probably not worth introducing something so complex yet at least (we could not use a `produce` function/argument, but then we would probably be making the callback less flexible)
it was just a thought, i think we can keep this as it is it's pretty easy to read and understand.
i agree with using `immer` later for this kind of operations later though, makes them a bit cleaner!

apedroferreira (approved):

packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -40,7 +40,7 @@ import {
   NodeRuntimeWrapper,
   ResetNodeErrorsKeyProvider,
 } from '@mui/toolpad-core/runtime';
-import { pick } from 'lodash-es';
+import * as _ from 'lodash-es';
```
bytasv, line 43: minor thing - just wondering if we are using something to treeshake such imports?
In the past I've used [babel-transform-imports](https://www.npmjs.com/package/babel-plugin-transform-imports) maybe in the future it would be worth to look into that 🤔

Janpot: It is tree shaken by the `swc` minifier in next.js. The import style doesn't influence the behavior.

bytasv: Oh nice, thanks, I didn't know that 👍

PR #729, opened 2022-08-03
Add RFC template

bytasv:

A template for RFCs


mnajdova:
Maybe we also need a section for examples? This is the easiest way for seeing new API in action.

.github/ISSUE_TEMPLATE/3.rfc.yml
```diff
@@ -0,0 +1,41 @@
...
+        This is the core of RFC, and its purpose is to help you think through the problem because [writing is thinking](https://medium.learningbyshipping.com/writing-is-thinking-an-annotated-twitter-thread-2a75fe07fade).*
+        
+        Consider:
+        - using diagrams to help illustrate your ideas
+        - including code examples if you're proposing an interface or system contract
+        - linking to project briefs or wireframes that are relevant
+  
+  - type: textarea
+    attributes:
+      label: Relevant resources 🔗
```
mnajdova, line 40:
```suggestion
      label: Relevant resources/benchmarks 🔗
```

mnajdova: For example, this is how other products have solved this problem.

bytasv:
> Maybe we also need a section for examples? This is the easiest way for seeing new API in action.

@mnajdova in the `Proposed option` section we have this line:

> - including code examples if you're proposing an interface or system contract

Do you think we should add `Examples` as a standalone section? I'm thinking that if I were to open an RFC I would aim to add examples in `Proposed option` and `What are our options?` sections anyways 🤔 
Maybe it would be enough to extend `What are our options?` section and also mentioned that examples could be useful?

.github/ISSUE_TEMPLATE/3.rfc.yml
```diff
@@ -0,0 +1,41 @@
...
+      description: Provide a list of requirements that should be met by the accepted proposal
+
+  - type: textarea
+    attributes:
+      label: What are our options? 💡
+      description: Have you considered all options that we can achieve same/similar thing? It doesn't have to be explored in same detail as the main proposal, but having alternative options that you can present can help strengthen main proposal.
+
+  - type: textarea
+    attributes:
+      label: Proposed option 🟢
```
mnajdova, line 29:
```suggestion
      label: Proposed solution🟢
```

Sounds better :)

mnajdova (approved):
All good from my end :)

DanailH (approved):
Sorry for the delay. I think it's solid. Should you convert it from a Draft to a PR?

Issue #752, opened 2022-08-06
Deleting component doesn't work when just added to the page


oliviertassinari:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

I have to unselect/reselect to have the <kbd>Backspace</kbd> key delete the component.

### Expected behavior 🤔

I can press the <kbd>Backspace</kbd> key as soon as I add the component to remove it.

### Steps to reproduce 🕹

https://user-images.githubusercontent.com/3165635/183259732-7c0478ad-2eba-4360-8755-596bee610bca.mov

### Context 🔦

I **systematically** confuse the TextField with the Typography component. Hopefully, #598 would fix this. This is how I found the bug.
I'm only reporting it to collect upvotes, to not fix yet.

Janpot:
I think it makes sense to at least look into what's going on here. And fix now only if it's straightforward

apedroferreira:
hopefully should be a quick one, i can take a look

PR #771, opened 2022-08-10
Fix deleting newly placed components with Backspace key

apedroferreira:
Fixes issue where it was not possible to delete a component from the editor with the "Backspace" key just after placing it, without clicking it first.
My solution to this issue was to refocus on the element that catches `keydown` events after placing a new element, because dragging it from outside the editor was moving the focus outside of it.

Closes https://github.com/mui/mui-toolpad/issues/752

- [ ] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/material-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).

### DIFF
--- a/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/RenderOverlay.tsx
+++ b/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/RenderOverlay.tsx
@@ -783,6 +783,8 @@ export default function RenderOverlay({ canvasHostRef }: RenderOverlayProps) {
     ],
   );
 
+  const overlayRef = React.useRef<HTMLDivElement | null>(null);
+
   const handleNodeDrop = React.useCallback(
     (event: React.DragEvent<Element>) => {
       const cursorPos = canvasHostRef.current?.getViewCoordinates(event.clientX, event.clientY);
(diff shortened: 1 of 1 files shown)


Janpot (approved):
👍 makes sense, nice catch!

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/RenderOverlay.tsx
```diff
@@ -1011,14 +1013,20 @@ export default function RenderOverlay({ canvasHostRef }: RenderOverlayProps) {
...
+      api.dragEnd();
+
       if (newNode) {
         api.select(newNode.id);
+
+        // Refocus on overlay so that keyboard events can keep being caught by it
+        const overlayElement = overlayRef.current;
+        if (overlayElement) {
+          overlayElement.focus();
+        }
```
oliviertassinari, line 1029: Do we need the defensive check here? I would assume React guarantees it's defined, hence:
```suggestion
        overlayRef.current!.focus();
```

would be better, so we catch regressions at runtime

Janpot: In [some other places](https://github.com/mui/mui-toolpad/blob/a1794f5c61e160c3df82a9fabe77728d3181fab4/packages/toolpad-app/src/components/MonacoEditor.tsx#L279) we do:

```suggestion
        invariant(overlayRef.current, 'Ref not bound');
        overlayRef.current.focus();
```

This communicates the intent a bit clearer and avoids usage of `!`

apedroferreira: good points, thanks! i used an `invariant` check here and also in a couple other places where it seemed right to use it

PR #772, opened 2022-08-10
Prevent node HUD from getting cut-off

apedroferreira:
In the top row, every node's HUD was being cut-off, as the following image shows:

Screen Shot 2022-08-10 at 19 55 45

This PR fixes the issue by showing the HUD at the bottom when there is not enough space for it at the top.

https://user-images.githubusercontent.com/10789765/183995142-4ba5b6ca-d230-4046-9026-88c3a7eb1f8d.mov

### DIFF
--- a/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/RenderOverlay.tsx
+++ b/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/RenderOverlay.tsx
@@ -1365,8 +1365,6 @@ export default function RenderOverlay({ canvasHostRef }: RenderOverlayProps) {
 
         const isResizing = Boolean(draggedEdge) && node.id === draggedNodeId;
 
-        const hasBottomHud = nodeRect.y <= 32;
-
         return (
           <React.Fragment key={node.id}>
             {!isPageNode ? (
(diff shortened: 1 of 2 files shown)


packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/RenderOverlay.tsx
```diff
@@ -1365,6 +1365,8 @@ export default function RenderOverlay({ canvasHostRef }: RenderOverlayProps) {
 
         const isResizing = Boolean(draggedEdge) && node.id === draggedNodeId;
 
+        const hasBottomHud = nodeRect.y <= 32;
```
Janpot, line 1368: `nodeRect` is already passed to `NodeHud`, can't it do the calculation internally?

apedroferreira: that's correct, i was thinking of another approach first so i didn't consider this - i'll do everything in `NodeHud`

apedroferreira: done, there's other stuff like this that i can move from `RenderOverlay` into `NodeHud` - i have those noted to do it in the future

Janpot (approved):
Is there a way we can infer this 32 y-value somewhere? Or else perhaps make it a constant?

apedroferreira:
> Is there a way we can infer this 32 y-value somewhere? Or else perhaps make it a constant?

it's a bit higher than the height of the HUD, but from what i saw it wasn't being set by a constant value. 
we can set a fixed height in the HUD and use a constant.

Janpot:
> we can set a fixed height in the HUD and use a constant.

Just using a constant is fine for me as well. It's just about reducing magic numbers

PR #774, opened 2022-08-11
Fix missing key warning

Janpot:
The `Link` was added afterwards, forgetting to move the key property

### DIFF
--- a/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/ComponentCatalog.tsx
+++ b/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/ComponentCatalog.tsx
@@ -163,8 +163,13 @@ export default function ComponentCatalog({ className }: ComponentCatalogProps) {
                 })}
               {FUTURE_COMPONENTS.map((futureComponent, i) => {
                 return (
-                  <Link href={futureComponent.githubLink} underline="none" target="_blank">
-                    <ComponentCatalogItem key={`futureComponent[${i}]`}>
+                  <Link
+                    key={`futureComponent[${i}]`}
+                    href={futureComponent.githubLink}
+                    underline="none"
+                    target="_blank"
+                  >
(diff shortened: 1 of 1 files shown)


packages/toolpad-app/src/toolpad/AppEditor/PageEditor/ComponentCatalog.tsx
```diff
@@ -163,8 +163,13 @@ export default function ComponentCatalog({ className }: ComponentCatalogProps) {
                 })}
               {FUTURE_COMPONENTS.map((futureComponent, i) => {
                 return (
-                  <Link href={futureComponent.githubLink} underline="none" target="_blank">
-                    <ComponentCatalogItem key={`futureComponent[${i}]`}>
+                  <Link
+                    key={`futureComponent[${i}]`}
```
bytasv, line 167: Could we use `futureComponent.githubLink` instead? 🤔 or aren't those links unique? (just thinking if we can avoid using index as a key)

Janpot: The list is purely static, never reordered, so using the index should be perfectly fine to use here. We could use the `futureComponent.displayName` perhaps. But there are no hard contracts in place that guarantee uniqueness neither

bytasv: What about `futureComponent.githubLink` wouldn't that be the uniq thing? 🤔

Janpot: in practice? probably yes, but the result would be no more safe against bugs than using the index. Nothing is preventing a developer from duplicating any of these links. https://github.com/mui/mui-toolpad/pull/775 however, does guarantee those safeguards.

PR #775, opened 2022-08-11
Improve handling of React keys in ComponentCatalog

Janpot:
Refactor component catalog to guarantee unique future component keys

Continuation of https://github.com/mui/mui-toolpad/pull/774 (which was auto-merged)

Also rename `githubLink` to `url`. It doesn't matter where the url points, so let's use a more generic name.

### DIFF
--- a/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/ComponentCatalog.tsx
+++ b/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/ComponentCatalog.tsx
@@ -161,10 +161,10 @@ export default function ComponentCatalog({ className }: ComponentCatalogProps) {
                     </ComponentCatalogItem>
                   );
                 })}
-              {FUTURE_COMPONENTS.map((futureComponent, i) => {
+              {FUTURE_COMPONENTS.map((futureComponent) => {
                 return (
                   <Link
-                    key={`futureComponent[${i}]`}
+                    key={`futureComponent.${futureComponent.displayName}`}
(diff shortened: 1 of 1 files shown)


packages/toolpad-app/src/toolpad/AppEditor/PageEditor/ComponentCatalog.tsx
```diff
@@ -15,44 +15,20 @@ import {
...
-    displayName: 'DatePicker',
-    githubLink: 'https://github.com/mui/mui-toolpad/issues/743',
-  },
-  {
-    displayName: 'Checkbox',
-    githubLink: 'https://github.com/mui/mui-toolpad/issues/742',
-  },
-];
+const FUTURE_COMPONENTS = new Map<string, FutureComponentSpec>([
+  ['Form', { url: 'https://github.com/mui/mui-toolpad/issues/749' }],
```
bytasv, line 23: Just curious - why would we omit `displayName` key and use array form instead? 🤔 Seems like even in the code below it could be bit simpler:
```
FUTURE_COMPONENTS = [
 { name, link }, ...
]
//...

Array.from(FUTURE_COMPONENTS, ({ name, link} => <Link key={name} href={link} />)
```

Janpot: because a map is a good code construct to communicate that these are intended to be unique, and a good runtime construct to guarantee it as well

bytasv (approved):

PR #776, opened 2022-08-11
Handle appDom versioning

### LINKED ISSUE #741: Toolpad App DOM versioning and migrations
As we evolve our application format, we sometimes need to make breaking changes. As users upgrade their Toolpad instance their apps shouldn't break. 

When we can't add changes to the app definition in a backwards compatible way, we will have to migrate it.

I propose:
*  we add a version number to the app dom
* A migration function for each version which takes the dom from previous to current version
* On every read from the db, we immediately compare the dom version with the version of toolpad and run necessary migrations
* The application itself will always use the latest version, we reduce migrations to a single point in Toolpad
* Therefore every write will be using the latest versions, as soon as the dom is saved once, the application is migrated on disk.
* We also migrate snapshots on read
* New snapshots will be written using the latest versions
* Old snapshots are never overwritten, migrate on read. In the future we can prompt users to create a new snapshot. not part of this issue

This provides us a good basis to keep working, we can add UI and scripts for migration workflows as the need arises.

bharatkashyap:

- Fixes #741


Janpot:
Adding the pseudocode we talked about in last 1-on-1 as a potential starting point

```tsx
// ./migrations/version1.ts
export default {
  up (dom) {
    invariant(!dom.version || dom.version === '0', 'Can only migrate dom of version 0');
    return dom;
  }
}


// ./migrations/version2.ts
export default {
  up (dom) {
    invariant(dom.version === '1', 'Can only migrate dom of version 1');
    return dom;
  },
  down (dom) {
    invariant(dom.version === '2', 'Can only migrate dom of version 2');
    return dom;
  }
}

// ./migrations/index.ts
import version1 from './version1'
import version2 from './version2'

const versions = new Map([
  ['1', version1],
  ['2', version2],
]);

const latestVersion = Array.from(versions.keys()).pop();

function migrate(dom, version = latestVersion) {
  return //...
}
```

bharatkashyap:
Should the migrated dom be written back to the app in the database? Should we save the versions of the dom and keep an audit log?

packages/toolpad-app/src/appDomMigrations/v1.ts
```diff
@@ -0,0 +1,91 @@
...
+        }),
+      });
+    }
+  }
+  return node;
+}
+
+export default {
+  up(dom: AppDom): AppDom {
+    const { nodes, version } = dom;
```
Janpot, line 72:
```suggestion
    const { nodes, version = 0 } = dom;
```

packages/toolpad-app/src/appDomMigrations/v1.ts
```diff
@@ -0,0 +1,87 @@
...
+            connectionId: createConst(ref(node.attributes.query.value.connectionId.value)),
+          }),
+        }),
+      });
+    }
+  }
+  return node;
+}
+
+function migrateLegacyNavigationAction(node: ElementNode<any>): ElementNode<any> {
```
bharatkashyap, line 53: Should we remove https://github.com/mui/mui-toolpad/blob/master/packages/toolpad-app/src/appDom.ts#L903 and https://github.com/mui/mui-toolpad/blob/master/packages/toolpad-app/src/appDom.ts#L890 as well?

Janpot: these need to be moved in a migration

bharatkashyap: Yep, have added these to the first migration in this PR

packages/toolpad-app/src/appDomMigrations/index.ts
```diff
@@ -0,0 +1,10 @@
+import invariant from 'invariant';
+import v1 from './v1';
+
+const versions = new Map([[1, v1]]);
+
+export const latestVersion = Array.from(versions.keys()).pop();
+
+invariant(latestVersion, 'No migrations found');
```
Janpot, line 8:
```suggestion
```

bharatkashyap: This helps with the type error on the `versions.get(latestVersion)` in the last line, but I agree with this

Janpot (commented):
👍

PR #778, opened 2022-08-11
Store the dom in a json column instead of normalizing it

Janpot:
From a bird's eye perspective, a stored dom is a collection of dom nodes with associated attributes. This is currently normalized in two tables, dom nodes, and dom attributes. The intent was to have a good datastructure to atomically update single dom nodes. The idea being that it would allow patching the dom on a per-node basis. As our application definition becomes more and more complex, this way of storing the dom becomes a blocker (it's becoming challenging to represent complex request configurations for https://github.com/mui/mui-toolpad/pull/721). We could achieve more a complex structures and flexibility by just storing the data as JSON.

This PR introduces a dom property on the App entity to store the stores the dom in there instead of normalizing it to those two tables. It's backwards compatible as on read it falls back to the old datastrutures if no dom is found. (We can remove the old tables later)

* The [proposed solution](https://www.prisma.io/docs/concepts/components/prisma-client/working-with-fields/working-with-json-fields#advanced-example-update-a-nested-json-key-value) to update JSOn fields is not great, it can't update only parts of the field. If we want to write an efficient dom update, we're going to have to write a custom query. See https://github.com/prisma/prisma/issues/5057
* This will also align the structure more to the way we store snapshots, and therefor reduce the chance of having discrepancies in both dom loading strategies

This PR will also help for the [migration feature](https://github.com/mui/mui-toolpad/pull/776) @bharatkashyap is working on as it takes away the necessity to store a version number somewhere separately


bharatkashyap (approved):
Makes sense! I imagine that the migration can then be run in the `loadPreviewDom` method, whereas the `loadPreviewDomLegacy` method can return a dom version 0 by default?

Janpot:
the DomNode and DomNodeAttribute tables won't be updated again, so if we load the dom from there it will be at version `0`, always.

apedroferreira (approved):
looks much more mantainable!

Janpot:
It will also remove the need for globally unique dom node ids. Which means we may:
* Use much shorter ids, as they only need to be unique within an application. e.g. we may use 5 characters. Or a variable length id
* Simplify duplicating, importing and sharing applications as we won't require globally unique ids, we don't need to reindex an application

Issue #709, opened 2022-07-29
Opening an application without changing marks it as edited


Janpot:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

1. From the overview page, open an app that is not at the first position.
2. Move back to the overview page
3. The app you just opened has moved to the first position

### Expected behavior 🤔

If the app is not edited, it shouldn't be marked as such

apedroferreira:
this might be because of the row resizing logic, i can look into it

apedroferreira:
ok i've investigated and doesn't seem to be anything in the `RenderPanel` after all... 🤔

PR #779, opened 2022-08-12
Make sure the application isn't saved during mounting

Janpot:
Initial dom is compared with `null`, which triggers an update. Solution: instead, initialize the ref that stores last saved application with the initial application, this makes it not detect a difference.

Fixes https://github.com/mui/mui-toolpad/issues/709


bharatkashyap (approved):
👍

PR #780, opened 2022-08-12
Make sure old style references are backwards compatible

Janpot:
TheseSome properties that used to be represented by a string `'xyz'` are now represented by a `{ $$ref: 'xyz' }`. This allows us to detect them in our data structure and treat them as their own datatype. There was no backwards compatibility put in place and we can't do migrations yet, so this will fix older apps that are now broken

### DIFF
--- a/packages/toolpad-app/src/appDom.ts
+++ b/packages/toolpad-app/src/appDom.ts
@@ -822,6 +822,13 @@ export function deref(nodeRef: NodeReference): NodeId;
 export function deref(nodeRef: null | undefined): null;
 export function deref(nodeRef: Maybe<NodeReference>): NodeId | null;
 export function deref(nodeRef: Maybe<NodeReference>): NodeId | null {
+  if (typeof nodeRef === 'string') {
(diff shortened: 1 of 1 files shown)


bharatkashyap (approved):
Makes sense

bytasv:
@Janpot I was wondering - would it make sense to tackle backwards compatibility / migrations problem a little bit more? I mean going forward I assume we will have more and more such cases where we want to tweak data structure. We could open a GH issue and maybe figure out if there is a pattern that we could reuse, or at least start gather info where we needed that so we can build a better picture 🤔

Janpot:
@bytasv yes, absolutely, it's becoming more and more a blocker as the app gets more complex. Datamodels that we started out with are becoming too simplistic for newer features. There is the following ticket https://github.com/mui/mui-toolpad/issues/741
The idea is to version our application datamodel, and provide migrations upwards for each version. then we'll migrate the model immediately when we load it from the db, so that the application will always work with the latest version.
@bharatkashyap Is starting an effort on that in https://github.com/mui/mui-toolpad/pull/776

PR #782, opened 2022-08-12
Try out sharing vscode setup

Janpot:
One way of solving the onboarding problem. This should configure formatting code correctly on save. An alternative could be to run prettier as an eslint plugin


.vscode/extensions.json
```diff
@@ -0,0 +1,3 @@
+{
+  "recommendations": ["dbaeumer.vscode-eslint", "esbenp.prettier-vscode"]
```
apedroferreira, line 2: pretty cool, didn't know about this

Issue #538, opened 2022-06-08
Select options editor enhancements


Janpot:
* Editing an option, but only the value results in creation of an option with empty label. Empty label should be considered as "no label". clearing the label input should be regarded as "deleting the label"
* You can add an option using the "enter" key, but there should be a "+" button to the side as well, for discoverability

prakhargupta1:
@Janpot Can you add reference screenshot for this task ?

Janpot:
👍  done

bharatkashyap:
> * You can add an option using the "enter" key, but there should be a "+" button to the side as well, for discoverability

A "+" button on the side of the input? Does the helper text on the input not make the "enter" feature discoverable?

Janpot:
> A "+" button on the side of the input? Does the helper text on the input not make the "enter" feature discoverable?

As a thought exercise, imagine a search input. Which one has the better UX?

- The one that has a magnifying glass icon inside
- The one that has a helper text underneath that says "press Enter to search"

PR #788, opened 2022-08-12
Fix scrollbar when selection options

oliviertassinari:
**Before** (most visible on Windows)

Screenshot 2022-08-13 at 00 37 48

**After**

Screenshot 2022-08-13 at 00 37 27

This seems to have been introduced in #443.

---

@bharatkashyap Could you configure your macOS to match how Windows behaves with a scrollbar? This way, you will experience [how most](https://www.notion.so/mui-org/Analytics-f2e2576203564e6c8dc3429cf305a36f#8a38433d21814a15acfb99ce2eac7d6b) of the developers will experience the interface

Screenshot 2022-08-13 at 00 38 48


Janpot (approved):
nice 👍 
This was also reported as part of https://github.com/mui/mui-toolpad/issues/538

Issue #790, opened 2022-08-13
Throw an exception when creating a custom component with a hyphen


oliviertassinari:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

It throws.

### Expected behavior 🤔

It works.

### Steps to reproduce 🕹

Steps:

1. Open https://master--toolpad.mui.com/_toolpad/app/cl42qt2jx01249xnr14xjce8b/editor/pages/cl42qt2kt00029xnr5rta2nog
2. <img width="479" alt="Screenshot 2022-08-13 at 13 40 37" src="https://user-images.githubusercontent.com/3165635/184492576-c976e507-1368-4788-bc3b-f4f85ab224bd.png">
3. <img width="527" alt="Screenshot 2022-08-13 at 13 40 50" src="https://user-images.githubusercontent.com/3165635/184492575-bc9f6034-d808-46fd-a0be-7a03f45f9c1c.png">

### Context 🔦

It's never clear to me what capitalization and word separation I should use in the different text fields of Toolpad.

Janpot:
We should add validation to this input, and the create button should be disabled as long as the input is invalid

oliviertassinari:
> the create button should be disabled

I would be in favor of never disabling a submit button, but to make it a noop and to show the error when pressed.

Janpot:
> I would be in favor of never disabling a submit button, but to make it a noop and to show the error when pressed.

I think it makes sense to decide on a case-by-case basis. 
- A single input + submit button in a dialog: we should show a validation message as soon as invalid input is typed, and disable the button to communicate that nothing will happen when clicked.
- a large form with multiple inputs that can scroll: I think not disabling the button and showing the error when pressed (as you described) makes sense

PR #802, opened 2022-08-16
Validate the name when creating code components

Janpot:
Proposing this as a solution for this dialog:

Screenshot 2022-08-16 at 16 39 20

Closes https://github.com/mui/mui-toolpad/issues/790

### DIFF
--- a/packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/CreateCodeComponentNodeDialog.tsx
+++ b/packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/CreateCodeComponentNodeDialog.tsx
@@ -62,7 +62,10 @@ export default function CreateCodeComponentDialog({
     event.target.select();
   }, []);
 
-  const inputErrorMsg = name ? appDom.validateNodeName(name, 'a code component name') : null;
+  const inputErrorMsg = React.useMemo(
+    () => (name ? appDom.validateNodeName(name, 'a code component name') : null),
+    [name],
+  );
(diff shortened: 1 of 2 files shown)


packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/CreateCodeComponentNodeDialog.tsx
```diff
@@ -62,6 +62,9 @@ export default function CreateCodeComponentDialog({
     event.target.select();
   }, []);
 
+  const inputErrorMsg = name ? appDom.validateNodeName(name, 'a code component name') : null;
```
apedroferreira, line 65: i guess it might be a bit more semantic to validate on change or in an effect that only runs if `name` changes, instead of running it on every render. shouldn't be much an issue for now though, of course

apedroferreira:
looks like a good solution to me!

packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/CreateCodeComponentNodeDialog.tsx
Janpot: wrapped it with a `useMemo`

apedroferreira: that works, thanks!

apedroferreira (approved):

Issue #712, opened 2022-07-29
Extract/automate logic for the `Box` that wraps every component


apedroferreira:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

With the new drag & drop system, every component needs to be wrapped by a `Box` in order to work as expected.
As of now, these boxes have been added manually in each component's file.

2 things need to be done to improve this and prevent new components from malfunctioning:
- [ ] Make it so that every component is wrapped by a `Box` automatically without it needing to be added manually to each component
- [ ] Extra: Make it possible to access the values of the bounding rectangle for both the container box and the element immediately inside it - this will allow us in the future to show a much more intuitive UI for resizing elements, for example.

PR #804, opened 2022-08-16
Automatically wrap non-layout components in box containers

apedroferreira:
Issue https://github.com/mui/mui-toolpad/issues/712

Screen Shot 2022-08-16 at 20 20 18

Not 100% convinced about this UI for the box alignment (which might just be a temporary thing until we improve resizing anyway), but just looking for feedback on where to place it, or this might be okay after all.

### DIFF
--- a/packages/toolpad-app/src/runtime/ToolpadApp.tsx
+++ b/packages/toolpad-app/src/runtime/ToolpadApp.tsx
@@ -257,7 +257,7 @@ function RenderedNodeContent({ node, childNodeGroups, Component }: RenderedNodeC
         return null;
       }
 
-      const action = node.props?.[key];
+      const action = (node as appDom.ElementNode).props?.[key];
 
       if (action?.type === 'navigationAction') {
         const handler = () => {
@@ -484,7 +484,8 @@ function parseBindings(
       if (!isPageLayoutComponent(elm)) {
         for (const [propName, argType] of Object.entries(layoutBoxArgTypes)) {
           const binding =
-            elm.layout?.[propName] || appDom.createConst(argType?.defaultValue ?? undefined);
+            elm.layout?.[propName as keyof typeof layoutBoxArgTypes] ||
+            appDom.createConst(argType?.defaultValue ?? undefined);
           const bindingId = `${elm.id}.layout.${propName}`;
           const scopePath = `${elm.name}.${propName}`;
           parsedBindingsMap.set(bindingId, parseBinding(binding, { scopePath }));
--- a/packages/toolpad-app/src/runtime/ToolpadApp.tsx
+++ b/packages/toolpad-app/src/runtime/ToolpadApp.tsx
@@ -63,7 +63,10 @@ import usePageTitle from '../utils/usePageTitle';
 import ComponentsContext, { useComponents, useComponent } from './ComponentsContext';
 import { AppModulesProvider, useAppModules } from './AppModulesProvider';
 import Pre from '../components/Pre';
-import { boxAlignArgTypeDef, boxJustifyArgTypeDef } from '../toolpadComponents/layoutBox';
+import {
+  layoutBoxAlignArgTypeDef,
+  layoutBoxJustifyArgTypeDef,
+} from '../toolpadComponents/layoutBox';
 
 const USE_DATA_QUERY_CONFIG_KEYS: readonly (keyof UseDataQueryConfig)[] = [
   'enabled',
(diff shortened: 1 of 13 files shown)


packages/toolpad-core/src/types.ts
```diff
@@ -227,8 +227,16 @@ export interface ComponentConfig<P> {
    */
   loadingProp?: keyof P & string;
   /**
-   * Designates a property as "the resizable height property". If Toolpad detects any vertical resizing of
-   * the component it will forward it to this property.
+   * Enables controlling the aligment of the component container box.
+   */
+  hasLayoutBoxAlign?: boolean;
```
Janpot, line 232: Which components don't we want this for?

apedroferreira: DataGrid, for example, as it takes up the whole box?
(btw i noticed bindings are still broken and fixing it now, sorry about that)

Janpot: This is public API, I'm not sure about the interface
```
hasLayoutBoxAlign
hasLayoutBoxJustify
```

* It's a bit verbose for the current use-case. e.g. might as well get away with something like a `layout: 'static' | 'flex'` to mark whether a component that sizes with its box, or has a static size.
* Would it make sense to be more explicit about direction? i.e. use vertical/horizontal or something instead of "align" and "justify". `layout: { vertical: 'flex', horizontal: 'static' }`
* It's not powerful enough to build upon. e.g. it might depend on component state. like a TextField that is in mode `fullWidth` may want to have its justification disabled.



```
resizableHeightProp
```
I'm not sure we should rely on a custom component correctly implementing the height prop. Can't we remove this `resizableHeightProp` and instead allow changing the height of the box container of every component?

apedroferreira: just fixed the bindings, i hadn't messed with bindings before but seems good now

Janpot: 🤔 I'm not sure whether we should make these bindable, at least the height prop shouldn't be I think, it would create too much layout shift.

apedroferreira: great points, i agree with all of them
- something simpler like the `layout` prop you mentioned should cover all our current use cases, no point yet even in align/justify being separate things, i'll change it to something like that
- justify/align naming is being confusing even to me so i'll change them to something more explicit
- about making the resizable height be the height of the container box, now that i abstracted the box container and am creating more "layout" props this should be easier to do and makes complete sense - this i probably will do in a separate PR though

apedroferreira: > 🤔 I'm not sure whether we should make these bindable, at least the height prop shouldn't be I think, it would create too much layout shift.

I ended up doing it because i noticed bindings were setting the default values in the controls. I don't see any cons to doing it for the alignment/justify even if it's not super useful (even though these might just be temporary controls until we allow for controlling margin/padding a bit better).
I can make the height not be bindable though - as elements in the same row always take up the same height anyway i guess it shouldn't be too useful...

Janpot: Some other options

```tsx
{
  flex: 'vertical' | 'horizontal' | 'both',
  flex: 'row' | 'column' | 'all',
  flex: 'x' | 'y' | 'xy'
}
```

Letting the imagination free, we could in the future expand that to `(props) => 'vertical' | 'horizontal' | 'both'`. e.g.

```tsx
{
  flex: ({ fullWidth }) => fullWidth ? 'vertical' : 'both'
}
```

apedroferreira: > Some other options
> 
> ```tsx
> {
>   flex: 'vertical' | 'horizontal' | 'both',
>   flex: 'row' | 'column' | 'all',
>   flex: 'x' | 'y' | 'xy'
> }
> ```
> 
> Letting the imagination free, we could in the future expand that to `(props) => 'vertical' | 'horizontal' | 'both'`. e.g.
> 
> ```tsx
> {
>   flex: ({ fullWidth }) => fullWidth ? 'vertical' : 'both'
> }
> ```

we can make it a function that takes the props, just not sure how useful it would really be to make this behavior change according to stateful values.
i wouldn't mind that a "horizontal alignment" control was enabled even if a component was taking up the full-width -  wouldn't be wrong, it just wouldn't be doing anything in that particular case.
but there's probably in no harm in making this more powerful in case we see the need for it down the line.

apedroferreira:
@Janpot how about this?

packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -437,6 +477,17 @@ function parseBindings(
...
         }
       }
+
+      if (!isPageLayoutComponent(elm)) {
+        for (const [propName, argType] of Object.entries(layoutBoxArgTypes)) {
+          const binding =
+            elm.layout?.[propName as keyof typeof layoutBoxArgTypes] ||
+            appDom.createConst(argType?.defaultValue ?? undefined);
+          const bindingId = `${elm.id}.layout.${propName}`;
+          const scopePath = `${elm.name}.@layout.${propName}`;
```
apedroferreira, line 487: i want to make sure this scope path can't collide with any normal prop name. any other suggestions?

apedroferreira:
merging, if there are any more suggestions i can add later.

packages/toolpad-app/src/runtime/ToolpadApp.tsx
Janpot: It shouldn't be in teh scope, you can remove the scope path

apedroferreira: ok let's make these not be bindable for now - i'll just have to set the default value in a different way if i remember correctly

PR #809, opened 2022-08-17
[website] Landing page

### LINKED ISSUE #477: [website] Landing page
Build a landing page in mui core repo, similar to MUI X

## Implementation

- [x] Add a redirection rule in https://github.com/mui/material-ui/blob/be24437157376588d5feb17e72dc369c15125669/docs/public/_redirects#L329 like for MUI X to redirect all the traffic of https://mui.com/toolpad to our docs https://github.com/mui/material-ui/pull/33287
- [x] Setup a `/docs` folder in this report on port 3003 (https://www.notion.so/mui-org/Ports-b113d0d4c76c4db9a0ef99c703d0084c)

### LINKED ISSUE #806: [website] Integrate Google Analytics
- Integrate Google Analytics to measure traction on the marketing website

bharatkashyap:
- Changed the theme to the finally selected version
- Closes #477 
- Closes #806 
- Also to merge https://github.com/mui/material-ui/pull/33937 
- Created a separate PR so that we can have the orange version live, for a live [comparison (blue)](https://deploy-preview-809--mui-toolpad-docs.netlify.app/toolpad/) with the [previous (orange)](https://deploy-preview-701--mui-toolpad-docs.netlify.app/toolpad/) version

Preview: https://deploy-preview-809--mui-toolpad-docs.netlify.app/toolpad/


prakhargupta1:
When adding email ID, the text box doesn't look good. Can we keep the original color ?

Screenshot 2022-08-19 at 12 06 00 PM

bharatkashyap:
> When adding email ID, the text box doesn't look good. Can we keep the original color ?
> 
> Screenshot 2022-08-19 at 12 06 00 PM

Is this happening when you select an email through autocomplete, or in all cases?

prakhargupta1:
> Is this happening when you select an email through autocomplete, or in all cases?

Just checked, only happening during autocomplete.

gerdadesign:
This is looking great so far! Just some small differences I'm noticing in design vs preview: 
| Design | Preview | Notes |
| ------------- | ------------- | ------------- |
| Figma Design sign up |  Preview sign up | I noticed the email signup in the bottom footer had 10px spacing between the input and the button. Just updated this (from 20px) in design! After seeing the live page, I'm also wondering if this CTA should be in pure white instead of grey in dark mode to give it a little more contrast? |
| Figma Design video  | Preview video  | The video in design has rounded corners with a border and is anchored to the bottom of the section below it. This is mimicking the design pattern of the [previews](https://mui.com/x/#data-grid-full) in other sections of the site. Perhaps this adjustment may help with the gap with the above section as well? |
| Use cases | Use Cases | From our [style guide](https://www.notion.so/mui-org/Style-guide-2a957a4168a54d47b14bae026d06a24b#d8e9cfb2423c4f83844a9f29d2032e3f) and the example on the [Templates page](https://mui.com/templates/), this should be lower case |
| Figma Design borders | Preview borders | The borders at the bottom of the rows right above the row header, should not be visible. The current [pricing](https://mui.com/pricing/) page does this. |
| no Commercial column background | Commercial column background | I had originally removed this because there were only 2 plans columns at the moment, so it seemed like it didn't need the additional separation indicator. I'm open to keeping the background though! In its center position on the Pricing page, the borders look fine, but they seem a little messy when the column is on the edge. Perhaps it may be a little cleaner without a border? |

bharatkashyap:
> | Design | Preview | Notes |
> | ------------- | ------------- | ------------- |
> | Figma Design sign up |  Preview sign up | I noticed the email signup in the bottom footer had 10px spacing between the input and the button. Just updated this (from 20px) in design! After seeing the live page, I'm also wondering if this CTA should be in pure white instead of grey in dark mode to give it a little more contrast? |
> | Use cases | Use Cases | From our [style guide](https://www.notion.so/mui-org/Style-guide-2a957a4168a54d47b14bae026d06a24b#d8e9cfb2423c4f83844a9f29d2032e3f) and the example on the [Templates page](https://mui.com/templates/), this should be lower case |
> | Figma Design borders | Preview borders | The borders at the bottom of the rows right above the row header, should not be visible. The current [pricing](https://mui.com/pricing/) page does this. |
> | no Commercial column background | Commercial column background | I had originally removed this because there were only 2 plans columns at the moment, so it seemed like it didn't need the additional separation indicator. I'm open to keeping the background though! In its center position on the Pricing page, the borders look fine, but they seem a little messy when the column is on the edge. Perhaps it may be a little cleaner without a border? |
I've addressed these, do check them out.

> | Design | Preview | Notes |
> | ------------- | ------------- | ------------- |
> | Figma Design video  | Preview video  | The video in design has rounded corners with a border and is anchored to the bottom of the section below it. This is mimicking the design pattern of the [previews](https://mui.com/x/#data-grid-full) in other sections of the site. |

Given the peculiarities of the HTML `video` element, I think the border of the video might have to be a part of the video itself - although I will look into making this work in the same way as the demos. 

> Perhaps this adjustment may help with the gap with the above section as well? 

I don't understand how anchoring the video to the bottom of the section below it would help in reducing the gap with the above section though.

gerdadesign:
This update looks great! 👏  Thank you 😀 

Only other question I have is in when looking on mobile, the Use cases cards and features extend to full width as expected, but they do not when I'm on desktop and pull in the width of the screen.
Screen Shot 2022-08-22 at 7 31 53 PM

docs/src/components/landing/Demo.js
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
+import Box from '@mui/material/Box';
+import dynamic from 'next/dynamic';
+
+const DynamicVideo = dynamic(() => import('./DemoVideo'), {
```
oliviertassinari, line 5: What's the use case for lazy loading this component? `DemoVideo` is mostly empty.

docs/src/components/landing/Hero.js
```diff
@@ -0,0 +1,62 @@
+import * as React from 'react';
+import Box from '@mui/material/Box';
+import Typography from '@mui/material/Typography';
+import IconImage from 'docs/src/components/icon/IconImage';
+import GradientText from './GradientText';
+import SignUp from './SignUp';
+import ToolpadHeroContainer from '../../layouts/ToolpadHeroContainer';
+import Demo from './Demo';
```
oliviertassinari, line 8: Since we are not using absolute import paths, I think that we should have file name as long as possible, otherwise, it will be a pain to find who is importing Demo (there could be others in the codebase)

```suggestion
import HeroDemo from './HeroDemo';
```

docs/src/components/landing/SignUp.js
```diff
@@ -0,0 +1,160 @@
...
+      ) : null}
+    </Form>
+  );
+};
+
+SignUp.propTypes = {
+  sx: PropTypes.object,
+};
+
+export default SignUp;
```
oliviertassinari, line 160: For the next components, I think that we will be better off writing them in TypeScript.

docs/src/components/landing/Hero.js
```diff
@@ -0,0 +1,62 @@
...
+
+export default function Hero() {
+  return (
+    <ToolpadHeroContainer>
+      <Box sx={{ minHeight: { xs: '0px', sm: '20px' } }} />
```
oliviertassinari, line 13: We should probably use the padding or the margin instead of creating a new DOM node

```suggestion
```

docs/src/components/landing/Hero.js
```diff
@@ -0,0 +1,62 @@
...
+              flexDirection: { xs: 'column', sm: 'row' },
+              justifyContent: 'center',
+              maxWidth: 'initial',
+            },
+          }}
+        />
+      </Box>
+      <Box sx={{ display: 'flex', flexDirection: 'row', justifyContent: 'center' }}>
+        <Demo />
+      </Box>
```
oliviertassinari, line 59: It looks strange on a small screen:

Screenshot 2022-08-23 at 13 06 38

docs/src/components/landing/Demo.js
```diff
@@ -0,0 +1,32 @@
...
+  suspense: true,
+});
+
+export default function HeroDemo() {
```
oliviertassinari, line 9: Something is off in the way we handle the intermediary loading state: it's 🎄. See the filmstrip: https://www.webpagetest.org/video/compare.php?tests=220823_AiDcK0_869-r:1-c:0

docs/src/components/landing/DemoVideo.js
```diff
@@ -0,0 +1,31 @@
...
+  return (
+    <Video
+      sx={{ height: { sm: 600 }, borderWidth: { xs: '10px 10px 0 10px', sm: '20px 20px 0 20px' } }}
+      playsInline
+      autoPlay
+      muted
+      loop
+      poster="/static/toolpad/marketing/hero-screenshot.svg"
+    >
+      <source src="/static/toolpad/marketing/hero-video.mp4" type="video/mp4" />
```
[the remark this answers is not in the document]
oliviertassinari: Also, the videos as strange borders, could we remove them? Here is the video, without any CSS borders

Screenshot 2022-08-23 at 13 17 10

oliviertassinari (commented):
We might want to merge this PR as is and deploy, it so we can be LIVE on https://mui.com/ as soon as possible, and continue to clean up the page in follow-up PRs.

Screenshot 2022-08-23 at 13 19 35

docs/src/components/landing/SignUp.js
```diff
@@ -0,0 +1,160 @@
...
+    <Form onSubmit={handleSubmit} sx={sx}>
+      <Box
+        sx={{
+          display: 'flex',
+          gap: 0,
+          width: { xs: '100%', sm: 'auto' },
+          maxWidth: 320,
+        }}
+      >
+        <InputBase
```
oliviertassinari, line 86: [WAVE](https://chrome.google.com/webstore/detail/wave-evaluation-tool/jbbplnpkjmmeebjpijfedlgcdilocofh) is reporting that we are missing a label

Screenshot 2022-08-23 at 13 22 59

docs/src/route.ts
```diff
@@ -1,5 +1,10 @@
 const ROUTES = {
-  toolpadDocs: '/toolpad/getting-started/setup',
+  TOOLPAD_DOCS: '/toolpad/getting-started/setup',
```
oliviertassinari, line 2: Why not match the convention in https://github.com/mui/material-ui/blob/1c9b514738037e81fcbc1b434191cab128eb64ed/docs/src/route.ts#L3?

```suggestion
  toolpadDocs: '/toolpad/getting-started/setup',
```

docs/src/route.ts
```diff
@@ -1,5 +1,10 @@
...
+  // https://docs.netlify.com/site-deploys/overview/#deploy-contexts
+  TOOLPAD_SIGN_UP_URL:
+    process.env.CONTEXT !== 'production' || process.env.NODE_ENV !== 'production'
```
oliviertassinari, line 5: Why not the same logic as https://github.com/mui/material-ui/blob/1c9b514738037e81fcbc1b434191cab128eb64ed/docs/src/config.ts#L3?

But then, we also have https://github.com/mui/material-ui/blob/1c9b514738037e81fcbc1b434191cab128eb64ed/docs/next.config.js#L197, why not even trash DEPLOY_CONTEXT to use STAGING?

docs/src/components/landing/Hero.js
Janpot: One of the reasons why I prefer to avoid custom aliasing, it makes it easy for all other developer tools (like vscode) to statically analyze the code and build a dependency tree of the code files:

https://user-images.githubusercontent.com/2109932/186153317-23a1de5a-43fb-4e1c-8b21-d8e101e49e6e.mov

no extra configuration needed, it just works out of the box

oliviertassinari: Interesting, I'm used to searching the whole filename, e.g. `import Link from 'docs/src/modules/components/Link';` in the codebase. First time I see this reference feature. I didn't know this exsited.

docs/src/components/landing/DemoVideo.js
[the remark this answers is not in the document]
bharatkashyap: Makes sense, I will replace with the compressed video

@gerdadesign On the video borders, is this a result of the software that's producing the video, or something that has been added when downloading/sharing?

docs/src/components/landing/Demo.js
bharatkashyap: This was intended to help in the performance by having a `fallback` image and then lazy load the `video`; it helps (marginally) when I compared the performance after removing the dynamic component: https://www.webpagetest.org/result/220824_AiDc00_3WF/

docs/src/route.ts
bharatkashyap: I feel that the `CONTEXT` variable is better suited to determining what the environment is ("production", "deploy-preview" or "branch-deploy") and it is documented clearly with this purpose as well: https://docs.netlify.com/configure-builds/environment-variables/#build-metadata, whereas the `PULL_REQUEST` variable is documented as being useful to check "whether the build is from a pull/merge request (true) or not (false).": https://docs.netlify.com/configure-builds/environment-variables/#git-metadata

docs/src/components/landing/Hero.js
bharatkashyap: I will change the names regardless, although this is a nifty feature that I use (sometimes) as well

docs/pages/toolpad/index.js
```diff
@@ -1,25 +1,58 @@
...
         card="/static/social-previews/toolpad-preview.jpg"
       />
+      <NoSsr>
+        <SignUpToast />
+      </NoSsr>
       <CssBaseline />
       <AppHeaderBanner />
       <AppHeader />
-      <main>
+      <main id="main-content">
```
Janpot, line 42: Is this `id` being used anywhere?

docs/src/components/landing/Demo.js
Janpot: Wouldn't this cause layout shift? Perhaps we can look into [lazy loading techniques](https://web.dev/lazy-loading-video/#video-gif-replacement) that preserve the markup?

docs/src/components/landing/DemoVideo.js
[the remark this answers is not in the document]
Janpot: It looks to me as it's part of the video, I don't see it when I try with other videos. The narrower I make the video the fainter the lines become, which is also a good indicator that it's in the video.

docs/pages/toolpad/index.js
oliviertassinari: It's used by the "Skip navigation".

Janpot: 👍  got it

docs/src/components/landing/Demo.js
oliviertassinari: I'm not even sure we need to look into lazy loading. The video is one of the first element that we want people to see.

Janpot: > Do we even need lazy loading? The video is one of the first element that we want people to see.

I can see the value if it improves the FCP, we can still autoplay it. Returning visitors may care less about the video than about the rest of the content.

bharatkashyap: Agreed; better initial loading performance without the video is perhaps meaningless. We can remove the `dynamic` . There is already a `poster` attribute on the `video` element, I think adding a default minimum width to it will also fix the 🎄 intermediary state problem.

oliviertassinari: The `poster` could help with video indexation: https://developers.google.com/search/docs/advanced/guidelines/video#provide-a-high-quality-thumbnail

docs/src/route.ts
oliviertassinari: We frequently use `CONTEXT="branch-deploy"` for hosting content in production. e.g. https://v4.mui.com/. So I would consider this as production. My main concern is that we don't have 10 different ways to determine production, it's too painful to debug down the line. I think that we should standardize on one way.

docs/src/components/landing/Demo.js
bharatkashyap: Adding a fixed width to the element fixes this: https://www.webpagetest.org/video/compare.php?tests=220824_BiDc9B_85X-r:2-c:0-e:filmstrip

docs/src/route.ts
bharatkashyap: Which way do you propose to be adopted as the standard?

oliviertassinari: > Which way do you propose to be adopted as the standard?

First, I would look at what we already use to determine production:

1. `process.env.NODE_ENV === 'production' && !process.env.PULL_REQUEST` [in config.js](https://github.com/mui/material-ui/blob/1c9b514738037e81fcbc1b434191cab128eb64ed/docs/src/config.ts#L3)
2. `process.env.NODE_ENV === 'production' && !process.env.PULL_REQUEST` [in document.js](https://github.com/mui/material-ui/blob/1c9b514738037e81fcbc1b434191cab128eb64ed/docs/pages/_document.js#L32)
3. `process.env.REPOSITORY_URL !== undefined && /mui-org\/material-ui-docs$/.test(process.env.REPOSITORY_URL);` [in next.js](https://github.com/mui/material-ui/blob/1c9b514738037e81fcbc1b434191cab128eb64ed/docs/next.config.js#L197)

I believe 3. is wrong, we should be using the environment variables to fill these links, and that's it:

Screenshot 2022-08-25 at 11 32 20

And that's it. So I think that we should pick 1. and 2.

oliviertassinari: Now, I think that the challenge is how do we make this a standard? I think that we should add a new variable inside the `env: {` path of the `next.config.js`. Maybe having an `DEPLOY_ENV` would make sense. Something that would replicate the [CONTEXT](https://docs.netlify.com/configure-builds/environment-variables/#build-metadata) but tailored for our use case. I would imagine 3 values: `production`, `staging`, `development`.

docs/src/components/landing/Hero.js
```diff
@@ -0,0 +1,73 @@
+import * as React from 'react';
+import Box from '@mui/material/Box';
+import Typography from '@mui/material/Typography';
+import IconImage from 'docs/src/components/icon/IconImage';
+import GradientText from './GradientText';
```
oliviertassinari, line 5: Why not import from?

```suggestion
import GradientText from 'docs/src/components/typography/GradientText';
```

The dark mode look different.

docs/src/components/landing/Marquee.js
```diff
@@ -0,0 +1,60 @@
...
+        py: { xs: 4, sm: 6, md: 8 },
+        display: 'flex',
+        flexDirection: 'column',
+        alignItems: 'center',
+        // TODO: Figure out why blueDark.700 does not work here
+        borderImage: `conic-gradient(#132F4C 0deg 360deg) fill 1/ /0 50vw`,
+      }}
+    >
+      <Typography
+        color={'grey.50'}
```
oliviertassinari, line 22:
```suggestion
        color="grey.50"
```

docs/src/content/features.js
```diff
@@ -0,0 +1,111 @@
...
+import LinkIcon from '@mui/icons-material/Link';
+import Link from '@mui/material/Link';
+import ViewModuleIcon from '@mui/icons-material/ViewModule';
+import ThumbUpOutlinedIcon from '@mui/icons-material/ThumbUpOutlined';
+import GppGoodIcon from '@mui/icons-material/GppGood';
+import FastForwardIcon from '@mui/icons-material/FastForward';
+import EditIcon from '@mui/icons-material/Edit';
+import AutoAwesomeIcon from '@mui/icons-material/AutoAwesome';
+import Typography from '@mui/material/Typography';
+import GradientText from '../components/landing/GradientText';
```
oliviertassinari, line 13:
```suggestion
import GradientText from 'docs/src/components/typography/GradientText';
```

docs/src/content/useCases.js
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react';
+import Typography from '@mui/material/Typography';
+import GradientText from '../components/landing/GradientText';
```
oliviertassinari, line 3:
```suggestion
import GradientText from 'docs/src/components/typography/GradientText';
```

docs/src/content/pricing.js
```diff
@@ -0,0 +1,267 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import Box from '@mui/material/Box';
+import Typography from '@mui/material/Typography';
+import Tooltip from '@mui/material/Tooltip';
+import InfoOutlinedIcon from '@mui/icons-material/InfoOutlined';
+import Link from 'docs/src/modules/components/Link';
+import IconImage from 'docs/src/components/icon/IconImage';
+import LaunchRounded from '@mui/icons-material/LaunchRounded';
+import GradientText from '../components/landing/GradientText';
```
oliviertassinari, line 10:
```suggestion
import GradientText from 'docs/src/components/typography/GradientText';
```

docs/src/components/landing/GradientText.ts
```diff
@@ -0,0 +1,10 @@
+import { styled } from '@mui/material/styles';
+
+const GradientText = styled('span')(({ sx }) => ({
+  background: `linear-gradient(315deg, #0059B3 0%, #007FFF 100%);`,
+  WebkitBackgroundClip: 'text',
+  WebkitTextFillColor: 'transparent',
+  sx,
+}));
+
+export default GradientText;
```
oliviertassinari, line 10:
```suggestion
```

docs/src/components/landing/DemoVideo.js
```diff
@@ -0,0 +1,37 @@
...
+  borderStyle: 'solid',
+  borderColor:
+    theme.palette.mode === 'dark' ? theme.palette.primaryDark[900] : theme.palette.grey[100],
+  borderRadius: '10px 10px 0 0 ',
+  overflow: 'hidden',
+  objectFit: 'cover',
+  objectPosition: 'top',
+  outline: `1px solid ${
+    theme.palette.mode === 'dark' ? theme.palette.primaryDark[400] : theme.palette.grey[300]
+  }`,
```
oliviertassinari, line 14: Note that Safari doesn't support `outline` border-radius. It looks like this:

Screenshot 2022-08-25 at 12 49 12

docs/src/components/landing/DemoVideo.js
```diff
@@ -0,0 +1,37 @@
...
+
+export default function DemoVideo() {
+  return (
+    <React.Fragment>
+      <Video
+        sx={{
+          height: { lg: 655 },
+          borderWidth: { xs: '10px 10px 0 10px', sm: '20px 20px 0 20px' },
+        }}
+        width={1400}
```
oliviertassinari, line 25: Do we need this width attribute?

```suggestion
```

If we do, the actual value with the largest viewport seems to be:

```suggestion
        width={1100}
```

docs/src/components/landing/DemoVideo.js
```diff
@@ -0,0 +1,37 @@
...
+    theme.palette.mode === 'dark' ? theme.palette.primaryDark[400] : theme.palette.grey[300]
+  }`,
+}));
+
+export default function DemoVideo() {
+  return (
+    <React.Fragment>
+      <Video
+        sx={{
+          height: { lg: 655 },
```
oliviertassinari, line 22: This leads to a truncated video:

Screenshot 2022-08-25 at 12 54 10

Without it:

```suggestion
```

Screenshot 2022-08-25 at 12 54 36

oliviertassinari: If it was added to avoid layout shift, I think that we could use this trick: https://stackoverflow.com/questions/600743/how-to-get-div-height-to-auto-adjust-to-background-size.

oliviertassinari (commented):
It looks good, I think that we can deploy this 👍

docs/src/route.ts
Janpot: In general, my recommendation would be to not designate "environments", but instead use "flags" with sensible defaults, and supply the right flags when you deploy. i.e. don't do

```tsx
if (process.env.DEPLOY_TARGET === 'pull-request') {
  doFoo()
  doBar()
}
```
but do
```tsx
if (process.env.DO_FOO) {
  (()
}
if (process.env.DO_BAR) {
  doBar()
}
```

This essentially decouples the definition of how an environment behaves from the actual code. Some of the advantages are that it's much faster to reconfigure and introducing a new environment doesn't require a code change.

`NODE_ENV` is a widely accepted standard to designate between production and not production, and can be kept for minification and enabling debugging features.

oliviertassinari: @Janpot Sure, we could add this abstraction. However, How would we define `DO_FOO`? I assume we need `DEPLOY_TARGET` beforehand.

Janpot: > However, how would we define DO_FOO? I assume this variable would be infered from DEPLOY_TARGET.

Yes, eventually the same thing is happening, somewhere there needs to be a translation between a target identifier, and the features it unlocks. The question is: where and when do we want this translation to take place. Is it in the code? Or external to the code? At build time? Or at deploy time?

docs/src/components/landing/DemoVideo.js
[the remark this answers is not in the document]
gerdadesign: Updated video to address the borders! 

| Full res (9.4 MB) | Compressed (867 KB) | 
| ---------------- | ---------------- | 
| https://user-images.githubusercontent.com/106991024/186701340-c37dc592-8c9e-4882-993a-6e7fcceeaa56.mp4 | https://user-images.githubusercontent.com/106991024/186701339-9f4798e7-7d57-47a9-a402-315db7edeb86.mp4 |

docs/src/components/landing/DemoVideo.js
bharatkashyap: Without the `height`, the video gets vertically truncated; I've tried to add a new approach with specific heights for all breakpoints which feels like it works the best across all viewport sizes

gerdadesign:
How do we feel about this update we discussed of moving the Upvote CTA banner into the features section? 

It breaks convention of the alternating background colors, which may make more sene when we add more sections. Until then, this break in convention may be a nice way to still group the overall features vs the pricing vs the hero. 

Screen Shot 2022-08-29 at 10 28 05 AM

bharatkashyap:
> How do we feel about this update we discussed of moving the Upvote CTA banner into the features section?
> 
> It breaks convention of the alternating background colors, which may make more sene when we add more sections. Until then, this break in convention may be a nice way to still group the overall features vs the pricing vs the hero.
> 
> Screen Shot 2022-08-29 at 10 28 05 AM

Looks good! Based on what we discussed today, I think it's safe to make this change in a follow-up PR along with some other changes to the cards (adding a "Learn more" link/Upvote link per card). 
We should be good to merge this from my side pending a review of the content from @prakhargupta1.

Issue #793, opened 2022-08-15
Add support for PostgreSQL data source


oliviertassinari:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Motivation 🔦

A continuation of https://github.com/mui/mui-toolpad/issues/36.

prakhargupta1:
Same as https://github.com/mui/mui-toolpad/issues/36

PR #811, opened 2022-08-17
Create postgres dataSource prototype

Janpot:
Allows for building postgres queries

* You can define parameters that you can pass from the page. postgres only support positional parameters `$1`, `$2`,... but this PR also rewrites the query to support interpollating the variable by name `$varname`
* It doesn't limit amount of returned rows yet (we should probably protect users against themselves?)
* editor has syntax highlighting but no autocomplete

This allows you to do some inception with the Toolpad database 🙂 

https://user-images.githubusercontent.com/2109932/185631917-592cf4b4-970f-4396-961d-aa941ad4eef6.mov

An exploration of https://github.com/mui/mui-toolpad/issues/793

closes https://github.com/mui/mui-toolpad/issues/793
closes https://github.com/mui/mui-toolpad/issues/833


apedroferreira:
really cool!
UX comments (i saw the video, can't see postgres as an option in the preview):
- how to i pass in the database url?
- autocomplete like you said would be cool, or maybe even some view that shows a list of tables/fields in the database?

but overall this seems almost like i would expect this feature to work, i can't think of anything important that's missing

apedroferreira (approved):

Janpot:
> * how to i pass in the database url?

Under when you create the connection you can edit the parameters. (Just realized I should add a "test connection" button there)

bytasv:
This looks really amazing, great job!

I was wondering where we pass around `$appName::TEXT` - why do we need the `::TEXT` part? Is it possible to get rid of that or it hints our app or postgres that something special needs to be done? 🤔 

I think this is a great POC, I assume the major thing for this to be completed would be to provide a way to connect to external DBs?

Janpot:
>  Is it possible to get rid of that or it hints our app or postgres that something special needs to be done?

This uses postgres prepared statements under the hood. We use prepared statements to protect against sql injection. Since postgres datatypes don't always map to javascript datatypes, it is not always possible for the driver to specify the datatype of a parameter in the prepared statement. In those cases postgres server tries to do a best effort when it can to infer the datatype when it parses the query, but it isn't able to in this specific instance since `CONCAT` can accept more datatypes than just strings. This is common with prepared statements. 

As an example: for the following query postgres is able to infer datatype correctly:
```SQL
SELECT * FROM "App"
WHERE name = $appName
```
It's because it conclusively knows the datatype of the column it's comparing the parameter against.

There are multiple ways to solve this:
* We can help postgres a hand by explicitly casting the parameter, like I did in the video.
* You could change the binding to `` `%${textField1.value}%` `` and the query to `SELECT * FROM "App" WHERE name LIKE $appName`.

If you google for "postgres could not determine data type of parameter" you will see that it's not uncommon for this to show up with prepared statements. e.g. in [retool forums](https://community.retool.com/t/how-to-use-a-state-variable-in-a-like-postgres-query/4446). 

> I assume the major thing for this to be completed would be to provide a way to connect to external DBs?

This can already connect to any db it has network access to, I didn't show the connection flow in this video, which you get when you create the postgres connection. It's self-explanatory, just provide host,user,password,.... I just created a connection for the internal db to test with as we can just treat it as any other postgres database.

Screenshot 2022-08-22 at 11 40 32

I'm finishing this up with a button to test the connection in that screen

bytasv:
> I'm finishing this up with a button to test the connection in that screen

That's really awesome, great job again! 🙇 

re: `$appName::TEXT`
Thanks for a thorough explanation!
At first I was thinking that maybe we could implicitly add `::TEXT` to the variable that we are passing, but I guess there are cases when we wouldn't be using `TEXT` and that could be something else? 🤔

Janpot:
> but I guess there are cases when we wouldn't be using TEXT and that could be something else?

exactly, it could be a number, boolean (perhaps a date or set or so, some day), postgres has multiple datatypes that can describe values that in JS only have one datatype (e.g. multiple numerical types, multiple date/time types). It can infer the type you meant in some cases (e.g. when you compare something to a column, which has an associated type), but not in other cases (e.g. when you pass a parameter to a polymorphic function, like `CONCAT`)

PR #814, opened 2022-08-17
Add property for child rectangle in DOM node info

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/712

Include an `innerRect` property with the exact position and dimensions of a component, excluding its box container.
This will be useful later for showing a better UI for resizing, for example, and closes this issue I had open along with the other PR.

### DIFF
--- a/packages/toolpad-app/src/pageViewState.ts
+++ b/packages/toolpad-app/src/pageViewState.ts
@@ -1,6 +1,7 @@
 import { FiberNode, Hook } from 'react-devtools-inline';
 import { NodeId, RUNTIME_PROP_NODE_ID, RUNTIME_PROP_SLOTS, SlotType } from '@mui/toolpad-core';
 import { NodeFiberHostProps } from '@mui/toolpad-core/runtime';
+import invariant from 'invariant';
 import { PageViewState, NodesInfo, NodeInfo, FlowDirection } from './types';
 import { getRelativeBoundingRect, getRelativeOuterRect } from './utils/geometry';
 
@@ -21,11 +22,16 @@ function getNodeViewInfo(
     const rect = getRelativeOuterRect(viewElm, elm);
     const props = fiber.child?.memoizedProps ?? {};
 
+    const innerElm = elm.firstElementChild;
+    invariant(innerElm, 'Node has no inner element');
+    const innerRect = getRelativeOuterRect(viewElm, innerElm);
+
     return {
       nodeId,
       error: fiberHostProps.nodeError,
       componentConfig: fiberHostProps.componentConfig,
       rect,
+      innerRect,
       slots: {},
       props,
     };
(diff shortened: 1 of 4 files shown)


packages/toolpad-app/src/pageViewState.ts
```diff
@@ -21,11 +22,16 @@ function getNodeViewInfo(
     const rect = getRelativeOuterRect(viewElm, elm);
     const props = fiber.child?.memoizedProps ?? {};
 
+    const innerElm = elm.firstElementChild;
```
Janpot, line 25: I'm a bit hesitant adding extra dom for measuring the contentBox. Can't we calculate it using the computed properties of `elm`?

apedroferreira: yeah, i originally thought about it like this but you're right, no need to do it in advance for every element as this rectangle will probably only be used when resizing

apedroferreira: i need access to the `element` though, so not sure how else to do this, i can give it a try

apedroferreira: sorry, not sure if i understood your point correctly at first - i added a `div` around each component just to cover the case where people add custom components that could be composed of multiple elements - a div would guarantee that any component inside the container is always wrapped by a single element

if it wasn't for those specific cases, nothing new would need to be added to the dom in this PR

other than that, if i didn't add any new elements to the DOM and just created this new property for each node, would doing things in a different way still be more efficient somehow?

Janpot: > a div would guarantee that any component inside the container is always wrapped by a single element

Shouldn't we add a box container around custom components?

apedroferreira: > > a div would guarantee that any component inside the container is always wrapped by a single element
> 
> Shouldn't we add a box container around custom components?

We do, but my idea for the inner div was for its bounding box rectangle to be possible to use for resizing the component inside the container later, for example, or to be able to show the margin between the edges of the container and the edges of the component if that becomes customizable later.

apedroferreira: We can keep this PR on hold until I possibly have to use it, it's not necessary yet anyway.

Janpot: > We do, but my idea for the inner div was for its bounding box rectangle to be possible to use for resizing the component inside the container later, for example, or to be able to show the margin between the edges of the container and the edges of the component if that becomes customizable later.

But that rectangle can also be calculated using a `getRelativeInnerRect` on `elm`, instead of `getRelativeOuterRect` of the child of `elm`

apedroferreira: > > We do, but my idea for the inner div was for its bounding box rectangle to be possible to use for resizing the component inside the container later, for example, or to be able to show the margin between the edges of the container and the edges of the component if that becomes customizable later.
> 
> But that rectangle can also be calculated using a `getRelativeInnerRect` on `elm`, instead of `getRelativeOuterRect` of the child of `elm`

ok i can give it a try, i was thinking with flexbox as we have it in container it wouldn't be possible, but even if it's not maybe we can switch to a box model with margin and padding instead of flexbox, once we add resizing

apedroferreira:
@Janpot how about this? as we don't use a box model for the containers it's not possible to get the inner rect just from margins/paddings, but from what i've tested this function should work for every possible case.

apedroferreira:
Also closing as no use for this for now, but if we ever see the need to get the smallest rectangle that wraps any element we can get this back.

PR #820, opened 2022-08-18
App renaming integration test

Janpot:
- [x] Integration test for https://github.com/mui/mui-toolpad/pull/816
- [x] Deduplicate logic around app menu + fix accessibility issues
- [x] Add db backup/restore around integration tests so your dev setup doesn't get polluted


test/integration/appRename.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  await createApp(page, appName1);
+
+  await page.goto('/');
+  await createApp(page, appName2);
+
+  await page.goto('/');
+
+  await page.click(`${locators.toolpadHomeAppRow(appName1)} >> [aria-label="Application menu"]`);
+
+  await page.click('[role="menuitem"]:has-text("Rename"):visible');
```
bharatkashyap, line 30: Should all the locators used in this test also be present in `locators.ts`?

packages/toolpad-app/src/toolpad/Home.tsx
```diff
@@ -264,64 +265,43 @@ function AppOpenButton({ app, activeDeployment }: AppOpenButtonProps) {
...
+      <IconButton {...buttonProps} aria-label="Application menu">
+        <MoreVertIcon />
+      </IconButton>
+      <Menu {...menuProps}>
+        <MenuItem onClick={handleRenameClick}>
+          <ListItemIcon>
+            <DriveFileRenameOutlineIcon />
+          </ListItemIcon>
+          <ListItemText>Rename</ListItemText>
+        </MenuItem>
```
bharatkashyap, line 296: Should we leave the accessibility issue link in there (assuming the issue persists of course)?

Janpot: I don't understand the comment. Isn't this how `MenuItem` is supposed to be used? If it's forcing us to do  a workaround in another part of the code then I think the comment belongs at that location.

bharatkashyap: This was intended to be a reference to the bug report which talks about how `MenuItem` doesn't work as a button when used with a keyboard; we can remove it since we aren't doing anything specifically to tackle it

PR #821, opened 2022-08-18
Add hook to ease menu implementation

Janpot:
- [ ] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/material-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).


apedroferreira (commented):
could these not be 2 separate reusable components instead?
edit: or maybe have the components coming from the hook already, anyway not a big deal deal this seems good

Janpot:
> could these not be 2 separate reusable components instead?

`'aria-controls'` and `'aria-labelledby'` are cross-referencing each other

apedroferreira:
yeah makes sense, first i was thinking those props could come from the outside but since it's always the same pattern a hook works well

apedroferreira (approved):

PR #823, opened 2022-08-18
Use prettier CLI + pretty-quick

Janpot:
Avoid too much drift with prettier CLI. Let's see how well this works for us (`pretty-quick` feels a bit faster)


bytasv (approved):

oliviertassinari:
If it works well, could we update the mui/material-ui? I think that this mono-repository should be our source of truth for tooling.

Janpot:
Giving it a try here: https://github.com/mui/material-ui/pull/34062

The only problem I ran into is the handling of translations: https://github.com/mui/material-ui/blob/df891841e991a4c5e84cd20b79352b18b04fba84/scripts/prettier.js#L64-L67
It seems like we don't enforce any formatting on files coming from crowdin. prettier finds 85 ranslated files with formatting problems. e.g. there's [the codeblock](https://github.com/mui/material-ui/blob/master/docs/data/material/customization/color/color-pt.md#exemplos) containing `purple. A200;`. It seems to me it would be better to enforce correct formatting in crowdin (maybe we can block merging these translations if prettier fails?).

PR #829, opened 2022-08-19
Tweak release instructions

bytasv:

I've added a few clarifications which weren't immediately obvious when I first did a release:
- How to pass GH token
- Link where `master` CI job can be found to check on docker build status
- Direct link to create a release
- Minor formatting for release making steps


CONTRIBUTING.md
```diff
@@ -81,7 +81,7 @@ If you would like to hack on MUI Toolpad or want to run the latest version, you
 1. Generate the changelog using:
 
    ```sh
-   yarn release:changelog
```
Janpot, line 84: I think we should let the developer decide whether they use it this way or use a variable in the `.env` file. Maybe instead of this, just add a note somewhere that you can add a `GITHUB_TOKEN` variable in the `.env` file

bytasv: Thanks, didn't think about that. Added a note and will set it up for my self for next time 🙇

Janpot: I prefer to leave `--githubToken YOUR_TOKEN` off the command here as we would be using this guide as a script as well and might want to copy paste from here

apedroferreira (approved):
lgtm other than Jan's suggestions!

CONTRIBUTING.md
bytasv: Sorry I'm not sure I follow, could you please clarify what you mean? I would assume that whether you copy entire command with `--githubToken` or not is up to you and whether this extra piece is in the code or not doesn't change much?

Janpot: I want to nudge people to use

```sh
yarn release:changelog
```

instead of

```sh
yarn release:changelog--githubToken YOUR_TOKEN
```

Janpot (approved):

Issue #758, opened 2022-08-08
Add integration test for the application editor


Janpot:
Test basic functionality in a playwright test:

- Can we test drag & drop? (In a way that doesn't regress constantly)
- selection and default property values https://github.com/mui/mui-toolpad/issues/755
- deleting elements
- updating property value
- ...

oliviertassinari:
> Can we test drag & drop?

MUI X has a few tests about this https://github.com/mui/mui-x/blob/249969c611a2d89ff895bdb639551b6dae0d5e3d/test/e2e/index.test.ts#L175

PR #831, opened 2022-08-19
Editor integration tests

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/758

Refactored to use Playright's Page Object Model as we intended.


packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -307,7 +307,7 @@ function PageRoot({ children }: PageRootProps) {
   return (
     <Container>
       <Stack
-        data-testid="page-root"
+        data-test-id="page-root"
```
apedroferreira, line 310: this one really tripped me up :D

Janpot: The reason it's called that way is so we can also use it for `getByTestId`: https://testing-library.com/docs/queries/bytestid/#api

apedroferreira: yeah, there another in another file that was `data-test-id` so i was trying that first, i guess we can use `data-testid` everywhere

apedroferreira:
Looks like we have to use the overlay element to capture clicks/drags, as it's the one that captures them. It doesn't work with other elements in the same position.
Also I don't plan to add lots of test ids but I think that for these specific elements it was the best choice.

Let me know if you think anything should be different in terms of how i reorganized the files and their structure a bit too!

apedroferreira:
Will add a couple of tests for prop controls in a separate PR now too.

test/integration/editor.spec.ts
```diff
@@ -0,0 +1,154 @@
+import { test, expect, Page, Locator } from '@playwright/test';
+import createApp from '../utils/createApp';
+import generateId from '../utils/generateId';
+import { canvasFrame, componentCatalog, pageRoot, pageOverlay } from '../utils/locators';
+import twoFieldsDomTemplate from '../domTemplates/twoTextFields.json';
```
Janpot, line 5: I prefer not to organize this using a big `/domTemplates/...` folder containing doms for all the tests, that doesn't scale well. The json should just live right next to the test. Can we rather do something like the following?

```
test
  integration
    appRename
      index.spec.ts
    editor
      index.spec.ts
      input.json
    smoke
      index.spec.ts
```

test/integration/editor.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  const canvasPageOverlayBoundingBox = await canvasPageOverlayLocator.boundingBox();
+
+  const secondTextFieldLocator = canvasInputLocator.nth(1);
+  const secondTextFieldBoundingBox = await secondTextFieldLocator.boundingBox();
+
+  expect(canvasPageOverlayBoundingBox).toBeDefined();
+  expect(secondTextFieldBoundingBox).toBeDefined();
+
+  await canvasMoveElementHandleLocator.dragTo(canvasPageOverlayLocator, {
```
Janpot, line 101: Can we use mouse events to simulate drag and drop instead? Did you try the strategy from https://github.com/mui/mui-toolpad/issues/758#issuecomment-1207914180 ?

test/integration/editor.spec.ts
apedroferreira: sure, that sounds fine too

test/integration/editor.spec.ts
apedroferreira: Probably can do done like those, I just thought `dragTo` would be better since it encapsulates the whole drag interaction?
Also I can add a comment here explaining that it's dragging to the right side of the other element.

Janpot: It seems to me that the method as used in the core is much closer to real user interaction, and less dependant on implementation details

apedroferreira: ok, ill also add a method about getting an (x,y) position relative to the overlay, it should make this kind of code more readable.

Janpot: Best would be to avoid the overlay whenever we can and rather detect/calculate based these things on the underlying app running in the iframe. The overlay is an implementation detail, we may want to change its implementation in the future. In fact I'd say it's very likely we will heavily change this implementation at some point.

apedroferreira: Yeah, I agree, the problem is that the overlay is handling all the events, I tried with other elements in the same position but it did not work. I'd be glad to do it in an alternative way but would have to look more into what else is possible.

apedroferreira: alright i hadn't realized we could use the Playwright's page object and do everything relative to it - that means i don't have to use the overlay after all!
thanks for the feedback, it was super helpful, this should be better now, lmk if you see anything else that can be improved

.circleci/config.yml
```diff
@@ -170,7 +170,7 @@ jobs:
       - run:
           command: |
             # See https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-shell-command
-            TAG=$CIRCLE_SHA1 docker-compose -f ./docker/compose/docker-compose.yml up -d
+            TAG=$CIRCLE_SHA1 docker-compose -f ./docker/compose/docker-compose.test.yml up -d
```
Janpot, line 173: perhaps we should move this top-level, next to docker-compose.test.yml. We could reserve `./docker/compose/` for public examples?

test/integration/editor/index.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+  await page.mouse.up();
+}
+
+test('can place new components from catalog', async ({ page }) => {
+  const appId = generateId();
+
+  await page.goto('/');
+  await createApp(page, `App ${appId}`);
+
+  const canvasFrameLocator = page.frameLocator(canvasFrame);
```
Janpot, line 35: For the future we could think of an [Object model](https://playwright.dev/docs/test-pom) for this. Been trying that out in https://github.com/mui/mui-toolpad/pull/838
Doesn't have to be change for this PR

test/utils/createApp.ts
```diff
@@ -0,0 +1,15 @@
+import { Page } from '@playwright/test';
+
+export default async function createApp(page: Page, name: string, dom?: string) {
+  await page.locator('button:has-text("create new")').click();
+
+  await page.fill('[role="dialog"] label:has-text("name")', name);
```
Janpot, line 6: We could improve dialog selectors a bit: https://github.com/mui/mui-toolpad/pull/838/files#diff-e537b99227068da9665d973e6573a1cd6da8d36be778d2aba43b159890a43d99R30

But we can merge this as is and use the ToolpadHome model as an optimization later

Janpot (approved):
I'd move the compose file top-level, but other than that we can merge this 👍

apedroferreira:
> I'd move the compose file top-level, but other than that we can merge this 👍
> 
> _edit_ 🤔 wait, the test doesn't pass?

yeah, the CI with the env var isn't working so something must be wrong there, and there's a separate issue with the Firefox tests that include drag & drop not passing

test/integration/editor/index.spec.ts
apedroferreira: didn't know about it, yeah we could use to abstract the things we end up reusing a lot
i might still include it in this PR when we take a look at the firefox stuff again, shouldn't be too time consuming

test/utils/createApp.ts
apedroferreira: sure, i just took this function from the other tests where it was
we can use that model if we merge it first

docker/compose/docker-compose.test.yml
```diff
@@ -0,0 +1,30 @@
+version: '3.7'
```
Janpot, line 1: This one is not needed anymore I think

test/integration/propControls/index.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    .elementHandle();
+  const propControlLabelFor = await propControlLabelHandle?.getAttribute('for');
+
+  return selectedNodeEditorLocator.locator(`input[id="${propControlLabelFor}"]`);
+}
+
+async function getInputElementLabelLocator(page: Page, inputLocator: Locator) {
+  const canvasFrameLocator = page.frameLocator(canvasFrame);
+
+  const inputHandle = await inputLocator.elementHandle();
```
Janpot, line 22: One potential way of avoiding relying on these implementation details is to add `Typography` elements in the same row and [layout](https://playwright.dev/docs/selectors#selecting-elements-based-on-layout) selectors to find the associated input, regardless of the label

test/utils/locators.ts
```diff
@@ -1,3 +1,9 @@
 export function toolpadHomeAppRow(appName: string): string {
```
Janpot, line 1: We should get rid of this file and move the locators inside of the `ToolpadHome` and `ToolpadEditor` models

docker/compose/docker-compose.test.yml
apedroferreira: yeah i didn't clean this up yet, just rebased

test/integration/propControls/index.spec.ts
apedroferreira: i can do that, i though using a label's `for` and the input `id` was actually better

test/utils/locators.ts
apedroferreira: yeah im gonna use those new objects, just havent done it yet

test/integration/propControls/index.spec.ts
apedroferreira: this is using accessibility props that should always match so i think it's a good approach (nothing implementation specific imo), we can change it if you insist though

apedroferreira: but the label should always match the id, right? that's how html inputs should work.

apedroferreira: or if it's an issue that we are kind of assuming the MUI inputs and labels have accessibility properly set up we can do it in another way - but it feels safe to assume that those accessibility props work.

Janpot: Yeah, I was thinking about the edge-cases
* can an input have an `aria-label`?
* if the `id` changes over the lifetime of the page, this locator will stop working as it only considers the id of the first time it executed

but that may be too far fetched

apedroferreira: i'll give locators another quick try, it's a more readable approach at least - for some reason i had some issues when i first tried due to the iframe, but now i understand things a bit better so maybe it's easy

Janpot: Yes, I haven't fully internalized them yet. The way I understand them is that they are like `document.queryAll`, only it's sort of a live collection. So if the underlying DOM changes, the content of the locators change with it. I feel like I haven't fully harnassed their power yet, it seems to sort of require you to think differently about selectors.

apedroferreira: layout selectors aren't working, maybe it's because the label intersects with the input, they don't seem very predictable at all... let's probably keep this approach for now, i don't think it should ever cause any issues.

apedroferreira:
@Janpot 
CI passed 3 times in a row, maybe it's ok to merge this now?
i had to dispatch events manually in Firefox, couldn't find a better way.
If these tests start being flaky later though we can skip them temporarily and i can work on them a bit more - locally they always pass for me but CI seems a bit more unpredictable.

Janpot:
Is there a ticket in the playwright repo that we can link to that describes the faulty behavior in firefox? If not, have you tried creating a minimal reproduction? Perhaps you can open a ticket and link to it?

I don't like to add hacks for bugs without adding documentation that can tell us more about the status of said bug. (example: https://github.com/mui/mui-toolpad/blob/4f71c00b28c8e57d9bad5ff99ede85f1d1aecc45/packages/toolpad-app/src/components/SplitPane.tsx#L58)

apedroferreira:
> Is there a ticket in the playwright repo that we can link to that describes the faulty behavior in firefox? If not, have you tried creating a minimal reproduction? Perhaps you can open a ticket and link to it?
> 
> I don't like to add hacks for bugs without adding documentation that can tell us more about the status of said bug. (example:
> 
> https://github.com/mui/mui-toolpad/blob/4f71c00b28c8e57d9bad5ff99ede85f1d1aecc45/packages/toolpad-app/src/components/SplitPane.tsx#L58
> 
> )
> When there's a proper way forward towards solving the firefox issue in the future, then I'm fine merging this.

i couldn't find a proper related issue in the Playwright repo, and yeah i thought about create one myself, just takes a bit of work to create a reproduction. i'm working on it, there's probably no rush to merge this so we can wait

apedroferreira:
> Is there a ticket in the playwright repo that we can link to that describes the faulty behavior in firefox? If not, have you tried creating a minimal reproduction? Perhaps you can open a ticket and link to it?
> 
> I don't like to add hacks for bugs without adding documentation that can tell us more about the status of said bug. (example:
> 
> https://github.com/mui/mui-toolpad/blob/4f71c00b28c8e57d9bad5ff99ede85f1d1aecc45/packages/toolpad-app/src/components/SplitPane.tsx#L58
> 
> )
> When there's a proper way forward towards solving the firefox issue in the future, then I'm fine merging this.

I've managed to reproduce the bug in a minimal reproduction:
https://github.com/microsoft/playwright/issues/17441

So it's confirmed, seems like there's a specific issue with triggering HTML5 drag & drop events in iFrames, in Firefox only.

I've included a link to the issue I created now, so will merge this PR if the tests seem stable in CI.

Janpot:
great 👌

Janpot:
@apedroferreira Following up on the playwright issue, it seems that it's closed due to the closing of https://github.com/microsoft/playwright/issues/17153 (would have been nice of them to reference that issue). Does calling `hover()` twice fix the issue?

apedroferreira:
> @apedroferreira Following up on the playwright issue, it seems that it's closed due to the closing of [microsoft/playwright#17153](https://github.com/microsoft/playwright/issues/17153) (would have been nice of them to reference that issue). Does calling `hover()` twice fix the issue?

looks like they're supposed to have fixed it, i'll try removing the Firefox specific logic and see if things work now

apedroferreira:
or i guess they might not have fixed it after all?
i'll try things from that issue then, maybe using `hover` twice works

PR #836, opened 2022-08-22
Nudge PRs to have linked issues, videos

bharatkashyap:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,7 @@
 <!-- Thanks so much for your PR, your contribution is appreciated! ❤️ -->
 
-- [ ] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/material-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).
+- Fixes <!-- Issue hyperlink -->
+
+- Visual demonstration <!-- Screenshot/Video link -->
+
+- Please look at the [PR section of the contributing guide](https://github.com/mui/material-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request) and also MUI's broad [guidelines](https://www.notion.so/mui-org/GitHub-PRs-7112d03a6c4346168090b29a970c0154#258c9e963dc445348562a52a62052941) on Pull Requests.


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,3 +1,7 @@
 <!-- Thanks so much for your PR, your contribution is appreciated! ❤️ -->
 
-- [ ] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/material-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).
+- Fixes <!-- Issue hyperlink -->
```
bytasv, line 3: Should we add checkboxes? We could later introduce a PR check which validates that all boxes are checked.

There were concerns about increased friction, but IMO adding a `x` between `[ ]` should not be such a huge friction

bytasv (approved):

.github/PULL_REQUEST_TEMPLATE.md
bharatkashyap: I feel like adding the checkbox adds very little value, since I would rather just add the image instead of checking a box

bytasv: I brought it up only because it allows us to use PR checks to validate if action was done, however it's fine if we skip that for now since we don't have checks anyways 👍

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,3 +1,7 @@
...
+
+- Visual demonstration <!-- Screenshot/Video link -->
+
+- Please look at the [PR section of the contributing guide](https://github.com/mui/material-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request) and also MUI's broad [guidelines](https://www.notion.so/mui-org/GitHub-PRs-7112d03a6c4346168090b29a970c0154#258c9e963dc445348562a52a62052941) on Pull Requests.
```
oliviertassinari, line 7: I'm not a fan of having multiple links here because it's something that we will need to keep up to date. At the same time, I believe that we need the same information in https://github.com/mui/mui-toolpad/blob/master/CONTRIBUTING.md.

So to solve this problem, I would propose to link the relevant section in https://github.com/mui/mui-toolpad/blob/master/CONTRIBUTING.md, that could in turn, defer to the notion and material ui docs.

.github/PULL_REQUEST_TEMPLATE.md
oliviertassinari, line 3: Maybe 

```suggestion
- Closes <!-- Issue hyperlink -->
```

would be more generic. Fixing a feature request feels a bit off.

Janpot (approved):

Issue #414, opened 2022-05-13
Add default value support to components


oliviertassinari:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Add the ability to pick the default value for a component

### Motivation 🔦

I would like the license key generator to default to a subscription mode. I'm using the Select component, in which it's impossible to implement.

<img width="260" alt="Screenshot 2022-05-14 at 01 25 38" src="https://user-images.githubusercontent.com/3165635/168401005-c8cde8e0-9f65-4a3d-9cb3-d75cf5e62921.png">

Janpot:
It's currently possible (as a test I've set your license model to "perpetual" by default), but the UI currently doesn't update with it. The reason is that default values are set on the first render only. I'd consider this a bug.

oliviertassinari:
I have found a new use case for that defaultValue with a TextField. I would like it to be `new Date(new Date().getTime() + 1000 * 366 * 60 * 60 * 24).getTime()`.

oliviertassinari:
I have landed again on this problem. It behaves strangely. Here is a different angle:

1. Add a `TextField` on the page
2. Set the value to `Foo`
3. 💥 the value is not visible on the text field
4. 💥 same in the binding view

<img width="221" alt="Screenshot 2022-08-06 at 19 28 13" src="https://user-images.githubusercontent.com/3165635/183259565-f2c2babf-54f2-4ecd-b422-b2a3851e5f70.png">

5. Reload: ✅

Janpot:
> It behaves strangely.

The reason it behaves like this is that default value is only used during the first render. When the default value is changed while the component is already being rendered, it doesn't have an effect. But I agree it's strange behavior, we should probably make an exception to it for when the value is changed by the user in the editor.

There may also be a benefit to having a separate `defaultValue` property.

oliviertassinari:
> The reason it behaves like this is that default value is only used during the first render.

+1 to try to experiment with remounting the component when the `value` changes. Effectively, to perform what a browser reload would do. If it's not enough, then `defaultValue`, might be needed.

oliviertassinari:
I have found a new use case. I would like in https://master--toolpad.mui.com/_toolpad/app/cl4hla83p01949xoizo5uxf2a/editor/pages/cl4hla87800029xoihl7tf4qz to be able to use a page parameter.

[...]

PR #838, opened 2022-08-22
Support bindable default value for inputs

Janpot:
split out default value in a separate prop and make it bindable to e.g. page parameters

Fixes https://github.com/mui/mui-toolpad/issues/414


Janpot:
@apedroferreira Trying out applying the methods from https://playwright.dev/docs/test-pom here

test/integration/runtime-basic/basicDom.json
```diff
@@ -0,0 +1,176 @@
+{
```
apedroferreira, line 1: in my PR i gave the file a more explicit name like `twoTextFields`, maybe it helps understand better the contents of the DOM without having to look at the JSON content itself.

test/integration/runtime-basic/index.spec.ts
```diff
@@ -0,0 +1,31 @@
+import { ToolpadHome } from '../../models/ToolpadHome';
+import { test, expect } from '../../playwright/test';
+import basicDom from './basicDom.json';
+
+const PAGE_ID = 'vlpwdwr';
```
apedroferreira, line 5: would it be worth getting this directly from the DOM json?

apedroferreira (approved):
seems good! just added some comments about testing
lmk when you get the runtime env variable running in CI

apedroferreira:
One more comment about UX: is it intentional that the prop control is completely disable on the right? I thought we wanted to have two-way binding, but not sure how easy or doable it is for now.

packages/toolpad-app/src/config.ts
```diff
@@ -65,11 +63,11 @@ function getBrowsersideRuntimeConfig(): RuntimeConfig {
   return maybeRuntimeConfig;
 }
 
-export const runtimeConfig: RuntimeConfig =
+const runtimeConfig: RuntimeConfig =
   typeof window === 'undefined'
     ? {
         // Define runtime config here
-        foo: process.env.TOOLPAD_FOO,
+        integrationTest: !!process.env.TOOLPAD_INTEGRATION_TEST,
```
apedroferreira, line 70: was something like this what i was missing for CI to work?

Janpot: We were using built-time configuration for this, and we build the docker image for production, so it was set to false

PR #839, opened 2022-08-22
[app] Add an update banner

### LINKED ISSUE #734: Show a banner to notify about the release of a new version
### Duplicates

### Latest version

### Summary 💡

On their self-hosted instances, users should be notified that the new version is out.
They should also be able to:

1. close the banner (this would make it less intrusive)
2. opt-out from the service
3. View change log
4. Update now CTA

Benchmark
Screenshot 2022-08-04 at 3 56 30 PM

### Examples 🌈

_No response_

### Motivation 🔦

_No response_

bharatkashyap:
- Fixes #734 

- Visual Demonstration
<video src="https://user-images.githubusercontent.com/19550456/189048549-13350ffa-0ddd-4517-ad31-f429ead0778a.mov"></video>

In case of a long list of apps:
Screenshot 2022-09-08 at 11 44 46 AM

- The close button hides the banner for all instances of the current version of toolpad, so it is in between a full opt-out and a temporary hide. Providing an option to opt-out felt overkill, but showing the banner on every refresh is also unnecessary, so this seems to me the most appropriate middle path.


packages/toolpad-app/src/toolpad/ToolpadShell.tsx
```diff
@@ -78,6 +86,81 @@ function UserFeedback() {
   );
 }
 
+function UpdateBanner() {
+  const [latestVersion, setLatestVersion] = React.useState(CURRENT_RELEASE_VERSION);
+  const [dismissedBanner, setDismissedBanner] = useLocalStorageState<boolean>(
+    `${CURRENT_RELEASE_VERSION}-update-banner-dismissed`,
```
Janpot, line 92: Hadn't thought about this problem yet, but this will keep pushing unused items in the local storage indefinitely

packages/toolpad-app/src/toolpad/ToolpadShell.tsx
```diff
@@ -78,6 +86,81 @@ function UserFeedback() {
...
+              component="a"
+              sx={{ mr: 1 }}
+              target="_blank"
+              href={DOCUMENTATION_URL}
+            >
+              Update
+            </Button>
+            <Button
+              aria-label="close"
+              size="small"
```
Janpot, line 130: The theme should take care of the sizing already

packages/toolpad-app/src/toolpad/ToolpadShell.tsx
```diff
@@ -78,6 +86,81 @@ function UserFeedback() {
...
+    fetchLatestRelease();
+  }, []);
+
+  return (
+    <Collapse in={showBanner}>
+      <Alert
+        action={
+          <React.Fragment>
+            <Button
+              aria-label="close"
```
Janpot, line 117: these labels don't seem right

packages/toolpad-app/src/toolpad/ToolpadShell.tsx
```diff
@@ -78,6 +86,81 @@ function UserFeedback() {
...
+              size="small"
+              color="inherit"
+              endIcon={<OpenInNewIcon fontSize="inherit" />}
+              component="a"
+              sx={{ mr: 1 }}
```
Janpot, line 122: Is it possible to use `gap` on the parent instead?

packages/toolpad-app/src/toolpad/ToolpadShell.tsx
```diff
@@ -13,14 +16,19 @@ import {
   Tooltip,
 } from '@mui/material';
 import HomeIcon from '@mui/icons-material/Home';
+import CloseIcon from '@mui/icons-material/Close';
 import HelpOutlinedIcon from '@mui/icons-material/HelpOutlined';
 import OpenInNewIcon from '@mui/icons-material/OpenInNew';
 import useMenu from '../utils/useMenu';
+import useLocalStorageState from '../utils/useLocalStorageState';
 
 const DOCUMENTATION_URL = 'https://mui.com/toolpad/getting-started/setup/';
```
apedroferreira, line 25: do we have specific instructions on how to update besides the setup instructions? should we?

bharatkashyap: We could create a dialog within the app containing the instructions instead of redirecting?

Janpot: I think we can start with a link to the documentation. There's not much to be done from within the app to update Toolpad.

packages/toolpad-app/src/server/data.ts
```diff
@@ -455,3 +461,43 @@ export async function loadDom(appId: string, version: VersionOrPreview = 'previe
 export async function loadRenderTree(appId: string, version: VersionOrPreview = 'preview') {
   return appDom.createRenderTree(await loadDom(appId, version));
 }
+
+const LATEST_RELEASE_API_URL = 'https://api.github.com/repos/mui/mui-toolpad/releases/latest';
```
Janpot, line 465: Let's create a separate file for this `updateCheck`. And let's move everything that's below here inside of it?

packages/toolpad-app/src/server/data.ts
```diff
@@ -455,3 +461,43 @@ export async function loadDom(appId: string, version: VersionOrPreview = 'previe
...
+
+const latestReleaseCache: GithubReleaseCache = {
```
Janpot, line 467: No need to create an object, you can just create two mutable variables

packages/toolpad-app/src/types.ts
```diff
@@ -165,3 +165,13 @@ export interface AppTheme {
 }
 
 export type VersionOrPreview = 'preview' | number;
+
+export interface GithubRelease {
+  tag_name: string;
+  html_url: string;
+}
+
+export interface GithubReleaseCache {
```
Janpot, line 174: I don't believe there's a need to have types like this shareable. Interfaces that are local to an implementation can stay there. (In this case, it's probably not even needed)

packages/toolpad-app/src/server/data.ts
bharatkashyap: Makes sense. 
`getLatestRelease` as a filename feels clearer to me. What do you think?

packages/toolpad-app/src/server/data.ts
```diff
@@ -455,3 +461,43 @@ export async function loadDom(appId: string, version: VersionOrPreview = 'previe
...
+
+  // Abort the request after 30 seconds
+  const controller = new AbortController();
+  const timeout = setTimeout(() => controller.abort(), 30_000);
+
+  /* Fetch the latest release from the Github API
+   * https://developer.github.com/v3/repos/releases/#get-the-latest-release
+   */
+
+  const response = await fetch(LATEST_RELEASE_API_URL, {
```
Janpot, line 486: This won't work, while this promise is being awaited, other requests may come in, and the `nextFetchAllowedAt` has not updated yet. Resulting in multiple fetches. Instead, you need to cache the promise. For example:

```tsx
async function fetchRelease() {
  const response = await fetch(LATEST_RELEASE_API_URL, {
    // @ts-expect-error See https://github.com/microsoft/TypeScript/issues/48003
    signal: AbortSignal.timeout(30000),
  });

  if (response.ok) {
    const { tag_name: tag, html_url: url } = await response.json();
    return { tag, url };
  }

  throw new Error(`Failed to fetch latest release from Github API: ${response.status}`);
}

let nextFetchAllowedAt = 0;
let cached;

export async function getLatestToolpadRelease(): Promise<GithubRelease> {
  const now = Date.now();
  if (nextFetchAllowedAt < now || !cached) {
    cached = Promise.resolve().then(async () => {
      try {
        return await fetchRelease();
      } catch (error) {
        // retry in 30 seconds
        nextFetchAllowedAt = Date.now() + 1000 * 30;
        throw error;
      }
    });

    nextFetchAllowedAt = now + 1000 * 60 * 10;
  }

  return cached;
}
```

Janpot (approved):

PR #840, opened 2022-08-22
Add prop controls integration test

apedroferreira:
Add integration test for prop controls.
The test is failing at the end because of an actual bug - changing an input's value in the prop controls does not reflect the changes in the DOM - so need to fix that for the tests to pass.


Janpot:
As it currently stands, the input value is a controlled property, and setting a value actually means setting the default value of the prop. That's why it doesn't update as it already has been initialized. I'm [working on improving the behavior](https://github.com/mui/mui-toolpad/pull/838) around values and default values. I'd suggest that you test another property than value for now. I'll be adding integration tests around value/defaultValue as part of my changes.

Also see https://github.com/mui/mui-toolpad/issues/414

apedroferreira:
> As it currently stands, the input value is a controlled property, and setting a value on those actually means setting the default value of the prop. That's why it doesn't update as its value already has been initialized. I'm [working on improving the behavior](https://github.com/mui/mui-toolpad/pull/838) around values and default values. I'd suggest that you test another property than value for now. I'll be adding integration tests around value/defaultValue as part of my changes.
> 
> Also see #414

Ok sounds good! I was using value and text inputs because they're so easy to read from the DOM, i can use something else.

PR #842, opened 2022-08-23
Fix missing key warning

Janpot:
I added some testing infra that catches these warnings as well. It also caught a warning on the `Home` for `Text nodes cannot appear as a child of <tbody>.` which I also fix in this PR. cc @bharatkashyap

### DIFF
--- a/packages/toolpad-app/src/toolpad/Home.tsx
+++ b/packages/toolpad-app/src/toolpad/Home.tsx
@@ -45,6 +45,7 @@ import getReadableDuration from '../utils/readableDuration';
 import EditableText from '../components/EditableText';
 import type { AppMeta } from '../server/data';
 import useMenu from '../utils/useMenu';
+import ErrorAlert from './AppEditor/PageEditor/ErrorAlert';
 
 export interface CreateAppDialogProps {
   open: boolean;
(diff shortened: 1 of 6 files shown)


apedroferreira:
> Should we add a check for certain warnings in the integration tests?

Shouldn't eslint catch that kind of stuff if we add the proper rules?

Janpot:
> Shouldn't eslint catch that kind of stuff if we add the proper rules?

It seems to miss this specific case. It is enabled for the file though, I can trigger it using another pattern (`const x = [<div />, <div />]`).

An integration test could catch these and possible more React warnings

apedroferreira:
I've never used integration tests for code related issues but if you think it's a valid approach I would be curious to see how it would work :) I guess it could be useful to catch the issues that eslint can't.

Janpot:
> I would be curious to see how it would work

We can capture console messages in playwright and throw when we detect certain patterns

apedroferreira:
> > I would be curious to see how it would work
> 
> We can capture console messages in playwright and throw when we detect certain patterns

ok that's what i was thinking, sounds good!

packages/toolpad-app/src/toolpad/Home.tsx
```diff
@@ -463,33 +464,32 @@ function AppsListView({
...
-                        onDelete={() => setDeletedApp(app)}
-                      />
-                    );
-                  })
-                : 'No apps yet';
-            default:
-              return '';
-          }
-        })()}
+        {status === 'loading' ? (
```
apedroferreira, line 478: just nitpicking but maybe it's slightly easier to understand if this loading check is done above like the error check, and then here we use the ternary operator to check if there are no apps or more?

test/integration/smoke.spec.ts
```diff
@@ -1,8 +1,8 @@
-import { test, expect, Request } from '@playwright/test';
+import { test, expect, Request } from '../playwright/test';
 import generateId from '../utils/generateId';
 import * as locators from '../utils/locators';
 
-test('basic app creation flow', async ({ page }) => {
+test.only('basic app creation flow', async ({ page }) => {
```
apedroferreira, line 5: we should have an eslint rule for these if we don't, i missed it before in my PRs too

packages/toolpad-app/src/toolpad/Home.tsx
Janpot: I refactored it a bit for readability

test/integration/smoke.spec.ts
Janpot: I added the `--forbid-only` flag in CI

apedroferreira (approved):

PR #851, opened 2022-08-25
Fix missing jsx key warning

Janpot:
Fix missing key warning, and debug why eslint isn't catching this


apedroferreira (approved):
thanks, this one should be pretty helpful!

Issue #663, opened 2022-07-19
Custom components: add a console for debugging


Janpot:
Just like in the function query editor, we could add a browser console in the bottom right to help debugging custom components.

<img width="1188" alt="Screenshot 2022-07-19 at 15 07 54" src="https://user-images.githubusercontent.com/2109932/179757844-f0bcf329-1999-49d9-96b2-e664c60d74ba.png">

benchmark:
- [codesandbox](https://codesandbox.io/)

bytasv:
As per initial investigation done in https://github.com/mui/mui-toolpad/pull/711 looks like it's not a straight forward thing to do.
Seems that, we need to refactor current custom component editor so that the code gets evaluated in an isolated iframe context. Once that's done we should be able to hook into `console.*` calls and pass them to our `Console` component.
Let's keep this issue open and decide on a priority among other things for now

prakhargupta1:
I guess we can close this? @Janpot

PR #853, opened 2022-08-25
Code component console logging

Janpot:
- [x] run code components preview in an iframe in its own react root. We'll be reusing our Application preview for this to be able to align as close as possible.
- [ ] intercept console messages
- [ ] add console to code component editor

Fixes https://github.com/mui/mui-toolpad/issues/663


Janpot:
This isn't the right way. It's logging multiple render. It's also making logs harder to trace back to a location.

Next up will be trying out running the app within a [SES Compartment](https://www.npmjs.com/package/ses#compartment)

Janpot:
This is a dead end. Let's start thinking about building a new runtime on top of code generation and SES

Issue #819, opened 2022-08-18
In the right pane, props and component name should be visually separated


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Currently, the right panel shows everything together. Should we have some visual separation based on the type of items?

Moreover, we can also look into separating UI props from data props.
<img width="301" alt="Screenshot 2022-08-18 at 1 50 04 PM" src="https://user-images.githubusercontent.com/92228082/185347082-9ba6917c-5acb-4e71-8055-3b22e0b59382.png">

apedroferreira:
i'm going to separate some "layout" props in this PR https://github.com/mui/mui-toolpad/pull/804, but it still doesn't look great, I can still at least try to improve the spacing there.

apedroferreira:
> Moreover, we can also look into separating UI props from data props.

We would need something like property categories to do something like this, so we could group them by category. But this would mean every prop would need to have its own category, which I'm not sure would make sense in every case.

So for now I just tried to make the order of the props in each component a bit more logical, but if we want to categorize and split thing I guess we can do it separately.

PR #855, opened 2022-08-25
Props panel sections

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/819

- Separate component name/id section from layout properties section and component properties section - by using dividers, spacing and trying to improve the typography hierarchy a bit
- Reordered properties of each component to try to make them a bit more logical - if we want to split them more and categorize them in groups we can probably do it separately as I mentioned in the issue (https://github.com/mui/mui-toolpad/issues/819)

Component with layout props:

Screen Shot 2022-08-26 at 16 13 05

Component without layout props:

Screen Shot 2022-08-26 at 16 13 12


apedroferreira:
looks like i messed up the component name input, will fix

apedroferreira:
@prakhargupta1 does this work for now?

PR #857, opened 2022-08-26
Avoid state update during layout effect in NoSsr

Janpot:
Trying out an alternative implementation  of MUI `NoSsr` that avoids setting state in a layout effect.


apedroferreira (approved):

packages/toolpad-app/src/components/NoSsr.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react';
+
+export function useIsSsr(defer: boolean = false): boolean {
+  const isSsrInitialValue = React.useSyncExternalStore(
+    () => () => {},
+    () => false,
+    () => true,
```
eps1lon, line 7: This will cause a mismatch between client and server in React 18 which means it'll fallback to client-rendering. Client and server snapshot **must** match.

Ideally you should use React 18 in which case you rarely need `NoSSR`. And in 17 you almost always want to use `<NoSsr defer />` to avoid state updates in layout effects. You're not rendering it on the server so there will always be new content on the client anyway. No need to push that during layout.

Janpot: > Client and server snapshot must match.

I wasn't aware of this limitation, as far as I understood `getServerSnapshot` is used for hydration as well, so it will never mismatch as long as `getServerSnapshot` returns the same thing server side and client side, which it does here.

The main use case for us currently would be for importing components that access DOM APIs in top-level scope (`monaco-editor`). I initially assumed we could use `React.lazy` for this, but experimenting with it showed me that it doesn't prevent rendering the component on the server.

eps1lon: getServerSnapshot is used for hydration. Otherwise React wouldn't know how to reconcile client and server.

Why does NoSsr with defer not fit the use case you described?

Have you checked why the monaco editor can't be rendered on the server? Sounds like it should and that hydration should just make it interactive

Janpot: > Why does NoSsr with defer not fit the use case you described?

I now know it fits the use-case, but I initially misunderstood how it behaves during client-side mount. As a result I wrote this version of `NoSsr`, which, as it turns out, also doesn't synchronously rerender in a layout effect. I since took a closer look at the MUI `NoSsr` and noticed it behaves mostly the same. One could argue for replacing the MUI implementation with this one.

> Have you checked why the monaco editor can't be rendered on the server? Sounds like it should and that hydration should just make it interactive

[It doesn't](https://github.com/microsoft/monaco-editor/issues/420), and it I don't believe it needs to for our use-case

eps1lon: > [It doesn't](https://github.com/microsoft/monaco-editor/issues/420), and it I don't believe it needs to for our use-case

If you have to do a multi-render pass i.e. use this hook/NoSsr then there's a clear use case for why it should.

> One could argue for replacing the MUI implementation with this one.

What would be the argument?

Janpot: > If you have to do a multi-render pass i.e. use this hook/NoSsr then there's a clear use case for why it should.

In our use-case it always gets rendered inside a Dialog. There is no clear use-case for SSR here for us.

> What would be the argument?

Do you think any of these two implementations has advantages over the other?

eps1lon: I just completely blanked. If `getServerSnapshot` is called during hydration then we don't get a mismatch. I somehow misread that as "getClientSnapshot is called during hydration" which would be bad.

Janpot: For context, I requested your review because I don't feel 100% confident yet about my own understanding of React suspense, and Olivier mentioned this would be the kind of problem you'd be interested in weighing in on.

eps1lon: > Do you think any of these two implementations has advantages over the other? If so, which would that be, and why?

I'm not sure what you make of this. Since you don't want to explain why you need this I'm adding my assumptions here  to use later. It seems like you want to ensure the bundle is split at a certain point that will only be reached after user-interaction. But you don't want to use `React.lazy` because this will trigger a fallback the first time the user triggers that interaction. But this hook doesn't actually ensure any bundle splitting so it is probably another use case. You're talking about a Dialog but for Dialogs you don't need NoSSR since Dialogs aren't triggered until explicitly opened. So maybe it's dialog that is rendered on the server at which point it's irrelevant whether it is a dialog or not. It's something you render on the server.

Janpot: > Since you don't want to explain why you need this

I thought I made it clear when I said "The main use case for us currently would be for importing components that access DOM APIs in top-level scope". Rereading that, I can understand that I could have given a bit more body to my explanation. Consider the following [codeandbox](https://codesandbox.io/s/funny-silence-jrnowt?file=/pages/index.js). You will see three links, all three will open the same page rendering a component, if you refresh each page, you'll notice that only the last one actually doesn't crash. The imported module accesses `window` in its module scope. The built-in code splitting doesn't prevent the module from being executed on the server. Mind you that this last page doesn't render fallback IF the component is mounted client-side AND the component has been lazy loaded before.

It's a bit similar to what the `ssr` option is for in [`next/dynamic`](https://nextjs.org/docs/advanced-features/dynamic-import#with-no-ssr).

PR #863, opened 2022-08-26
Allow vertical resizing of container for any component

apedroferreira:
WIP, will continue this later - need access to element inner rects as in https://github.com/mui/mui-toolpad/pull/814


apedroferreira:
Outdated, and not sure if we would use this approach for vertical resizing in the future, so closing.

PR #866, opened 2022-08-28
[core] Fix scroll restoration in the docs

oliviertassinari:
https://deploy-preview-866--mui-toolpad-docs.netlify.app/toolpad/getting-started/overview/

- Fix the bug reported in https://github.com/mui/material-ui/pull/34037
- Remove duplication of Next.js configuration. We should be able to abstract most of the docs infrastructure in the mono-repo. We are one step closer to having our own MUI https://github.com/facebook/docusaurus.

### DIFF
--- a/docs/next.config.js
+++ b/docs/next.config.js
@@ -1,4 +1,5 @@
 const path = require('path');
+const withDocsInfra = require('@mui/monorepo/docs/nextConfigDocsInfra');
 
 const withTM = require('next-transpile-modules')(['@mui/monorepo'], {
   __unstable_matcher: (matchedPath) => {
@@ -8,25 +9,6 @@ const withTM = require('next-transpile-modules')(['@mui/monorepo'], {
 
 const pkg = require('../package.json');
 const { findPages } = require('./src/modules/utils/find');
-const { LANGUAGES, LANGUAGES_SSR } = require('./src/modules/constants');
-
-let DEPLOY_ENV = 'development';
-
-// Same as process.env.PULL_REQUEST === 'true'
-if (process.env.CONTEXT === 'deploy-preview') {
-  DEPLOY_ENV = 'pull-request';
-}
-
-if (process.env.CONTEXT === 'production' || process.env.CONTEXT === 'branch-deploy') {
-  DEPLOY_ENV = 'production';
-}
-
-if (
-  process.env.CONTEXT === 'branch-deploy' &&
-  (process.env.HEAD === 'master' || process.env.HEAD === 'next')
-) {
-  DEPLOY_ENV = 'staging';
-}
 
 const MONORPO_PATH = path.resolve(__dirname, './node_modules/@mui/monorepo');
 const MONOREPO_PACKAGES = {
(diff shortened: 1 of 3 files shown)


docs/package.json
```diff
@@ -61,7 +61,7 @@
     "lz-string": "^1.4.4",
     "markdown-to-jsx": "^7.1.7",
     "marked": "^4.0.17",
-    "next": "12.2.0",
+    "next": "^12.2.5",
```
oliviertassinari, line 64: We have two different Next.js app, I'm keeping this one in sync with the docs-infra

packages/toolpad-core/src/runtime.tsx
```diff
@@ -29,11 +29,8 @@ export const NodeRuntimeContext = React.createContext<string | null>(null);
 // React elements to the fibers.
 interface SlotsWrapperProps {
   children?: React.ReactNode;
-  // eslint-disable-next-line react/no-unused-prop-types
```
oliviertassinari, line 32: This is related to the update of `@mui/monorepo`. I don't know why, but it seems to be a good thing. Maybe they changed the default eslint rules.

docs/package.json
Janpot: Ideally we'd decouple `@mui/monorepo` as well

packages/toolpad-core/src/runtime.tsx
Janpot: 🤔 I'm no sure, these properties are deliberately not used in a way that is not statically analyzable. This looks more like the rule got inadvertently disabled.

Trying to add the following in the file as a test, and it doesn't seem to catch the unused prop, it does on master.
```tsx

interface Props {
  firstname: string;
  middlename: string; // middlename is never used by the Hello component
  lastname: string;
};

export function Hello({ firstname, lastname }: Props) {
  return (
    <div>
      Hello {firstname} {lastname}
    </div>
  );
}
```

Perhaps something with us importing the overrides in eslintrc? I removed it in https://github.com/mui/mui-toolpad/pull/835

Janpot: Updating the branch to verify this

oliviertassinari:
I have found an issue with this approach, it leads to a warning on the core's side. We need to change the import a bit.

packages/toolpad-core/src/runtime.tsx
oliviertassinari: I have seen the same in https://github.com/mui/mui-toolpad/pull/809

Janpot: Just learnt about `eslint --print-config` option. The definitely rule is configured differently for master:

```json
    "react/no-unused-prop-types": [
      "error",
      {
        "customValidators": [],
        "skipShapeProps": true
      }
    ],
```

vs this PR

```json

    "react/no-unused-prop-types": [
      "off",
      {
        "customValidators": [],
        "skipShapeProps": true
      }
    ],
```

Looks like it got turned off in monorepo eslintrc: https://github.com/mui/material-ui/pull/33027/files#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5R140
If it was disabled on purpose then we can probably re-enable it for Toolpad for now.

oliviertassinari: Ok great, so I think that we should revert this change in the main repo.

Janpot: Perhaps we can turn the rule off in javascript files only? ideally for toolpad I'd love to disallow `.propTypes` and rely on typescipt only, but that [doesn't seem possible right now](https://github.com/jsx-eslint/eslint-plugin-react/issues/3107).

docs/next.config.js
```diff
@@ -49,16 +50,9 @@ if (reactStrictMode) {
 const isDeployment = process.env.NETLIFY === 'true';
 
 module.exports = withTM({
-  eslint: {
-    ignoreDuringBuilds: true,
-  },
+  ...nextConfig.baseline,
```
Janpot, line 53: Would it make sense to do this more plugin style?

```tsx
withTM(withBaseline({
  // next config ...
}))
```

It would allow us to move the responsibility of merging the configs into the plugin.

oliviertassinari: It sounds good. We could almost have a flow-right helper too:

```jsx
flowRight([
  withTM,
  withBaseline,
])({})
```

Janpot (approved):

Issue #826, opened 2022-08-18
Components overlapping when horizontally resized


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

https://user-images.githubusercontent.com/92228082/185468526-d88441c6-4315-430a-b295-98078f3be059.mov

Not sure if this is the expected behavior as of now.
But sometimes the component is pushing the adjacent one, sometimes it is not.

apedroferreira:
thanks for finding this Prakhar!
are you able to send me the DOM object from the app where you have this issue? that might help me reproduce it and understand it better

apedroferreira:
oh wait, i think the only issue here is that the resizing is not aware that the text input has a minimum width, so it does not respect it.
this might be a tough issue to fix, but i'll see how we can improve this

Janpot:
We can make it always `fullWidth` and remove that property? How would we size a texfield that is the only component on a row?

prakhargupta1:
@apedroferreira In case it helps, here is the link to the app: https://master--toolpad.mui.com/_toolpad/app/cl1on1bd80041a1pdrrfdt0rq/editor/pages/cl6ysktjz00023b6g8xkizxr8

apedroferreira:
> How would we size a texfield that is the only component on a row?

with some improvements to resizing we could make it possible to adjust the left/right margin in the container to any value, as well as resize the text input horizontally to any possible size. right now it's very basic though so users can't do it yet.

> We can make it always `fullWidth` and remove that property? 

as things are for now if we had this type of component always be full width i guess we wouldn't be able to add any margins? 
and i guess any imposed minimum width in any component would still create visual issues like these?

Issue #461, opened 2022-05-24
Set minimum and maximum value for component number attributes


apedroferreira:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Right now, for component properties that have a number value, any integer value can be set, including negative numbers.
For some for these attributes, we want there to be a minimum and maximum value that can be set.
This will make it less likely for the layout to break, or any other unintended behaviors to happen.

prakhargupta1:
Is it related to all or some particular components? 
Where do we show 'component number attributes'?

apedroferreira:
> Is it related to all or some particular components? Where do we show 'component number attributes'?

I could help on any UI controls for components that use numbers, even new custom components - for example if a row has a giant `gap` value like `9999` it would probably break the layout pretty bad.
This probably isn't high priority for now as we don't have that many cases/components where something like this can have bad consequences.

<img width="283" alt="Screen Shot 2022-08-18 at 16 36 38" src="https://user-images.githubusercontent.com/10789765/185436061-e0a7b15a-30af-48a6-89a3-98f08c75c723.png">
<img width="297" alt="Screen Shot 2022-08-18 at 16 36 49" src="https://user-images.githubusercontent.com/10789765/185436052-05a2e8bf-b947-4939-bd52-6edf1d32e48e.png">

PR #871, opened 2022-08-30
Allow minimum and maximum value for component number props

apedroferreira:
Initially just intended for this PR to be prep for fixing this issue https://github.com/mui/mui-toolpad/issues/826, by using a single source of truth for the minimum component dimensions. But it seems like it's actually a solution, so it closes https://github.com/mui/mui-toolpad/issues/826.
Also closes https://github.com/mui/mui-toolpad/issues/461

Using it already for `Select` component minimum width as an example.

https://user-images.githubusercontent.com/10789765/187513853-ed12ff95-b256-4cf0-90e4-aa616ca933ec.mov


Janpot:
proptype is based on json-schema, I think  that can describe min and max values. perhaps we should just reuse that?

Janpot:
I'm still not understanding the minimum width use case and how it should behave. Suppose minimum width is 50px, and I put 100 Selects in one row, what happens? what overflows or wraps?

apedroferreira:
> proptype is based on json-schema, I think that can describe min and max values. perhaps we should just reuse that?

didn't know about that, i'll take a look

apedroferreira:
Looks like I already solved the whole issue in this PR after all, slightly unintentionally:

https://user-images.githubusercontent.com/10789765/187966124-55b8dea7-7fbc-4192-addb-ad59e64dcc36.mov

I've tested a bit and basically as long as we don't set a minimum width in the component CSS this kind of issue does not happen, the components always adjust to the width of the container.
 If we make the screen very very narrow and have a lot of components they do overflow as it's inevitable, but I think the best option is to just let that happen rather than hiding the overflow.

Screen Shot 2022-09-01 at 17 10 39

In the future we can probably improve support for small device screens.

So what do you think? In my opinion we can merge this.
  If you don’t agree with adding the new `width` prop it’s also not strictly necessary, we can remove it and use whatever default values the MUI component uses. 
In my opinion this `width` prop is useful without being detrimental - it gives the user more control over how the element looks, and it sets a minimum width that doesn’t look so bad when the select is empty, just like the minimum width before this PR.
But the `width` prop is just an extra thing.

apedroferreira:
Also if you like this idea we can disable the width control when full-width is enabled, I just don't want to do it ahead of time if we don't want to go in this direction.

Janpot (approved):
👌

prakhargupta1:
👌

PR #879, opened 2022-09-01
Editor and save state tweaks

apedroferreira:
Implements PR feedback from https://github.com/mui/mui-toolpad/pull/861.
No debug logs, no color in logs, less log messages, no hiding of save state icon
Also adjusted a section title in the page properties editor, to improve consistency with changes in the component properties editor.


apedroferreira:
The console logs are breaking the CI. Any suggestions since we don't use the env variable?
Should i try to disabled console logs in tests?

Janpot:
> The console logs are making the CI tests fail. 

only the use of `console.error` should make it fail

apedroferreira:
sounds good i probably misunderstood why the ci failed then

Janpot:
creating https://github.com/mui/mui-toolpad/pull/881 for more insight in these messages

Janpot (approved):

Issue #876, opened 2022-09-01
Running SQL on PostgresDB not showing results


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

App link: https://master--toolpad.mui.com/_toolpad/app/cl1on1bd80041a1pdrrfdt0rq/editor/pages/b013gge

DB: https://rnacentral.org/help/public-database

Using the same resource I am able to get results in other tools.

PR #880, opened 2022-09-01
Force node 16 on render.com

Janpot:
`String.prototype.replaceAll` is only available on node >=v15, we're already using 16 for the docker image
Closes https://github.com/mui/mui-toolpad/issues/876


bharatkashyap (commented):
👌

PR #890, opened 2022-09-03
[core] Add release step for the docs

oliviertassinari:
- As far as I know, we can't have https://mui.com/toolpad to map with https://github.com/mui/mui-toolpad/commits/master. This would lead to one structural problem: Documenting new features that developers can't use because not been released yet. There are also other considerations like making it harder to batch the release of PRs together, but it's probably secondary. I have changed the configuration in Netlify and now intend to document it.
- Github -> GitHub

---

As side note, I think that it would be better to move the release step to a different markdown than the contributing guide, the community doesn't care about how we release, at best, it's noise for them and worse, we overwhelm/confuse them. I would recommend the same pattern as https://github.com/mui/mui-x/blob/master/scripts/README.md.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -88,7 +88,7 @@ Please have a look at our general guidelines for sending pull requests [here](ht
    yarn release:changelog
    ```
 
-   Running this command requires a [Github personal access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) with the `public_repo` scope.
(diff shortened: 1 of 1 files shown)


Janpot:
> Documenting new features that developers can't use because not been released yet.

I believe over time we could aim at versioning our docs as well. One version per toolpad version. By default the user is directed to the last released version. There would be a "next" version or something that contains whatever is on the master branch.

> I think that it would be better to move the release step to a different markdown than the contributing guide

I'm fine with making that part of this PR. Otherwise will do after this gets merged.

apedroferreira:
+ 1 for separating the release instructions

oliviertassinari:
Ok, so it seems that this workflow proposal is accepted, I'm merging.

PR #898, opened 2022-09-05
Editable text component batched fixes

bharatkashyap:
- Visual demonstration 
  Before (Text cropped on Safari)
   Safari_bug

    After
   Screenshot 2022-09-06 at 2 48 58 AM
   
- Also replaced the use of string classes with the experiment class name generator.


packages/toolpad-app/src/components/EditableText.tsx
```diff
@@ -121,6 +129,7 @@ const EditableText = React.forwardRef<HTMLInputElement, EditableTextProps>(
             whiteSpace: 'nowrap',
             fontSize: theme.typography[typographyVariant].fontSize,
             height: theme.typography[typographyVariant].fontSize,
```
Janpot, line 131: Isn't the cropping due to the height set to the fontsize here?

bharatkashyap: 😄 Good catch!

apedroferreira (approved):

packages/toolpad-app/src/components/EditableText.tsx
```diff
@@ -120,7 +128,7 @@ const EditableText = React.forwardRef<HTMLInputElement, EditableTextProps>(
             overflow: 'hidden',
             whiteSpace: 'nowrap',
             fontSize: theme.typography[typographyVariant].fontSize,
-            height: theme.typography[typographyVariant].fontSize,
+            height: `calc(${theme.typography[typographyVariant].fontSize} + 2px)`,
```
Janpot, line 131: Isn't this what `em` units are for?

apedroferreira: `em` is based on the font size of the parent i think? so might work here but not sure.
also this could probably be written as `${theme.typography[typographyVariant].fontSize + 2 }px` if it sounds preferrable not to use the `calc()`

Janpot: > `em` is based on the font size of the parent i think?

Yes for properties like `font-size`. For other properties like `height` it's based on the fontsize of the element itself

Janpot (approved):

PR #899, opened 2022-09-05
[website] Fix GA events going to development

bharatkashyap:
- This is to fix Google Analytics issues going to the "development" instance of Google Analytics

See https://github.com/mui/material-ui/blob/master/docs/pages/_document.js#L34 and 
Screenshot 2022-09-06 at 3 07 20 AM

### DIFF
--- a/docs/next.config.js
+++ b/docs/next.config.js
@@ -83,6 +83,7 @@ module.exports = withTM({
     LIB_VERSION: pkg.version,
     REACT_STRICT_MODE: reactStrictMode,
     FEEDBACK_URL: process.env.FEEDBACK_URL,
+    PULL_REQUEST: process.env.PULL_REQUEST === 'true',
     // #default-branch-switch
     SOURCE_CODE_ROOT_URL: 'https://github.com/mui/mui-toolpad/blob/master',
     SOURCE_CODE_REPO: 'https://github.com/mui/mui-toolpad',
--- a/docs/pages/_document.js
+++ b/docs/pages/_document.js
@@ -1,10 +1,3 @@
 import MyDocument from '@mui/monorepo/docs/pages/_document';
 
-// eslint-disable-next-line no-console
-console.log(
-  process.env.PULL_REQUEST,
-  typeof process.env.PULL_REQUEST,
-  JSON.stringify(process.env.PULL_REQUEST),
-);
-
 export default MyDocument;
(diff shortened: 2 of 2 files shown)


bharatkashyap:
https://app.netlify.com/teams/mui/builds/63166d6c07c39500080bdfc2

Value of 

```ts
process.env.PULL_REQUEST,
  typeof process.env.PULL_REQUEST,
  JSON.stringify(process.env.PULL_REQUEST),
  ```
  in the deploy preview:

Screenshot 2022-09-06 at 10 36 41 AM

Janpot:
Ok, I understand now. During SSR it always uses the environment variables available on the server. netlify passes `PULL_REQUEST=false` during production build. We are not converting `process.env.PULL_REQUEST` to a boolean in `next.config.js` (like the monorepo does). So it gets interpolated as `"false"`, which is a truthy value, thus the Toolpad docs thinks it's always building a PR preview.

Solution, either:
* do `PULL_REQUEST: process.env.PULL_REQUEST === 'true'` like in the monorepo. (This is kind of brittle, it makes the value of `PULL_REQUEST` a string in some places, and a boolean in other. We're also relying on a variable that is defined by the CI provider)
* just pass `process.env.GA_TAG` and pass the correct value in CI

bharatkashyap:
> Ok, I understand now. During SSR it always uses the environment variables available on the server. netlify passes `PULL_REQUEST=false` during production build. We are not converting `process.env.PULL_REQUEST` to a boolean in `next.config.js` (like the monorepo does). So it gets interpolated as `"false"`, which is a truthy value, thus the Toolpad docs thinks it's always building a PR preview.
> 
> Solution, either:
> 
> * do `PULL_REQUEST: process.env.PULL_REQUEST === 'true'` like in the monorepo. (This is kind of brittle, it makes the value of `PULL_REQUEST` a string in some places, and a boolean in other. We're also relying on a variable that is defined by the CI provider)
> * just pass `process.env.GA_ID` and pass the correct value in CI

Let's go ahead and take the first route for now so that we can defer to the existing way of doing things in the immediate term

Janpot (approved):

oliviertassinari:
https://github.com/mui/material-ui/pull/34259 should prevent this bug from happening again, making the docs infra less coupled to Toolpad.

PR #906, opened 2022-09-06
[website] Landing Page follow up

bharatkashyap:
https://deploy-preview-906--mui-toolpad-docs.netlify.app/toolpad/

- Visual demonstration 
- [x] Visually group Upvote button with features:
Screenshot 2022-09-12 at 2 50 33 PM

https://user-images.githubusercontent.com/19550456/189080567-58f5348a-aca0-49eb-b7cb-01f80df67df1.mov


apedroferreira:
if this is about putting the rectangle about "upvoting" closer to that grid with 9 rectangles, maybe it could be a bit closer to the grid? still seems a bit far in my opinion, almost as far as it is from the next section below

docs/src/components/landing/Banner.js
```diff
@@ -10,54 +11,67 @@ import KeyboardArrowRightRounded from '@mui/icons-material/KeyboardArrowRightRou
...
-          fullWidth
-          endIcon={<KeyboardArrowRightRounded />}
+    <Container
+      sx={{
+        pb: { xs: 2, sm: 4, md: 8 },
+        pt: { xs: 0, sm: 1, md: 2 },
+        bgcolor: (theme) =>
+          theme.palette.mode === 'dark' ? theme.palette.primaryDark[900] : theme.palette.grey[50],
+      }}
+      maxWidth="100vw"
```
apedroferreira, line 21: is it possible to use something more relative to the parent elements instead of viewport units?

docs/src/components/landing/Banner.js
```diff
@@ -10,54 +11,67 @@ import KeyboardArrowRightRounded from '@mui/icons-material/KeyboardArrowRightRou
...
+    >
+      <Container>
```
apedroferreira, line 23: do we really have to nest 2 containers? i feel like it should be possible to do it with just 1 unless this is more complicated that it seems at first glance...

bharatkashyap: The issue is about having the element to be constrained within the normal `Container` width, while having the background color span the entire screen width

apedroferreira: Ah I see, I was assuming we had some kind of root container that already limited the max width to the width of the page, and all sections were children of it. If not, and this is difficult to do, this isn't too bad though.

apedroferreira (approved):
thanks for looking into the changes!

oliviertassinari:
> Visual demonstration

I think that screenshots would have been much better in this context

PR #908, opened 2022-09-06
Bug template: Move reproduction above current behavior

Janpot:
Doesn't this logically belong before "current behavior"? I mean, how can I explain the current behavior without first explaining how to produce the behavior?
I repeatedly find myself explaining the problem in "current behavior" and then arriving at "reproduction steps" and realizing I already gave them in the earlier section.


apedroferreira:
not sure, maybe it makes sense to first explain the issue in general and then provide the reproduction as it can be more specific

oliviertassinari:
This PR makes me think of https://github.com/mui/mui-design-kits/pull/73#discussion_r945648768

oliviertassinari:
One pain point I have with the current bug template is the duplication of the effort between "Actual behavior" and "Reproduction". Most of the time, I almost give the full reproduction on the "Actual behavior" step. It seems to be close to the pain point raised by @Janpot 

> I repeatedly find myself explaining the problem in "current behavior" and then arriving at "reproduction steps" and realizing I already gave them in the earlier section.

So I think that it could make sense to ask for the reproduction directly inside the "Actual behavior", or just after. I would only be careful with the change not reducing the compliance rate of the developers with the instructions.

Otherwise, I personally think that this is a change that should be first experiemented with http://github.com/mui/material-ui as we will learn faster about either it works or not.

oliviertassinari (commented):
👍 it could indeed work better

Janpot:
> Otherwise, I personally think that this is a change that should be first experimented with [http://github.com/mui/material-ui](https://github.com/mui/material-ui) as we will learn faster about whether it's better or worse.

👍  opening https://github.com/mui/material-ui/pull/34279

oliviertassinari:
I suspect that we could move forward with this change 👍

PR #909, opened 2022-09-07
Toolpad-app dependencies update

Janpot:


apedroferreira (approved):

packages/toolpad-core/package.json
```diff
@@ -33,8 +33,8 @@
     "url": "https://github.com/mui/mui-toolpad/issues"
   },
   "dependencies": {
-    "@mui/x-data-grid-pro": "^5.15.2",
-    "@tanstack/react-query": "^4.1.0",
+    "@mui/x-data-grid-pro": "^5.17.1",
```
apedroferreira, line 36: @Janpot i think this broke the datagrids

apedroferreira: we should add some integration test to see if datagrid work sometime soon

Janpot: In what sense? They seemed to work for me. Can you open a ticket?

apedroferreira: i've created an issue for the bug, hopefully it's not just some local issue i'm having but didn't seem like it
bug: https://github.com/mui/mui-toolpad/issues/921
integration test issue: https://github.com/mui/mui-toolpad/issues/920

Issue #513, opened 2022-06-01
Datagrid: recreate/refresh columns


Janpot:
Add a button in the datagrid columns control to infer again all columns based on the data it received. This would help when the user has changed the underlying datasource.

PR #914, opened 2022-09-07
Allow recreating data grid columns from existing data

apedroferreira:
Allow for recreating columns in Data Grid elements, from existing data.
Replaces all columns with the inferred columns when the button is pressed.

- Closes https://github.com/mui/mui-toolpad/issues/513

https://user-images.githubusercontent.com/10789765/188915514-123656bf-1498-4124-b43e-e2490f6e1a51.mov


packages/toolpad-app/src/toolpad/propertyControls/GridColumns.tsx
```diff
@@ -176,6 +188,13 @@ function GridColumnsPropEditor({
           <React.Fragment>
             <DialogTitle>Edit columns</DialogTitle>
             <DialogContent>
+              <IconButton
+                aria-label="Recreate columns"
```
Janpot, line 192: We should probably add a tooltip as well.

@prakhargupta1 @gerdadesign nominating this component as well for a redesign. Perhaps we could even put this inline in the inspector, instead of in a dialog?

apedroferreira: good idea, i'll add

apedroferreira: I've added tooltips to the "Recreate columns" and the "Add column" buttons

packages/toolpad-app/src/toolpad/propertyControls/GridColumns.tsx
```diff
@@ -188,16 +189,20 @@ function GridColumnsPropEditor({
...
-                disabled={!hasColumnSuggestions}
-              >
-                <RefreshIcon />
-              </IconButton>
-              <IconButton aria-label="Add column" onClick={handleMenuClick} disabled={disabled}>
-                <AddIcon />
-              </IconButton>
+              <Tooltip title="Recreate columns">
+                <IconButton
+                  aria-label="Recreate columns"
```
Janpot, line 194: Tooltip [should add the label](https://mui.com/material-ui/react-tooltip/#accessibility)

apedroferreira: cool i didn't know, i've removed the unnecessary props + added a span so that the tooltip shows when the "recreate columns" button is disabled (as recommended by the MUI console warnings)

packages/toolpad-app/src/toolpad/propertyControls/GridColumns.tsx
```diff
@@ -190,16 +190,14 @@ function GridColumnsPropEditor({
...
             <DialogContent>
               <Tooltip title="Recreate columns">
-                <IconButton
-                  aria-label="Recreate columns"
-                  onClick={handleRecreateColumns}
-                  disabled={!hasColumnSuggestions}
-                >
-                  <RefreshIcon />
-                </IconButton>
+                <span>
```
Janpot, line 193: this prevents it from passing the aria label to the button, right?

apedroferreira: it's passed to the span, actually that might not be ok good point

apedroferreira: i added `describeChild` to the tooltip and `aria-label` in the button, seems like the right way to do it

Janpot (approved):

Issue #902, opened 2022-09-06
Provide 2 basic prebuilt apps as examples


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

The apps will help self-host users to examine steps required to build an app.
 
1. COVID app
2. Dog app

apedroferreira:
i'll take this if it's up for grabs

prakhargupta1:
I'll share DOM code of the apps with you.

apedroferreira:
So which direction do we want to go with this?
1. Show the apps in the beginning for every new user, and they can delete them if they wish
2. Allow choosing some app templates in this panel:

<img width="1792" alt="Screen Shot 2022-09-07 at 15 13 12" src="https://user-images.githubusercontent.com/10789765/188900721-ea8a41bb-ad33-49d2-bcc9-126acfa7b201.png">

I guess we opted for option 2 but just confirming. Both have their pros and cons but option 2 may be a useful feature in itself, as it allows us to provide some "starters" for users.

Janpot:
We will do 2.

prakhargupta1:
App 1: https://toolpad-pr-940.onrender.com/_toolpad/app/cl7t6cpxb0106axq11s54imas/editor/pages/xo1a1q7
App 2: https://master--toolpad.mui.com/_toolpad/app/cl27ajqfu0403a0oz236ay3pj/editor/pages/cl27arb1i00043g693pzhex2r

apedroferreira:
> App 1: https://toolpad-pr-940.onrender.com/_toolpad/app/cl7t6cpxb0106axq11s54imas/editor/pages/xo1a1q7 App 2: https://master--toolpad.mui.com/_toolpad/app/cl27ajqfu0403a0oz236ay3pj/editor/pages/cl27arb1i00043g693pzhex2r

thanks Prakhar! the first link stopped working for some reason (the covid stats app), it says `Not found`

prakhargupta1:
Updated the link, it should work now.

apedroferreira:
> https://master--toolpad.mui.com/_toolpad/app/cl7tbkukv0265awtoxfauiywo/editor/pages/l11ao9l

it worked, thanks!

WangLarry:
need password to open this app

prakhargupta1:
Hi Wang,
Currently it is in development and being used internally. For now, you can self-host to try it out.

PR #926, opened 2022-09-07
Prebuilt app templates

apedroferreira:
Allow using app templates when creating a new app.

- Closes https://github.com/mui/mui-toolpad/issues/902

**Images:**

https://user-images.githubusercontent.com/10789765/189967429-6e04595e-0761-41ba-a41e-4934c3cab599.mov

**Statistics:**

https://user-images.githubusercontent.com/10789765/189967479-4a4793a2-e960-407e-8e10-9424c259b6be.mov


packages/toolpad-app/src/toolpad/Home.tsx
```diff
@@ -95,9 +116,28 @@ function CreateAppDialog({ onClose, ...props }: CreateAppDialogProps) {
                 setName(event.target.value);
               }}
             />
+
+            <FormControl sx={{ my: 1 }} fullWidth>
```
Janpot, line 120: Did you know you can turn a `TextField` into a `Select` using `<TextField select>`? It will take care of the `InputLabel` and `FormControl`.

packages/toolpad-app/src/toolpad/Home.tsx
```diff
@@ -77,7 +93,12 @@ function CreateAppDialog({ onClose, ...props }: CreateAppDialogProps) {
         <DialogForm
           onSubmit={(event) => {
             event.preventDefault();
-            createAppMutation.mutate([name, { dom: dom.trim() ? JSON.parse(dom) : null }]);
+
+            const selectedAppTemplateDom =
+              APP_TEMPLATE_OPTIONS.find(({ value }) => value === appTemplate)?.dom || null;
+            const appDom = dom.trim() ? JSON.parse(dom) : selectedAppTemplateDom;
+
+            createAppMutation.mutate([name, { dom: appDom }]);
```
Janpot, line 101: I feel like these templates should live on the backend. So that we don't pull all of them in the frontend bundles. This will definitely become problematic as we add more of them.

I've been wondering whether we should update the `createApp` method and give it a `from` option, which would contain a discriminated union specifying what we want to start from. Either a provided dom, or a template:

```tsx
export interface CreateAppOptions {
  from?:
    | {
        kind: 'dom';
        dom: appDom.AppDom;
      }
    | {
        kind: 'template';
        id: string;
      }
    | null;
}

export async function createApp(name: string, opts: CreateAppOptions = {}): Promise<prisma.App>;
```

@bharatkashyap This could also incorporate the app duplication if we add
```tsx
    | {
        kind: 'app';
        id: string;
      }
```
to the discriminated union.

bharatkashyap: > @bharatkashyap This could also incorporate the app duplication if we add
> 
> ```tsx
>     | {
>         kind: 'app';
>         id: string;
>       }
> ```
> 
> to the discriminated union.

I feel there's a difference between a `kind: 'dom'` and every other type of `kind`, since `template` and `app` are fetching  a `dom` and then passing them into `createApp`.

In that sense, the `kind`s do not seem to be on the same level of abstraction. For me, it is okay for `createApp` with `dom` to continue being the lowest-level abstraction. 

A `from` option might suit a `fetchDom` function which should be callable inside wrapper functions we create to duplicate apps, create them from templates and so on.

So, 
```ts
export async function duplicateApp(id: string, name: string): Promise<AppMeta> {
  const dom = await fetchDom({ from: { kind: 'app', id: id}})
  const newApp = await createApp(`${name} (copy)`, { dom });
  return newApp;
}
```

and,
```ts
export async function createTemplateApp(id: string, name: string): Promise<AppMeta> {
  const dom = await fetchDom({ from: { kind: 'template', id: id}})
  const newApp = await createApp(name, { dom });
  return newApp;
}
```

Janpot: > In that sense, the kinds do not seem to be on the same level of abstraction.

They're all creating an new application based on an initializer. the initializer can be another "app", a predefined "template", or an inline "dom". That's the proposed abstraction.

As for `fetchDom`, I think we should figure out the right abstractions as the need arises. I think we're starting to see that DOM on its own doesn't have a lot of meaning. And it will have less so in the context of a template when we make them parametrizable.

bharatkashyap: > > In that sense, the kinds do not seem to be on the same level of abstraction.
> 
> They're all creating a new application based on an initializer. the initializer can be another "app", a predefined "template", or an inline "dom". That's the proposed abstraction.
> 
Would you not say that the predefined "template" and another "app" are essentially an abstraction over fetching the dom attached to the respective template/app? That's the only issue in my head.

Janpot: An "abstraction" hides implementation details. Would you not say that the method by which a dom is obtained is an implementation detail? 
Suppose some day we want to add a way to create an app from JSON hosted under a remote url. Wouldn't that fit under the same abstraction?

bharatkashyap: > Suppose some day we want to add a way to create an app from JSON hosted under a remote url. Wouldn't that fit under the same abstraction?

Yep, I agree it would. To me it would fit under another kind of "fetching the DOM" (via URL) and not "creating an app" (which is always through a dom). Even if none is passed, we generate a "default" dom. 

In essence, I'm trying to say that the "kinds" in the example fit better as implementation details of fetching the DOM and not creating an app. 

In the future, if there starts to exist another way of creating an app (say integrating with an existing codebase, which doesn't require generating a dom) that would feel more fitting as another "kind" of `createApp`.

packages/toolpad-app/src/toolpad/Home.tsx
apedroferreira: oh... i'll use that then

packages/toolpad-app/src/toolpad/Home.tsx
apedroferreira: I don't think any of the 2 prosposals are bad, but a single `createApp` sounds a bit easier to extend/mantain in the long run in my opinion, in case we end up having many different ways of creating apps. i'll start with that and move the templates to the backend.
Shouldn't be too hard to adjust it in the future if we see the need to.

apedroferreira:
I've updated the videos in this PR with the latest changes:
- made some changes to the template selector - feel free to suggest changes to the copy or anything else
- tried to clean up up example apps, also added a heading - not sure if my changes are good so feel free to leave feedback on these
- store template DOMs in backend as discussed in review (i assumed they won't be in the bundle if they're in the server folder?)

@prakhargupta1 feel free to give feedback too if you can access the preview

prakhargupta1:
It looks perfect! 👍

packages/toolpad-app/src/server/appTemplateDoms/doms.ts
```diff
@@ -0,0 +1,11 @@
+import * as appDom from '../../appDom';
+
+import statisticsAppDom from './statistics.json';
```
Janpot, line 3: This will likely not scale well with the typechecking. Can't we read them using `fs` instead?

packages/toolpad-app/src/server/data.ts
```diff
@@ -175,17 +176,36 @@ function createDefaultDom(): appDom.AppDom {
...
+      dom?: appDom.AppDom | null;
+    }
+  | {
+      type: 'template';
+      name: AppTemplateName;
+    };
+
+export async function createApp(
+  name: string,
+  opts: CreateAppOptions = { type: 'dom' },
```
Janpot, line 191:
```ts
export interface CreateAppOptions {
  from?:
    | {
        kind: 'dom';
        dom: appDom.AppDom;
      }
    | {
        kind: 'template';
        id: string;
      };
}
```

Represents the options better. `from: undefined` represents "just create an empty app". and `{ kind: 'dom' }` becomes illegal. The type system helps you to not accidentally forget to pass a dom.

I also prefer to move more towards the `kind` nomenclature instead of `type` for discriminated unions.

bharatkashyap (approved):

bharatkashyap:
Great improvement to the app!

packages/toolpad-app/src/server/appTemplateDoms/doms.ts
apedroferreira: sure, i'll use `fs`

packages/toolpad-app/src/server/data.ts
apedroferreira: sounds good! let's do it like that then

packages/toolpad-app/src/server/appTemplateDoms/doms.ts
```diff
@@ -1,11 +1,19 @@
...
 
-import statisticsAppDom from './statistics.json';
-import imagesAppDom from './images.json';
-import { AppTemplateName } from '../../types';
+import { AppTemplateId } from '../../types';
+import { readJsonFile } from '../../utils/fs';
 
-export const APP_TEMPLATE_DOMS: Record<AppTemplateName, appDom.AppDom | null> = {
+// __dirname does not work, see: https://github.com/vercel/next.js/discussions/14341
+const DOMS_DIR_PATH = './src/server/appTemplateDoms';
```
apedroferreira, line 8: not ideal but i could not find a way to use relative paths as `__dirname` resolves to `pages/api`
i've added a test to ensure selecting templates works as expected

packages/toolpad-app/src/toolpad/Home.tsx
```diff
@@ -95,9 +129,29 @@ function CreateAppDialog({ onClose, ...props }: CreateAppDialogProps) {
...
+                    <Typography variant="caption">{description || ''}</Typography>
+                  </span>
+                </MenuItem>
+              ))}
+            </TextField>
+
             {config.enableCreateByDom ? (
               <TextField
-                label="seed DOM"
+                sx={{ my: 1 }}
```
Janpot, line 153: Alternatively `sx={{ gap: 1 }}` on the parent would make this a little DRYer

packages/toolpad-app/src/toolpad/Home.tsx
```diff
@@ -48,18 +48,41 @@ import useLocalStorageState from '../utils/useLocalStorageState';
 import ErrorAlert from './AppEditor/PageEditor/ErrorAlert';
 import { ConfirmDialog } from '../components/SystemDialogs';
 import config from '../config';
+import { AppTemplateId } from '../types';
 import { errorFrom } from '../utils/errors';
 
+export const APP_TEMPLATE_OPTIONS = {
```
Janpot, line 54: It would probably make sense to load the templates list from the server as well

apedroferreira: this a different thing though, it's a list with the texts we show in the UI, so meant for the browser-side only

apedroferreira: could be more type safe though if it's not, i thought i had made sure the key names matched the templates
i'll take another look at that

Janpot: Also, prefer to use a `Map<string, TemplateDef>` for these sort of things. Just avoids issues when someone decides to create a `'toString'` template or something.

Issue #920, opened 2022-09-07
Datagrid Integration Tests


apedroferreira:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Write a basic integration test for placing Datagrid components in the app canvas, and making sure they render successfully.

### Examples 🌈

- Test that Datagrid renders correctly, and does not render its error state

### Motivation 🔦

We've unintentionally broken Datagrid elements a few times so far, so a test that covers whether Datagrid components are functioning well at a basic level would probably be useful.

Janpot:
By extension, I think we need a test that just has every component on a page and verifies they all render.

PR #938, opened 2022-09-08
Add components integration test

Janpot:
This just verifies they all render in the runtime. Fixes https://github.com/mui/mui-toolpad/issues/920

### DIFF
--- a/test/integration/components/index.spec.ts
+++ b/test/integration/components/index.spec.ts
@@ -1,19 +1,18 @@
 import * as path from 'path';
 import { ToolpadHome } from '../../models/ToolpadHome';
+import { ToolpadRuntime } from '../../models/ToolpadRuntime';
 import { test } from '../../playwright/test';
 import { readJsonFile } from '../../utils/fs';
 
 test('components', async ({ page }) => {
   const dom = await readJsonFile(path.resolve(__dirname, './componentsDom.json'));
-  const { id: pageId } = Object.values(dom.nodes).find(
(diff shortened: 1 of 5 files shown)


test/integration/components/index.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+  const dom = await readJsonFile(path.resolve(__dirname, './componentsDom.json'));
+  const { id: pageId } = Object.values(dom.nodes).find(
+    (node: any) => node.name === 'components',
+  ) as any;
+
+  const homeModel = new ToolpadHome(page);
+  await homeModel.goto();
+  const app = await homeModel.createApplication({ dom });
+
+  await page.goto(`/app/${app.id}/preview/pages/${pageId}`);
```
apedroferreira, line 16: should we have one of those Playwright Object Model classes for the runtime as we have for the home page and the editor? or is it maybe not worth it for the runtime?

test/integration/components/index.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  await page.locator('text="foo button"').waitFor({ state: 'visible' });
```
apedroferreira, line 18: this works but maybe it should be more easy to understand which component each check is testing? maybe comments will do?

apedroferreira: nevermind i see the labels, didn't see it immediately though

apedroferreira (approved):

PR #942, opened 2022-09-08
Format function source on save

Janpot:
Prettify the code when saving, and support Cmd+S, just like code components.


apedroferreira (approved):
awesome!

PR #943, opened 2022-09-08
Pass datasource errors through to the page

Janpot:
Currently it's hard to get insight into errors happening in datasources as they get logged on the server and a generic 500 is sent to the client. This passes the error through to the client so a meaningful message can be shown


Janpot:
Splitted into https://github.com/mui/mui-toolpad/pull/946 and https://github.com/mui/mui-toolpad/pull/948

Issue #523, opened 2022-06-02
Debuggability: Visualize dependants of a node


Janpot:
* Allow for tracing back the dependants of a certain UI element/query/... 
* In the same spirit we should warn before deleting an element that is being depended on
*

prakhargupta1:
Does the View DOM option solve this?

![Image](https://user-images.githubusercontent.com/92228082/185236287-b395c815-17a3-4704-9fa9-53ae62976162.png)

Janpot:
no, this issue is about warning when you delete something that something else still depends on. e.g. when you delete a connection that still has queries on a page. And possible also going further, showing the chain of dependencies across a page.

PR #946, opened 2022-09-09
Make errors visible in the runtime

Janpot:
Make sure we surface the real underlying errors in the frontend, and not the generic 500 that toolpad creates:

Before:
Screenshot 2022-09-09 at 10 52 38

After:
Screenshot 2022-09-09 at 10 53 53

Extracted from https://github.com/mui/mui-toolpad/pull/943 which addresses a few more problems in the runtime error handling.

PR #948, opened 2022-09-09
Preserve javascript semantics when dependencies fail

Janpot:
Currently when an underlying value in a binding expression fails or is loading, the whole expression evaluates to undefined. As can be seen in the following:

Screenshot 2022-09-09 at 11 21 52

the second line is bound to the error message of a failing query
the third line is bound to the data of a failing query

Both of them show the default value "Text" of the Typography component. This PR corrects the behavior and makes sure we can bind to the error message, or the data (`undefined`) when a query fails. The error information is still preserved. 

Screenshot 2022-09-09 at 11 29 15

The runtime engine now tracks dependencies of expressions. We can build further on top of this to provide more insight into how the toolpad nodes are interconnected for https://github.com/mui/mui-toolpad/issues/523


Janpot:
🤔 Didn't mean to merge this yet, I thought I was on https://github.com/mui/mui-toolpad/pull/946

PR #986, opened 2022-09-12
Header layout towards UI revamp

Janpot:
- [x] Move preview and deploy to the center
- [x] Update save status icons
- [x]  Move toolpad ("Community") target into help menu
- [x] logo
- [x] ~theme~ See https://github.com/mui/mui-toolpad/pull/988. Just updated the primary main color

Screenshot 2022-09-13 at 18 53 31


packages/toolpad-app/src/toolpad/AppEditor/AppEditorShell.tsx
```diff
@@ -130,42 +129,37 @@ export default function AppEditorShell({ appId, ...props }: ToolpadShellProps) {
...
-          >
+          <Button variant="outlined" color="primary" onClick={handleCreateReleaseDialogOpen}>
             Deploy
           </Button>
         </Stack>
       }
+      status={
+        <Tooltip title={getSaveStateMessage(isSaving, hasUnsavedChanges)}>
+          <Box sx={{ display: 'flex', flexDirection: 'row', alignItems: 'center', mr: 3 }}>
+            {isSaving ? (
```
apedroferreira, line 155: This isn't really about these specific changes, but just to maybe discuss: Could we assume that if there are unsaved changes the app is already saving, and show the loader?

In which scenarios could there be unsaved changes, but the app isn't saving? And what could the user do about it in those cases not to lose progress?
I though about that already when I made changes to this, but I wasn't sure.

packages/toolpad-app/src/toolpad/AppEditor/AppEditorShell.tsx
```diff
@@ -130,42 +129,37 @@ export default function AppEditorShell({ appId, ...props }: ToolpadShellProps) {
...
-            color="inherit"
-            onClick={handleCreateReleaseDialogOpen}
-            endIcon={<RocketLaunchIcon />}
-          >
+          <Button variant="outlined" color="primary" onClick={handleCreateReleaseDialogOpen}>
             Deploy
           </Button>
         </Stack>
       }
+      status={
```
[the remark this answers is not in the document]
Janpot: The idea is to define an area where derived UI can place components "actions area", "status area". if we do `saveStatus` we may as well just make that a boolean and build the whole UI with icon directly in this file.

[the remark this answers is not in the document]
apedroferreira: ah so we can add more things in the future - `status` is fine then!

packages/toolpad-app/src/toolpad/AppEditor/AppEditorShell.tsx
Janpot: > Could we assume that if there are unsaved changes the app is already saving, and show the loader?

it's a three state:

1. all progress saved
2. local changes, not synced
3. syncing

I don't see what would be the problem with visualizing the three states, and what we would gain by merging 2. and 3.

> In which scenarios could there be unsaved changes, but the app isn't saving?

It's an edge-case, but maybe: make a change, then undo before the change was synced to the server

bharatkashyap:
Personally, I gained a lot from the visual distinction between "Preview" and "Deploy" due to the presence of the icons. I would miss them if they were removed.

packages/toolpad-app/src/toolpad/AppEditor/AppEditorShell.tsx
apedroferreira: I was just wondering if the icon changes could be confusing for the user, since most users might only care about whether the app is saved or saving - and get worried if it's neither saved or saving, especially if there's no way to save manually.
But it should be fine to keep this as is and keep it if it never causes any real issues - and the current behavior might work better with future changes. Just wanted to suggest the idea just in case.

apedroferreira:
> Personally, I gained a lot from the visual distinction between "Preview" and "Deploy" due to the presence of the icons. I would miss them if they were removed.

i think the rocket icon did help a bit visually

packages/toolpad-app/src/toolpad/AppEditor/AppEditorShell.tsx
Janpot: Updated

apedroferreira (approved):

packages/toolpad-app/src/toolpad/AppEditor/AppEditorShell.tsx
gerdadesign: I want to make sure I understand this convo — is the question here if it is valuable to show 3 icons (or other indicators) for the 3 states? 
🟢 All changes saved & synced
🟡 Saved changes not synced
🔴 Unsaved changes not synced

Could another case for 2 🟡 be an interrupted internet connection for the theoretical cloud version? Could this potentially happen in the self hosted version? 

----

On a related note, I'm not a huge fan of the icon that material has for cloud sync vs cloud done since it scales the cloud down a lot. I was also looking at just the regular sync icon instead — it's even more of a visual change but at least a similar scale. 
Sync

Then there's both a Sync Disabled or Sync Problem icon to indicate an issue like what number 2 describes. 
Sync Disabled Sync Problem

Janpot: > is the question here if it is valuable to show 3 icons

No, the conversation is about

1. all changes synced
2. there are local changes ready to be synced
3. actually busy syncing changes

> Could another case for 2 🟡 be an interrupted internet connection for the theoretical cloud version?

Good point, that's an additional state we should visualize.

> Could this potentially happen in the self hosted version?

Yes, it represents saving to the backend

PR #988, opened 2022-09-13
[ui] Add branding theme to Toolpad app

### LINKED ISSUE #516: Convert UI to MUI dark theme

### LINKED ISSUE #900: [EPIC] Toolpad UI needs to look visually appealing
### Duplicates

### Latest version

Related issues:

1. https://github.com/mui/mui-toolpad/issues/873
2. https://github.com/mui/mui-toolpad/issues/598
3. #516  
---
4. #870 
5. https://github.com/mui/mui-toolpad/pull/914#discussion_r964982827
Feedback & iteration history in [notion](https://www.notion.so/mui-org/Toolpad-UI-UX-revamp-d927c98f3540429f80400ecf7265fa32).

bharatkashyap:
- Closes #516 
- Closes #900 
- Visual demonstration 

https://user-images.githubusercontent.com/19550456/190144480-16596e89-e298-42d4-aa84-c3a0db5a236c.mov


Janpot:
1. How important is it to keep the theming in sync with the docs. We copy+paste, which means from now on there will be only divergence. Fixes to the theme in the docs won't get ported to toolpad, necessary overrides and fixes for toolpad won't make it upstream. Is having a shareable theme something to consider?
2. It looks like it's exporting things like `getMetaThemeColor` and `getDesignTokens`. It's has a bunch of light mode specific styles. If we're not going to support a light theme, should we remove these things?

bharatkashyap:
> 1. How important is it to keep the theming in sync with the docs. We copy+paste, which means from now on there will be only divergence. Fixes to the theme in the docs won't get ported to toolpad, necessary overrides and fixes for toolpad won't make it upstream. Is having a shareable theme something to consider?

> 2. It looks like it's exporting things like `getMetaThemeColor` and `getDesignTokens`. It also has a bunch of light mode specific styles in the component overrides. If we're not going to support a light theme, should we remove these things?

@gerdadesign can probably answer both. 

In my opinion:
For one, we should probably separate the theme from the docs theme since we probably don't want to create an external dependency with something as critical as the theme of the app.

For two, I think we might want to support light mode as well at some point, but given that it is bloating the size of the file right now, we can potentially remove the light theme overrides from this PR and add it in a follow on?

gerdadesign:
Good questions! 

1. This came up in a convo I had with @danilo-leal. It makes sense to me to separate the two — certain components may require a different styling in the context of Toolpad's dense UI vs in the context of the site's more spaced-out marketing needs. I think using this theme is a great starting point, but I think we currently have and eventually will have more customizations that the site simply doesn't need. 
👉   If we are still considering adjusting the brand of MUI Toolpad in the future, a separate theme could make that transition easier. 
👉   If we make adjustments to color / font on the MUI theme, that may require duplication here, but I would think that type of change would be rare. 

3. Agreed with @bharatkashyap. The [original issue](https://github.com/mui/mui-toolpad/issues/516) seems to advocate for dark mode and I think it provides a nice separation between the UI and the canvas, even when the canvas is in dark mode. Would def be open to adding a UI light mode in the future.

packages/toolpad-app/src/theme.ts
```diff
@@ -1,104 +1,723 @@
-import { createTheme } from '@mui/material/styles';
-import { red } from '@mui/material/colors';
-import type {} from '@mui/x-data-grid-pro/themeAugmentation';
-import type {} from '@mui/lab/themeAugmentation';
+import { deepmerge } from '@mui/utils';
+import ArrowDropDownRounded from '@mui/icons-material/ArrowDropDownRounded';
+import { createTheme, ThemeOptions, Theme, alpha } from '@mui/material/styles';
```
oliviertassinari, line 4: One bug, we miss a `color-schema`

Screenshot 2022-09-27 at 15 54 06

https://toolpad-pr-988.onrender.com/_toolpad/app/cl44kcwjj34669yofihq14yr0/pages/w713g5e

bharatkashyap: I've added a `color-scheme` CSS variable to our `DataGrid`

packages/toolpad-components/src/DataGrid.tsx
```diff
@@ -359,7 +359,10 @@ const DataGridComponent = React.forwardRef(function DataGridComponent(
   );
 
   return (
-    <div ref={ref} style={{ height: heightProp, minHeight: '100%', width: '100%' }}>
+    <div
+      ref={ref}
+      style={{ height: heightProp, minHeight: '100%', width: '100%', colorScheme: 'auto' }}
+    >
```
oliviertassinari, line 365: We are not solving https://github.com/mui/mui-toolpad/pull/988#discussion_r981281610

For example https://toolpad-pr-988.onrender.com/_toolpad/app/cl6xm8w6n0616arpkpgmr5219/pages/cl6xm8w9u0001arpkdgfndg7n

Screenshot 2022-10-06 at 00 32 34

```suggestion
    <div ref={ref} style={{ height: heightProp, minHeight: '100%', width: '100%' }}>
```

Plus, 'auto' is not an accepted value: https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme (Chrome might not yet autocomplete the CSS property correctly). It's like you had set the initial value, so `normal`: https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme#formal_definition.

oliviertassinari: For solving the root cause, I think that we miss the use of `<CssBaseline enableColorScheme />` for the theme inside the runtime toolpad app.

Janpot: > For solving the root cause, I think that we miss the use of <CssBaseline enableColorScheme /> for the theme inside the runtime toolpad app.

Opening https://github.com/mui/mui-toolpad/pull/1100 for that

gerdadesign (approved):
Great job so far! This is looking really good to me as a starting point and design-wise I think fine to merge and we can work together on some enhancements to tighten it up (tiny things like spacing, icon size, padding).

packages/toolpad-app/src/components/HarViewer.tsx
```diff
@@ -11,6 +11,22 @@ function fixLinks(elm: Element) {
   elm.querySelectorAll('a').forEach((link) => link.setAttribute('target', '_blank'));
 }
 
+function forceDarkMode(elm: Element) {
+  elm.querySelectorAll('svg').forEach((svg) => {
+    svg.setAttribute('fill', 'white');
+    const fullLabel = svg.getElementsByClassName('rect.label-full-bg');
+    if (fullLabel && fullLabel.length > 0) {
+      (fullLabel[0] as SVGElement).style.fill = '#5090D3'; // theme.palette.primary.main
```
oliviertassinari, line 19: I'm curious, why not import the theme? or at the very least `darkBlue[500]`?

bharatkashyap: Agreed, this is how it should have been done. Fixed

bharatkashyap:
Based on the discussion in https://github.com/mui/mui-toolpad/issues/516, I've added a theme mode toggle switch to the app header:

https://user-images.githubusercontent.com/19550456/195031255-42943abb-5b1b-4dc1-a988-976ccdd81a8a.mov

packages/toolpad-app/src/toolpad/ToolpadShell.tsx
```diff
@@ -178,6 +214,9 @@ function Header({ actions, status }: HeaderProps) {
           }}
         >
           {status}
+          {mode !== null ? (
```
apedroferreira, line 217: can the `mode` ever be `null`? from the code here looks like it can't

packages/toolpad-app/src/ThemeContext.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { unstable_useEnhancedEffect as useEnhancedEffect } from '@mui/material/utils';
+import { getDesignTokens, getThemedComponents, getMetaThemeColor } from './theme';
+
+interface ThemeProviderProps {
+  children?: React.ReactNode;
+}
+
+export type PaletteModeOptions = 'light' | 'dark';
+
+export const DispatchContext = React.createContext<
```
Janpot, line 14: You may want to use `createProvidedContext`?

packages/toolpad-app/src/ThemeContext.tsx
```diff
@@ -0,0 +1,75 @@
...
+export function ThemeProvider(props: ThemeProviderProps) {
+  const { children } = props;
+  const prefersDarkMode = useMediaQuery('(prefers-color-scheme: dark)');
+  const preferredMode = prefersDarkMode ? 'dark' : 'light';
+  const [paletteMode, setPaletteMode] = React.useState<PaletteModeOptions>(preferredMode);
+
+  React.useEffect(() => {
+    if (typeof window !== 'undefined') {
+      setPaletteMode(
+        JSON.parse(localStorage.getItem('paletteMode') as PaletteModeOptions) || preferredMode,
```
Janpot, line 29: Can we use our `useLocalStorage` hook?

packages/toolpad-app/src/ThemeContext.tsx
```diff
@@ -0,0 +1,75 @@
...
+      );
+    }
+  }, [preferredMode]);
+
+  React.useEffect(() => {
+    const metas = document.querySelectorAll('meta[name="theme-color"]');
+    metas.forEach((meta) => {
+      meta.setAttribute('content', getMetaThemeColor(paletteMode));
```
Janpot, line 37: Can [`next/head`](https://nextjs.org/docs/api-reference/next/head) be used for this purpose?

packages/toolpad-app/src/inspectorTheme.ts
```diff
@@ -1,15 +1,14 @@
...
-
-  BASE_BACKGROUND_COLOR: 'inherit',
-  TREENODE_FONT_FAMILY: 'inherit',
-  TREENODE_FONT_SIZE: 'inherit',
-  ARROW_FONT_SIZE: 'inherit',
-  TREENODE_LINE_HEIGHT: 'inherit',
-};
-
-export default inspectorTheme;
+function getInspectorTheme(mode: 'dark' | 'light'): InspectorTheme {
```
Janpot, line 3: Let's create a hook `useInspectorTheme` that internally calls MUI `useTheme`?

packages/toolpad-app/src/monacoEditorTheme.ts
```diff
@@ -1,5 +1,6 @@
-import theme from './theme';
+import { Theme } from '@mui/material';
 
-const monacoEditorTheme: string = theme.palette.mode === 'dark' ? 'vs-dark' : 'vs';
+const getMonacoEditorTheme = (theme: Theme): string =>
+  theme.palette.mode === 'dark' ? 'vs-dark' : 'vs';
 
-export default monacoEditorTheme;
+export default getMonacoEditorTheme;
```
Janpot, line 6: same, how about a `useMonacoTheme`?

packages/toolpad-app/src/ThemeContext.tsx
```diff
@@ -0,0 +1,75 @@
...
+  }, [theme.palette.mode]);
+
+  return (
+    <MuiThemeProvider theme={theme}>
+      <DispatchContext.Provider value={setPaletteMode}>{children}</DispatchContext.Provider>
+    </MuiThemeProvider>
+  );
+}
+
+export function useChangeTheme() {
```
Janpot, line 72: Can we simplify to something closer to

```ts
export function` useToolpadThemeMode () {
  const prefersDarkMode = useMediaQuery('(prefers-color-scheme: dark)');
  const preferredMode = prefersDarkMode ? 'dark' : 'light';
  return useLocalStorageState('...choose a key', preferredMode)
}
```

And use it for the toggle, as well as up above in the theme provider?

packages/toolpad-app/src/ThemeContext.tsx
bharatkashyap: That is what is being used to set the value; if I'm not wrong there's no good way to just get the value of a local storage variable without setting anything to it using the hook - so I went ahead with using this.

packages/toolpad-app/src/ThemeContext.tsx
```diff
@@ -0,0 +1,75 @@
...
+      },
+    });
+    nextTheme = deepmerge(nextTheme, getThemedComponents(nextTheme));
+
+    return nextTheme;
+  }, [paletteMode]);
+
+  useEnhancedEffect(() => {
+    if (theme.palette.mode === 'dark') {
+      document.body.classList.remove('mode-light');
```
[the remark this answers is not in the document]
bharatkashyap: Nope, missed removing this

packages/toolpad-app/src/ThemeContext.tsx
Janpot: The hook is not setting anything in localstorage until you actually call its returned `setState` function.
The hook will be a lot smarter than what this effect does. (e.g. sync across tabs, deal better with SSR)

packages/toolpad-app/src/ThemeContext.tsx
bharatkashyap: Do you suggest something like importing a `useTheme` hook into `ToolpadShell` and using 

```tsx
<Head> 
   <meta
            name="theme-color"
            content={getMetaThemeColor(theme.palette.mode)}
            media=`(prefers-color-scheme: ${theme.palette.mode})`
          />
</Head>
 ```

Comment (Janpot) [reply packages/toolpad-app/src/ThemeContext.tsx:37]: I believe it can just go here in `Themeprovider`, and we can then remove it from `_document.ts`

File: packages/toolpad-app/src/ThemeContext.tsx (hunk 5 above)
Comment (bharatkashyap) [reply packages/toolpad-app/src/ThemeContext.tsx:14]: I imagine that `createProvidedContext` and `useToolpadThemeMode` won't both be required together, since:

```ts
export const [useThemeModeDispatch, ThemeModeDispatchContextProvider] = createProvidedContext<React.Dispatch<React.SetStateAction<PaletteModeOptions>>>(
    'ThemeModeDispatch',
  );
 ```
 
 creates a `useThemeModeDispatch` which isn't needed?

Comment (Janpot) [reply packages/toolpad-app/src/ThemeContext.tsx:14]: Yeah, if we use `useLocalStorageState`, I think we may not need this context at all?

File: packages/toolpad-app/src/toolpad/ToolpadShell.tsx (hunk 4 above)
Comment (bharatkashyap) [reply packages/toolpad-app/src/toolpad/ToolpadShell.tsx:217]: I agree, the check is not needed. I think the type can remain the same, since (currently) we're not setting it to be "system" anywhere

File: packages/toolpad-app/src/ThemeContext.tsx (hunk 6 above)
Comment (bharatkashyap) [reply packages/toolpad-app/src/ThemeContext.tsx:29]: You're right - this works well. I was probably making an unknown error in the first implementation I did with `useLocalStorageState` (I imagine the issue was defining it twice, instead of wrapping it in one function and importing it in both places which is smarter)

File: packages/toolpad-app/src/ThemeContext.tsx (hunk 10 above)
Comment (bharatkashyap) [reply packages/toolpad-app/src/ThemeContext.tsx:72]: Agreed, works better

**Comment by bharatkashyap** [comment]
> > I've added a theme mode toggle switch to the app header
> 
> Thanks! Could we perhaps support the `prefers-color-scheme` MediaQuery like we do in the docs?
> 
> https://github.com/mui/material-ui/blob/master/docs/src/modules/components/AppSettingsDrawer.js#L55-L56 https://github.com/mui/material-ui/blob/master/docs/src/modules/components/AppSettingsDrawer.js#L62-L81
> 
> Since we don't (yet) have a user settings panel or menu, perhaps it could be an icon menu?

Makes sense, I've added support for system preference mode, and turned it into an icon menu where the icon is the theme mode that is currently selected, with a tooltip/`aria-label` which says "Change theme":


https://user-images.githubusercontent.com/19550456/195189216-f4fe11c7-109e-4724-be81-afee2e0890d8.mov

**Comment by gerdadesign** [comment]
Wondering if we think it's confusing to have two theme toggles in close proximity to one another or if it's clear which theme is referring to which?  
ChangeTheme

It may be as simple as changing the Theme tab (that controls the app) title to "App theme" or "Styles"? 

Or, as Matt mentioned, we may eventually have a "User settings" that this would fit under. When there are a lot of options, there may be a separate Settings icon, but if this is the only option for now, perhaps it can sit within the help menu? 

The tradeoff may be discoverability, but we might be able to increase clarity of which toggle refers to which theme.
Screen Shot 2022-10-13 at 2 56 37 PM

**Comment by bharatkashyap** [comment]
This is what I went with while implementing this, but I think the "Help and resources" nature of that menu gets lost if it starts to have a theme toggle, in which case it becomes a "Settings" menu.

> The tradeoff may be discoverability, but we might be able to increase clarity of which toggle refers to which theme. Screen Shot 2022-10-13 at 2 56 37 PM



> Wondering if we think it's confusing to have two theme toggles in close proximity to one another or if it's clear which theme is referring to which? ChangeTheme

I think this should be clear enough to someone using the editor, given that everything inside the sidebar menu is associated with the page being edited – and even if not, it should be abundantly clear when toggling them to be able to know which is which?

**Comment by gerdadesign** [comment]
>  I think the "Help and resources" nature of that menu gets lost if it starts to have a theme toggle, in which case it becomes a "Settings" menu.

Agree. What if we just went ahead and made it a settings button? 
Screen Shot 2022-10-13 at 4 30 15 PM

> it should be abundantly clear when toggling them to be able to know which is which

I do think it will become clear when they interact with the buttons! But it is extra work to get that understanding.

**Comment by prakhargupta1** [comment]
I think it makes sense to have Settings button instead of a theme icon as it is more relatable. Also, eventually theme will have to reside inside the Settings.

I don't think at this point the app theme and Toolpad theme should create confusion among the users.

**Comment by bharatkashyap** [comment]
> Agree. What if we just went ahead and made it a settings button? Screen Shot 2022-10-13 at 4 30 15 PM

Are you proposing this kind of a sidebar on click of the settings icon:
Screenshot 2022-10-14 at 2 03 31 PM

**Comment by Janpot** [comment]
> Are you proposing this kind of a sidebar on click of the settings icon:

I'm not sure that makes sense in the context of toolpad. Perhaps just a separate page?

**Comment by bharatkashyap** [comment]
> I'm not sure that makes sense in the context of toolpad. Perhaps just a separate page? I'm also fine making this a separate effort from this PR.

Agreed, we should take the theme mode toggle question into a follow up. For me, the current experience is okay to start and a broader settings menu is a suitable final destination for it live in.

What do you say @gerdadesign?

**Comment by gerdadesign** [comment]
Ahh, I meant we could go ahead and tuck theme under settings, but if you think it's better to wait that's fine too. 
rough sketch

**Comment by Janpot** [comment]
> Ahh, I meant we could go ahead and tuck theme under settings, but if you think it's better to wait that's fine too.

By separate effort, I mean in a second PR, to make it easier for reviewers. We can wait, but that was not my intention.

**Comment by bharatkashyap** [comment]
> > Ahh, I meant we could go ahead and tuck theme under settings, but if you think it's better to wait that's fine too.
> 
> By separate effort, I mean in a second PR, to make it easier for reviewers. We can wait, but that was not my intention when I said that.

Agree with moving the settings menu effort to a separate PR. To me, the theme toggle is an acceptable state for now to be able to merge this PR.

File: packages/toolpad-app/src/ThemeContext.tsx [hunk 12]
```diff
@@ -0,0 +1,60 @@
...
+import { deepmerge } from '@mui/utils';
+import useMediaQuery from '@mui/material/useMediaQuery';
+import { getDesignTokens, getThemedComponents, getMetaThemeColor } from './theme';
+import useLocalStorageState from './utils/useLocalStorageState';
+
+interface ThemeProviderProps {
+  children?: React.ReactNode;
+}
+
+type PaletteMode = 'light' | 'dark';
```
Janpot, line 13: I believe we can import this?
```tsx
import { PaletteMode } from '@mui/material';
```

packages/toolpad-app/src/ThemeContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+export type ThemeMode = PaletteMode | 'system';
+
+function usePreferredMode(): PaletteMode {
+  const prefersDarkMode = useMediaQuery('(prefers-color-scheme: dark)');
+  return prefersDarkMode ? 'dark' : 'light';
+}
+
+export function useThemeMode() {
+  return useLocalStorageState<ThemeMode>('toolpad-palette-mode', 'system');
```
Janpot, line 22: To make it more self documenting for consumers of this hook, maybe we should have it return a

```suggestion
  const [theme, setTheme] = useLocalStorageState<ThemeMode>('toolpad-palette-mode', 'system');
  return { theme, setTheme };
```
?

Janpot (approved):
👍 Two small tweaks proposed, then this is good to merge

PR #995, opened 2022-09-14
Self-document integration test mode

Janpot:
- Rename `TOOLPAD_INTEGRATION_TEST` to `TOOLPAD_ENABLE_CREATE_BY_DOM`: This signals the exact purpose to maintainers, to prevent abusing it for other purposes. e.g. to make a test pass. 
- This also allows us to configure our PR previews with this variable to make it easier to test production setups on PR previews.
- Throw a meaningful error when integration tests are run when this variable is not set. (https://github.com/mui/mui-toolpad/pull/658#discussion_r970665327)
- Make sure all integration tests use the same code path to create their applications, so they perform the check of this PR.


bharatkashyap (commented):
Got it!

apedroferreira (approved):

Issue #592, opened 2022-06-23
[EPIC] Online demo version


Janpot:
Allow evaluation of Toolpad through an online demo. Toolpad doesn't support authentication yet. The goal is to demonstrate a subset of the app building experience.

The demo version will be unauthenticated. We will disable all possibilities of entering sensitive data.

### To Do:
- [x] Introduce `TOOLPAD_DEMO` configuration to enable demo version
  - [x] remove the applications list from the overview page. Just show the app creation dialog
  - [x] clearly mark in the UI that demo is enabled.
  - [x] clearly mark that the app is ephemeral and may be deleted at any time without notice. Don't use for anything else than evaluation purposes.
  - [x] disable all datasources
    - ~enable only the REST datasource, but disable headers/authentication. Enforce a baseurl?~
    - [x] Expand the fake movies datasource and enable it in demo only
  - [x] add captcha when creating application
- [x] Currently saving the application deletes and recreates the whole app in the db, replace it with an incremental update to reduce the throughput on the database.
- [x] host using render.com
- [x] usage metrics: Add Google analytics?
- [x] developer metrics: 
    - database workload (Can we monitor this sufficiently on render.com)
    - application logs
    - application errors (Sentry?)
    - monitor REST dataSource requests
- [x] Disable the "make application public" feature for the demo mode. All apps are public
- [x] App names have to be unique. Let's not complicate the `createApp` logic too much. We either add a unique id to name the user gives, or we just generate a unique name and don't let them specify one. They can't list their apps so what'd be the purpose of naming it anyway?
- [ ] ...

apedroferreira:
Will do in 3 separate PRs:
- application changes
- set up metrics
- deployment

prakhargupta1:
Correct me if I am wrong, we will provide 3 pre-enabled connections:
1. Stats app public API (Template 1)
2. Dog app API (Template 2)
3. Fake movies API


@bytasv Regarding examples in our docs, is there any other connection that you think we should provide?

bytasv:
I think it really depends how much freedom we want to give. There is also google sheets and postgres that must use connection. But it seems that for demo purposes all we gonna allow us a short list of allowed URLs that users can try to use for fetching in fetch or function datasource?

apedroferreira:
It shouldn't be too difficult to add more APIs similar to movies just to enable more options, as long as they have about the same functionality.

Janpot:
[...]

PR #1002, opened 2022-09-14
Online demo

apedroferreira:
Online demo version of the app:

Related issue: https://github.com/mui/mui-toolpad/issues/592

- show demo badge
- hide list of apps on app start, with alert about demo version
- Only allow fake movies connection (works with local data), or fetch connection with the base URLs from the prebuilt apps we're going to show (COVID and dog APIs)
- Restrict other fetch query options in the UI and also server-side so that demo users can't access them

App start:

https://user-images.githubusercontent.com/10789765/191091167-3240d3d2-cbca-491f-94a8-4524d2904ac8.mov

Fake movies connection:

https://user-images.githubusercontent.com/10789765/191091195-50fc7fe8-0204-4afc-b49e-27a5ddd03abd.mov

Limited fetch connection:

https://user-images.githubusercontent.com/10789765/191091250-a67f6bb0-4130-4975-84b3-cd4e16633381.mov

I've also ran into this bug with connections, which we might want to fix before launching the demo app: https://github.com/mui/mui-toolpad/issues/1032


packages/toolpad-app/src/config.ts
```diff
@@ -66,6 +68,7 @@ const runtimeConfig: RuntimeConfig =
     ? {
         // Define runtime config here
         integrationTest: !!process.env.TOOLPAD_INTEGRATION_TEST,
+        isDemo: !!process.env.TOOLPAD_DEMO,
```
Janpot, line 71: Why can't this be build time config?

Janpot: we can use build time config in the UI. e.g. TOOLPAD_VERSION

packages/toolpad-app/src/toolpadDataSources/rest/server.ts
```diff
@@ -138,6 +139,21 @@ async function execBase(
     resolveBindableEntries(fetchQuery.headers || [], params),
   ]);
 
+  if (process.env.TOOLPAD_DEMO) {
+    const demoUrls = DEMO_BASE_URLS.map((baseUrl) => baseUrl.url);
+
+    const hasNonDemoConnectionParams =
+      !demoUrls.includes(connection?.baseUrl || '') ||
+      (connection?.headers && connection?.headers.length > 0) ||
```
bytasv, line 147: Would `connection?.headers?.length > 0` work the same way? 🤔

packages/toolpad-app/src/toolpadDataSources/rest/server.ts
```diff
@@ -138,6 +139,21 @@ async function execBase(
...
+      !!connection?.authentication;
+    const hasNonDemoQueryParams =
+      fetchQuery.method !== 'GET' || (fetchQuery.headers && fetchQuery.headers.length > 0);
+
+    if (hasNonDemoConnectionParams || hasNonDemoQueryParams) {
+      throw new Error(`Cannot use unsupported features in demo version.`);
```
bytasv, line 153: This sounds like we do not support these features at all, maybe better wording would be something like `Can not use these features in demo version`?

packages/toolpad-app/src/toolpadDataSources/rest/server.ts
bytasv: It's not doing `undefined > 0` it's "short circuiting" and simply returning `undefined` if `headers` are not defined, so should be doing pretty much the same thing just gonna be shorted: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

IMO it's weird to use that on `connection` but not use that on `headers` since that's why the operator exists

apedroferreira: ah my bad, i think `connection?.headers.length > 0` is also not good...
thanks for pointing this out, i'll improve it.

apedroferreira: i've changed it, i think it should be better now?

bytasv: I'm not sure I understand, what exactly has been changed? addition of `!!` in front? That shouldn't really change anything since `undefined` is falsy value anyways and doing `!!` on undefined would be same in all cases. I suggest reading on optional chaining and how it works it might clear up the confusion here or you could see what is the result of compiled `connection?.headers?.length > 0` [here in babel repl](https://babeljs.io/repl/#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=MYewdmCmwC4JbgPwDoAWkCGATSAnAzigDaRgDmMqABAHxUAMQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-2&prettier=false&targets=&version=7.19.2&externalPlugins=&assumptions=%7B%7D)

apedroferreira: i know about optional chaining, i just added `!!` so that the variable is a `boolean` instead of `undefined` just for it to be cleaner, but that change was not to address your comment.

what changed is that i got rid of the `?` in `connection?.headers.length > 0` because it results in `undefined > 0` which does result to `false` but isn't good logic, and the optional chaining wasn't needed there.

bytasv: Not exactly sure what is meant by "isn't good logic" - do you see where it could break or behave in a way that is not expected if you used optional chaining? Because if left side evaluates to `undefined` the entire condition will always be false. The only case when it would be true is if everything was defined and length was more than 0 - exactly what you want.
Anyways if you prefer to keep it as is I'm fine, I just wanted to point out that same can be achieved with `connection?.headers?.length > 0` which is shorter and (I personally find it) easier to understand way. Unless I'm totally wrong here and it can actually break under some circumstances that I'm not familiar with?

apedroferreira: yeah i just didn't want to use "javascript magic" and have more explicit types, but doesn't mean anything would break - just personal preference. 
doing it like you say is less verbose for sure though - probably a bit easier to read.
i don't mind changing it to your suggestion, it's just a style difference really.

apedroferreira: actually just tired your suggestion and Typescript complains about it so let's keep it as is if it's ok!

bytasv: Sure, let's keep it as is 👍 . Would you mind sharing what sort of error/warning TS threw? Wanna know what to look for in the future 🤔 
I'm trying to guess that it complains comparing `undefined to number` 🤔  In that case something like that could be used:
```
(connection?.headers?.length || 0) > 0
```

bytasv (approved):

packages/toolpad-app/src/toolpadDataSources/rest/server.ts
apedroferreira: Screen Shot 2022-09-20 at 18 59 35

bytasv: Thanks! My bad, didn't compile TS in my head 😅 anyways, it was an interesting find (regarding TS) - seems that there is even an open issue discussing this on TS repo: https://github.com/microsoft/TypeScript/issues/45543 maybe in the future this will change who knows :)

prakhargupta1:
Duplication in app name will be very common as we won't show list of apps and also won't allow creating a duplicate entry. Can we do something about it?

apedroferreira:
@prakhargupta1 @bytasv I've added logic to deal with duplicate app names in the demo, it will add "(#2)", "(#3)", etc as more apps with the same name I created. 
I also tried to make it as secure as possible, so people can't create issues when messing with app names.

Feel free to try it when it's live in the preview, or to review the commit with this change!

PR #1004, opened 2022-09-15
Add instruction for testing a random commit on master

Janpot:


CONTRIBUTING.md
```diff
@@ -104,7 +104,13 @@ Please have a look at our general guidelines for sending pull requests [here](ht
 
 1. Review/merge the PR — once the PR is merged to `master` a Docker image is built for the commit, tagged with the commit SHA and then pushed to Docker Hub, just like for any other commit in `master`.
 
-1. Wait for the Docker build to finish. (You can find a job [here](https://app.circleci.com/pipelines/github/mui/mui-toolpad?branch=master))
+1. Wait for the Docker build to finish. (You can find the job in the [CircleCI pipelines](https://app.circleci.com/pipelines/github/mui/mui-toolpad?branch=master)). Note that you can always verify the built docker image by running it as
```
bytasv, line 107: you **can** or you **should** verify? Does it mean manually testing the app in random places? It's a bit unclear what exactly is verification, is it enough that the docker image starts without errors?

Janpot: > you can or you should verify?

You can. We have some smoke tests that should catch most of the mundane changes, but you may have added a change that affects the docker build or push specifically and want more tools to verify everything is where it's supposed to be.

bytasv: Is it possible to define conditions when that should be done? i.e. when I'm doing release looking at commits I cant tell whether this check is needed or not, how do I know without asking around?

Janpot: Not really, there is no condition. it's mostly about **knowing this functionality exists** so that you can apply it in certain circumstances:

* If you think there's a change in there that may affect the docker image in the registry that won't be caught by the smoke test on CI. I guess that happens mostly when you're doing work on the release process itself.
* Or for debugging purposes, if you want to bisect commits to find where a bug was introduced.
* Or we can communicate it to customers to test a fix.
* ...

In general during the weekly release, **you** probably won't need to do it. But for the team members that work more on maintaining the release process itself it's interesting to have this option.

PR #1007, opened 2022-09-15
Stronger integration test for duplication

bharatkashyap:

### DIFF
--- a/test/integration/appDuplicate.spec.ts
+++ b/test/integration/appDuplicate.spec.ts
@@ -10,10 +10,18 @@ test('duplicate app from home flow', async ({ page }) => {
   await homeModel.goto();
   const app = await homeModel.createApplication({ dom });
   await homeModel.goto();
-  await homeModel.duplicateApplication(app.name);
+  const duplicateApp = await homeModel.duplicateApplication(app.name);
   await homeModel.goto();
   await expect(page.locator(locators.toolpadHomeAppRow(`${app.name} (copy)`))).toBeVisible();
   await homeModel.duplicateApplication(app.name);
   await homeModel.goto();
   await expect(page.locator(locators.toolpadHomeAppRow(`${app.name} (copy 2)`))).toBeVisible();
+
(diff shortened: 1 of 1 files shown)


test/integration/appDuplicate.spec.ts
```diff
@@ -10,10 +10,18 @@ test('duplicate app from home flow', async ({ page }) => {
   await homeModel.goto();
   const app = await homeModel.createApplication({ dom });
   await homeModel.goto();
-  await homeModel.duplicateApplication(app.name);
+  const duplicateApp = await homeModel.duplicateApplication(app.name);
   await homeModel.goto();
   await expect(page.locator(locators.toolpadHomeAppRow(`${app.name} (copy)`))).toBeVisible();
   await homeModel.duplicateApplication(app.name);
   await homeModel.goto();
```
apedroferreira, line 17: maybe no need to check twice? unless there's something specific about the logic we're trying to cover here

test/integration/appDuplicate.spec.ts
```diff
@@ -10,10 +10,18 @@ test('duplicate app from home flow', async ({ page }) => {
...
   await expect(page.locator(locators.toolpadHomeAppRow(`${app.name} (copy 2)`))).toBeVisible();
+
+  await page.goto(`/_toolpad/app/${duplicateApp.id}/preview/pages`);
```
apedroferreira, line 20: the `ToolpadRuntime` model might be useful for this part
checkout out the `components` integration test, maybe this test could do something similar (if it's not working)

test/integration/appDuplicate.spec.ts
bharatkashyap: Trying to check if the second duplicate is appropriately named

test/integration/appDuplicate.spec.ts
bharatkashyap: Thanks! I'll check that out

test/integration/appDuplicate.spec.ts
apedroferreira: got it, makes sense then!

Issue #664, opened 2022-07-19
Support making applications public


Janpot:
Support an option that makes a Toolpad application hosted publicly on a toolpad instance. By "public" is meant "not requiring authentication".

PR #1009, opened 2022-09-15
Support public applications

Janpot:
Closes https://github.com/mui/mui-toolpad/issues/664

https://user-images.githubusercontent.com/2109932/190437201-1ddd07fb-ae69-44d1-9f6d-829ba2dc94ed.mov


packages/toolpad-app/middleware.ts
```diff
@@ -1,24 +1,27 @@
...
-    const [user, pwd] = atob(auth).toString().split(':');
-
-    if (user === config.basicAuthUser && pwd === config.basicAuthPassword) {
-      return NextResponse.next();
-    }
+  if (checkBasicAuthHeader(req.headers.get('authorization'))) {
+    return NextResponse.next();
   }
 
   return new Response(null, {
```
apedroferreira, line 27: we might need to change this middleware according to some changes here:
https://nextjs.org/docs/messages/middleware-upgrade-guide

basically using `rewrite/redirect` for the response, and maybe we can use the `config` export for the whitelisting?

example here: https://github.com/vercel/examples/tree/main/edge-functions/basic-auth-password

packages/toolpad-app/pages/deploy/[appId]/[[...path]].tsx
```diff
@@ -14,6 +17,20 @@ export const getServerSideProps: GetServerSideProps<ToolpadAppProps> = async (co
...
 
+  const app = await getApp(appId);
+  if (!app) {
+    return {
+      notFound: true,
+    };
+  }
+
+  if (!app.public) {
+    if (!checkBasicAuth(context.req, context.res)) {
```
Comment (apedroferreira) [inline packages/toolpad-app/pages/deploy/[appId]/[[...path]].tsx:28]: would it be possible to throw an error and make it trigger the proper response?

Comment (apedroferreira) [reply packages/toolpad-app/pages/deploy/[appId]/[[...path]].tsx:28]: maybe nevermind this comment I made - i guess throwing errors is not the right away of returning error responses in Next.js serverless functions.

Comment (Janpot) [reply packages/toolpad-app/pages/deploy/[appId]/[[...path]].tsx:28]: I'm planning to split this into two functions. one that checks the request headers and one that sets the response headers. that should make it explicit enough.

packages/toolpad-app/middleware.ts
Janpot: Empty body responses are still be allowed. This code has already been passed through the upgrade guide. (See this example where an empty body response is created: https://nextjs.org/docs/messages/middleware-upgrade-guide#after-1)

packages/toolpad-app/src/components/DialogForm.tsx
```diff
@@ -4,5 +4,4 @@ export default styled('form')({
   overflowY: 'auto',
   display: 'flex',
   flexDirection: 'column',
-  width: 400,
```
Janpot, line 7: @apedroferreira I'm reverting this, I don't think it's the right place to define dialog width, it's problematic when fullWidth is used on the dialog that contains this form. The component is intended to be reusable regardless of width of the form (it's in the components folder). Instead we should control dialog width using `maxWidth`  and `fullWidth` on the Dialog component

packages/toolpad-app/middleware.ts
apedroferreira: ok so empty body still works, my bad
about the config sounds good too

packages/toolpad-app/src/components/DialogForm.tsx
apedroferreira: ok, i remember trying other things before but for some reason there was a lot of shifting in the dialog
might just be unfamiliarity with MUI components
feel free to remove it

PR #1013, opened 2022-09-16
[security] Improve the HTTP headers for security

oliviertassinari:
I'm working on improving the perceived security of MUI to solve some of the friction that we have with the sales on MUI X. Some companies are using https://www.riskrecon.com/:

Screenshot 2022-09-17 at 00 43 36

we are the worst. In this PR, I'm going after:

Screenshot 2022-09-17 at 00 44 35

This PR implements most of https://nextjs.org/docs/advanced-features/security-headers.

### DIFF
--- a/packages/toolpad-app/next.config.mjs
+++ b/packages/toolpad-app/next.config.mjs
@@ -57,19 +57,46 @@ const regexEqual = (x, y) => {
   );
 };
 
+const securityHeaders = [
+  {
+    key: 'Strict-Transport-Security',
+    value: 'max-age=31536000; includeSubDomains; preload',
+  },
+  {
+    key: 'X-Frame-Options',
+    value: 'SAMEORIGIN',
+  },
+  {
+    // Force the browser to trust the Content-Type header
+    // https://stackoverflow.com/questions/18337630/what-is-x-content-type-options-nosniff
+    key: 'X-Content-Type-Options',
+    value: 'nosniff',
+  },
+  {
+    key: 'X-XSS-Protection',
+    value: '1; mode=block',
+  },
+  {
+    key: 'X-XSS-Protection',
+    value: '1; mode=block',
+  },
+  {
+    key: 'Referrer-Policy',
+    value: 'strict-origin-when-cross-origin',
+  },
+];
+
 const NEVER = () => false;
 
 export default /** @type {import('next').NextConfig} */ ({
   reactStrictMode: true,
-
+  poweredByHeader: false,
   eslint: {
     // We're running this as part of the monorepo eslint
     ignoreDuringBuilds: true,
   },
-
   // build-time env vars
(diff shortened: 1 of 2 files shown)


packages/toolpad-app/next.config.mjs
```diff
@@ -57,19 +57,46 @@ const regexEqual = (x, y) => {
...
+    value: 'strict-origin-when-cross-origin',
+  },
+];
+
 const NEVER = () => false;
 
 export default /** @type {import('next').NextConfig} */ ({
   reactStrictMode: true,
-
+  poweredByHeader: false,
```
oliviertassinari, line 93: Removes 

Screenshot 2022-09-17 at 00 47 37

packages/toolpad-app/next.config.mjs
```diff
@@ -57,19 +57,46 @@ const regexEqual = (x, y) => {
   );
 };
 
+const securityHeaders = [
```
Janpot, line 60:
```suggestion
// See https://nextjs.org/docs/advanced-features/security-headers
const securityHeaders = [
```

packages/toolpad-app/next.config.mjs
```diff
@@ -57,19 +57,46 @@ const regexEqual = (x, y) => {
...
+  {
+    key: 'Strict-Transport-Security',
```
Janpot, line 62: this one needs to get optional for self hosters

Janpot: in fact I'd even argue the responsibility for this one doesn't belong in the application. I'm on my phone so i can't verify, but I'd be surprised if render.com doesn't already set it.

oliviertassinari: Ok, I'm removing it for this PR, we can move more progressively on this header.

oliviertassinari: > but I'd be surprised if render.com doesn't already set it.

They don't https://toolpad-gfpr.onrender.com/. But their certificate is issued by Cloudflare while on https://master--toolpad.mui.com/ the certificate is issued by MUI

packages/toolpad-app/next.config.mjs
```diff
@@ -57,19 +57,39 @@ const regexEqual = (x, y) => {
   );
 };
 
+// See https://nextjs.org/docs/advanced-features/security-headers
+const securityHeaders = [
+  {
+    key: 'X-Frame-Options',
+    value: 'SAMEORIGIN',
+  },
```
oliviertassinari, line 65: This might also create issues for self-hosters. I wouldn't be surprised if we get requests to remove this header. Now, it's probably safer to have it.

Janpot: agreed, we can add an option to whitelist /deploy/* over time of there's interest

Janpot (approved):

Issue #993, opened 2022-09-14
Allow nesting inside of the Paper component


gerdadesign:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Currently, after placing the Paper component onto the canvas, you cannot add more than one other component onto the paper. 
<img width="300" alt="Screen Shot 2022-09-13 at 3 13 04 PM" src="https://user-images.githubusercontent.com/106991024/189994298-e6d1e335-b59d-4528-96e3-0ec7b3cea5fa.png">

The paper component should allow infinite nesting inside. 

### Examples 🌈

A mockup of many nested components within paper components. 
![Nested Components in Paper](https://user-images.githubusercontent.com/106991024/190038363-b9e23b21-da4d-4fa4-945f-ab83fccad1ef.png)

WangLarry:
RenderOverlay is very complex. It is hard to supply this.

PR #1031, opened 2022-09-19
Improve Paper component

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/993

- Allow building layouts inside Paper component with the new `layout` slot type
- Force Paper to take full-width of its container to keep resizing logic predictable (as it would be in the page itself)

I made Paper full-width because the resizing breaks easily and becomes very finicky if the Paper width changes when new components are added inside. To be able to resize widths inside an inner layout with content-dependent width would add a lot of edge cases to the resizing logic and make it quite more complicated, so not sure if it's worth the effort of looking into, it might be a very difficult problem to solve or may not even have a good solution. I could try if we find it really important, but I feel like it would take significant time and effort.

https://user-images.githubusercontent.com/10789765/197014848-2538e6b6-e9c8-4859-a49d-0b95fb8de28f.mov


apedroferreira:
I think this is good now, it was only pending because of a bug we've fixed in the meantime.

packages/toolpad-components/src/Paper.tsx
```diff
@@ -1,24 +1,43 @@
...
 }
 
 export default createComponent(Paper, {
+  layoutDirection: 'both',
   argTypes: {
     elevation: {
       typeDef: { type: 'number', minimum: 0 },
       defaultValue: 1,
     },
+    direction: {
```
Janpot, line 27: Shouldn't we allow users to build UI in a paper just like it is in a page? If you drag at the bottom it creates an extra row, if you drag to the side it creates an extra column?

I feel like these properties `direction` and `gap` are a step back.

bytasv: I agree, I'd expect the behaviour that Jan described

apedroferreira: I thought about that, I guess I had tried it before but gave up on it for some reason. 
Might not be super easy but I'm trying again, I think it should be possible.

apedroferreira: I've made it so we can build layouts in certain slots, in this case inside the Paper component.
For that I created a new slot type, `layout`.
Let me know what you think!

bytasv:
Hey @apedroferreira I tested this around and I think I found a weird glitch 🤔 seems that everything is working as long as you are adding items, but I tried removing item and after that I could put anything into the paper component anymore 👇 

https://user-images.githubusercontent.com/437214/197123767-bf13b990-cab2-4fb1-a670-1425ae64279f.mov

apedroferreira:
> Hey @apedroferreira I tested this around and I think I found a weird glitch 🤔 seems that everything is working as long as you are adding items, but I tried removing item and after that I could put anything into the paper component anymore 👇
> 
>  Screen.Recording.2022-10-21.at.09.01.25.mov

There must be a few more cases where the logic depends on something being a page but now I need to account for inner layouts too.
I'll figure it out, thanks for finding this!

apedroferreira:
> Hey @apedroferreira I tested this around and I think I found a weird glitch

I've added a fix, seems good now

bytasv (approved):
Seems to be working well now 👍

Issue #998, opened 2022-09-14
Add error reporting


apedroferreira:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

From https://github.com/mui/mui-toolpad/issues/592.
Add automatic error reporting to Toolpad.

### Examples 🌈

I'm most familiar with Sentry but feel free to suggest others.
The basic Next.js integration should probably be enough to start with?

Janpot:
I'm fine with Sentry. I think we can disable it for the self-hosted version for now. I'd keep telemetry for self-hosted Toolpad out of scope for this demo feature.

oliviertassinari:
We have Sentry set up for https://mui.com/store/ https://www.notion.so/mui-org/Sentry-5784d0f006b9464b847953c3b23e5311. We also have the credentials here https://www.notion.so/mui-org/Accounts-945f0ccb9f7749ccb9e84b9eb9b13ec7#db24eea458014567a130b0849229356b.

oliviertassinari:
@apedroferreira Regarding the `website` label, it's described at:

<img width="316" alt="Screenshot 2022-09-16 at 01 28 53" src="https://user-images.githubusercontent.com/3165635/190525953-caee7682-c719-4fc5-9697-57d9f1eae844.png">

So, right now, it's about one page: https://mui.com/toolpad. However, It seems that we will use inside toolpad, the app.

apedroferreira:
> @apedroferreira Regarding the `website` label, it's described as:
> 
> <img alt="Screenshot 2022-09-16 at 01 28 53" width="316" src="https://user-images.githubusercontent.com/3165635/190525953-caee7682-c719-4fc5-9697-57d9f1eae844.png">
> 
> https://www.notion.so/mui-org/GitHub-labels-af1ef18d3f534f1dbc086aac87b819e8
> 
> So, right now, it's about one page: https://mui.com/toolpad/. It seems that we are tracking the addition of Sentry inside toolpad, the product.

Oh, I didn't notice you had removed the label, sorry!
I originally took the label from the original issue at https://github.com/mui/mui-toolpad/issues/592, but yeah I guess it's not the best label, it's probably not the best for other issues I've added it to either...

PR #1043, opened 2022-09-20
Add Sentry

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/998

Adds Sentry to Toolpad - reporting errors in client and server.

~~Most of this configuration should be correct, but I cannot get any errors to create issues in the MUI Sentry account... 
Not sure what might be wrong, I've tried with a personal account and error reporting seems to work, so the problem seems to be related with the MUI Sentry account in specific.~~ (Nevermind, I found and fixed the problem)

- [x] API route errors create issues correctly
- [x] Frontend errors create issues correctly (without being stopped by error boundaries, for example)
- [x] Datasource/query errors create issues correctly?


apedroferreira:
Instructions to try to debug issue with MUI Sentry:
1. Create a blank next project with `npx create-next-app@latest --typescript`
2. Inside the project, install Sentry with `yarn add @sentry/nextjs`
3. Inside the project, run `npx @sentry/wizard -i nextjs`
4. Reconcile the `next.config.js` generated file with the main `next.config.js` in the app
5. In `pages/index.tsx`, for example, before returning inside the component add `throw new Error("error test");`

These instructions work fine if I use a personal Sentry account, but not the MUI account.
Would appreciate some help in figuring this out!

apedroferreira:
@oliviertassinari Any ideas on why errors do not create issues in the MUI Sentry account? 
I checked our plan quotas and there should still be room for a few more issues.

bharatkashyap (approved):
Looks good! Is there a way to test out whether we're receiving the logs? 
Also, do we want to see logs in deploy previews?

apedroferreira:
> Looks good! Is there a way to test out whether we're receiving the logs?

Should be working in the preview now, I've set up the environment variables.
Our credentials for Sentry are in https://www.notion.so/mui-org/Sentry-5784d0f006b9464b847953c3b23e5311.

> Also, do we want to see logs in deploy previews?

I think the plan for now is to enable Sentry in demo only.
I've enabled it for this specific preview though with the `IS_SENTRY_ENABLED` environment variable.

Janpot:
> I've enabled it for this specific preview though with the IS_SENTRY_ENABLED environment variable

not defining the TOOLPAD_SENTRY_DSN variable means disabling sentry

packages/toolpad-app/src/config.ts
```diff
@@ -30,7 +30,9 @@ export type BuildEnvVars = Record<
   // The current Toolpad version
   | 'TOOLPAD_VERSION'
   // The current Toolpad build number
-  | 'TOOLPAD_BUILD',
+  | 'TOOLPAD_BUILD'
+  | 'IS_SENTRY_ENABLED'
+  | 'SENTRY_DSN',
```
Janpot, line 35: sure this is build time only? i.e. we bake it in the image? self hosters can't add their own sentry?

packages/toolpad-app/src/toolpad/Home/index.tsx
```diff
@@ -557,7 +557,13 @@ export default function Home() {
         <Typography variant="h2">Apps</Typography>
         <CreateAppDialog open={createDialogOpen} onClose={() => setCreateDialogOpen(false)} />
         <Toolbar variant={'dense'} disableGutters sx={{ justifyContent: 'space-between' }}>
-          <Button onClick={() => setCreateDialogOpen(true)}>Create New</Button>
+          <Button
+            onClick={() => {
+              setCreateDialogOpen(true);
```
Janpot, line 562: is this intended?

packages/toolpad-app/sentry.client.config.js
```diff
@@ -0,0 +1,17 @@
+// This file configures the initialization of Sentry on the browser.
+// The config you add here will be used whenever a page is visited.
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/
+
+import * as Sentry from '@sentry/nextjs';
+
+const SENTRY_DSN = process.env.SENTRY_DSN || process.env.NEXT_PUBLIC_SENTRY_DSN;
```
Janpot, line 7:
```suggestion
const SENTRY_DSN = process.env.SENTRY_DSN;
```

packages/toolpad-app/src/config.ts
apedroferreira: > not defining the TOOLPAD_SENTRY_DSN variable means disabling sentry. no need to add more redundant variables. also, I'd keep the convention of prefixing Toolpad variables with TOOLPAD_

Makes sense, I was thinking of this env variable as more like an on/off toggle but you're right, it's redundant, better to just use SENTRY_DSN.

packages/toolpad-app/src/toolpad/Home/index.tsx
apedroferreira: Nope, I changed it back to how it was before, was just trying some things earlier.

packages/toolpad-app/sentry.client.config.js
apedroferreira: What's the right way to have a client-side readable environment variable in our project, if it's not a build-time variable? Is it using the runtime config, like I've changed it to now?

packages/toolpad-app/next.config.mjs
```diff
@@ -79,94 +80,123 @@ const securityHeaders = [
...
+const sentryWebpackPluginOptions = {
+  // Additional config options for the Sentry Webpack plugin. Keep in mind that
+  // the following options are set automatically, and overriding them is not
+  // recommended:
+  //   release, url, org, project, authToken, configFile, stripPrefix,
+  //   urlPrefix, include, ignore
+
+  silent: true, // Suppresses all logs
+
+  dryRun: !process.env.TOOLPAD_SENTRY_DSN,
```
Janpot, line 92: Hmm, I'm not sure this would work well for us. This only gets executed at build time. Would it still work if we don't upload sourcemaps and instead rely on production sourcemap urls? https://nextjs.org/docs/advanced-features/source-maps#configuration-flag

apedroferreira: Which thing in specific would not work well?
I added this `dryRun` condition because if `dryRun` has its default value Sentry seems to require an auth token, but an auth token should not be required if someone is not using Sentry.
This logic seems to fix that issue both locally and in production.

apedroferreira: i could add a comment making it more clear though if that's the confusion

apedroferreira: if it's about `sourceMaps` and there's a chance they could still be visible in the browser, we can try not uploading them and setting them manually in Sentry just for safety

Janpot: I don't mind them being visible in the browser. This is an OS project after all 🙂 .
My point is basically, if we have sourcemaps publicly available, hosted from the toolpad application, do we still need to upload them for Sentry? Will Sentry find the sourcemaps if our js files have `//# sourceMappingURL` directives? If yes, then I'd just disable uploading them altogether.

bytasv: Seems that we should be fine having soruce-maps hosted with app itself: https://docs.sentry.io/platforms/javascript/sourcemaps/#2-provide-source-maps-to-sentry

PR #1045, opened 2022-09-21
[docs] Add POC to capture screenshots for documentation

bytasv:
I went ahead and experimented a bit with playwright and it's screenshot capturing capabilities. I did manage to capture screenshots of a few most common cases:
1. Regular full screen
2. Wait for Iframe full screen
3. Wait for dialog only element

<del>The problem that I couldn't figure out yet was whether it's possible to capture screenshots in a higher resolution. After long searches I put that aside and asked a question on playwright repo.</del>
edit: I managed to finally generate high quality screenshots

IMO the overhead to generate screenshots is really minimal especially if we have certain scenarios in place. Even if we don't or don't want to use existing ones we can create lightweight scenarios just for screenshot capturing


oliviertassinari:
@siriwatknp Wrote this https://github.com/mui/material-ui/blob/e8c2875908514ddb6e9e9ed036d959fddb26ae21/docs/scripts/generateTemplateScreenshots.ts, it might help

bytasv:
@oliviertassinari seems that those images are "regular" quality as well, also I don't think there is something in the playwright code that would make it generate higher quality images 🤔

oliviertassinari:
@bytasv yes, there are, it would be great to fix it. 

The main value I find in this prior art is that the script is not run as part of the e2e tests, but as a standalone script to more easily update the screenshot with a PR when needed.

bytasv:
> The main value I find in this prior art of Jun is that the script is not run as part of the e2e tests

Oh for sure, this shouldn't run as part of testing. I just didn't want to invest more time into creating separate scripts if we decide not to move forward with this and throw all away. If we would move forward, then definitely should do it properly

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/NodeDropArea.tsx
```diff
@@ -362,7 +362,9 @@ export default function NodeDropArea({
       {isEmptySlot && slotRect ? (
         <EmptySlot style={absolutePositionCss(slotRect)}>
           <AddCircleOutlineIcon fontSize="inherit" sx={{ mb: 0.2 }} />
-          <Typography variant="caption">Drop component here</Typography>
+          <Typography variant="caption" data-test-id="drop-area">
```
apedroferreira, line 365: maybe this was just for the POC but as a general practice we're trying to avoid using `data-test-id` unless strictly necessary - we've decided it's a better practice to use more visual properties such as the text content of the element to identify elements, or accessibility props

this is according to Playright's suggested best practices too: https://playwright.dev/docs/selectors#best-practices

apedroferreira:
this seems cool and easy to use!
it's just a shame that the images are a bit small, but maybe they can still be used in some cases

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/NodeDropArea.tsx
bytasv: Oh yes, I did struggle for a while until I figured how to select within iframe... but it should be doable with just text selector 👍

bytasv:
> it's just a shame that the images are a bit small, but maybe they can still be used in some cases

Trying to figure out if and how it's possible to make them higher resolution, I guess that's the only main thing keeping this less optimal solution atm

bytasv:
@mui/toolpad after getting support on Playwright repo, I managed to generate high quality screenshots. Seems that the main blocker for this is gone. So now the question is whether we want to proceed with this approach forward

oliviertassinari:
If I understand this correctly the bet we are making with this is that:

> writing the script + keeping the screenshot script up to date

will be less work than:

> taking the screenshots manually + retaking them when the UI changes

I have no idea if this will turn out to be true, but I think that it can make a lot of sense to have some pages of the docs that are depending on images built with this method. The value is that the next time that we need to update the screenshots (e.g. redesign), we would then see what works best between the two options.

Another thought, we might be able to minimize the cost of keeping the screenshot script up to date by taking them during the e2e tests. I mean, I think that we should take screenshots during the e2e for visual regressions tests, by pushing them to Argos, but maybe it would make sense to share some of the logic for the docs.

A closing thought, I suspect that the docs will scale by having technical writers, people that would not have the skill to keep the script up to date, while the people that have the skills to do it, will be focusing on customer-facing products, hence we might be left with nobody with the bandwidth.

Janpot:
> Another thought, we might be able to minimize the cost of keeping the screenshot script up to date by taking them during the e2e tests. I mean, I think that we should take screenshots during the e2e for visual regressions tests (by pushing them to Argos CI), but maybe it would make sense to share some of the logic for the docs screenshots.

Yes, I also believe the script will go out of date and break real quick of we don't add any sort of visual regression testing for these screenshots. Literally any banal change can break the script. When time comes when you need to run it again you may waste more time fixing the regressions than retaking the screenshots manually.

> A closing thought, I suspect that the docs will scale (2 years from now?) by having technical writers, people that might not have the skill to keep the script up to date

If it's part of visual regression testing, the burden of keeping the screenshots up-to-date is on the maintainer. If we keep a manual escape hatch, technical writers won't be blocked on maintainer time to add new screenshots.

> About using it for Toolpad, I think it clearly works and saves time. So, we should use it.

It saves time as a first-order effect i.e. one technical writer doesn't have to spend time manually taking screenshots. The time wasted as a higher-order effect (e.g. it needs to be maintained on every other PR) isn't 100% clear to me. But I'm not against using a technique like this.

Some things I noticed around the dialog screenshot:
- it seems to darken space outside of rounded corners
- for some UI elements (like the dialogs) you may want to capture the drop shadow as well.

Janpot:
Just to note, visual regression testing of these screenshots can potentially be as straightforward as accompanying every `locator.screenshot()` call with a `expect(locator).toHaveScreenshot()` call.

bytasv:
> visual regression testing of these screenshots can potentially be as straightforward as accompanying every

Yes, the idea was to do similar to that, have specs that run on a regular basis and when they are run with special flag capture screenshots.

> a way to capture the dropshadow could be by injecting a transparent div in the page, absolutely positioning it over the desired element with an offset and capturing screenshot of that div.

I haven't investigated much but I think it might be possible to add "padding" to the screenshot element, so hopefully we wouldn't need to artificially add more nodes when running tests

> alternatively, we could inject global styling that turns off dialog dropshadows temporarily while generating a screenshot

I think we might need these sort of global styles, at least for transitions as I already had an issue with that

Janpot:
> I haven't investigated much but I think it might be possible to add "padding" to the screenshot element, so hopefully we wouldn't need to artificially add more nodes when running tests

It doesn't seem like there's a feature in playwright to add an offset to the screenshot (Might want to 👍  the feature request https://github.com/microsoft/playwright/issues/16928).

> at least for transitions

There is an `animations` option that may be of help here: https://playwright.dev/docs/test-assertions#locator-assertions-to-have-screenshot-2

PR #1046, opened 2022-09-21
[core] Remove outdated docsearch.js dependency

oliviertassinari:
In the future, we should look into how we can have the docs-infra control it's own npm dependencies so https://github.com/mui/material-ui/pull/34421 is enough.


apedroferreira (approved):

oliviertassinari:
Actually, we need that: https://twitter.com/olivtassinari/status/1576154683252813825

PR #1048, opened 2022-09-22
[core] Add default data to DataGrid, Image, Select components

### LINKED ISSUE #841: Components should have some pre-filled dummy data when dragged on canvas
### Duplicates

### Latest version

### Summary 💡

This would assist in getting started with the components. Component names:
1. Data grid
2. Select
3. Image

Also, for Select and Textfield components Label="Label" should be there by default.

bharatkashyap:
- Closes #841 
- Visual demonstration 

- The DataGrid will have default data (similar to this https://mui.com/x/react-data-grid/#mit-version  in the demos). 

- It will automatically re-infer columns if the rows are changed from the default ones, but not after that (To match the current behaviour)

- Added some clean up of the columns (adding a readable `headerName` and a default minimum width)

https://user-images.githubusercontent.com/19550456/191744094-c35166ef-b291-4b0d-b832-8d9304868efc.mov

- The Image and Select will have default data as well

https://user-images.githubusercontent.com/19550456/191744162-8558ca87-b31f-4f4b-bf00-e5bcff069d32.mov

### DIFF
--- a/packages/toolpad-components/src/DataGrid.tsx
+++ b/packages/toolpad-components/src/DataGrid.tsx
@@ -154,6 +154,11 @@ export function inferColumns(rows: GridRowsProp): SerializableGridColumns {
   return Object.entries(firstRow).map(([field, value]) => {
     return {
       field,
+      headerName:
+        // turn field into a human readable name by capitalizing the first letter and replacing camel case with spaces
+        field[0].toUpperCase() + field.slice(1).replace(/([A-Z])/g, ' $1'),
+      // Add minimum width to the columns
+      width: 150,
       type: inferColumnType(value),
     };
   });
@@ -167,7 +172,52 @@ export function parseColumns(columns: SerializableGridColumns): GridColumns {
   }));
 }
 
-const EMPTY_ROWS: GridRowsProp = [];
+const DEFAULT_COLUMNS: (Omit<GridColDef, 'type'> & { type: string })[] = [
+  { field: 'id', headerName: 'ID', type: 'number', width: 90 },
+  {
+    field: 'firstName',
+    headerName: 'First name',
+    type: 'string',
+    width: 150,
+    editable: true,
+  },
+  {
+    field: 'lastName',
+    headerName: 'Last name',
+    type: 'string',
+    width: 150,
+    editable: true,
+  },
+  {
+    field: 'age',
+    headerName: 'Age',
+    type: 'number',
+    width: 110,
+    editable: true,
+  },
+  {
+    field: 'fullName',
+    headerName: 'Full name',
+    description: 'This column has a value getter and is not sortable.',
+    type: 'string',
+    sortable: false,
+    width: 160,
+    valueGetter: (params: GridValueGetterParams) =>
+      `${params.row.firstName || ''} ${params.row.lastName || ''}`,
+  },
+];
+
+const DEFAULT_ROWS: GridRowsProp = [
+  { id: 1, lastName: 'Snow', firstName: 'Jon', age: 35 },
+  { id: 2, lastName: 'Lannister', firstName: 'Cersei', age: 42 },
+  { id: 3, lastName: 'Lannister', firstName: 'Jaime', age: 45 },
+  { id: 4, lastName: 'Stark', firstName: 'Arya', age: 16 },
+  { id: 5, lastName: 'Targaryen', firstName: 'Daenerys', age: null },
+  { id: 6, lastName: 'Melisandre', firstName: null, age: 150 },
+  { id: 7, lastName: 'Clifford', firstName: 'Ferrara', age: 44 },
+  { id: 8, lastName: 'Frances', firstName: 'Rossini', age: 36 },
+  { id: 9, lastName: 'Roxie', firstName: 'Harvey', age: 65 },
+];
 
 interface Selection {
   id?: any;
@@ -242,7 +292,7 @@ const DataGridComponent = React.forwardRef(function DataGridComponent(
   );
   React.useEffect(() => handleColumnOrderChange.clear(), [handleColumnOrderChange]);
 
-  const rowsInput = rowsProp || EMPTY_ROWS;
+  const rowsInput = rowsProp || DEFAULT_ROWS;
 
   const hasExplicitRowId: boolean = React.useMemo(() => {
     const hasRowIdField: boolean = !!(rowIdFieldProp && rowIdFieldProp !== 'id');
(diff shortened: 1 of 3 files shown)


packages/toolpad-components/src/Select.tsx
```diff
@@ -11,6 +11,11 @@ export type SelectProps = TextFieldProps & {
   options: (string | SelectOption)[];
 };
 
+const DEFAULT_OPTIONS = {
+  options: ['January', 'February', 'March', 'April', 'May', 'June'],
+  label: 'Months',
```
apedroferreira, line 16: maybe just `Month`?

packages/toolpad-components/src/Select.tsx
```diff
@@ -11,6 +11,11 @@ export type SelectProps = TextFieldProps & {
   options: (string | SelectOption)[];
 };
 
+const DEFAULT_OPTIONS = {
+  options: ['January', 'February', 'March', 'April', 'May', 'June'],
```
apedroferreira, line 15: why not all of them? :D

packages/toolpad-components/src/Image.tsx
```diff
@@ -12,6 +12,8 @@ export interface ImageProps {
   fit: 'contain' | 'cover' | 'fill' | 'none' | 'scale-down';
 }
 
+const DEFAULT_IMG_SRC = 'https://mui.com/static/branding/mui-x/Mocktable-dark.png';
```
apedroferreira, line 15: i found this image a bit dark and hard to understand - maybe we can use something bigger and more photographic?

packages/toolpad-components/src/DataGrid.tsx
```diff
@@ -154,6 +154,11 @@ export function inferColumns(rows: GridRowsProp): SerializableGridColumns {
   return Object.entries(firstRow).map(([field, value]) => {
     return {
       field,
+      headerName:
+        // turn field into a human readable name by capitalizing the first letter and replacing camel case with spaces
+        field[0].toUpperCase() + field.slice(1).replace(/([A-Z])/g, ' $1'),
```
apedroferreira, line 159: can we rely on this for every type of data? what if an API returns column names that aren't camel case?

packages/toolpad-components/src/DataGrid.tsx
```diff
@@ -242,7 +292,7 @@ const DataGridComponent = React.forwardRef(function DataGridComponent(
   );
   React.useEffect(() => handleColumnOrderChange.clear(), [handleColumnOrderChange]);
 
-  const rowsInput = rowsProp || EMPTY_ROWS;
+  const rowsInput = rowsProp || DEFAULT_ROWS;
```
apedroferreira, line 295: is there a way to set the default value here automatically according to the `defaultValue` in `argTypes`?

packages/toolpad-components/src/DataGrid.tsx
```diff
@@ -256,10 +306,14 @@ const DataGridComponent = React.forwardRef(function DataGridComponent(
   );
 
   const columnsInitRef = React.useRef(false);
-  const hasColumnsDefined = columnsProp && columnsProp.length > 0;
+  const isUsingDefaultData = rows === DEFAULT_ROWS;
```
apedroferreira, line 309: i don't think you can compare arrays with `===`
for example: `[1,2,3] === [1,2,3]` returns `false`
you need to use some way to check for deep equality

apedroferreira (changes requested):
left some comments and suggestions!

packages/toolpad-components/src/DataGrid.tsx
bharatkashyap: I thought this wouldn't work as well, but for some reason it does. I don't know yet why, so I am trying to wait for a case that breaks it

packages/toolpad-components/src/DataGrid.tsx
bharatkashyap: If the name isn't in camel case it would return the name with the first letter capitalized

apedroferreira: yeah you're right. sounds good!

packages/toolpad-components/src/DataGrid.tsx
apedroferreira: i mean you're already passing `DEFAULT_ROWS` to `defaultValue` below, so not 100% how this logic works but maybe it could be set automatically as the initial value of that prop

apedroferreira: i just think it's a bit unpredictable if the default value for the props is not guaranteed to match the value we set in the `argTypes`... but if it's not something we're doing yet we can do it later separately.

packages/toolpad-components/src/DataGrid.tsx
bharatkashyap: Got it! Thanks for clearing that up

packages/toolpad-components/src/Image.tsx
bharatkashyap: Something from https://unsplash.com/ potentially?

packages/toolpad-components/src/DataGrid.tsx
apedroferreira: ah, i just checked and i see that the default value already works like i was describing, i wasn't sure. so this is kind of a fallback? 
would keeping `EMPTY_ROWS` as the fallback value work?
if not, and this fallback is necessary, we can keep it.

packages/toolpad-components/src/Image.tsx
apedroferreira: those seem nice!

packages/toolpad-components/src/DataGrid.tsx
bharatkashyap: >  would keeping `EMPTY_ROWS` as the fallback value work? 

Yep, I think `EMPTY_ROWS` should remain as well as the fallback for cases when the `defaultValue` isn't the value of the `prop` yet

packages/toolpad-components/src/Image.tsx
bharatkashyap: How about an image of [ El Teide ](https://unsplash.com/photos/S12z98cqAzM)? :D

apedroferreira (approved):
looks great! thanks!

apedroferreira:
@bharatkashyap looks like one of the editor tests I added some time ago is getting a bit flaky... - don't worry about it I'll look into it

Issue #997, opened 2022-09-14
Add Google Analytics


apedroferreira:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

From https://github.com/mui/mui-toolpad/issues/592.
Add Google Analytics to Toolpad for usage reporting.

### Examples 🌈

For now we can probably just do something like in this example?
https://github.com/vercel/next.js/tree/canary/examples/with-google-analytics

prakhargupta1:
I think this issue can be delayed. At this point, the only GA event we'll be interested would be no. of apps built. But I think there would be more ways to measure it. 
Acc. to me, we should do this after the overhaul. Let me know what do you think?

apedroferreira:
> I think this issue can be delayed. At this point, the only GA event we'll be interested would be no. of apps built. But I think there would be more ways to measure it. Acc. to me, we should do this after the overhaul. Let me know what do you think?

It could maybe still be useful to track the amount of traffic and such, but we can also delay it if you think it's best.

prakhargupta1:
Ok, let's do the basic implementation then, I think following events would be good to track:
1. Clicks on View demo button
2. Count of demo views
3. Number of apps created (may not be related to GA)

prakhargupta1:
@apedroferreira Let's track the demo related GA events through this old issue.

apedroferreira:
> @apedroferreira Let's track the demo related GA events through this old issue.

Sounds good, I guess for now we're adding:
- custom event for when an app is successfully created

PR #1049, opened 2022-09-23
Add Google Analytics

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/997

I think this basic integration should cover all the data we need for now.

### DIFF
--- a/packages/toolpad-app/next.config.mjs
+++ b/packages/toolpad-app/next.config.mjs
@@ -167,6 +167,6 @@ export default /** @type {import('next').NextConfig} */ ({
         source: '/:path*',
         headers: securityHeaders,
       },
-    ]
+    ];
   },
 });
--- a/packages/toolpad-app/pages/_app.tsx
+++ b/packages/toolpad-app/pages/_app.tsx
@@ -1,20 +1,26 @@
 import * as React from 'react';
 import Head from 'next/head';
-import App, { AppContext, AppProps } from 'next/app';
+import App, { AppContext, AppProps, NextWebVitalsMetric } from 'next/app';
 import { ThemeProvider } from '@mui/material/styles';
 import CssBaseline from '@mui/material/CssBaseline';
 import { CacheProvider, EmotionCache } from '@emotion/react';
 import { QueryClientProvider } from '@tanstack/react-query';
 import { LicenseInfo } from '@mui/x-data-grid-pro';
+import { useRouter } from 'next/router';
 import theme from '../src/theme';
 import createEmotionCache from '../src/createEmotionCache';
 import { MUI_X_PRO_LICENSE } from '../src/constants';
 import { queryClient } from '../src/api';
+import { reportWebVitalsToGA, setGAPage } from '../src/utils/ga';
 
 LicenseInfo.setLicenseKey(MUI_X_PRO_LICENSE);
 
 // Client-side cache, shared for the whole session of the user in the browser.
 const clientSideEmotionCache = createEmotionCache();
+
+export const reportWebVitals = (metric: NextWebVitalsMetric): void => {
+  reportWebVitalsToGA(metric);
+};
 
 interface MyAppProps extends AppProps {
   emotionCache?: EmotionCache;
@@ -22,6 +28,19 @@ interface MyAppProps extends AppProps {
 
 export default function MyApp(props: MyAppProps) {
   const { Component, emotionCache = clientSideEmotionCache, pageProps } = props;
+
+  const router = useRouter();
+
+  React.useEffect(() => {
+    router.events.on('routeChangeComplete', setGAPage);
+    router.events.on('hashChangeComplete', setGAPage);
+
+    return () => {
+      router.events.off('routeChangeComplete', setGAPage);
+      router.events.off('hashChangeComplete', setGAPage);
+    };
+  }, [router.events]);
+
   return (
     <CacheProvider value={emotionCache}>
       <Head>
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "@swc/helpers": "^0.4.11",
     "@testing-library/jest-dom": "^5.16.5",
     "@testing-library/react": "^13.4.0",
+    "@types/gtag.js": "^0.0.11",
     "@types/jest": "^29.0.0",
     "@types/node": "^18.7.16",
     "@types/rimraf": "^3.0.2",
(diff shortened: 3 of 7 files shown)


bytasv (commented):
Wondering if it made sense to use something like https://www.npmjs.com/package/nextjs-google-analytics ? - if I understood correctly we get pretty much the same functionality there

apedroferreira:
It doesn't

> Wondering if it made sense to use something like https://www.npmjs.com/package/nextjs-google-analytics ? - if I understood correctly we get pretty much the same functionality there

I didn't know about this library, it doesn't seem very popular, maybe because it really isn't that much simpler/ easier to use than this integration, and does not provide any meaningful advantages?
Also I would argue that not relying on another external dependency gives us a bit more control and knowledge about what our integration does? Plus that way we don't have to rely on that dependency being actively mantained.

bytasv:
> > Wondering if it made sense to use something like https://www.npmjs.com/package/nextjs-google-analytics ? - if I understood correctly we get pretty much the same functionality there
> 
> I didn't know about this library, it doesn't seem very popular, maybe because it really isn't that much simpler/ easier to use than this integration, and does not provide any meaningful advantages? Also I would argue that not relying on another external dependency gives us a bit more control and knowledge about what our integration does? Plus that way we don't have to rely on that dependency being actively mantained.

That's fair. I just looked around and this popped up, I don't know about the popularity, but seems that it is somewhat actively maintained and from the initial impression it already does all what you had to implement so my thinking was that we might be just able to use that out of the box.
As for more control and knowledge about it - maybe you could be a little bit more specific? For me it would be enough to know that whatever the library is it does what I want it to do and if I start seeing some issues with it like maybe lack of functionality, that's then I would consider switching to home-made solution. Anyways these are just 2 cents from me and how I ended up there

apedroferreira:
> That's fair. I just looked around and this popped up, I don't know about the popularity, but seems that it is somewhat actively maintained and from the initial impression it already does all what you had to implement so my thinking was that we might be just able to use that out of the box. As for more control and knowledge about it - maybe you could be a little bit more specific? For me it would be enough to know that whatever the library is it does what I want it to do and if I start seeing some issues with it like maybe lack of functionality, that's then I would consider switching to home-made solution. Anyways these are just 2 cents from me and how I ended up there

This implementation I made isn't too complex in my opinion, just analytics + page tracking + web vitals (not that web vitals are important for now, but they were just easy to add), so I don't think we'll have any issues maintaining it either.
It's not that the library you proposed has any big disadvantages, but as I've already implemented things this way I think we could keep it, as I don't see any advantages in using the library either, and there would be the effort in switching over.

The kind of issues that could come from using the library isn't too probable in the short-term but would be something like:
- Google Analytics makes changes to their API
- The library does not get updated / stops being actively maintained
- We're stuck with the library unless we change to a manual implementation

bytasv:
As I mentioned that's fair enough, I'm not pushing to replace implementation, let's keep it as is, I just wanted to share how and why I was thinking of a library in the first place.
While your mentioned concerns can be valid in the situations that you described if the lib did what we need to do and it worked right now, none of that would matter and in the worse case scenario we could switch to home-made solution.
IMO this is something that we could/should keep an eye for in the future when we need to implement very common functionality and ask our selves whether we really want to write and maintain code ourselves (ideally that code should also be tested) if we could use something of the shelf (supposingly we could). I think it could sometimes be deceptive at first to treat that this is the only functionality that we would need for such thing and find ourselves adding more and more features just to end up recreating the same thing in house.

bytasv (approved):

apedroferreira:
> As I mentioned that's fair enough, I'm not pushing to replace implementation, let's keep it as is, I just wanted to share how and why I was thinking of a library in the first place. While your mentioned concerns can be valid in the situations that you described if the lib did what we need to do and it worked right now, none of that would matter and in the worse case scenario we could switch to home-made solution. IMO this is something that we could/should keep an eye for in the future when we need to implement very common functionality and ask our selves whether we really want to write and maintain code ourselves (ideally that code should also be tested) if we could use something of the shelf (supposingly we could). I think it could sometimes be deceptive at first to treat that this is the only functionality that we would need for such thing and find ourselves adding more and more features just to end up recreating the same thing in house.

Makes sense, let's keep an eye out for that sort of thing in the future.
Thanks for the feedback!

packages/toolpad-app/next.config.mjs
```diff
@@ -167,6 +167,6 @@ export default /** @type {import('next').NextConfig} */ ({
         source: '/:path*',
         headers: securityHeaders,
       },
-    ]
```
Janpot, line 170: @apedroferreira Any idea on how this passed the `prettier` CI check up until now?

apedroferreira: not really, maybe if you saved it before it's not running for files with the `.mjs` extension?
i probably just saved the file and prettier added this

packages/toolpad-app/pages/_app.tsx
```diff
@@ -1,27 +1,46 @@
...
 interface MyAppProps extends AppProps {
   emotionCache?: EmotionCache;
 }
 
 export default function MyApp(props: MyAppProps) {
   const { Component, emotionCache = clientSideEmotionCache, pageProps } = props;
+
+  const router = useRouter();
+
+  React.useEffect(() => {
```
Janpot, line 34: @apedroferreira 
- How useful would this really be? We also have `react-router` on the editor, so the data wouldn't be exactly a good representation of user navigation throughout the app.
- Did you verify we only send one event on hard reload?

packages/toolpad-app/next.config.mjs
Janpot: Did `prettier` get updated recently? CI only checks for changed files I believe. Perhaps we should run `prettier:all` in CI?

packages/toolpad-app/pages/_app.tsx
apedroferreira: I took it from this example https://github.com/vercel/next.js/tree/canary/examples/with-google-analytics, just rechecked again and as far as I can tell:
- hard reload only triggers one page view
- client site router changes (creating a new connection for example) trigger page views too, probably because of `hashChangeComplete`

apedroferreira: seems to run for the iframe too, not sure if that's helpful

Janpot: might make sense to disable for the app canvas

apedroferreira: ok actually it turns out this effect is doing nothing at all... the page view events are being sent automatically.
i'll remove this effect and see if i can filter out events from the iframe

apedroferreira: and because the page view events are sent automatically doesn't seem simple to filter out the ones from the canvas... - we can just ignore that data in the google analytics dashboard itself if we want to use page view data, im gonna make a PR that just removes the effect

Issue #1010, opened 2022-09-15
Add captcha to online demo


apedroferreira:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

From https://github.com/mui/mui-toolpad/issues/592
We may want to have a captcha in the demo, to be solved before users can create a new application.

### Examples 🌈

Possible solutions:
- https://www.google.com/recaptcha/about/ ?

(I'm not very familiar with captcha solutions)

### Motivation 🔦

Reduce bot traffic in demo, or application spam.

bharatkashyap:
I've used https://www.hcaptcha.com/ before, did the job well

PR #1054, opened 2022-09-26
Add Google reCAPTCHA v3 to online demo

apedroferreira:
- Closes https://github.com/mui/mui-toolpad/issues/1010

Use Google reCAPTCHA v3 to verify users upon creating an app.
This behavior is optional and enabled by the environment variables for the reCAPTCHA.

Google's reCAPTCHA v3 uses a score to try to determine whether a user is a person or a bot - it does not show any puzzle to solve or anything of the sort, which is what captcha v2 used to do. I set the minimum acceptable score as 0.5 (it goes from 0-1), which seems to be recommended as a reasonable default value.

**UI changes:**
Using the CAPTCHA caused a small badge to show up in the lower-right corner of the website. Because I decided to hide it with CSS, it is required to add the following disclaimer in the app creation form when we use the captcha:
https://stackoverflow.com/questions/44543157/how-to-hide-the-google-invisible-recaptcha-badge
Screen Shot 2022-09-26 at 18 35 54

**Create app mutation error changes:**
I've also allowed showing custom errors when creating an app, as the current logic for is it to always show `An app with this name already exists`, no matter what the error is. I've made it so that we have to throw errors with a certain message to be shown.


packages/toolpad-app/src/utils/recaptchaV3.ts
```diff
@@ -0,0 +1,30 @@
...
+): Promise<boolean> => {
+  const recaptchaParams = new URLSearchParams({
+    secret: secretKey,
+    response: token,
+  });
+
+  const recaptchaResponse = await fetch(`${SITE_VERIFY_URL}?${recaptchaParams}`, {
+    method: 'POST',
+  });
+  const recaptchaReponseJson = await recaptchaResponse.json();
```
bytasv, line 23: minor suggestion:
```
const { success, score } = await recaptchaResponse.json()

if (!success || score < SCORE_THRESHOLD) {
```

bytasv (approved):

packages/toolpad-app/src/toolpad/Home/index.tsx
```diff
@@ -124,7 +132,7 @@ function CreateAppDialog({ onClose, ...props }: CreateAppDialogProps) {
               label="Name"
               value={name}
               error={createAppMutation.isError}
-              helperText={createAppMutation.isError ? `An app named "${name}" already exists` : ''}
+              helperText={(createAppMutation.error as Error)?.message || ''}
```
bharatkashyap, line 135:
```suggestion
              helperText={(createAppMutation.isError ? (createAppMutation.error as Error)?.message : ''}
```

bharatkashyap (commented):
Looks good! I was under the impression that this was only for the demo version, but looks like this is being added across environments. Is that intended?

packages/toolpad-app/src/toolpad/Home/index.tsx
apedroferreira: does this have any particular advantage? 
i thought that checking just for the error would be correct, is there any case where that could have issues?

apedroferreira:
> Looks good! I was under the impression that this was only for the demo version, but looks like this is being added across environments. Is that intended?

We can enable or disabled by including the environment variables for the captcha or not - so there will be no captcha unless you add it.
I think that for now the plan is for this feature to be enabled in the demo version of Toolpad only, even though it could technically be a separate feature.

oliviertassinari:
> Google reCAPTCHA v3

I'm curious to see how well this will work.

4 years ago, I was facing an issue with the integration of Stripe. We were processing a lot of $1 payments, for which Stripe detected a lot of bot testing credit cards with. They asked us to implement an anti-boat protection. I started with Google reCAPTCHA v3 but it was blocking actual users and letting bot go thought 😅. We had to move to the v2. On the other hand https://strapi.io/demo uses the v3 (as you can see from the required banner) so 🤷

packages/toolpad-app/src/utils/recaptcha.ts
```diff
@@ -0,0 +1,31 @@
...
+  await new Promise<void>((resolve) => grecaptcha.ready(resolve));
+  return grecaptcha.execute(siteKey, {
+    action: 'submit',
+  });
+};
+
+const SITE_VERIFY_URL = 'https://www.google.com/recaptcha/api/siteverify';
+const SCORE_THRESHOLD = 0.5;
+
+export const validateRecaptchaToken = async (
```
Janpot, line 11: @apedroferreira This file is mixing serverside-only code with client side code.   You might want to split that

apedroferreira: added the changes here https://github.com/mui/mui-toolpad/pull/1087/commits/8ec6c3dbff03c1e4201a1be183e12c0db71a87d1

PR #1055, opened 2022-09-27
[app] Select options editor enhancements

### LINKED ISSUE #538: Select options editor enhancements

bharatkashyap:
- Closes #538 
- Visual demonstration 

https://user-images.githubusercontent.com/19550456/192457051-27a533b4-9936-4408-8f45-06a26ac003b6.mov

### DIFF
--- a/packages/toolpad-app/src/toolpad/propertyControls/SelectOptions.tsx
+++ b/packages/toolpad-app/src/toolpad/propertyControls/SelectOptions.tsx
@@ -30,6 +30,7 @@ function SelectOptionsPropEditor({
   const [editOptionsDialogOpen, setEditOptionsDialogOpen] = React.useState(false);
   const optionInputRef = React.useRef<HTMLInputElement | null>(null);
   const [editingIndex, setEditingIndex] = React.useState<number | null>(null);
+  const [optionTextInputInvalid, setOptionTextInputInvalid] = React.useState(false);
 
   const editingOption: SelectOption | null = React.useMemo(() => {
     if (typeof editingIndex === 'number') {
@@ -48,11 +49,18 @@ function SelectOptionsPropEditor({
   const handleOptionTextInput = React.useCallback(
     (event: React.KeyboardEvent<HTMLDivElement>) => {
       if (event.key === 'Enter') {
-        onChange([...value, (event.target as HTMLInputElement).value]);
+        const inputText = (event.target as HTMLInputElement).value;
+        if (!inputText) {
+          setOptionTextInputInvalid(true);
+          return;
+        }
 
+        onChange([...value, (event.target as HTMLInputElement).value]);
         if (optionInputRef.current) {
           optionInputRef.current.value = '';
         }
+      } else {
+        setOptionTextInputInvalid(false);
       }
     },
     [onChange, value],
@@ -79,7 +87,12 @@ function SelectOptionsPropEditor({
 
   const handleOptionChange = React.useCallback(
     (newValue: SelectOption) => {
-      onChange(value.map((option, i) => (i === editingIndex ? newValue : option)));
+      let newOption: string | SelectOption = newValue;
+      setOptionTextInputInvalid(Boolean(!newValue.value));
+      if (!newValue.label) {
+        newOption = newValue.value;
+      }
+      onChange(value.map((option, i) => (i === editingIndex ? newOption : option)));
     },
     [editingIndex, onChange, value],
   );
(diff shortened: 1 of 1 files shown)


packages/toolpad-app/src/toolpad/propertyControls/SelectOptions.tsx
```diff
@@ -48,11 +49,18 @@ function SelectOptionsPropEditor({
   const handleOptionTextInput = React.useCallback(
     (event: React.KeyboardEvent<HTMLDivElement>) => {
       if (event.key === 'Enter') {
-        onChange([...value, (event.target as HTMLInputElement).value]);
+        const inputText = (event.target as HTMLInputElement).value;
```
apedroferreira, line 52: Would the correct event type above in this function not be `React.KeyboardEvent<HTMLDivElement>`? That way you shouldn't need to do these type castings.

packages/toolpad-app/src/toolpad/propertyControls/SelectOptions.tsx
```diff
@@ -72,14 +80,20 @@ function SelectOptionsPropEditor({
 
   const handleOptionItemClick = React.useCallback(
     (index: number) => () => {
+      setOptionTextInputInvalid(false);
```
apedroferreira, line 83: if all this state does for now is check if the empty is empty, maybe it should be called `optionTextInputEmpty` instead?

apedroferreira: or what if we use a `optionTextInputError` state and you can set there any error message as a string?

packages/toolpad-app/src/toolpad/propertyControls/SelectOptions.tsx
```diff
@@ -72,14 +80,20 @@ function SelectOptionsPropEditor({
...
       setEditingIndex(index);
     },
     [],
   );
 
   const handleOptionChange = React.useCallback(
     (newValue: SelectOption) => {
-      onChange(value.map((option, i) => (i === editingIndex ? newValue : option)));
+      let newOption: string | SelectOption = newValue;
```
apedroferreira, line 91: `newValue.value` sounds a bit confusing to read - what about calling it `newOption` instead?

packages/toolpad-app/src/toolpad/propertyControls/SelectOptions.tsx
```diff
@@ -95,19 +109,30 @@ function SelectOptionsPropEditor({
...
+        }}
       >
         {label}
       </Button>
       <Dialog
         fullWidth
         open={editOptionsDialogOpen}
-        onClose={() => setEditOptionsDialogOpen(false)}
+        onClose={() => {
+          if (optionTextInputInvalid) {
```
apedroferreira, line 122: Are we blocking closing the dialog? Maybe could be frustrating for the user, we could maybe make some assumption instead and close the dialog.

apedroferreira (changes requested):
left some suggestions and ideas

packages/toolpad-app/src/toolpad/propertyControls/SelectOptions.tsx
bharatkashyap: The typecasing would continue to happen regardless; the issue is with `React.KeyboardEvent`. If that type is removed, then forced typecasting will be needed on the `event.key` line - so the problem exists regardless

packages/toolpad-app/src/toolpad/propertyControls/SelectOptions.tsx
bharatkashyap: Makes sense, I can actually get rid of that variable itself and just use type narrowing

packages/toolpad-app/src/toolpad/propertyControls/SelectOptions.tsx
bharatkashyap: Makes sense. No need to add friction unnecessarily - present behaviour is to just have an undefined value in there and that should be fine

packages/toolpad-app/src/toolpad/propertyControls/SelectOptions.tsx
apedroferreira: ah ok, we can leave it like this then

apedroferreira (approved):
seems good, thanks for the changes!

Issue #999, opened 2022-09-14
[RFC] Add logging to Toolpad


apedroferreira:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

From https://github.com/mui/mui-toolpad/issues/592
Add logging to Toolpad.

**What should we log?**
- Frontend application logs
- Request tracing
- REST datasource requests

**What logging aggregation service should we use, if any?**
Some popular ones are: Datadog, New Relic, Grafana Loki (more complicated to set up)

Feel free to suggest others.

bytasv:
I'm wondering what exactly we would like to see at this moment? Maybe error logging would be enough for now until we know better?

apedroferreira:
> I'm wondering what exactly we would like to see at this moment? Maybe error logging would be enough for now until we know better?

Maybe the logs Nextjs gives us out of the box are enough to start with. I think that would also include any errors if the application crashes for example.

Janpot:
I want to be able to get insight into the requests that are made from the backend to external hosts. To be able to gain insight into load and spot abuse.

bharatkashyap:
> **What logging service should we use?** Some popular ones are: Datadog, New Relic, Grafana Loki (more complicated to set up)
> 
> Feel free to suggest others.

https://github.com/winstonjs/winston

Janpot:
Ideally

1. we log to stdout, the host environment is responsible for collecting logs. This is important for self-hosters.
2. each log line is a json message, in a standard format containing timestamp, message, severity and metadata. I'm familiar with `bunyan` and `pino`.
3. logs are searchable, filterable, can be aggregated and dashboards can be built based on metadata. I'm familiar with elasticsearch/kibana for this purpose.


regarding 1. it seems that render.com has more or less an all-or-nothing approach with regards to log shipping: https://render.com/docs/log-streams. This may be a quick and easy solution for us. I'm not quite sure about predicting the cost. alternatively we'll have to ship logs from within the application but I don't want to support that option towards self-hosters.

regarding 2. maybe not top priority, but towards self hosters we may have to provide an option to format logs. For me it's important that we standardize on log messages (e.g. standard fields to attach errors, req, res objects with their own formatters,...), and be able to attach metadata.

regarding 3. I've seen powerful things being done with the ELK stack, it has a learning curve though. Alternatively we can look into any of the [options](https://render.com/docs/log-streams) render.com proposes

[...]

PR #1066, opened 2022-09-29
Add API and datasource logging

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/999

Logging for:
- RPC server requests and responses
- Other server request and responses (e.g.: using datasources)
- Captcha validations
- Support sending logs to Elastic Search


bharatkashyap (approved):

packages/toolpad-app/src/server/withReqResLogs.ts
```diff
@@ -0,0 +1,87 @@
...
+    const requestId = uuidv4();
+
+    const logMessageInfo = getLogMessageInfo(req);
+
+    logger.info({
+      msg: `API logs: request${logMessageInfo ? `(${logMessageInfo})` : ''}`,
+      requestId,
+      url: req.url,
+      method: req.method,
+      ...(!_.isEmpty(req.query) ? { query: req.query } : {}),
```
apedroferreira, line 76: this empty check is looking weird to me as this might be a string? i'll take another look in a bit

packages/toolpad-app/src/server/withReqResLogs.ts
```diff
@@ -0,0 +1,87 @@
...
+  const oldWrite = res.write;
+  const oldEnd = res.end;
+
+  const chunks: Buffer[] = [];
+
+  res.write = (...restArgs: any[]) => {
+    chunks.push(Buffer.from(restArgs[0]));
+    return oldWrite.apply(
+      res,
+      restArgs as [
```
bytasv, line 32: Would it be useful to define types outside the methods? (not inline)

apedroferreira: this part was very hard to type because some of those functions have different call signatures - i'll see if i can remove that duplication, there's a slight difference there, not sure if it's the reason why i didn't reuse the type. i'll give it another look!

packages/toolpad-app/src/server/withReqResLogs.ts
```diff
@@ -0,0 +1,83 @@
...
+  callback?: ((error?: Error | null) => void) | undefined,
+];
+
+function logRequestResponse(req: NextApiRequest, res: NextApiResponse, requestId: string) {
+  const oldWrite = res.write;
+  const oldEnd = res.end;
+
+  const chunks: Buffer[] = [];
+
+  res.write = (...restArgs: any[]) => {
```
bytasv, line 34: Could you use `RestArgs` type instead of `any[]`? would it allow us to avoid type-hinting with `as`? 🤔 
```ts
res.write = (...restArgs: RestArgs) => {
  chunks.push(Buffer.from(restArgs[0]));
  return oldWrite.apply(res, restArgs);
}
```

bytasv (approved):

bharatkashyap:
Might be a bit late for this but I came across this yesterday: https://www.axiom.co/

apedroferreira:
> Might be a bit late for this but I came across this yesterday: https://www.axiom.co/

I've used this, it has the best pricing I've seen but I didn't find many good options for ingesting the logs - we would probably have to set up and run our own proxy or something like that.

packages/toolpad-app/src/server/withReqResLogs.ts
apedroferreira: It turns out I could just just remove the `any[]`, I just did.
About the `as` assertion, I think it's fine here as the restArgs can be an array with 1-3 elements so we would always need logic to assert each type anyway. And for this usecase it doesn't seem like this logic would ever break.

packages/toolpad-app/src/server/withReqResLogs.ts
```diff
@@ -0,0 +1,83 @@
...
+      // params: req.body.params,
+      ...getBaseLogProperties(req),
+    });
+
+    logRequestResponse(req, res, requestId);
+
+    return apiHandler(req, res);
+  };
+
+export default withReqResLogs;
```
Janpot, line 83: Any way we can do this in `middleware.ts` instead?

packages/toolpad-app/src/utils/uuid.ts
```diff
@@ -1,8 +1,10 @@
+import { webcrypto } from 'crypto';
+
 // credit: https://stackoverflow.com/a/2117523
 export function uuidv4() {
   // @ts-ignore
   return ([1e7] + -1e3 + -4e3 + -8e3 + -1e11).replace(/[018]/g, (c) =>
     // eslint-disable-next-line no-bitwise
-    (c ^ (crypto.getRandomValues(new Uint8Array(1))[0] & (15 >> (c / 4)))).toString(16),
```
Janpot, line 6: `crypto` seems to work as a global for me in node.js (`node -e 'console.log(!!crypto)'  `). Maybe you're on an outdated version?

Janpot: Adding https://github.com/mui/mui-toolpad/pull/1141

packages/toolpad-app/src/server/withReqResLogs.ts
apedroferreira: Never did it but was also a possibility, I'll look into it.

packages/toolpad-app/src/utils/uuid.ts
apedroferreira: Yeah I'm on Node 16, maybe the issue isn't there in 18.
I can start using 18, should be fine.

apedroferreira:
I've merged all logging related PRs into this branch.
Re-requesting reviews but most feedback has already been addressed in the other PRs, so just to be sure it can be merged!

apedroferreira:
I'm gonna merge this if everyone's ok with it, if there's any further feedback down the line just let me know!

PR #1069, opened 2022-09-30
Attempt to fix render.yaml

apedroferreira:


Janpot:
you can remove the variable completely if it creates trouble

apedroferreira:
> you can remove the variable completely if it creates trouble

Isn't it useful for the previews at least?
I think this should fix the blueprint in render. Anyway i think we don't need to use the blueprint for the demo app.

PR #1081, opened 2022-10-02
[core] Don't build on render for docs changes

oliviertassinari:
For example, we shouldn't need https://github.com/mui/mui-toolpad/pull/1053#issuecomment-1257644551 when making a change in the docs.

The setting used is documented here: https://render.com/docs/blueprint-spec#build-filters.

### DIFF
--- a/render.yaml
+++ b/render.yaml
@@ -21,11 +21,15 @@ services:
           name: toolpad-db
           property: connectionString
       - fromGroup: toolpad-settings
+    ignoredPaths:
+     - docs/**
 
 databases:
   - name: toolpad-db
     ipAllowList: []
     plan: starter
+    ignoredPaths:
+     - docs/**
 
 envVarGroups:
   - name: toolpad-settings


oliviertassinari:
It might help with https://groups.google.com/a/mui.com/g/toolpad/c/ZeZz6QoK_3M/m/2e3uGFr3BgAJ

render.yaml
```diff
@@ -21,11 +21,15 @@ services:
...
       - fromGroup: toolpad-settings
+    ignoredPaths:
+     - docs/**
 
 databases:
   - name: toolpad-db
     ipAllowList: []
     plan: starter
+    ignoredPaths:
+     - docs/**
```
oliviertassinari, line 32: I wonder if this even makes sense. I added it just in case 😅

Janpot: Actually, I'm rereviewing this PR, I think there's a problem. Shouldn't it specify the `buildFilter` property containing a `ignoredPaths` property?

PR #1085, opened 2022-10-03
[core] Fix buildFilter property in render.yml

Janpot:
I believe it should live under a `buildFilter` property, like [this sample](https://render.com/docs/monorepo-support#example-blueprint-spec-1).

### DIFF
--- a/render.yaml
+++ b/render.yaml
@@ -21,8 +21,9 @@ services:
           name: toolpad-db
           property: connectionString
       - fromGroup: toolpad-settings
-    ignoredPaths:
-     - docs/**
+    buildFilter:
+      ignoredPaths:
+        - docs/**
 
 databases:
   - name: toolpad-db
--- a/render.yaml
+++ b/render.yaml
@@ -24,6 +24,7 @@ services:
     buildFilter:
       ignoredPaths:
         - docs/**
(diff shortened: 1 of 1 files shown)


oliviertassinari:
oops 🤦‍♂️

oliviertassinari (approved):
Same in https://render.com/docs/blueprint-spec#build-filters

render.yaml
```diff
@@ -21,8 +21,9 @@ services:
           name: toolpad-db
           property: connectionString
       - fromGroup: toolpad-settings
-    ignoredPaths:
-     - docs/**
+    buildFilter:
+      ignoredPaths:
+        - docs/**
```
oliviertassinari, line 26: While we are at it?

```suggestion
        - docs/**
        - **/*.spec.ts
```

Curious we are not calling them "test.ts" 😁

Janpot: > Curious we are not calling them "x.test.ts"

Muscle memory I think. I'm happy to rename, I don't have a strong opinion. I believe I saw both flavors in `@mui/material`

oliviertassinari: >  I believe I saw both flavors in @mui/material

True, the spec.tsx files are TypeScript tests.

Issue #907, opened 2022-09-06
Custom component crashes Toolpad


Janpot:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

Create a custom component and paste the following code:

```tsx
import * as React from "react";
import { createComponent } from '@mui/toolpad-core';

export default function Hello() {
  return (
    <div />
  );
}

export default createComponent(Hello, {});
```

Observe Toolpad crash with
```
Cannot convert undefined or null to object
```

### Expected behavior 🤔

Toolpad not crashing

### Steps to reproduce 🕹

Steps:

1. Create a custom component and paste the following code:
2. paste the code from above

PR #1088, opened 2022-10-04
Allow empty argTypes object in component config

Janpot:
Fixes https://github.com/mui/mui-toolpad/issues/907


bharatkashyap (commented):
Makes sense!

PR #1089, opened 2022-10-04
Add a few more cache folders in circleci

Janpot:
- Closes 
- Visual demonstration 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,7 +71,15 @@ commands:
             # Keep path in sync with "Set yarn cache folder"
             # Can't use environment variables for `save_cache` paths (tested in https://app.circleci.com/pipelines/github/mui/material-ui/37813/workflows/5b1e207f-ac8b-44e7-9ba4-d0f9a01f5c55/jobs/223370)
             - /tmp/yarn-cache
+            # monorepo node_modules folders (paths doesn't support globbing, see https://circleci.com/docs/caching#saving-cache)
             - node_modules
+            - docs/node_modules
+            - packages/eslint-plugin-material-ui/node_modules
+            - packages/toolpad-app/node_modules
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -71,7 +71,15 @@ commands:
             # Keep path in sync with "Set yarn cache folder"
             # Can't use environment variables for `save_cache` paths (tested in https://app.circleci.com/pipelines/github/mui/material-ui/37813/workflows/5b1e207f-ac8b-44e7-9ba4-d0f9a01f5c55/jobs/223370)
             - /tmp/yarn-cache
+            # monorepo node_modules folders ("paths" doesn't support globbing, see https://circleci.com/docs/caching#saving-cache)
             - node_modules
```
oliviertassinari, line 75: What's the use case for caching this folder? It doesn't seem mentioned in https://circleci.com/docs/caching#basic-example-of-yarn-package-manager-caching.

Janpot: When I run `yarn` locally it tends to be way faster with installed `node_modules` present vs not. I assume it doesn't unpack and copy from cache when the module is already installed? That way avoiding a ton of IO unnecessary. I'm not 100% sure, though.
🤔 The win may be nullified by downloading the node_nodules folder from circleci cache though.

oliviertassinari: > The win may be nullified by downloading the node_nodules folder from circleci cache though.

I wonder about this 👍 to try and see how it goes.

oliviertassinari: Looking at the CircleCI logs, it doesn't seem to make the build run faster.

Janpot: Yeah, this PR wasn't recreating the cache (lock file didn't change). I created a new PR https://github.com/mui/mui-toolpad/pull/1104

PR #1091, opened 2022-10-04
Alternative fix for 1050

Janpot:
This is proposed as an alternative to https://github.com/mui/mui-toolpad/pull/1061
The advantage is that besides const bindings, this PR should also work for javascript bindings.

The issue is caused by [this effect](https://github.com/mui/mui-toolpad/blob/712642e7a23e0c9db6f0885854810bbf89e81cbe/packages/toolpad-app/src/runtime/ToolpadApp.tsx#L730) that updates the bindings only on the second render when the page changes. By forcing a remount the state gets reset when the page loads.

To reviewers: for extra clarity I recommend reviewing the second commit on its own https://github.com/mui/mui-toolpad/pull/1091/commits/7e40adb038b21659189e664fec431182dcd07876. The first commit just reverts the previous fix

### DIFF
--- a/packages/toolpad-app/src/runtime/ToolpadApp.tsx
+++ b/packages/toolpad-app/src/runtime/ToolpadApp.tsx
@@ -703,7 +703,16 @@ function RenderedPage({ nodeId }: RenderedNodeProps) {
   const [pageBindings, setPageBindings] =
     React.useState<Record<string, ParsedBinding>>(parsedBindings);
 
+  const prevDom = React.useRef(dom);
   React.useEffect(() => {
+    if (dom === prevDom.current) {
+      // Ignore this effect if there are no dom updates.
+      // IMPORTANT!!! This assumes the `RenderedPage` component is remounted when the `nodeId` changes
+      //  <RenderedPage nodeId={someId} key={someId} />
+      return;
+    }
+    prevDom.current = dom;
+
     setPageBindings((existingBindings) => {
       // Make sure to patch page bindings after dom nodes have been added or removed
       const updated: Record<string, ParsedBinding> = {};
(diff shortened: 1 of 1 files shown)


Janpot:
cc @WangLarry In case you'd like to review

apedroferreira (approved):
Seems much better! It handles the problem in a much simpler and more general way.

WangLarry:
@Janpot 
I did not test your fix yet. 
But I think:

There are three kinds of 'value':
1. default value:   is defined in component function, is default if designer did not set.
2. const value:   is in dom, which is set by designer, is initial value in runtime.
3. bonded value:   is in dom, which is set by designer, but need to evaluate in runtime.

When 'if (typeof hookResult[propName] === "undefined")',  means:
1. this component with bonded value, but has not evaluated yet.
2. this component with const value, should be inited by const value
3. this component with default value, should be inited by default value

Janpot:
@WangLarry 

> 1. this component with bonded value, but has not evaluated yet.

This was the exact case before this PR. bindings are evaluated on the first render. It's just that the `RenderedPage` component was reused when switching pages, so it didn't mount again. instead an effect patches the bindings for the second render after switching the page. There was just this one render during page change where the old page bindings were still there, but the new page content was rendered. This PR fixes it by forcing a remount when pages change, thus having the correct evaluated bindings available on the first render of the page content.

> 2. this component with const value, should be inited by const value

const values are part of the evaluation so now also handled by the same fix above.

WangLarry:
> @WangLarry
> 
> > 1. this component with bonded value, but has not evaluated yet.
> 
> This was the exact case before this PR. bindings are evaluated on the first render. It's just that the `RenderedPage` component was reused when switching pages, so it didn't mount again. instead an effect patches the bindings for the second render after switching the page. There was just this one render during page change where the old page bindings were still there, but the new page content was rendered. This PR fixes it by forcing a remount when pages change, thus having the correct evaluated bindings available on the first render of the page content.
> 
> > 2. this component with const value, should be inited by const value
> 
> const values are part of the evaluation so now also handled by the same fix above.

"const values are part of the evaluation"!
I have tested your fix, looks work very well.

Issue #1056, opened 2022-09-27
Online demo hosting


apedroferreira:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

From https://github.com/mui/mui-toolpad/issues/592.

Host online demo of Toolpad - with error reporting, analytics and captcha enabled.

Some questions to solve:

**- Where should we host it?**
render.com as that's what we've been using for the Toolpad app.

**- How should deployments be made?**
Render can deploy based on certain branches, should we try to use it or for example have a completely manual process for deploying the demo?

**- What domain should we host on?**
From the discussion so far it seems like we're leaning towards hosting as a subdomain of toolpad.io - maybe demo.toolpad.io?
Another suggestion was toolpad-demo.mui.com

### Motivation 🔦

By hosting the online demo users can finally access it and use it.

apedroferreira:
@prakhargupta1 we may want to move this into the board to give it more visibility?

prakhargupta1:
> What domain should we host on?

I am assuming we can't use mui.com/toolpad/demo for some reason? 
Maybe we should use: demo.toolpad.io

bytasv:
I think we should be able to host even on mui.com/toolpad/demo we would just to make sure that there are no issues because of what's the base url is 🤔

apedroferreira:
Probably `mui.com/toolpad/demo` is too much trouble for what it's worth?
This is a completely separate app so I think it would be good to host it in its own subdomain.
I think `demo.toolpad.io` sounds great too if we can use that new domain.

prakhargupta1:
@oliviertassinari What do you suggest?

oliviertassinari:
> Render can deploy based on certain branches, should we try to use it or for example have a completely manual process for deploying the demo?

I think that a new branch here would be more than enough.

> What do you suggest?

I think that using a subfolder of the TLD (https://mui.com/) leads to "pollutions". I would be afraid of:

- SEO content that gets indexed by the community
- Community member that writes spamming logic, running on the backend, that gets our domain blacklisted

But even then, from what I understand, when self-hosted, and in the cloud, it will always be `/` so I don't think that it would make business sense to invest resources to support a base path different from `/`.  So, I think that we should remove this option from the list. 

> maybe demo.toolpad.io

Why not. Strapi uses https://api-iax04.strapidemo.com/. We also have toolpad.app and toolpad.dev available for us.

[...]

PR #1092, opened 2022-10-04
Set up demo with Docker

apedroferreira:
Possible solution for https://github.com/mui/mui-toolpad/issues/1056


apedroferreira:
@Janpot would this be a correct way to do things using Docker?
I tried to deploy https://dashboard.render.com/web/srv-ccu728t3t398codsgi90/events based on https://github.com/mui/mui-toolpad/pull/426 but it seems to run out of memory too.

Janpot:
This wouldn't use out docker image. It's rebuilding it using our dockerfile. 🤔 Perhaps we need to create a dockerfile that has the desired image as base?

```docker
# Dockerfile.demo
FROM muicom/toolpad:123455
```

apedroferreira:
Got it, I wasn't fully understanding last week that we wanted to use the remote image - that should be great, I'll try it out, probably on a separate repo.

apedroferreira:
@Janpot I tried Render with the remote image and still got an out of memory error https://dashboard.render.com/web/srv-cd26q32en0hm1tsfqoh0
I wonder if I did something wrong or if our memory problems could be elsewhere?

Janpot:
Ah right, we're hitting https://github.com/prisma/prisma/issues/11413 again

docker/images/toolpad/Dockerfile.demo
```diff
@@ -0,0 +1 @@
+FROM muicom/toolpad:latest
```
Janpot, line 1: Shall we pin the version

apedroferreira: I pinned it to v0.0.21, will just have to manually keep it updated.
Other than that does this flow with the `demo` branch look good to you if it works? Or would you still prefer we use a separate repo for the demo?

oliviertassinari:
A bit of topic, I just saw: 

Screenshot 2022-10-20 at 15 13 11

https://app.forestadmin.com/Live%20Demo/Production/Operations/data/806052/index

It's could be interesting to have something like this (I haven't seen the demo app UX yet, and I changed the color contrast because OMG dark seriously!)

render.yaml
```diff
@@ -27,11 +27,38 @@ services:
         - docs/**
         - '**/*.spec.ts'
 
+  - type: web
+    name: toolpad-demo
+    env: docker
+    branch: demo
+    dockerfilePath: ./docker/images/toolpad/Dockerfile.demo
+    dockerContext: ./
```
Janpot, line 35: Perhaps we can set the context to 
```yml
dockerContext: ./docker/images/toolpad/
```
? That should speed up the build as it doesn't have to upload the whole repository as a context. The docker file doesn't (and shouldn't) use anything from the repository

apedroferreira: I'll keep the path just in case?

render.yaml
```diff
@@ -27,11 +27,38 @@ services:
...
+  - type: web
+    name: toolpad-demo
+    env: docker
+    branch: demo
+    dockerfilePath: ./docker/images/toolpad-demo/Dockerfile
+    dockerContext: ./docker/images/toolpad-demo/
+    healthCheckPath: /health-check
+    plan: starter plus
+    pullRequestPreviewsEnabled: false
+    envVars:
```
Janpot, line 39: Shouldn't we set the demo mode env var?

apedroferreira: Yeah good one, I'll set every env var except the secret ones

apedroferreira: I guess just `TOOLPAD_DEMO` should do

Janpot: Yep, as much infrastructure expressed in code as possible

Janpot (approved):

Issue #882, opened 2022-09-02
QueryEditor bindings should use name parameters


Janpot:
For consistency and discoverability:

* function datasource: the `params` property should be named `parameters`
* rest datasource: the `query` global binding should be named `parameters`

Let's keep the old names around, but mark their types with `@deprecated` in the Toolpad editor

Issue #497, opened 2022-05-29
Show more info on global variables in binding editor


Janpot:
From https://github.com/mui/mui-toolpad/issues/309

> DX: In the "Bind a property" view, it would be interesting to know the nature of the global variable. Is it a component or an API state? It could help more quickly navigate what's available. We might even want to group them.
>
> ![image](https://user-images.githubusercontent.com/2109932/170872550-013eba96-65a1-41a4-a970-837663a99a53.png)

PR #1096, opened 2022-10-05
[core] Rename query and params to parameters in fetch and function editors

bytasv:
- Closes https://github.com/mui/mui-toolpad/issues/882

image

image

1. As per issue's description I've renamed `params` and `query` to `parameters`
2. I've done it only to the parts that are relevant to end-user, kept every other variable named same way as the changes to all `params` variable names seems to bring a lot of unnecessary/noisy changes. Lemme know if we should also change related variables
3. I'll leave comments in PR for the `@deprecated` part - I need some guidance to understand what would be the best way to mark them deprecated (cc @Janpot )
4. I've got `function` argument type updated and it shows dev hints in the editor correctly, however, I couldn't figure out how should I change type for `fetch` so it doesn't come up as `Query` type:
image

### DIFF
--- a/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/QueryEditor.tsx
+++ b/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/QueryEditor.tsx
@@ -235,16 +235,19 @@ function QueryNodeEditorDialog<Q>({
 
   const connectionParams = connection?.attributes.params.value;
 
-  const queryModel = React.useMemo<QueryEditorModel<any>>(
-    () => ({
+  const queryModel = React.useMemo<QueryEditorModel<any>>(() => {
+    const params =
+      (Object.entries(inputParams).filter(([, value]) => Boolean(value)) as BindableAttrEntries) ||
+      [];
+
+    return {
       query: input.attributes.query.value,
-      params:
-        (Object.entries(inputParams).filter(([, value]) =>
-          Boolean(value),
-        ) as BindableAttrEntries) || [],
-    }),
-    [input.attributes.query.value, inputParams],
-  );
+
+      // TODO mark params as @deprecated
+      params,
+      parameters: params,
+    };
+  }, [input.attributes.query.value, inputParams]);
 
   const handleQueryModelChange = React.useCallback(
     (model: QueryEditorModel<Q>) => {
--- a/packages/toolpad-app/src/toolpadDataSources/function/client.tsx
+++ b/packages/toolpad-app/src/toolpadDataSources/function/client.tsx
@@ -81,8 +81,8 @@ function ConnectionParamsInput({
   );
 }
 
-const DEFAULT_MODULE = `export default async function ({ params }: ${EVENT_INTERFACE_IDENTIFIER}) {
-  console.info('Executing function with params:', params);
+const DEFAULT_MODULE = `export default async function ({ parameters }: ${EVENT_INTERFACE_IDENTIFIER}) {
+  console.info('Executing function with parameters:', parameters);
   const url = new URL('https://gist.githubusercontent.com/saniyusuf/406b843afdfb9c6a86e25753fe2761f4/raw/523c324c7fcc36efab8224f9ebb7556c09b69a14/Film.JSON');
   url.searchParams.set('timestamp', String(Date.now()));
   const response = await fetch(String(url));
@@ -141,7 +141,7 @@ function QueryEditor({
 
     const content = `
       interface ${EVENT_INTERFACE_IDENTIFIER} {       
-        params: {
+        parameters: {
           ${paramsMembers}
         }
         secrets: {
--- a/packages/toolpad-app/src/toolpadDataSources/function/execFunction.ts
+++ b/packages/toolpad-app/src/toolpadDataSources/function/execFunction.ts
@@ -134,7 +134,7 @@ export default async function execFunction(
     userModule.evaluateSync({ timeout: 30000 });
 
     const defaultExport = await userModule.namespace.get('default', { reference: true });
-    data = await defaultExport.apply(null, [{ params, secrets }], {
+    data = await defaultExport.apply(null, [{ params, parameters: params, secrets }], {
       arguments: { copy: true },
       result: { copy: true, promise: true },
       timeout: 30000,
--- a/packages/toolpad-app/src/toolpadDataSources/rest/client.tsx
+++ b/packages/toolpad-app/src/toolpadDataSources/rest/client.tsx
@@ -279,7 +279,9 @@ function QueryEditor({
   );
 
   const queryScope = {
+    // TODO mark query as @deprecated
     query: previewParams,
+    parameters: previewParams,
   };
 
   const liveUrl: LiveBinding = useEvaluateLiveBinding({


packages/toolpad-app/src/toolpad/AppEditor/PageEditor/QueryEditor.tsx
```diff
@@ -235,16 +235,19 @@ function QueryNodeEditorDialog<Q>({
...
       query: input.attributes.query.value,
-      params:
-        (Object.entries(inputParams).filter(([, value]) =>
-          Boolean(value),
-        ) as BindableAttrEntries) || [],
-    }),
-    [input.attributes.query.value, inputParams],
-  );
+
+      // TODO mark params as @deprecated
```
bytasv, line 246: @Janpot could you help me with this - you mentioned of adding deprecation comment for old values, but I couldn't figure out where and how should we do that, maybe you have something in mind already?

packages/toolpad-app/src/toolpadDataSources/rest/client.tsx
```diff
@@ -279,7 +279,9 @@ function QueryEditor({
   );
 
   const queryScope = {
+    // TODO mark query as @deprecated
```
bytasv, line 282: @Janpot same question as above, in addition - if you could give a hint how and if I should provide a different type so it doesn't come up as `Query` type when typed in the binding editor

packages/toolpad-app/src/toolpadDataSources/function/client.tsx
```diff
@@ -141,7 +141,7 @@ function QueryEditor({
 
     const content = `
       interface ${EVENT_INTERFACE_IDENTIFIER} {       
-        params: {
+        parameters: {
```
Janpot, line 144: Something like this is what I meant above. But to be honest, I'm fine with removing it as well.

```suggestion
        @deprecated Use parameters instead
        params: {
          ${paramsMembers}
        }
        parameters: {
```

packages/toolpad-app/src/toolpadDataSources/function/execFunction.ts
```diff
@@ -134,7 +134,7 @@ export default async function execFunction(
     userModule.evaluateSync({ timeout: 30000 });
 
     const defaultExport = await userModule.namespace.get('default', { reference: true });
-    data = await defaultExport.apply(null, [{ params, secrets }], {
+    data = await defaultExport.apply(null, [{ params, parameters: params, secrets }], {
```
Janpot, line 137: Let's add a comment to `params` that it's there only for legacy reasons?

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/QueryEditor.tsx
Janpot: Roughly what I have in mind is something like updating the `BindingEditor` with

```tsx
<BindableEditor
  globalScope={queryScope}
  globalScopeAnotations={{
    query: {
      deprecated: true,
    },
    mySelect: {
      comment: 'Builtin select component',
      properties: {
        value: {
          comment: 'the select value...',
        },
      },
    },
  }}
/>;
```

So that we can add extra meta info to the typings and the global scope explorer.

feel free to pick that up

packages/toolpad-app/src/toolpadDataSources/function/client.tsx
bytasv: Thanks! I just had to wrap it with comment, otherwise it didn't work
```
/** @deprecated Use parameters instead */
params: {
```

image

Janpot: Ah right, if we go this route, maybe we should also add something like "will be removed in Toolpad v1" or so?
(and to the comment down below something like `TODO: remove after v1`?)

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/QueryEditor.tsx
Janpot: > From what I understand if we want to pass custom typings then we would need to implement our own jsonToTs method? In other words we would have to manually generate typing info?

Sort of, right below we are declaring these generated types to global variables:

```tsx
    const globals = Object.keys(globalScope)
      .map((key) => `declare const ${key}: RootObject[${JSON.stringify(key)}];`)
      .join('\n');
```

We should be able to add a comment to these declarations. I guess with the current system we can only annotate the top-level keys.

bytasv: I'm trying to hardcode the annotations but I can't figure out how to make it pick up that it's deprecated 🤔 

image

any ideas?

Within the app it appears without any extra details
image

Janpot: I'm not sure whether jsdoc picks up comments like `//*`, did you try `/** @deprecated ... */`?

bytasv: Oh damn, you are right 🤦 I did the other one correctly and for whatever reason used `//*`...

image

So I've got some ideas now how it can work. I could try to tackle https://github.com/mui/mui-toolpad/issues/497 if you could provide more details how you think it should behave 🤔 

I think we should be able to do the example that you suggested:

```
<BindableEditor
  globalScope={queryScope}
  globalScopeMeta={{
    query: {
      deprecated: 'Use parameters instead',
    },
    mySelect: {
      comment: 'Builtin select component',
      properties: {
        value: {
          comment: 'the select value...',
        },
      },
    },
  }}
/>;
```
or at least something similar

Janpot: > I could try to tackle https://github.com/mui/mui-toolpad/issues/497 if you could provide more details how you think it should behave

I'm not sure, I think we should start by benchmarking which information retool/appsmith provide to get an idea of a baseline.

I guess Among the features we could consider:
- group the scope variables
- make them filterable with a searchbox
- label/color them (e.g. this is a component, this is a query)
- add longer form descriptions (perhaps in tooltips), links to documentation
- show type information
- add _copy path to clipboard_ so it can be pasted in the editor
- add longer form descriptions to the nested properties
- ...?

Not necessary all at once ofcourse 🙂 , and I guess @gerdadesign may want to be involved in some parts of UX/design.

Janpot (approved):

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/JsExpressionEditor.tsx
```diff
@@ -44,9 +44,17 @@ export function JsExpressionEditor({
...
+    const queryDeprecation = `
+      /** @deprecated use parameters */
+      declare const query: Query;
+    `;
+
     const content = `
       ${type.join('\n')}
 
+      // TODO: cleanup after Toolpad v1
+      ${globalScope.query && globalScope.parameters ? queryDeprecation : ''}
```
bytasv, line 57: @Janpot I know this isn't the most elegant way, but I'm not sure if we should invest more if we plan to improve/rework this part entirely?

bytasv:
I think I got it working, @Janpot please check the changes out:

image

and 

image

packages/toolpad-app/src/toolpadDataSources/rest/client.tsx
```diff
@@ -50,6 +50,15 @@ import DEMO_BASE_URLS from './demoBaseUrls';
 
 const HTTP_METHODS = ['GET', 'POST', 'PUT', 'DELETE', 'PATCH', 'HEAD'];
 
+const DEPRECATED_URL_VARIABLES: GlobalScopeMeta = {
+  query: {
+    deprecated: 'Use parameters variable instead',
+  },
+  parameters: {
+    description: 'Parameters that can be bound to app scope variables',
```
Janpot, line 58: 👌

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/JsExpressionEditor.tsx
```diff
@@ -41,25 +44,45 @@ export function JsExpressionEditor({
...
+        let declaration = ``;
+        const metaData = globalScopeMeta[key] || {};
 
-    // TODO: cleanup after Toolpad v1
-    const queryDeprecation = `
-      /** @deprecated use parameters */
-      declare const query: Query;
-    `;
+        if (metaData.description) {
+          declaration = `/** ${metaData.description} */`;
```
Janpot, line 52: I believe they all need to go in the same comment block to support the case of having both a description and a comment defined

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/JsExpressionEditor.tsx
```diff
@@ -41,25 +44,45 @@ export function JsExpressionEditor({
     const type = jsonToTs(globalScope);
 
     const globals = Object.keys(globalScope)
-      .map((key) => `declare const ${key}: RootObject[${JSON.stringify(key)}];`)
-      .join('\n');
+      .map((key) => {
+        let declaration = ``;
+        const metaData = globalScopeMeta[key] || {};
```
Janpot, line 49: Could this potentially give trouble when someone passes `{ toString: 1 }` to `globalScope`? Perhaps this needs to check with [`hasOwnProperty`](https://github.com/mui/mui-toolpad/blob/594e6c7750de34029cea2604983d165e81f9e35e/packages/toolpad-app/src/utils/collections.ts#L11)

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/JsExpressionEditor.tsx
bytasv: `deprecated` is defined as `boolean | string` so if you want to add description of the deprecation then you should use it like this:
```
{
  query: {
    deprecated: 'Reason of deprecation'
  }
}
```
or do you mean using deprecation message + description?
Would we even use both at the same time?

Janpot: > or do you mean using deprecation message + description?

yes

> Would we even use both at the same time?

The `description` would describe the global variable. The `deprecation` would describe why it's deprecated or what to use instead.

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/JsExpressionEditor.tsx
Janpot: In other instances of this component the `globalScope` is derived from user input, so we don't know what gets passed. e.g. someone could create a TextField named "toString".

packages/toolpad-app/src/toolpadDataSources/rest/client.tsx
```diff
@@ -50,6 +50,15 @@ import DEMO_BASE_URLS from './demoBaseUrls';
 
 const HTTP_METHODS = ['GET', 'POST', 'PUT', 'DELETE', 'PATCH', 'HEAD'];
 
+const DEPRECATED_URL_VARIABLES: GlobalScopeMeta = {
```
Janpot, line 53: I'd rename it to `GLOBAL_SCOPE_META` now though

prakhargupta1:
@bytasv Somehow using parameters doesn't show any preview. While using query does:

https://user-images.githubusercontent.com/92228082/195256716-b28c38f5-c7ca-4b40-9acd-1297a2c40d5b.mov

bytasv:
@prakhargupta1 thanks, looking into it

Issue #1095, opened 2022-10-05
[RFC] renovatebot


Janpot:
### What's the problem? 🤔

Too many PRs opened by renovatebot. This is not very manageable, plus it eats into our render.com budget.

### Proposed solution 🟢

We change the schedule. Instead of creating a PR of every dependency and groups for certain others we create one PR for all dependencies to be opened bi-weekly? And create groups for specific other dependencies to run on different schedules.

groups:
- MUI dependencies: open PR as soon as new version available
- MUI X dependencies: open PR as soon as new version available
- dependencies that patch known vulnerabilities: open PR asap
- all dev dependencies: bi-weekly
- all normal dependencies: bi-weekly

PR #1099, opened 2022-10-05
New renovatebot schedule

Janpot:
Implementing https://github.com/mui/mui-toolpad/issues/1095

from https://docs.renovatebot.com/configuration-options/#packagerules

> Important to know: Renovate will evaluate all packageRules and not stop once it gets a first match. You should order your `packageRules` in ascending order of importance so that more important rules come later and can override settings from earlier rules if needed.


Janpot:
cc @mui/code-infra

Janpot:
Getting this in before the weekend so we can see it at work on Sunday

PR #1105, opened 2022-10-06
Allow updating and deleting DataGrid rows

### LINKED ISSUE #994: Make data grid editable
### Duplicates

### Latest version

### Summary 💡

As a user I should be able to update a cell of a row (inline) This means that the entire column has to be editable. Eg: ID should not be editable, but address column can be.

### Examples 🌈

https://docs.retool.com/docs/data-in-tables#editable-table-columns

### Motivation 🔦

_No response_

### LINKED ISSUE #1039: Allow deleting a row from data grid
### Duplicates

### Latest version

### Summary 💡

We should allow deleting a row. Ways we could do it:

1. Delete single row: Through an action button as one of the column. Clicking on it should open a confirmation box
2. Delete multiple rows: Through checkboxes as one of the column.
    
Relates to #994

bharatkashyap:
- Closes #1039 
- Closes #994 
- Visual demonstration 

Delete rows:

https://user-images.githubusercontent.com/19550456/197758901-6c36cea9-01e7-47f3-bf07-65629d06a117.mov

Edit rows:

https://user-images.githubusercontent.com/19550456/197758950-c8e4f991-a346-4b48-a4d9-cc8c532f9121.mov


packages/toolpad-components/src/DataGrid.tsx
```diff
@@ -298,20 +309,58 @@ const DataGridComponent = React.forwardRef(function DataGridComponent(
...
+              onClick={(event) => {
+                onDelete({
+                  event,
+                  toolpadEvent: {
+                    type: 'delete',
+                    id,
+                    row,
+                  },
+                });
+                setRows((prevRows) => prevRows.filter((prevRow) => prevRow.id !== id));
```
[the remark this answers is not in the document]
bharatkashyap: I've updated the implementation a bit for this to not be needed anymore

bharatkashyap:
The DataGrid remounts thrice before showing an updated state with the row deleted - I am not exactly sure how to provide an optimal UX around that: Should we blur the grid while the operation completes?

Janpot:
> The DataGrid remounts thrice before showing an updated state with the row deleted

Why does it remount?

packages/toolpad-components/src/DataGrid.tsx
```diff
@@ -293,10 +317,55 @@ const DataGridComponent = React.forwardRef(function DataGridComponent(
     [selection?.id],
   );
 
-  const columns: GridColumns = React.useMemo(
-    () => (columnsProp ? parseColumns(columnsProp) : []),
-    [columnsProp],
-  );
+  const columns: GridColumns = React.useMemo(() => {
```
apedroferreira, line 320: not sure about using an effect to alter the dom from the component - is it not possible to set these columns as initial values according to whether there is an `onDelete` prop or not?

apedroferreira: but i guess this might be the best way, if the prop can be changed... just wondering if there's another way.

apedroferreira:
Looks good, just a few things I noticed:
- Do we have to refetch manually after the delete/update? Shouldn't there at least be a way to do it automatically then?
- I noticed in the video that the delete takes a while - maybe we could at least show a loading state while it'd deleting?
- Also I'm not sure how/if we're handling errors during update or delete

bharatkashyap:
> Looks good! Just a few things I noticed:
> 
> * Do we have to refetch manually after the delete/update? Shouldn't there at least be a way to do it automatically then?
> * I noticed in the video that the delete takes a while - maybe we could at least show a loading state while it's deleting? I also didn't dislike doing an optimistic update like I think you were doing before? An optimistic update would just need to be reverted if there's an error, for example.
> * Also I'm not sure how/if we're handling errors during update or delete

I've updated the implementation (and demonstration) a bit, but these questions are still relevant:

> Do we have to refetch manually after the delete/update? Shouldn't there at least be a way to do it automatically then?

I agree, I think we should follow up with adding a way to specify dependent queries when defining queries to preclude having to write an extra line of code each time to refetch items after any operation.

> I noticed in the video that the delete takes a while - maybe we could at least show a loading state while it's deleting? 

I used Toolpad itself to show a loading state when any of the queries are in operation. Perhaps we should make that behaviour the default, or let a user be able to do this themselves?

> I also didn't dislike doing an optimistic update like I think you were doing before? An optimistic update would just need to be reverted if there's an error, for example.

The optimistic update happens for the edit; I agree that it'll be better UX to make that happen for delete as well. I'll explore adding that behaviour.

apedroferreira (approved):

prakhargupta1:
@bharatkashyap As a developer can I choose which column I want to allow my ops team to edit?

bharatkashyap:
> @bharatkashyap As a developer can I choose which column I want to allow my ops team to edit?

In this state all columns are editable; I'm planning to add a feature to our column editor to set specific columns as editable in a follow up to this.

bharatkashyap:
Closing https://github.com/mui/mui-toolpad/pull/1227 and this in favour of the new direction of Toolpad

Issue #1086, opened 2022-10-03
Unable to fetch images in "Dog Images" template app


apedroferreira:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

It's not possible to fetch new dog images in the respective template app unless a sub breed is defined.
This seems to be because the " sub breed" input now defaults to "January" when it's empty, which is one of the default values for the Select component.

### Expected behavior 🤔

It should be possible to fetch new dog images in the respective template app even if sub-breed is empty, or has no options.

### Steps to reproduce 🕹

Steps:

1. Make new app from the "Fetch images" template (app that fetches dog images according to breeds).
2. Try to fetch an image.
3. No new image is fetched - the default image stays the same.

WangLarry:
When component rended at first time, binding value has not evaluated yet. At this time, should use it's initial value in dom.

WangLarry:
@Janpot 
https://github.com/mui/mui-toolpad/pull/1091#issuecomment-1267829977

bharatkashyap:
Fixed by #1106

PR #1106, opened 2022-10-07
Revert "[core] Add default data to DataGrid, Image, Select components"

Janpot:
Looks like the implementation of https://github.com/mui/mui-toolpad/pull/1048 still has a few problems:

- https://github.com/mui/mui-toolpad/issues/1086
- I'm noticing issues in the components integration test. Since this PR, when I open [that app](https://github.com/mui/mui-toolpad/blob/master/test/integration/components/componentsDom.json) in the editor, I immediately see "1 changes saved." in the console. I think it's also what's making the integration test flaky in https://github.com/mui/mui-toolpad/pull/1097 as it triggers a dialog when trying to navigate away from a page with unsaved changes

cc @bharatkashyap

### DIFF
--- a/packages/toolpad-components/src/DataGrid.tsx
+++ b/packages/toolpad-components/src/DataGrid.tsx
@@ -154,11 +154,6 @@ export function inferColumns(rows: GridRowsProp): SerializableGridColumns {
   return Object.entries(firstRow).map(([field, value]) => {
     return {
       field,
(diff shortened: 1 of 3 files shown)


bharatkashyap:
Agree with the revert and refine the approach

bharatkashyap (approved):

PR #1109, opened 2022-10-07
[core] Remove jsdoc eslint

Janpot:
This was specific for the MUI X documentation and bears no relevance for Toolpad. It's returning false positives on another PR I'm working on

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,7 +2,7 @@ const baseline = require('@mui/monorepo/.eslintrc');
 
 module.exports = {
   ...baseline,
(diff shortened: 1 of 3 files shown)


.eslintrc.js
```diff
@@ -2,7 +2,7 @@ const baseline = require('@mui/monorepo/.eslintrc');
 
 module.exports = {
   ...baseline,
-  plugins: [...baseline.plugins, 'jsdoc'],
+  plugins: baseline.plugins,
```
oliviertassinari, line 5: Simpler? 😁

```suggestion
```

Janpot: 😃

PR #1119, opened 2022-10-09
pin  node version for more reproducability in the builds

Janpot:


Janpot:
That's unrelated I think. That would set it for the docs as well, I think netlify setup can't handle that right now. 

But your comment made me realize we have to update render.yml as well https://github.com/mui/mui-toolpad/blob/f0148d388adab0c1a2b9164bf82b2b7f50cc7275/render.yaml#L16

Issue #1121, opened 2022-10-10
Query editor shows unsaved changes modal when options change


Janpot:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

https://user-images.githubusercontent.com/2109932/194822330-f9bf0d1d-eb90-460e-84c8-81198a706c33.mov

After changing one of the query options at the bottom, when pressing "save" it shows the modal that blocks closing on unsaved changes. That shouldn't happen when clicking "save". The changes do persist though, regardless of whether "Ok" or "Cancel" is clicked.

### Expected behavior 🤔

No modal is shown, the content is saved and the query editor is closed

### Steps to reproduce 🕹

See video

Issue #990, opened 2022-09-13
[RFC] Fold mutations into queries


Janpot:
### What's the problem? 🤔

Currently the concept of query vs. mutation is about how client-side fetching behaves:

- query: idempotent, there is no problem running them automatically or retry them on error
- mutation: mutates something in the backend, needs to be explicitly run by the user. Can not automatically be retried on error.

In general the difference is only in when it can be triggered. Almost all else about it is identical. I've noticed this leads to a lot of confusion with the less technical team members.

Also, with the current model it's hard to convert between one and the other. You have to decide up front when the query is executed and if you want to change your mind you'll have to start all over.

### Proposed solution 🟢

I propose to fold the `mutation` functionality into the query. We'll add a switch between "run manually" and "run automatically". This will make more clear what the behavioral difference is.
Users will be able to bind to them manually in a very similar way as current mutations. e.g. `await query.fetch()`.
We will also make sure to make the response available as `.data` so the page can bind to it.

### Resources and benchmarks 🔗

Retool:

![Screenshot 2022-09-13 at 14 02 14](https://user-images.githubusercontent.com/2109932/189895872-5cfe2061-8616-42ac-80d9-6b6b5388b94c.png)

Superblocks:
![Screenshot 2022-09-13 at 14 13 43](https://user-images.githubusercontent.com/2109932/189898148-67f283ae-7610-43dc-999b-e016e160a222.png)

Appsmith:
![Screenshot 2022-09-13 at 14 16 56](https://user-images.githubusercontent.com/2109932/189900281-3ea99861-5e10-46aa-a5b3-16821aaeb0ee.png)

bharatkashyap:
Agree with this completely.

apedroferreira:
+1, not seeing any real drawbacks to this at first glance

Janpot:
Copying to do list from https://github.com/mui/mui-toolpad/pull/1122 so I can work on this in multiple PRs

To Do list:
- [ ] Migrate existing mutations to queries (blocked on https://github.com/mui/mui-toolpad/pull/776)
- [ ] Remove current "migrations" feature in favor of queries
- [ ] https://github.com/mui/mui-toolpad/issues/1121
- [ ] Automatically switch to mutation mode when user selects non-idempotent HTTP method in the fetch datasource

PR #1122, opened 2022-10-10
Make queries capable of mutations

Janpot:
Step 1 of https://github.com/mui/mui-toolpad/issues/990

https://user-images.githubusercontent.com/2109932/194826423-aa7c8a62-731d-4fbe-a813-f85e88d9b28b.mov

To Do list:
- [ ] Migrate existing mutations to queries (blocked on https://github.com/mui/mui-toolpad/pull/776)
- [ ] Remove current "migrations" feature in favor of queries
- [ ] https://github.com/mui/mui-toolpad/issues/1121
- [ ] Automatically switch to mutation mode when user selects non-idempotent HTTP method in the fetch datasource

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,7 +2,7 @@ const baseline = require('@mui/monorepo/.eslintrc');
 
 module.exports = {
   ...baseline,
-  plugins: [...baseline.plugins, 'jsdoc'],
+  plugins: baseline.plugins,
   /**
    * Sorted alphanumerically within each group. built-in and each plugin form
    * their own groups.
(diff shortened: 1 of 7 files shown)


apedroferreira:
Just an idea, not sure if it's good, but what if "Fetch on page load" was just a checkbox too?

apedroferreira (approved):

Janpot:
Currently it kind of cycles between two modes. idempotent and non-idempotent queries. I feel like a checkbox doesn't clearly convey that it's two fundamentally different things. A checkbox to me tells me turning something on or off, and here so much more is happening. Somehow we need to make clear that in one mode it's just "fetch when you ask" and the other mode is "Toolpad can fetch however much they want, as long as they do whateevr they can to get the data on the page". "fetch on page load" still sort of communicates to me that there will be only one fetch happening, which is not true.

apedroferreira:
> Currently it kind of cycles between two modes. idempotent and non-idempotent queries. I feel like a checkbox doesn't clearly convey that it's two fundamentally different things. A checkbox to me tells me turning something on or off, and here so much more is happening. Somehow we need to make clear that in one mode it's just "fetch when you ask" and the other mode is "Toolpad can fetch however much they want, as long as they do whateevr they can to get the data on the page". "fetch on page load" still sort of communicates to me that there will be only one fetch happening, which is not true.

Got it, makes sense. Maybe the text in the selector options can make that behavior more obvious somehow? Maybe something like "update automatically" or "refetch on change", not sure...
But yeah also in case we end up having different modes I think a selector for the mode actually seems good.

Janpot:
@apedroferreira I changed it to "Fetch at any time to always be available on the page", what do you think?

PR #1123, opened 2022-10-10
disable react-router-dom update

Janpot:
* disable `react-router-dom` updates: `react-router` folks came back on their promise of moving `usePrompt` features from v5 to v6 (https://github.com/remix-run/react-router/issues/8139), yet we depend on it to prevent navigating away when there's unsaved changes. I'm not 100% convinced their proposed solution is the most sensible for us. Short term we're disabling updates to this dependencies. Long term we have the following choices:
    * Do nothing, wait for them to change their mind and support the future in `react-router` > 6. Basically stay stuck on an outdated version of the latest major, potentially forever.
    * Migrate back to v5 which supports the functionality. Basically staying forever on the last version of an outdated major.
    * Migrate towards a library that cares more about the migration path for their users. e.g. [`@tanstack/react-location`](https://www.npmjs.com/package/@tanstack/react-location)
    * Refactor our UX to not rely on dialogs that asks users to confirm they want to navigate away with unsaved changes.
    
    IMO only the last two of those are desirable long-term

 * replace the package matcher for MUI(X) packages with regex

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -27,28 +27,12 @@
     },
     {
       "groupName": "MUI Core",
-      "matchPackageNames": [
-        "@mui/material",
-        "@mui/joy",
-        "@mui/icons-material",
-        "@mui/base",
-        "@mui/system",
-        "@mui/styles",
-        "@mui/utils"
-      ],
+      "matchPackagePatterns": ["^@mui/"],
       "prCreation": "immediate"
     },
     {
       "groupName": "MUI X",
-      "matchPackageNames": [
-        "@mui/x-data-grid",
-        "@mui/x-data-grid",
-        "@mui/x-data-grid-pro",
-        "@mui/x-data-grid-premium",
-        "@mui/x-data-grid-generator",
-        "@mui/x-date-pickers",
-        "@mui/x-date-pickers-pro"
-      ],
+      "matchPackagePatterns": ["^@mui/x-"],
       "prCreation": "immediate"
     },
     {
@@ -58,7 +42,8 @@
         "node-fetch",
         "eslint-config-airbnb-typescript",
         "chalk",
-        "react-router"
+        "react-router",
+        "react-router-dom"
       ],
       "enabled": false
     }


Janpot:
The current use-case is for when moving away from the code component editor. One alternative UX could be to 
- When navigating away internally: leave the changes unsaved + add a _has unsaved changes_ indicator in explorer.
- When navigating away from Toolpad when there are unsaved changes, show prompt "hey there are unsaved changes, you sure you want to navigate away?" (this could live outside of `react-router`, we already do something similar when changes are still syncing in the background)
- Support Cmd+alt+S to save everything at once, for this workflow, should we have something like tabs to show "open editors", just like in vscode or something?

bytasv:
> Refactor our UX to not rely on dialogs that asks users to confirm they want to navigate away with unsaved changes.

Maybe that actually wouldn't be too bad 🤔 I mean we could explore if it's not too big of an overhead to have 2 versions (worse case) for the code that we are editing:
one version which represents the code that is broken
another version which is "the last working version" so that the app would not be in the broken state

apedroferreira:
I'm familiar with this issue too.
I think we can try to keep using the latest version of React Router and try to adjust by not relying on blocking navigation - but that's not always perfect and blocking navigation can be a useful feature...

I guess we could always save the current state of an editor to memory or sessionStorage, and restore those unsaved changes, if that's not potentially too much to store. But we would also need very visible indicators that there are unsaved changes when there are some.
On top of that, we could block navigation ourselves where possible, even if it's not 100% the same as blocking the browser navigation.

Janpot:
> I guess we could always save the current state of an editor to memory or sessionStorage/localStorage, and restore those unsaved changes

To clarify further, the dialog that warns for unsaved changes in the application model is not affected by this. I further investigated and the only affected one specifically is the warning that blocks moving from the code component editor with unsaved changes to anywhere else in the app.

bharatkashyap:
> I further investigated and the only affected one specifically is the warning that blocks moving from the code component editor with unsaved changes to anywhere else in the app.

Agree with this strategy, feels like it offers the most predictable experience for a user

Janpot:
They're reimplementing it https://github.com/remix-run/react-router/issues/8139#issuecomment-1332652167

PR #1125, opened 2022-10-10
Fix warning on save

Janpot:
Fixes https://github.com/mui/mui-toolpad/issues/1126

The warning was caused because of the way interaction was set up between query editor dialog and the query editors of the datasource. Tried multiple times to patch it but all attempts made it worse. This PR refactors it to have the datasource act as a straightforward controlled component again with an `onCommit` property to trigger save actions. The whole `QueryEditorShell` could be removed together with its context. It's all top-down rendered again from state as it should be. This made it easier to control on which actions a dialog needs to be triggered and on which it doesn't. The whole  resulted in a significant reduction of code as well.

Recommended to review with whitespace disabled.


packages/toolpad-app/src/toolpad/AppEditor/PageEditor/MutationEditor.tsx
```diff
@@ -13,79 +13,33 @@ import {
...
-        </DialogContent>
-      ) : (
-        <DialogContent>
-          <Alert severity="error">Datasource &quot;{dataSourceId}&quot; not found</Alert>
-        </DialogContent>
-      )}
+function createQueryModel<Q>(node: appDom.MutationNode<Q>): QueryEditorModel<Q> {
+  const inputParams = node.params || EMPTY_OBJECT;
+  const params =
+    (Object.entries(inputParams).filter(([, value]) => Boolean(value)) as BindableAttrEntries) ||
```
bytasv, line 36: NIT: if `key` is not used we could use `Object.values(...)` instead

packages/toolpad-app/src/toolpadDataSources/function/client.tsx
```diff
@@ -161,66 +161,51 @@ function QueryEditor({
...
-              onChange={(newParams) => setInput((existing) => ({ ...existing, params: newParams }))}
-              globalScope={globalScope}
-              liveValue={paramsEditorLiveValue}
+    <SplitPane split="vertical" size="50%" allowResize>
+      <SplitPane split="horizontal" size={85} primary="second" allowResize>
+        <QueryInputPanel onRunPreview={handleRunPreview}>
+          <Box sx={{ flex: 1, minHeight: 0 }}>
+            <TypescriptEditor
+              value={input.query.module}
+              onChange={(newValue) =>
```
bytasv, line 174: How do we usually decide whether we should wrap handler with `useCallback` (i.e. `handleCommit`) or can just pass them inline?

bytasv (approved):
> The whole resulted in a significant reduction of code as well.

The best code is removed code, right? 😃

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/MutationEditor.tsx
Janpot: We're operating on entries here, not values, because we want `params` to be a list of entries, not values

bharatkashyap (approved):

packages/toolpad-app/src/toolpadDataSources/function/client.tsx
Janpot: You're encouraged to always use `useCallback` (or our own custom `useEvent`). Though I don't want to make this a hard rule.

Mind you that the only thing that changed to this code is indentation. I didn't do a review of all the code that I reindented. I recommend reviewing with _whitespace only_ enabled. About 1/3 to half of the code has been reindented.

Janpot:
> The best code is removed code, right?

https://twitter.com/codewisdom/status/1156181903541776385?lang=en

Issue #1101, opened 2022-10-06
circleci test_static consumes OOM


Janpot:
### Current behavior 😯

Regularly seeing `test_static` circleci jobs fail because of [OOM](https://support.circleci.com/hc/en-us/articles/115014359648-Exit-Code-137-Out-of-Memory) issues.

example: https://app.circleci.com/pipelines/github/mui/mui-toolpad/3578/workflows/b14c82c8-b004-45c1-b1ca-5a30412b6154/jobs/13481

```
@mui/toolpad-app: [build:*next] Killed
@mui/toolpad-app: [build:*next] error Command failed with exit code 137.
@mui/toolpad-app: [build:*next] info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
@mui/toolpad-app: [build:*next] yarn run build:next exited with code 137
```

[Insights](https://app.circleci.com/insights/github/mui/mui-toolpad/workflows/pipeline/jobs) confirmed an increase in max memory consumption since Sept 30

![Screenshot 2022-10-06 at 10 07 03](https://user-images.githubusercontent.com/2109932/194257787-a87506cb-00b3-44cf-ba7a-4882ad51bbde.png)

Might be caused by https://github.com/mui/mui-toolpad/pull/1043. I'm putting my money on the [source map changes](https://github.com/mui/mui-toolpad/blob/f0b69faa369f344c4225592eedc06f08ca5dffa6/packages/toolpad-app/next.config.mjs#L173) 🙂 

_We were already quite close to 100%, so if we can't bring memory consumption down, scaling the instance may be in order._

bytasv:
Do we currently also generate sourcemaps when running out tests? Wondering if we could easily test the assumption about sourcemaps impact by opening PR where we disable sourcemaps generation and comparing with the one that has it enabled? 🤔

apedroferreira:
Let's not upload sourcemaps with Sentry, doesn't make sense for us to do it since we're open source. Hopefully that will help.

Janpot:
I believe it'll be sourcemap generation that will be heavy, not the uploading

apedroferreira:
> I believe it'll be sourcemap generation that will be heavy, not the uploading

Yep I expressed myself a bit wrong, if it's Sentry that's generating them during the build and we can skip that it should be fine.

bytasv:
[...]

PR #1129, opened 2022-10-11
Improvements to Sentry and sourcemaps

apedroferreira:
Try changes according to this discussion https://github.com/mui/mui-toolpad/issues/1101

Also remove usage of `withSentry` as it seems that server functions are now wrapped automatically.


Janpot:
> Also remove usage of `withSentry` as it seems that server functions are now wrapped automatically.

I saw it in the Next.js example still. Is that a recent change?

packages/toolpad-app/next.config.mjs
```diff
@@ -92,7 +92,8 @@ const sentryWebpackPluginOptions = {
 
   silent: true, // Suppresses all logs
 
-  dryRun: !process.env.TOOLPAD_SENTRY_DSN,
+  dryRun: true,
+  sourceMapReference: false,
```
Janpot, line 96: Can't seem to find docs on this, what does it do?

apedroferreira:
> I saw it in the Next.js example still. Is that a recent change?

It was telling me in the console that `withSentry` isn't needed anymore, it probably comes from a very recent version update.

packages/toolpad-app/next.config.mjs
apedroferreira: From https://github.com/getsentry/sentry-webpack-plugin#options:
`sourceMapReference` 
```Determines whether sentry-cli should attempt to link minified files with their corresponding maps. By default, it will match files and maps based on name, and add a Sourcemap header to each minified file for which it finds a map. Can be disabled if all minified files contain sourceMappingURL. Defaults to true.```

Maybe it's not worth disabling though, I just thought maybe it'll help Sentry have to do less work, but it's probably not significant...

apedroferreira:
I've added the flag manually just in case.

PR #1141, opened 2022-10-12
Mandate node version for toolpad-app

Janpot:
For maintainers running Toolpad on their locally installed node


Janpot:
🤔  right, netlify wants to install everything, even `toolpad-app` dependencies.

packages/toolpad-app/package.json
```diff
@@ -137,5 +137,8 @@
   "typings": "./index.d.ts",
   "publishConfig": {
     "access": "public"
+  },
+  "engines": {
+    "node": ">=18"
```
oliviertassinari, line 142: Why so high?

```jsx
$ node --version
v16.16.0
```

oliviertassinari:
> netlify wants to install everything, even toolpad-app dependencies.

https://github.com/mui/mui-toolpad/blob/3c603a5cc1abbfb9ef66572f184debcd45917246/netlify.toml#L11

packages/toolpad-app/package.json
Janpot: > Why so high?

We could treat it as any other dependency in the project, and keep it up to date through renovatebot.

oliviertassinari:
@Janpot what do you think about reopening to set Node.js 16 as the lower limit? This is what is recommended for most developers: https://nodejs.org/en/

Janpot:
> what do you think about reopening to set Node.js 16 as the lower limit?

Not all versions in 16.x support all the features we use. (e.g. I think `AbortSignal.timeout` doesn't work pre 16.14). How do you feel about setting it to 16.18? Would there be a case for not keeping it up to date?

oliviertassinari:
> How do you feel about setting it to 16.18? Would there be a case for not keeping it up to date?

It sounds great. It would fail upfront for developers, not having to wonder why AbortSignal.timeout is not defined.

Janpot:
I'll set it at 16.17 for now. There's a circleci job runing on the docker machine executor and it only has 16.17.1 installed. We can look into fixing this part of CI in a separate effort.

Issue #1058, opened 2022-09-27
Home page overflows horizontally


apedroferreira:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

When home page has many apps it overflows horizontally.

<img width="1792" alt="Screen Shot 2022-09-27 at 19 50 23" src="https://user-images.githubusercontent.com/10789765/192611460-19f217ef-6bac-48da-9bb6-770720747da7.png">
<img width="1792" alt="Screen Shot 2022-09-27 at 19 50 48" src="https://user-images.githubusercontent.com/10789765/192611474-d466dbeb-9cad-474b-8f2e-df7f1f771082.png">

### Expected behavior 🤔

When home page has many apps it should not overflow horizontally.

### Steps to reproduce 🕹

Steps:

1. Create many apps
2. Look at home page

apedroferreira:
> @apedroferreira can you check if you can still reproduce it?

Yep I still have the scrollbars.

bytasv:
> @bytasv Maybe you did already, but make sure to set _Show scrollbars: always_ on your Mac. (I can still reproduce)

Thanks! that's the missing "ingredient" 👍

PR #1154, opened 2022-10-13
Fix visible horizontal scroll

bytasv:
- Closes https://github.com/mui/mui-toolpad/issues/1058

Found this suggestions: https://stackoverflow.com/questions/23367345/100vw-causing-horizontal-overflow-but-only-if-more-than-one - if someone knows a better way or if this should be done differently please lemme know.

image

### DIFF
--- a/packages/toolpad-app/src/toolpad/ToolpadShell/index.tsx
+++ b/packages/toolpad-app/src/toolpad/ToolpadShell/index.tsx
@@ -10,6 +10,7 @@ export interface ToolpadShellProps {
 
 const ToolpadShellRoot = styled('div')({
   width: '100vw',
+  maxWidth: '100%',
   minHeight: '100vh',
   display: 'flex',
   flexDirection: 'column',


apedroferreira:
> Oh ok, just wondering, would it make sense then to make the width of this element `100%`, and the width of its parent (`body`?) `100vw`?

I was thinking maybe `width: 100%; max-width: 100vw;` might be more intuitive too.
I wonder why something with `width: 100vw` is overflowing though.

Janpot:
> I wonder why something with `width: 100vw` is overflowing though.

Well, I understand its parent is overflowing vertically, adding a scrollbar to the right, making the parent's width `(100vw minus scrollbar width)` while the child's `width` still is `100vw`. So intuitively I'd say we're making the wrong thing `100vw`. ofcourse the `max-width` solution (or `width: min(100vw, 100%)`) would work as well, it just doesn't feel 100% right to me. But I'm fine width 95% right if other tradeoffs come up.

apedroferreira:
> Well, I understand its parent is overflowing vertically, adding a scrollbar to the right

Ok so the scrollbar is not accounted for in `vw`, in that case Vytautas' solution is actually probably better than what I said.
In my opinion think this solution is fine, even if maybe there's a better way.

apedroferreira (approved):

bytasv:
> Oh ok, just wondering, would it make sense then to make the width of this element 100%, and the width of its parent (body?) 100vw?

body is the element that gets scrollbars so we would end up in the same situation (I also just tested it out to be sure)

Janpot:
🤔 when I just replace `width: 100vw` with `width: 100%` I also don't seem to get the scrollbars. I wonder why I made it `100vw` in the first place?

Issue #1052, opened 2022-09-25
Should't show query content to client


WangLarry:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

'app/[appId]/[version]' entry is ssr, which call loadRenderTree. It load dom that contain query node's attributes.query.value, which is perhaps contain secret infos. So we should hidden it for client.

```json
             "j2t3hpa": {
            "id": "j2t3hpa",
            "name": "query4",
            "type": "query",
            "parentId": "6k1pc4c",
            "attributes": {
              "query": {
                "type": "const",
                **"value": {
                  "module": "export default async function ({ params }: ToolpadFunctionEvent) {\n  console.info('Executing function with params:', params);\n  const url = new URL('https://gist.githubusercontent.com/saniyusuf/406b843afdfb9c6a86e25753fe2761f4/raw/523c324c7fcc36efab8224f9ebb7556c09b69a14/Film.JSON');\n  url.searchParams.set('timestamp', String(Date.now()));\n  const response = await fetch(String(url));\n  if (!response.ok) {\n    throw new Error(`HTTP ${response.status}: ${response.statusText}`);\n  }\n  return response.json();\n}"
                }**
              },
              "dataSource": {
                "type": "const",
                "value": "function"
              },
              "connectionId": {
                "type": "const",
                "value": null
              }
            },
            "parentProp": "queries",
            "parentIndex": "a5"
          },
```

### Steps to reproduce 🕹

Steps:

1.
2.
3.
4.

apedroferreira:
Thanks @WangLarry .
@Janpot would know better but I think that it's not yet secure to use secrets in Toolpad if the application can be accessed by people you don't trust. That functionality is in the plans though!

Janpot:
@WangLarry Would you be interested in opening a pull request?

To give you a starting point, the following should work:

```diff
index e3b73f85..7e0234ca 100644
--- a/packages/toolpad-app/src/appDom.ts
+++ b/packages/toolpad-app/src/appDom.ts
@@ -14,7 +14,7 @@ import { ConnectionStatus, AppTheme } from './types';
 import { omit, update, updateOrCreate } from './utils/immutability';
 import { camelCase, generateUniqueString, removeDiacritics } from './utils/strings';
 import { ExactEntriesOf, Maybe } from './utils/types';
-import { filterValues } from './utils/collections';
+import { mapProperties } from './utils/collections';
 
[...]

PR #1155, opened 2022-10-13
[runtime] hide queries in the browser

Janpot:
Fixes https://github.com/mui/mui-toolpad/issues/1052

### DIFF
--- a/packages/toolpad-app/src/appDom.ts
+++ b/packages/toolpad-app/src/appDom.ts
@@ -14,7 +14,7 @@ import { ConnectionStatus, AppTheme } from './types';
 import { omit, update, updateOrCreate } from './utils/immutability';
 import { camelCase, generateUniqueString, removeDiacritics } from './utils/strings';
 import { ExactEntriesOf, Maybe } from './utils/types';
-import { filterValues } from './utils/collections';
+import { mapProperties } from './utils/collections';
 
 export const RESERVED_NODE_PROPERTIES = [
   'id',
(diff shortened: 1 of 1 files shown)


test/integration/propControls/index.spec.ts
```diff
@@ -40,35 +24,28 @@ test('can control component prop values in properties control panel', async ({
   const firstInputLocator = canvasInputLocator.first();
   await clickCenter(page, firstInputLocator);
 
-  await editorModel.componentEditor.waitFor();
+  await editorModel.componentEditor.waitFor({ state: 'visible' });
+
+  const labelControlInput = editorModel.componentEditor.getByLabel('label', { exact: true });
```
apedroferreira, line 29: i guess i hadn't found this selector that gets the input for a label? seems pretty useful

test/integration/propControls/index.spec.ts
```diff
@@ -40,35 +24,28 @@ test('can control component prop values in properties control panel', async ({
...
+  await labelControlInput.fill(TEST_VALUE_2);
 
-  await expect(firstInputLabelLocator).not.toHaveText(TEST_VALUE_2);
-
-  const labelControlInputLocator = await getPropControlInputLocator(editorModel, 'label');
-  await labelControlInputLocator.fill('');
-  await labelControlInputLocator.fill(TEST_VALUE_2);
-
-  await expect(firstInputLabelLocator).toHaveText(TEST_VALUE_2);
+  await inputByLabel.waitFor({ state: 'visible' });
```
apedroferreira, line 50: So we're not checking if the value in the label matches in the end anymore? Was there any issue with that?

apedroferreira (approved):
Thanks for improving that test, it's definitely an improvement!
I was waiting to see if it was still flaky or if the issue that made it flaky was unrelated, it was seeming a bit less flaky lately.

Just left a question about one of the changes to the test.

test/integration/propControls/index.spec.ts
Janpot: Yes we are, if the value doesn't match this selector wouldn't locate elements

apedroferreira: oh it checks with that value already, i missed that. looks good then!

PR #1159, opened 2022-10-14
[docs] Improve markdownlint

oliviertassinari:
This change is based on https://github.com/mui/material-ui/pull/34686, simpler :)


.markdownlint-cli2.cjs
```diff
@@ -0,0 +1,3 @@
+const baseline = require('@mui/monorepo/.markdownlint-cli2.cjs');
```
Janpot, line 1: Would it make sense for the future to prefer ESM over commonjs?

Janpot (approved):

.markdownlint-cli2.cjs
oliviertassinari: Yes, but I didn't use the `.mjs` option because it wouldn't lot on the VSCode extension: https://github.com/DavidAnson/vscode-markdownlint/issues/248.

PR #1162, opened 2022-10-15
Allow iframes

oliviertassinari:
I'm doing a quick POC https://github.com/mui/mui-store/pull/224 to use Toolpad to render a custom extension for showing WooCommerce information on Zendesk. It would looks something like this: https://www.zendesk.com/marketplace/apps/support/273938/woocommerce-by-zenplates/. 

Screenshot 2022-10-15 at 21 10 34

https://mui.zendesk.com/agent/tickets/4354

However, I can't load the Toolpad app with an iframe. The alternative would be to render the app with React but I'm not aware that it's still something that we can do with Toolpad, so hence my PR.

### DIFF
--- a/packages/toolpad-app/next.config.mjs
+++ b/packages/toolpad-app/next.config.mjs
@@ -64,6 +64,10 @@ const regexEqual = (x, y) => {
 // See https://nextjs.org/docs/advanced-features/security-headers
 const securityHeaders = [
   {
+    key: 'X-Frame-Options',
+    value: 'DENY',
+  },
+  {
     // Force the browser to trust the Content-Type header
     // https://stackoverflow.com/questions/18337630/what-is-x-content-type-options-nosniff
     key: 'X-Content-Type-Options',
@@ -189,6 +193,24 @@ export default /** @type {import('next').NextConfig} */ withSentryConfig(
         {
           source: '/:path*',
           headers: securityHeaders,
+        },
+        {
+          source: '/reactDevtools/:path*',
+          headers: [
+            {
+              key: 'X-Frame-Options',
+              value: 'SAMEORIGIN',
+            },
+          ],
+        },
+        {
+          source: '/app-canvas/:path*',
+          headers: [
+            {
+              key: 'X-Frame-Options',
+              value: 'SAMEORIGIN',
+            },
+          ],
         },
       ];
     },
--- a/packages/toolpad-app/next.config.mjs
+++ b/packages/toolpad-app/next.config.mjs
@@ -64,10 +64,6 @@ const regexEqual = (x, y) => {
 // See https://nextjs.org/docs/advanced-features/security-headers
 const securityHeaders = [
   {
-    key: 'X-Frame-Options',
-    value: 'SAMEORIGIN',
-  },
-  {
     // Force the browser to trust the Content-Type header
     // https://stackoverflow.com/questions/18337630/what-is-x-content-type-options-nosniff
     key: 'X-Content-Type-Options',
--- a/packages/toolpad-app/next.config.mjs
+++ b/packages/toolpad-app/next.config.mjs
@@ -191,7 +191,7 @@ export default /** @type {import('next').NextConfig} */ withSentryConfig(
     headers: async () => {
       return [
         {
-          source: '/:path*',
+          source: '/((?!deploy/).*)',
           headers: securityHeaders,
         },
         {
--- a/packages/toolpad-app/next.config.mjs
+++ b/packages/toolpad-app/next.config.mjs
@@ -195,15 +195,6 @@ export default /** @type {import('next').NextConfig} */ withSentryConfig(
           headers: securityHeaders,
         },
         {
-          source: '/reactDevtools/:path*',
-          headers: [
-            {
-              key: 'X-Frame-Options',
-              value: 'SAMEORIGIN',
-            },
-          ],
-        },
-        {
           source: '/app-canvas/:path*',
           headers: [
             {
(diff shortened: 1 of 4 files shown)


oliviertassinari:
I have another use case for allowing iframes. I would like to have all the health rows filled, e.g. https://www.notion.so/mui-org/KPIs-1ce9658b85ce4628a2a2ed2ae74ff69c#08a756250cd7499a9400ffdef4c50b33 like GitLab has: 

Screenshot 2022-10-16 at 00 11 43

Metabase might help us, but they only support SQL, and it's not really the output/level of control I'm looking for:

Screenshot 2022-10-16 at 00 16 39

Janpot (approved):

oliviertassinari:
I'm making a change before merging, I only need to allow iframes on the app built, the editor shouldn't allow it.

Janpot:
yes, that would be better. or we make it a setting, like the public apps

oliviertassinari:
> yes, that would be better. or we make it a setting, like the public apps

Metabase removes `X-Frame-Options: DENY` when the app is public. However, in my case, I don't think that it can be made public, I would still like to authenticate the use inside Zendesk.

packages/toolpad-app/pages/deploy/[appId]/[[...path]].tsx
```diff
@@ -12,6 +12,8 @@ export const getServerSideProps: GetServerSideProps<ToolpadAppProps> = async (co
     import('../../../src/server/basicAuth'),
   ]);
 
+  context.res.removeHeader('X-Frame-Options');
```
Comment (oliviertassinari) [inline packages/toolpad-app/pages/deploy/[appId]/[[...path]].tsx:15]: I assumed that Next.js only starts streaming the response after `getServerSideProps` is called. Considering that you can change the status to be a 404, It seems to be safe.

Janpot:
> However, in my case, I don't think that it can be made public, I would still like to authenticate the use inside Zendesk

yes, I meant a second option, alongside the "make public" option

packages/toolpad-app/next.config.mjs
```diff
@@ -198,6 +198,24 @@ export default /** @type {import('next').NextConfig} */ withSentryConfig(
           source: '/:path*',
           headers: securityHeaders,
         },
+        {
+          source: '/reactDevtools/:path*',
```
Janpot, line 202: This surprises me, aren't there only scripts hosted under this path?

packages/toolpad-app/pages/deploy/[appId]/[[...path]].tsx
Comment (Janpot) [reply packages/toolpad-app/pages/deploy/[appId]/[[...path]].tsx:15]: Yes, this should be fine 👍

packages/toolpad-app/next.config.mjs
oliviertassinari: You are right, it works without, removed, let's keep it simple

packages/toolpad-app/pages/deploy/[appId]/[[...path]].tsx
Comment (Janpot) [reply packages/toolpad-app/pages/deploy/[appId]/[[...path]].tsx:15]: Ah that's a pity, the method of selectively removing the header makes it almost impossible to add optionality to this in the future without completely rewriting it. Another option is to use regex to exclude this header from `/deploy/*` routes:

```rsx
        {
          source: '/((?!deploy/).*)',
          headers: [
            {
              key: 'X-Frame-Options',
              value: 'SAMEORIGIN',
            },
          ],
        },
```

Comment (oliviertassinari) [reply packages/toolpad-app/pages/deploy/[appId]/[[...path]].tsx:15]: @Janpot I didn't understand the "optionality" argument, let me know the approach that you prefer, I will update the PR.

(Personally, I feel that the Next.js headers abstraction doesn't make a lot of sense for non static pages, I might be more intuitive with an express server that would serve the Next.js request handler for some or the pages and set all the headers, request handler per request handler).

Comment (Janpot) [reply packages/toolpad-app/pages/deploy/[appId]/[[...path]].tsx:15]: > I didn't understand the "optionality" argument

Users may not want to allow their apps embedded in iframes. So we could make this header optional. `DENY` by default and provide no header when the user sets the option "Allow embedding in iframes". For that to work, either the header needs to be removed after we fetched the app (meaning we don't see the 404 like you pointed out). Or we need to turn the logic around, where it's not present by default and we enable it conditionally, which is how we built the "public app" feature.

> (Personally, I feel that the Next.js headers abstraction doesn't make a lot of sense for non static pages, I might be more intuitive with an express server that would serve the Next.js request handler for some or the pages, so that we can set all the headers, request handler per request handler).

👍 I agree. With @apedroferreira recently been talking about bringing back the custom server from before. Mainly for different reasons (Next.js edge runtime is just not powerful enough). I don't believe it would solve our problem here. You'd still do something like
```tsx
app.use('*', (req, res, next) => {
  res.setHeader('X-Frame-Options', 'SAMEORIGIN')
  next()
})

app.use('/deploy/*', (req, res, next) => {
  res.removeHeader('X-Frame-Options')
  next()
})

app.use('*', nextHandler)
```
which would have the same effect as setting

```ts
        {
          source: '/((?!deploy/).*)',
          headers: [
            {
              key: 'X-Frame-Options',
              value: 'SAMEORIGIN',
            },
          ],
        },
```
in next config.

Comment (oliviertassinari) [reply packages/toolpad-app/pages/deploy/[appId]/[[...path]].tsx:15]: > Users may not want to allow their apps embedded in iframes. 

Ahh yeah ok. I have missed the part of it's not for everybody to allow iframes. Ok, I'm fixing this with a new commit.

Comment (oliviertassinari) [reply packages/toolpad-app/pages/deploy/[appId]/[[...path]].tsx:15]: I'm going the option that has the safest fail mode: you got the logic wrong, you can't use iframes.

Comment (Janpot) [reply packages/toolpad-app/pages/deploy/[appId]/[[...path]].tsx:15]: Just to be clear. I don't expect this PR to implement it being optional (feel free though 🙂 ). Just that we choose an implementation that has a logical path towards making it so.

Janpot (approved):

PR #1165, opened 2022-10-17
Fix fetch query preview

bytasv:
- Follow up for https://github.com/mui/mui-toolpad/pull/1153#discussion_r996274109

- Kept only logic which checks if untransformed data is undefined
- Rendering `undefined` as a returned value in `JsonView`

image

One thing I'm not sure if it's fine - when api returns empty object, preview just show `Object` which to me seems a bit confusing (I'd expect maybe to see `Object { }`.

image

It seems that this is effect of using custom `nodeRenderer`, if I removed `nodeRenderer`, I get result close to what I expect to see in console:
image

And with empty object:
image

I kept this part as is, lemme know if you think we should change it as well


Janpot:
> It seems that this is effect of using custom nodeRenderer, if I removed nodeRenderer, I get result close to what I expect to see in console:

I believe this was originally done to improve how that global scope explorer looks (It was the first thing we used this JsonView for I think). When just used as an output it doesn't work well.

I think if we redesign a little bit how the global scope explorer works, we can remove this `nodeRenderer`. By redesign I mean: instead of just showing a json view for the whole object, we render each global variable with a `JsonView` on its own.

packages/toolpad-app/src/toolpadDataSources/rest/client.tsx
```diff
@@ -223,35 +212,14 @@ function ResolvedPreview({ preview }: ResolvedPreviewProps): React.ReactElement
 
   const { untransformedData } = preview;
 
-  if (!untransformedData || isEmpty(untransformedData)) {
+  if (untransformedData === undefined) {
     return (
```
Janpot, line 216: Just an idea, but would it make sense to both show the `JsonView` with `undefined` **and** the Alert? Basically we'd always show the exact JSON output, but in some cases we show an extra warning alongside of it.

bytasv: You mean this alert? 👇 (sorry for code screenshot 🙈 )
image

Would you like to display that only when `undefined` is returned or some other conditions as well?

bytasv:
> I think if we redesign a little bit how the global scope explorer works, we can remove this nodeRenderer. By redesign I mean: instead of just showing a json view for the whole object, we render each global variable with a JsonView on its own. But we can do that in follow-up.

Are you talking about this part:
image

The logic would be that we iterate through `globalScope` keys and render each using `JsonView` without `nodeRenderer` defined?

packages/toolpad-app/src/toolpadDataSources/rest/client.tsx
bytasv: IMO all suggestions are easily doable right now. I could make the changes and just render same text that I've used before without auto tab switching. I can also add a link to a tab on `transform` word so user could quickly jump into that if needed

bytasv:
image

Added this warning back, but only if transform returns `undefined`, @Janpot lemme know if I understood your suggestions correctly

Janpot (approved):
👍 looks good

Issue #1163, opened 2022-10-15
Umbrella: feedback building GitHub supports tools


oliviertassinari:
This is a list of pain points that I faced while building for https://www.notion.so/mui-org/Set-GitHub-scope-labels-12a84fdf50e44595afc55343dac00fca. I thought that it might be interesting for the rest of the team to get a perspective on this. I have seen a lot of things that we could improve, but it doesn't mean that we should. I mean for a lot of them we need to consider **when** or even if it's something to fix 😁.

The apps:

1. https://master--toolpad.mui.com/_toolpad/app/cl6rqzry10009arlv9sto6qja/pages/ip23ggo
2. https://master--toolpad.mui.com/_toolpad/app/cl6rqzry10009arlv9sto6qja/pages/e943n9y

The pain points:

1. #1157
2. I have exhausted my GitHub with 5,000 requests/hour in almost no time. I'm not sure how. One of my assumptions is because of this option:

<img width="196" alt="Screenshot 2022-10-16 at 01 13 09" src="https://user-images.githubusercontent.com/3165635/196010870-dab9c401-1084-43e6-bf60-2dc15138812d.png">

I have now disabled it to see if it's the origin of the problem.

3. #385 The GitHub API returns a total number of results that can be e.g. 124 but would only return pages of 30 rows. This can be quite confusing in the UI:

<img width="200" alt="Screenshot 2022-10-16 at 01 14 54" src="https://user-images.githubusercontent.com/3165635/196010900-df1e51c3-10ad-44dc-b73c-db95baf8da4b.png">

4. #827 I'm using the data grid with cells that are links. It would be great if I could click them:

<img width="373" alt="Screenshot 2022-10-16 at 01 15 26" src="https://user-images.githubusercontent.com/3165635/196010912-2cdeaddf-edf1-4adb-a978-c3ae27e65d76.png">

Similar to how it works with Metabase or Notion.

5. #1174 The point 4. made me wonder if it wouldn't be great to support custom column type, where you could create custom components and use them there. For example, to render https://mui-plus.vercel.app/components/Sparkline inside the cells.
6. I have added a quick description below the header. Having #756 would help.
7. The default size of the button feels quite big. I'm missing a size option.
8. #1172 I had two create almost two identical queries. I wish I could have the option to duplicate them.
9. #1175 When you drop a Typography component in a page, it has a default variant of body1 but the UI doesn't reflect it:

[...]

Issue #1093, opened 2022-10-05
Add query cache control


Janpot:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Allow for setting the cache time of a query to control how long it takes before they get stale

### Examples 🌈

basically `reacty-query` `cacheTime` property: https://tanstack.com/query/v4/docs/guides/caching

Issue #385, opened 2022-05-09
Support entities


Janpot:
Recently we implemented support for server-side [pagination](https://mui.com/toolpad/concepts/data-providers/#pagination) through the concept of [data providers](https://mui.com/toolpad/concepts/data-providers). We should extend the capabilities to also include:

1. Filtering #2886
2. Sorting #2539 
3. Row editing #2887
4. Row creation #2888 
5. Row deletion #2889 

Relates to: https://github.com/mui/mui-x/issues/8179

prakhargupta1:
It looks essential but I am unable to visualize. Can you help me understand how it may function? Also, is there a similar concept in other low-code tools?

JerryWu1234:
@Janpot  these problems is very suitable for me.
https://github.com/mui/mui-toolpad/issues/2539
such as this one.
Could you offer more details?

Janpot:
@JerryWu1234 I've updated the issue and turned it into an umbrella for improvements on [Toolpad data providers](https://mui.com/toolpad/concepts/data-providers/). Did you get a chance already to try them out? The [documentation page](https://mui.com/toolpad/concepts/data-providers/) should get you right on track. We would like to closely align this feature with [MUI X](https://github.com/mui/mui-x/issues/8179) in the near future and add filtering/sorting/... capabilities.

JerryWu1234:
>Did you get a chance already to try them out?

I didn't, but I developed the pagination function and filtering before.

PR #1167, opened 2022-10-17
Add a staleTime to queries

Janpot:
Adding a `staleTime` of 10m to `react-query` to tame the amount of API calls a bit until we figure out how we should allow users to finetune this behavior.

Also adding a way to enable the `react-query` devtools in the application.

This also takes care of invalidating cache during editing. The method is straightforward. Alongside the dom we supply a map of hashes of the dom nodes that we have confirmed being saved on the server. We use this hash in the query key to invalidate its cache when serverside node changes. During deployed runtime we don't need to supply these hashes.


packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -890,6 +898,7 @@ const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
       retry: false,
+      staleTime: 60000,
```
oliviertassinari, line 901: It doesn't match the PR's description

packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -872,6 +878,7 @@ const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
       retry: false,
+      staleTime: 10 * 60 * 1000,
```
oliviertassinari, line 881: I think that we would be OK with:

```suggestion
      staleTime: 1000,
```

From what I understand, it would mean that we never query more than once every second, so 3600 requests per hour, below the 5000 limits of GitHub. It's not pretty, but I wouldn't expect anyone to want to refresh more than once per second. But for example in https://master--toolpad.mui.com/_toolpad/app/cl6rqzry10009arlv9sto6qja/pages/ip23ggo I would definitely want to use the refresh button after I label a few issues.

Janpot: > it would mean that we never query more than once every second

But it's not an unrealistic scenario to have more than one github queries on one page. Perhaps we can make it slightly longer, say 10s? I'm also planning to make this configurable at some point (https://github.com/mui/mui-toolpad/issues/1093).

oliviertassinari: From my perspective, when using a refresh button, we should feel like the results returned are up to date, **reliably**. I doubt that a value > 1s can allow this UX. A second concern is that if we set 10s, then we might get blind to the problem. What I think would be great is to first solve the over-fetching, and then consider caching, but to solve slow queries, not overfetching.

Janpot: If I understand correctly, when you call `refetch` it should ignore cache and always refetch. So the button should always work, regardless of how long we make the `staleTime`.

oliviertassinari: Oh yeah, this would be even better

Janpot: It should be already working this way though I'm not sure I explicitly tried this.

Janpot: @oliviertassinari I took care of the cache busting in the editor. Do you still think we should shorten the `staleTime` when we have the `refetch` function?

oliviertassinari: @Janpot Per https://github.com/mui/mui-toolpad/issues/1163#issuecomment-1285114257, I think that 1000 is enough, sorry, I might have put you on the wrong track.

Janpot: > e.g. when editing records

The current architecture doesn't help with that neither. You'd still have to call `refetch` after updating the row to make the table up to date. There is no way to do it automatic as the list/delete/update methods are not linked in any way. For that we'd need something like a collection that we can link as a whole to a datagrid (https://github.com/mui/mui-toolpad/issues/385)

> But to be honest, I don't fully understand what this property does so I might not be the best one to answer.

`staleTime` sets the time after which data is considered out of date. That doesn't mean `react-query` will automatically refetch it after this time. It just means that after this time has passed, when certain triggers fire (e.g. window focus) `react-query` will refetch the query in the background.

oliviertassinari: @Janpot Ok, then yes `10s` sounds great. It's large enough to have an impact and small enough to not completely go wrong.

Janpot: Adding feature request that could help with invalidation: https://github.com/mui/mui-toolpad/issues/1184

oliviertassinari (commented):
Looks good

Issue #1036, opened 2022-09-20
Allow duplicating components


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Stemmed from #437

Use case:
I have done data binding, styling in a component and want to reuse it multiple times in the app. I should be able to duplicate it.

### Benchmarks

- https://www.appsmith.com/blog/may-round-up-app-theming-new-copy-paste-experience-and-product-updates#block-4 

  <img width="726" alt="Screenshot 2022-10-19 at 00 24 05" src="https://user-images.githubusercontent.com/3165635/196556469-d7cc911c-fad3-49ab-8532-92300880e34c.png">

PR #1169, opened 2022-10-17
Implement component duplication logic

bytasv:
- Closes https://github.com/mui/mui-toolpad/issues/1036

- Duplication happens inline, maybe later we would like to improve the logic where the element is placed
- I've also added tooltips with higher than default enter delay

image


packages/toolpad-app/src/appDom/index.ts
```diff
@@ -727,6 +727,16 @@ export function moveNode<Parent extends AppDomNode, Child extends AppDomNode>(
...
 }
 
+export function duplicateNode<Child extends ElementNode>(dom: AppDom, node: Child) {
+  if (!node.parentId || !node.parentProp || !node.parentIndex) {
+    throw new Error(`Node: "${node.id}" can't be duplicated`);
+  }
+
+  const newNode = createElement(dom, node.attributes.component!.value, node.props, node.layout);
+
+  return setNodeParent(dom, newNode, node.parentId, node.parentProp, node.parentIndex);
```
apedroferreira, line 737: The parentIndex needs to be different. 
Parent indexes identify the ordered position of a component within a container, so because we want to add the new component **after** the old one, we need to make the parent index be between the old component and the component next to it - or if there is no component next to it, get the last parent index.
Probably the function `getNewParentIndexAfterNode` is all we need to use here (it does all that for us).

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/NodeHud.tsx
```diff
@@ -180,8 +183,20 @@ export default function NodeHud({
           >
             {component?.displayName || '<unknown>'}
             <DragIndicatorIcon color="inherit" />
+            <IconButton
+              aria-label="Duplicate element"
+              color="inherit"
+              onClick={onDuplicate}
+              onMouseDown={stopPropagationHandler}
```
apedroferreira, line 190: Probably no need for this `onMouseDown`? Unless you found some bug with this.
Also I see the delete button is using `onMouseUp` instead of `onClick` - that might be to avoid certain conflicts with the drag & drop / resizing, so not sure if could just use `onMouseUp`?

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/NodeHud.tsx
```diff
@@ -180,8 +183,20 @@ export default function NodeHud({
...
+            >
+              <Tooltip title="Duplicate element" enterDelay={400}>
+                <ContentCopy color="inherit" />
+              </Tooltip>
+            </IconButton>
             <IconButton aria-label="Remove element" color="inherit" onMouseUp={onDelete}>
-              <DeleteIcon color="inherit" />
+              <Tooltip title="Remove element" enterDelay={400}>
```
apedroferreira, line 197: Tooltips look good!

apedroferreira (changes requested):
Looks good in general, thanks!
Just a few comments.

packages/toolpad-app/src/appDom/index.ts
Janpot: I believe we can use the `addNode` function to do that for us?

packages/toolpad-app/src/appDom/index.ts
```diff
@@ -727,6 +727,16 @@ export function moveNode<Parent extends AppDomNode, Child extends AppDomNode>(
   return setNodeParent(dom, node, parent.id, parentProp, parentIndex);
 }
 
+export function duplicateNode<Child extends ElementNode>(dom: AppDom, node: Child) {
```
Janpot, line 730: Should this function also duplicate the children?

bytasv: Good catch! Thanks I didn't think about children, gonna adjust the code. I'm wondering is there any other component (or other situation) besides `Paper` that can have children?

packages/toolpad-app/src/appDom/index.ts
bytasv: I've just tried this and I'm getting TS error which complains that the prop I'm passing doesn't belong neither to child nor to parent:
image

Not really sure how to fix this, any tips?

packages/toolpad-app/src/appDom/index.ts
```diff
@@ -727,14 +727,34 @@ export function moveNode<Parent extends AppDomNode, Child extends AppDomNode>(
...
   }
 
+  const { children } = getChildNodes(dom, node);
+
   const newNode = createElement(dom, node.attributes.component!.value, node.props, node.layout);
 
-  return setNodeParent(dom, newNode, node.parentId, node.parentProp, node.parentIndex);
+  let updatedDom = dom;
+
+  if (Array.isArray(children)) {
```
apedroferreira: Which values could `children` have? If it could be an array or null/undefined maybe we could just check for if the value is truthy or falsy?

packages/toolpad-app/src/appDom/index.ts
```diff
@@ -727,14 +727,34 @@ export function moveNode<Parent extends AppDomNode, Child extends AppDomNode>(
...
+    children.forEach((childNode) => {
+      updatedDom = duplicateNode(updatedDom, childNode, newNode);
+    });
+  }
+
+  if (parent) {
```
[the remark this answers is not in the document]
apedroferreira: The naming might be confusing - parent index is actually a property of the child. It defines the position of that child inside its parent. So you could probably just use `addNode` instead of `seNodeParent` like Jan suggested.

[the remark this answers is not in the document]
bytasv: > Why do we have this check that still uses the same parent index?

Because when I duplicate children `parentIndex` remains the same, but `parent` node is a newly created node which is passed only when we duplicate children nodes

> I've noticed that if I duplicate an element inside a column for example, it's still being placed before/above the copied element, not after it, because it has the same parent index.

Can you clarify what you mean, because when I duplicate element it's placed after as can be seen in the recording attached. Maybe you are expecting that duplicated node would be selected afterwards? 🤔 Though technically I don't see a big difference as in this current flow I'd expect you'd still have to move new node into correct place

https://user-images.githubusercontent.com/437214/196437263-836b8a9c-f539-44d2-9fbb-8f8402571403.mov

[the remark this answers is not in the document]
apedroferreira: I see - but even duplicated components without children can have a parent, right? And for those we would want to set a new parent index...
I guess we might need to pass in a `parentIndex` to the `duplicateNode` function in order to be able to set if it should be a new index or not, according to each scenario? Unless you can think of a better solution, but this one sounds good to me at first glance.

About the video - I think that if a new component is added with the same parent index as another, if the components are inside a row it shows up to the right of the previous component, but inside a column it shows up above it.

https://user-images.githubusercontent.com/10789765/196443341-9a75837c-13d4-43d8-a5f2-000482961ebb.mov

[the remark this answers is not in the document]
apedroferreira: > Can you clarify what you mean, because when I duplicate element it's placed after as can be seen in the recording attached. Maybe you are expecting that duplicated node would be selected afterwards? 🤔 Though technically I don't see a big difference as in this current flow I'd expect you'd still have to move new node into correct place

I see what you mean here now too, yeah i thought the selected component was the new one.
Anyway I think this behavior is arbitrary even if it works visually... We should probably set a different parent index where possible.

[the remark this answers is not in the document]
bytasv: > (maybe it should be?)

I'm not really sure, I think it doesn't make much difference since it's identical 🤔 

> I guess we might need to pass in a parentIndex to the duplicateNode function in order to be able to set if it should be a new index or not, according to each scenario?

Well the way I understand how currently code works, we don't need to change `parentIndex` when we just duplicate children of that node, I mean those duplicated children has to have same index as the original, that's why it's using same.

> I think that if a new component is added with the same parent index as another, if the components are inside a row it shows up to the left of the previous component, but inside a column it shows up above it. Anyway this behavior is not intentional and we should not have components with the same parent index inside the same parent.

When a component (not children) is duplicated, it's `parentIndex` is set to new, so they aren't the same. As I explained above, only when we are duplicating children their `parentIndex` remains the same, because we aren't changing their order.

I may not understand if there is an issue right now with this code, it would be helpful if you could maybe then pull the local version to point exactly what might be a problem. As of now I think it's all working fine 🤔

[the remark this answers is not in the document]
apedroferreira: Actually it might be fine as it is, I guess I was just expecting the new component to be the selected one.

apedroferreira (approved):

PR #1180, opened 2022-10-19
Allow typing event arguments for autocomplete

Janpot:
Context: https://github.com/mui/mui-toolpad/pull/1105 is blocked on supplying autocomplete to arguments passed to event handlers

This PR adds an option `arguments` to `createComponent` argument type of `'event'` that allows specifying a TS interface, to be used for autocomplete. A special type `ThisComponent` will be available that allows you to use live values of other props. See https://github.com/mui/mui-toolpad/pull/1180/files#diff-b355b8b4105dc88c343f9bef59352a13b361f2f4152c6b77d58bbb6815361f3eR384 for an example

Additionally I had to tweak the global scope explorer to allow for displaying additional global variables. Currently they don't have any extra information on their content.

In order for that to work I had to change a few things and it now looks like:

Screenshot 2022-10-19 at 14 24 21

I propose we avoid expanding the scope of this PR and defer further refinement to https://github.com/mui/mui-toolpad/issues/497


Janpot:
@bytasv Was running into limitations with global scope explorer. My fix also removes the `nodeRenderer` (https://github.com/mui/mui-toolpad/pull/1165#issuecomment-1281034362)

bytasv (approved):
Looks good ✅ 

Few minor suggestions to consider:
1. with "virtual scroll" initially UI looks a bit "broken" - maybe it would be worth to consider something that could hint this is scrollable, i.e. using some scroll component that always shows "virtual type" scrollbars. Alternatively maybe add a shadow on the right side which could be hidden upon first scroll
image

2. I'm wondering if we need all that space for expression editor, maybe we could make scope explorer at least a bit wider (I tried setting it to 250px) and also reduce font size a bit so that more info fits into the same area (I tried with 12px)
image

3. Maybe we could also render object's content **below** it's name, so we would get even more space upon expanding it (might not need to make width of the container itself bigger)

image

Janpot:
1. I added borders which seems to be more the aesthetic of  the theme
2. made it 250, but we should probably just make this a resizable split pane
3. we're a bit constrained by what's possible with `react-inspector`. I've added the object inspector under the name for now. We may be able to do something a bit better if they fix https://github.com/storybookjs/react-inspector/issues/163. But not holding my breath for that 🙂

bharatkashyap (approved):

PR #1185, opened 2022-10-20
Add room for post transformations

oliviertassinari:
I was looking for a way to apply post transformations to my query, I got a bit confused about where this would happen. Once I remembered about how fetch queries worked, I could figure it out.

```jsx
    }
  }
}
      `,
    }),
  });

  if (!response.ok) {
    throw new Error(`HTTP ${response.status}: ${response.statusText}`);
  }
  const result = await response.json();
  return result.data.repository.issues.nodes.map((issue) => ({
    ...issue,
    timelineItems: issue.timelineItems.nodes.map((event) => ({
      label: event.label.name,
      actor: event.actor.login,
    })),
  }));
}
```

https://master--toolpad.mui.com/_toolpad/app/cl6rqzry10009arlv9sto6qja/pages/gt03n67


Janpot:
Do you mean response transformation for fetch queries? In that case we probably need to make it more clear what this does, maybe add some extra docs:

Screenshot 2022-10-20 at 15 55 17

Or do you mean to specifically add another editor to the function datasource to transform whatever the function returns? The latter seems to be a strange request, why not just do it in the function datasource code?

oliviertassinari:
> why not just do it in the function code?

It's what I'm trying to improve with this change.

Janpot (approved):
> It's what I'm trying to improve with this change.

🤦 omg, nevermind, this as a PR not an issue, ignore my answer, this is good 👍

oliviertassinari:
I'm not the only one https://docs.airplane.dev/views/calling-other-apis#javascript-task 

Screenshot 2022-10-20 at 20 06 37

😁

Issue #1164, opened 2022-10-16
Deploy without using the mouse


oliviertassinari:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

<kbd>⌘ Command</kbd>+<kbd>Enter</kbd> doesn't submits the forms

### Expected behavior 🤔

<kbd>⌘ Command</kbd>+<kbd>Enter</kbd> on macOS submits the forms. <kbd>Ctrl</kbd>+<kbd>Enter</kbd> on Windows.

### Steps to reproduce 🕹

1. Reach this screen

<img width="680" alt="Screenshot 2022-10-16 at 21 08 30" src="https://user-images.githubusercontent.com/3165635/196053386-e1c5b2b6-969c-4c50-9402-02c7e45bd563.png">

2. Try to deploy without using the mouse.

### Context 🔦

I believe this UX pattern is now everywhere, for example, this submitted this issue using the above shortcut. https://davidwalsh.name/command-enter-submit-forms. I have stopped counting the number of times I tried to deploy a new Toolpad version using the shortcut, so this time I open an issue 😁.

A proposed implementation:

```jsx
import * as React from "react";
import ReactDOM from "react-dom";

function handleSubmit(event) {
  console.log("submit");
}

function App() {
  return (
    <form
      onSubmit={(event) => {
        event.preventDefault();
        handleSubmit(event);
      }}
      onKeyDown={(event) => {
        if (
          event.key === "Enter" &&
          event.target.type === "textarea" &&
          !event.shiftKey &&
          (event.ctrlKey || event.metaKey)
        ) {
          event.stopPropagation();
          handleSubmit(event);
        }
      }}
    >
      <textarea></textarea>
      <button>Submit</button>
    </form>
  );
}

ReactDOM.render(<App />, document.querySelector("#app"));

```

https://codesandbox.io/s/magical-tree-9vkkgy?file=/index.js

### Your environment 🌎

v0.0.22

PR #1188, opened 2022-10-20
Deploy with keyboard enter

bytasv:
- Closes https://github.com/mui/mui-toolpad/issues/1164

https://user-images.githubusercontent.com/437214/196986160-a863dc7b-3e63-4ccf-8ee3-2a431cb88574.mov


packages/toolpad-app/src/toolpad/AppEditor/AppEditorShell.tsx
```diff
@@ -77,6 +84,7 @@ function CreateReleaseDialog({ appId, open, onClose }: CreateReleaseDialogProps)
                 multiline
                 rows={5}
                 {...register('description')}
+                onKeyDown={handleKeyDown}
```
Janpot, line 87: Would it make sense to capture the event higher up in the tree?

oliviertassinari:
> you have added condition to not trigger event when shift is down, I'm not sure why exactly that is needed though

@bytasv I have copied the logic from the data grid. I don't know why it's here. No objections to remove it.

packages/toolpad-app/src/toolpad/AppEditor/AppEditorShell.tsx
bytasv: I guess if we had multiple input fields, maybe, otherwise I'm not sure, would it? 🤔

Janpot: >  guess if we had multiple input fields

We don't have now, but we do have buttons that can be focused. 

> would it?

If the next contributor comes in and adds a checkbox, will the Cmd+Enter work out of the box or will they have to rebuild the feature from this PR?

Janpot:
> I'm not sure why exactly that is needed though

So that when someone somewhere sometime in the future defines the Cmd+Shift+Enter shortcut, this one doesn't accidentally fire?

oliviertassinari:
@Janpot It came from here precisely https://github.com/mui/mui-x/pull/5744/files#diff-f5bd889280d2db3f858e79916957e7f360ba3392b12b0c2902f9b5b4b9a84d3bR62 for this demo: https://mui.com/x/react-data-grid/recipes-editing/#multiline-editing. I had a look at the rest of the codebase, they usually don't ignore the shift key.

@cherniavskii What was the intent with the shift key, e.g. https://github.com/mui/mui-x/pull/5744#discussion_r944358320.

packages/toolpad-app/src/toolpad/AppEditor/AppEditorShell.tsx
bytasv: Ah, I think I never tried to submit a form that way while being focused on checkbox/radio, good point, I will make an adjustment 👍

bytasv:
> So that when someone somewhere sometime in the future defines the Cmd+Shift+Enter shortcut, this one doesn't accidentally fire?

In that case we should guard that **no other** key is pressed, because there could also be other shortcuts in the future, that we might not be aware of right now, cmd+option+enter, cmd+tab+enter, etc.. (whatever they would do)

packages/toolpad-app/src/toolpad/AppEditor/AppEditorShell.tsx
Janpot: And by extension, there are several dialog forms in the application. Is there a way we can make this uniform everywhere, now and for future dialog forms?

bytasv: Sure, I think I can adjust `DialogForm` component for that

bytasv:
@Janpot please take a look. I moved key handler to the form itself, as well as tweaked `DialogForm` component so we could get this functionality everywhere where `DialogForm` is used.

packages/toolpad-app/src/components/DialogForm.tsx
```diff
@@ -1,7 +1,36 @@
+import React from 'react';
```
oliviertassinari, line 1: The correct import path

```suggestion
import * as React from 'react';
```

packages/toolpad-app/src/components/DialogForm.tsx
```diff
@@ -1,7 +1,36 @@
...
   flexDirection: 'column',
 });
+
+export interface DialogFormProps extends React.FormHTMLAttributes<HTMLFormElement> {
+  children: React.ReactNode;
+  onKeyDown?: (event: React.KeyboardEvent) => void;
+}
+
+export default ({ children, onKeyDown = () => {}, ...rest }: DialogFormProps) => {
+  const handleKeyDown = React.useCallback((event: React.KeyboardEvent) => {
```
oliviertassinari, line 16: I think that it would be better without the useCallback. There isn't performance concerns here form what I understand.

packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/CreateCodeComponentNodeDialog.tsx
```diff
@@ -73,6 +73,10 @@ export default function CreateCodeComponentDialog({
       <DialogForm
         autoComplete="off"
         onSubmit={(e) => {
+          if (!name || isInvalid) {
+            return;
+          }
+
           e.preventDefault();
```
oliviertassinari, line 80: **Off-topic**

No shorthands please
```suggestion
          event.preventDefault();
```

packages/toolpad-app/src/components/DialogForm.tsx
bytasv: I'm using that as per Jan's suggestion in some previous PRs

packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/CreateCodeComponentNodeDialog.tsx
bytasv: I also don't prefer shorthands, but this should be a convention agreed upon with entire team as this seems to be the pattern through the code base.
If we have an agreement to change it, then IMO we should do it across entire codebase

cherniavskii:
@oliviertassinari 
> What was the intent with the Shift key, e.g. https://github.com/mui/mui-x/pull/5744#discussion_r944358320?

The intention was to for `Enter` and `Shift + Enter` to add new line and only exit editing on `Escape` or `Ctrl + Enter`

packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/CreateCodeComponentNodeDialog.tsx
Janpot: > If I understand correctly that's an unwritten convention that we should guard against during PR reviews as well?

Sure, feel free to point it out or fix when you notice it. But I wouldn't make this a bigger thing than it needs to be. IMO It's part of the practice of writing readable, defensive, maintainable, self-documenting code and if that was easily codified in simple rules then there wouldn't be any bad code out there.

bytasv: If we are talking about codified rules, maybe we could give a try with https://eslint.org/docs/latest/rules/id-length 🤔

Janpot: I feel like there are too many exceptions and too little benefit to make this rule worthwhile. e.g. 

loop variables:
```tsx
for (let i = 0; i < rows; i += 1) {
   for (let j = 0; j < columns; j += 1) {
```

math functions:
```tsx
const add = (a, b) => a + b;
```

coordinates:
```tsx
mouse.click(x, y);
```
...

packages/toolpad-app/src/components/DialogForm.tsx
```diff
@@ -1,7 +1,39 @@
...
   display: 'flex',
   flexDirection: 'column',
 });
+
+export interface DialogFormProps extends React.FormHTMLAttributes<HTMLFormElement> {
+  children: React.ReactNode;
+  onKeyDown?: (event: React.KeyboardEvent) => void;
+}
+
+export default ({ children, onKeyDown = () => {}, ...rest }: DialogFormProps) => {
```
Janpot, line 15: [Upcoming eslint changes](https://github.com/mui/material-ui/pull/34642) will mandate function declarations. We can already adhere to this style

```suggestion
export default function ({ children, onKeyDown = () => {}, ...rest }: DialogFormProps) {
```

packages/toolpad-app/src/components/DialogForm.tsx
```diff
@@ -1,7 +1,39 @@
...
+    (event: React.KeyboardEvent) => {
+      const modifierKey = (event.metaKey || event.ctrlKey) && !event.shiftKey;
+
+      if (event.key === 'Enter' && modifierKey) {
+        const submitEvent = new Event('submit', {
+          cancelable: true,
+          bubbles: true,
+        });
+
+        event.currentTarget.dispatchEvent(submitEvent);
```
Janpot, line 26: Great idea. One drawback I see is that it will submit the form, even when the submit button is disabled (I noticed you have to add several guards across the form to account for this). Would it be possible to instead of dispatching an event on the form, query the DOM for a submit button and fire a click event? e.g.

```tsx
ref.current.querySelector('[type=submit]').click()
```

Not 100% sure this would work.

bytasv: I don't believe this is a right approach, IMO submit logic shouldn't be guarded by the submit button, instead we should always guard logic with the right condition inside the handler. While the above should work, I don't think that it should be implemented for the reason mentioned above. It is also possible to have forms without submit button entirely which could be "remotely" submitted, in that case it wouldn't work at all

Janpot: > I don't believe this is a right approach, IMO submit logic shouldn't be guarded by the submit button, instead we should always guard logic with the right condition inside the handler.

From a UX point of view, the Cmd+Enter is a shortcut for the submit button. So it shouldn't submit the form when the button is disabled. Emulating a click on the submit button is the closest I can think of to bening a shortcut for that button.
If you feel strongly about guarding the logic inside the handlers, then at least use an `invariant`. The handler should never be called when the submit button is disabled, so when it is called I want to see an error because that signifies a there has been a programming mistake. And we want these to surface, not being ignored.

> It is also possible to have forms without submit button entirely which could be "remotely" submitted, in that case it wouldn't work at all

I think we can safely say that for a good UX, all dialogs that have a form will have a submit button. 

I don't think I'm familiar with "remotely submitting forms", can you point me to some documentation about this feature?

packages/toolpad-app/src/components/DialogForm.tsx
Janpot: > There is no performance concerns here from what I understand.

You often need implementation details to judge whether it'll be a performance concern or not. Always passing stable values errs on the side of caution. In this case the implementation of the component is very local to the consumer and we can see there are no performance concerns. I'm fine with doing it either way in this specific case. Especially if there's a chance the component will be extracted into its own file at some point it could be interesting to be defensive about this.

packages/toolpad-app/src/components/DialogForm.tsx
bytasv: > From a UX point of view, the Cmd+Enter is a shortcut for the submit button

Thanks, you are right. I checked how the form behaves with a regular `<input />` when enter is pressed and it does properly regard submit button state. I've changed implementation as per your suggestion and it seems to work as expected now 👍 

> If you feel strongly about guarding the logic inside the handlers, then at least use an invariant

I've changed to usage of `invariant`. It's not that I feel strongly about current behaviour, I just believe that validation should happen inside the handler itself.

> I think we can safely say that for a good UX, all dialogs that have a form will have a submit button.

yeh, that's absolutely correct

> I don't think I'm familiar with "remotely submitting forms", can you point me to some documentation about this feature?

As per previous comment, this won't probably be applicable to our `Dialogs`, but an example to illustrate what I meant could be something like this - you have a form fields rendered in the main content area, but submit button is "taken" out of form and placed in a navigation bar or some other dom element which doesn't have access to trigger submit event. Upon pressing submit you would programatically trigger submit on the form, kicking in whatever the submit logic is defined. I don't know actually if "remote form submit" is correct term though, it might be misleading.

---

While we are on this subject I'd just like to share a couple of resources which explain why using disabled form buttons in general isn't good idea: 
- [one](https://www.smashingmagazine.com/2021/08/frustrating-design-patterns-disabled-buttons/)
- [two](https://medium.com/@tu.bui/dont-use-disabled-buttons-7541d0586ca1)
- [three](https://uxdesign.cc/why-heuristics-are-only-rules-of-thumb-the-case-of-the-disabled-button-4824958627e9) 
- [four](https://uxplanet.org/disabled-buttons-in-user-interface-4dafda3e6fe7)

tl;dr; - it is considered better to allow users to make mistake and show error explaining what's wrong, rather than preventing them to make the mistake without understanding why they can't proceed

IMO in our case a good example could be a component creation form - when I remove the name there is nothing in UI indicating why I can't proceed. So unless I can implicitly understand that the name is required, I'd be stuck
image

Anyways, this is probably a topic for a different occasion

packages/toolpad-app/src/components/DialogForm.tsx
bytasv: I don't think that implementation details matter, IMO the only thing that matters in regards to performance optimisation is measuring and being sure that it is a problem and applying necessary optimisations.

> In this case the implementation of the component is very local to the consumer and we can see there are no performance concerns.

I don't think we can SEE, unless we measure. I'd argue that this is true for majority of the cases where we use this (or useMemo) optimisation. IMO unless we measure the performance it is really subjective to claim whether it's needed or not. I'd personally prefer if it was either used everywhere blindly or only in the places that are truly problematic and needs optimising, otherwise it can just lead to unnecessary speculation and discussions whether THIS IS the place where we need to use optimisation or not.

I'd add that in general I wouldn't be against being "defensive" about this, but react optimisations do make code look more cluttered/ugly than it could be, so I'd prefer to avoid that.

packages/toolpad-app/src/components/DialogForm.tsx
Janpot: > Anyways, this is probably a topic for a different occasion

I agree, it's a debate we should probably do outside of this ticket. A few notes:

1. Incidentally, I just fixed those node creation dialogs this morning, they now give proper feedback:
    
    Screenshot 2022-10-24 at 13 13 46

2. These blog posts touch on interesting points, but I don't think they're proposing the right UX neither. First of all, they present a false dichotomy, either have a disabled button and no validation on the inputs, or have a disabled button with validation on the inputs. But you can have validation as well as a disabled submit button. Second, my expectation when clicking a submit button is that "my work is done". It's very frustrating to click submit only to find I'm going to have to spend more time writing inputs. It's not that I don't want to spend the time, it's that I expect to be done when I click submit and can go on with the day, the button doesn't do what I expect it to do. For inputs that are offscreen, it's perfectly possible to accompany the submit button with a message, with a link that scrolls to the input.

bytasv: > But you can have validation as well as a disabled submit button

You can, but unless you interact with a field - validation message won't appear where with enabled submit button you provide a way to continue and "fail".

Anyways, a topic for another day..

oliviertassinari: From a UX perspective, I would 👍 on **a default** of making the submit button clickable, even when the form is invalid to be able to raise a clear error message.

However, I think that it also makes sense to disable the button if there is a strong use case for it. For example with very simple forms. This GitHub comment texterea I'm commenting on and submitting with <kbd>Ctrl</kbd>+<kbd>Enter</kbd>:

Screenshot 2022-10-25 at 19 12 53

PR #1191, opened 2022-10-20
Use public environment settings for demo

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/1056

Created a separate env group in Render for public applications, without `TOOLPAD_BASIC_AUTH_USER` and `TOOLPAD_BASIC_AUTH_PASSWORD`, and using that env group in demo.

Let me know if you think we should use a different solution to exclude basic auth from the demo.


apedroferreira:
> Would it make sense to remove basic auth from `toolpad-settings` and create a `toolpad-basic-auth` group that we add to the main deployment?

Yeah that would work too, it's just more of a breaking change but should be fine and has less redundancy.

Issue #845, opened 2022-08-23
Reduce every DOM operation to the minimum possible DOM updates


apedroferreira:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Right now there are multiple operations in the drag & drop editor that cause multiple updates in the DOM, when one or zero updates would be expected instead. We should make it so that those operations only update the DOM as little as necessary.

See also this issue for more relevant discussion and information - however, that issue is only about hiding the number of changes in the UI: https://github.com/mui/mui-toolpad/issues/781

### Examples 🌈

- Dragging an element to the position it's already in should show no changes, but it adds 1 change to the DOM
- Dragging in a component as shown in https://github.com/mui/mui-toolpad/issues/781

### Motivation 🔦

Fixing this will enable us to build undo/redo functionality a lot more easily in the future.

PR #1192, opened 2022-10-21
atomic domApi updates

Janpot:
POC

ref https://github.com/mui/mui-toolpad/issues/845

### DIFF
--- a/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/RenderOverlay.tsx
+++ b/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/RenderOverlay.tsx
@@ -864,7 +864,9 @@ export default function RenderOverlay({ canvasHostRef }: RenderOverlayProps) {
         originalParent && appDom.isElement(originalParent) ? isPageColumn(originalParent) : false;
 
       let addOrMoveNode = domApi.addNode;
+      let appDomAddOrMoveNode = appDom.addNode;
       if (selection) {
+        appDomAddOrMoveNode = appDom.moveNode;
         addOrMoveNode = domApi.moveNode;
       }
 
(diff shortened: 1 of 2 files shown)


packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/RenderOverlay.tsx
```diff
@@ -875,15 +877,29 @@ export default function RenderOverlay({ canvasHostRef }: RenderOverlayProps) {
             ? appDom.getNewFirstParentIndexInNode(dom, dragOverNode, 'children')
             : appDom.getNewLastParentIndexInNode(dom, dragOverNode, 'children');
 
-        if (!isPageRow(draggedNode)) {
-          const rowContainer = appDom.createElement(dom, PAGE_ROW_COMPONENT_ID, {});
-          domApi.addNode(rowContainer, dragOverNode, 'children', newParentIndex);
-          parent = rowContainer;
+        domApi.update((draft) => {
```
apedroferreira, line 880: Looks awesome, so we could use this kind of logic to batch changes under the same update without applying each one separately? Might also be a good way to abstract some of this logic, make it more readable and split it in different logical parts, maybe by moving these functions to a separate file?

apedroferreira: I'll probably go from this once we work on undo/redo, thanks for suggesting it!

Janpot: Also meshes quite well with `immer` if we start using it

Janpot: Another important advantage is that it takes away an annoying limitation of `domApi`, it's that it doesn't update the local `dom` variable until after the next render:

```tsx
const dom = useDom()
const domApi = useDomApi()
const doSomething = (someNode) => {
  domApi.addNode(someNode)
  // dom is not up to date so this returns false:
  if (!!appDom.getMaybeNode(dom, node.id)) {

  }
}

// versus

const dom = useDom()
const domApi = useDomApi()
const doSomething = (someNode) => {
  domApi.update(draft => {
    draft = appDom.addNode(draft, someNode)
    // draft is up to date so this returns true:
    if (!!appDom.getMaybeNode(draft, node.id)) {
      
    }
  })
}
```

apedroferreira: Okay, might require some changes in the drag & drop logic as in some places I've been assuming that the DOM is not updated yet, but anyway this should make that logic much more intuitive.

PR #1193, opened 2022-10-21
[ui] Add button size property

Janpot:


apedroferreira (approved):

apedroferreira:
Oh, just one note, I think the button starts as "medium" but has no default value, maybe the default value should be explicitly "medium"?

Janpot:
I was hoping that by keeping it undefined we could delegate it to the theme (once we make that support button sizes)

apedroferreira:
> I was hoping that by keeping it undefined we could delegate it to the theme (once we make that support button sizes)

Yeah that makes sense, would be more flexible so sounds good.

Issue #1126, opened 2022-10-10
Query/mutation name updates before pressing "save"


Janpot:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

When changing the name of a query in the query editor, it is immediately saved, before pressing save.

### Expected behavior 🤔

Editing this name should be part of the save/cancel flow.

### Steps to reproduce 🕹

Steps:

1. open a query editor
2. change the name in the top-left
3. use cancel to close the modal
4. no warning is displayed for unsaved changes. The query name has changed in the dom

PR #1194, opened 2022-10-21
[core] Relax the restrictions on node naming

Janpot:
Working up towards https://github.com/mui/mui-toolpad/issues/1126 this PR changes when unique names are enforced and relaxes the strictness on duplicate names. e.g. after this PR page 1 can have a query name "x" and page 2 as well.

* Nodes must be unique within their siblings, and in the case of pages within the descendants on the page
* Improved the validation of the creation dialogs and add default names for nodes
* Had to rewrite the duplication method so that it uses `addNode` which now is in charge of renaming if there's a duplicate name (the new version should also duplicate all children, not just `children` parent prop)
* Also closes https://github.com/mui/mui-toolpad/issues/1126

### DIFF
--- a/packages/toolpad-app/src/appDom/index.ts
+++ b/packages/toolpad-app/src/appDom/index.ts
@@ -933,29 +933,26 @@ export function duplicateNode(
   dom: AppDom,
   node: AppDomNode,
   parent: AppDomNode | null = getParent(dom, node),
-) {
-  if (!parent || !node.parentId || !node.parentProp || !node.parentIndex) {
+  parentprop: string | null = node.parentProp,
+): AppDom {
+  if (!parent || !parentprop) {
     throw new Error(`Node: "${node.id}" can't be duplicated`);
   }
 
-  const childNodes = getChildNodes(dom, node);
-  const existingNames = getExistingNamesForChildren(dom, parent);
-
   const newNode = createNode(dom, node.type, node);
+  const childNodes = getChildNodes(dom, node);
 
-  let updatedDom = dom;
+  dom = addNode<any, any>(dom, newNode, parent, parentprop);
 
-  children?.forEach((childNode) => {
-    updatedDom = duplicateNode(updatedDom, childNode, newNode);
-  });
-
-  if (parent) {
-    return setNodeParent(updatedDom, newNode, parent.id, node.parentProp, node.parentIndex);
+  for (const [childParentProp, children] of Object.entries(childNodes)) {
+    if (children) {
+      for (const child of children as AppDomNode[]) {
+        dom = duplicateNode(dom, child, newNode, childParentProp);
+      }
+    }
   }
 
-  const newParentIndex = getNewParentIndexAfterNode(updatedDom, node, node.parentProp);
-
-  return setNodeParent(updatedDom, newNode, node.parentId, node.parentProp, newParentIndex);
+  return dom;
 }
 
 const RENDERTREE_NODES = [
--- a/packages/toolpad-app/src/appDom/index.ts
+++ b/packages/toolpad-app/src/appDom/index.ts
@@ -595,7 +595,7 @@ export function getExistingNamesForChildren<Parent extends AppDomNode>(
     return new Set(pageDescendants.map((scopeNode) => scopeNode.name));
   }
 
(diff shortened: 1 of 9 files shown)


packages/toolpad-app/src/appDom/index.ts
```diff
@@ -402,20 +403,18 @@ export function getChildNodes<N extends AppDomNode>(dom: AppDom, parent: N): Nod
 }
 
 export function getParent<N extends AppDomNode>(dom: AppDom, child: N): ParentOf<N> {
+  // Make sure we're using the last version of child in the dom
+  child = getNode(dom, child.id, child.type) as N;
```
apedroferreira, line 407: Should this ideally not be necessary?
Nothing against this extra check but if we plan on the DOM always being up to date maybe we should have a TODO comment to get rid of it at some point.
Or maybe the function arguments should be `childId` and `childType` but that might be too much of a refactor...

packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/validation.ts
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react';
+import * as appDom from '../../../appDom';
+
+export function useNodeNameValidation(name: string, disallowedNames: Set<string>, kind: string) {
```
apedroferreira, line 4: If we use that scenario a lot, maybe we could have an option to already check for and exclude the existing names here?

packages/toolpad-app/src/appDom/index.ts
```diff
@@ -874,32 +945,30 @@ export function getNewParentIndexAfterNode(
...
-export function duplicateNode<Parent extends AppDomNode, Child extends ElementNode>(
+export function duplicateNode(
   dom: AppDom,
-  node: Child,
-  parent?: Parent,
-) {
-  if (!node.parentId || !node.parentProp || !node.parentIndex) {
+  node: AppDomNode,
+  parent: AppDomNode | null = getParent(dom, node),
+  parentprop: string | null = node.parentProp,
```
apedroferreira, line 952: `parentProp`? (camel case :D)

apedroferreira (approved):
Works great as far as I've tested! Just left a few ideas.

packages/toolpad-app/src/appDom/index.ts
Janpot: > Should this ideally not be necessary?

Ideally, yes, but this is caused by the drag & drop code, adding a page row and then the item to the row. In the second call, the row doesn't have a parent yet even though it's added to the dom, the node that was passed to this function is the created one, without a parent.
The easy solution seemed to be to just extract the latest version of the node from the dom. I want to avoid messing to much with the drag&drop code until the paper component updates are in.
Ideally these methods take a `NodeId` instead of an `AppNode` because the latter may be outdated.

packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/validation.ts
Janpot: I did it this way for now because most places this is used in also use the `existingNames` to propose a default value for the node. And I wanted to avoid calculating this list multiple times. 🤔 maybe I should memoize it like the `getChildNodes` function

packages/toolpad-app/src/appDom/index.ts
apedroferreira: Sounds good, yeah I was wondering if that kind of logic in the drag & drop is what this was dealing with.
Ok so with the upcoming changes this kind of thing should improve.

Janpot: > Ok so with the upcoming changes this kind of thing should improve.

Yes, additionally I'd like to switch this back to operating on `NodeId`s to make this problem completely go away.

packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/validation.ts
apedroferreira: Yeah that makes sense, we can keep it as it is in that case, wouldn't make sense to change the hook.

Issue #1132, opened 2022-10-11
Update demo video and docs screenshots


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Reminder task to be done before https://github.com/mui/mui-toolpad/issues/900

2. In the header nav bar, show Try demo and self-host CTAs. Also, directly link the Toolpad GitHub repo

bytasv:
While I can do this task, IMO it would be best if someone else took it as per our previous discussions about generating screenshots for each case individually. I think it's important that someone else besides me experiences same, maybe could come up with an alternative solution to screenshot capturing than the one I have proposed.

Janpot:
I don't think the solution you proposed is bad. It's just unfinished.

bytasv:
> I don't think the merging of new UI must necessarily be blocked by screenshots in the docs

I also think that it might make more sense to update screenshots AFTER the new UI is merged, so that the docs match how our UI looks, what's the reason to update screenshots before UI update?

prakhargupta1:
No issues, we can merge it later as well.

prakhargupta1:
@gerdadesign please use this video to update the existing one. Let me know if you suggest any changes. I have taken the light them for Toolpad and the app.
https://user-images.githubusercontent.com/92228082/197453558-5e9566d4-8c07-4fb8-8f6a-764a99cb1787.mov

@bytasv Can we import existing apps (to cloud) using DOM export? This will help in updating screenshots faster.

bytasv:
@prakhargupta1 we don't really have an import feature 🤔 I'll probably have to recreate those apps then so that anyone could pick it up in the future

Janpot:
@bytasv We have a hidden option `TOOLPAD_ENABLE_CREATE_BY_DOM` that allows you to create an app by importing an existing dom. Currently we use this for integration testing only (and PR previews). We'll need to add validation before we can make this user-facing.

bytasv:
@Janpot can this be used on master app?

gerdadesign:
@prakhargupta1 There are a lot of instances in this video where the mouse is going in one direction and then changes direction to the button that its intending to click. Any way you can do a cleaner recording? That will make the editing a lot smoother.

prakhargupta1:
New video:
https://user-images.githubusercontent.com/92228082/198187732-bd3b32fc-8de0-4ff8-9cc4-fc3b13dba116.mov

[...]

PR #1200, opened 2022-10-24
Part 1: Update overview + quickstart screenshots

bytasv:
- Partial of https://github.com/mui/mui-toolpad/issues/1132


Janpot (approved):
Are these taken with a differently sized browser window? Perhaps it makes sense to track somewhere the browser window size? e.g. `1440px` for Macbook 13" or so

apedroferreira:
> Are these taken with a differently sized browser window? Perhaps it makes sense to track somewhere the browser window size? e.g. `1440px` for Macbook 13" or so

I guess that as long as it's 16:9 resolution and it doesn't look blurry at the sizes we're showing the screenshots at we should make them as small as possible?
Also not sure if we're doing that but if not we could run these PNGs through some optimizer to see if we can reduce their size.

bytasv:
If you are asking because of the `width` dimension that is used for images - the answer is they are all taken on the same sized browser. The problem is that when I grab the screenshot I don't know it's dimensions, so I have to manually inspect and set width for each screenshot. Unless the screenshot is full screen, you have to do it for all to maintain that 2 pixels ratio

PR #1202, opened 2022-10-24
Part 2: Update queries docs

bytasv:
- Partial of https://github.com/mui/mui-toolpad/issues/1132


Janpot (commented):
👍 
Are these generated using playwright?

Janpot (approved):

bytasv:
> Are these generated using playwright?

No, I'm capturing them manually so we can have update docs faster, then I can look into automation

PR #1209, opened 2022-10-25
Add guidelines for PR labels

Janpot:
Follow up of https://github.com/mui/mui-toolpad/pull/849

* Starting a `MAINTAINERS.md` as a maintainers guide. Move the release process docs in there. Add a section about PR labelling. [Formatted text](https://github.com/mui/mui-toolpad/blob/9b740719745ca27236639995edc123159de63760/MAINTAINERS.md#pull-request-labels)
* Update the PR label checker to mandate at least `core`/`docs` and add a link to the maintenance guide

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -22,6 +22,7 @@ _We choose these labels according to the way we divide work in the repository, w
 
 - _no label_: Default.
 - `regression`: For bug fixes of regressions, generally, these PRs are accompanied by a test.
+- `dependencies`: Automatically applied by dependabot on dependency updates.
 
 ## Release process
 
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -1,4 +1,29 @@
-# Release process
+# Maintainers guide
+
+This document is intended for _maintainers_. Anyone with commit access to the Toolpad respository.
+
+## Pull request labels
+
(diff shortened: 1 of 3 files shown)


MAINTAINERS.md
```diff
@@ -1,4 +1,30 @@
...
+This document is intended for _maintainers_. Anyone with commit access to the Toolpad respository.
+
+## Pull request labels
+
+Each pull request must have a label applied. We apply labels in the following dimensions:
+
+### location
+
+- `docs`: For updates to the `./docs` folder.
+- `core`: For updates Toolpad application and its build environment.
```
oliviertassinari, line 12: This is not how core is documented in https://www.notion.so/mui-org/core-c4efa9cabeab49248ff39e583807a2aa. For example, I don't think that https://github.com/mui/mui-toolpad/pull/1218 is about `core` but it directly impacts users, but https://github.com/mui/mui-toolpad/pull/1211 is about `core` because users won't directly care.

MAINTAINERS.md
```diff
@@ -1,4 +1,30 @@
-# Release process
+# Maintainers guide
```
oliviertassinari, line 1: Personally, I think that only `## Release process` would be better in git because the content needs to be synced with the code. For the rest, I think that Notion is easier to keep up to date.

PR #1212, opened 2022-10-25
Disable making applications public in demo mode

apedroferreira:
From https://github.com/mui/mui-toolpad/issues/592.
Make all applications in demo mode work as public applications.


packages/toolpad-app/pages/deploy/[appId]/[[...path]].tsx
```diff
@@ -2,6 +2,7 @@ import type { GetServerSideProps, NextPage } from 'next';
 import * as React from 'react';
 import { asArray } from '../../../src/utils/collections';
 import ToolpadApp, { ToolpadAppProps } from '../../../src/runtime/ToolpadApp';
+import config from '../../../src/server/config';
```
Comment (apedroferreira) [inline packages/toolpad-app/pages/deploy/[appId]/[[...path]].tsx:5]: need to fix this import

packages/toolpad-app/pages/deploy/[appId]/[[...path]].tsx
```diff
@@ -26,7 +28,7 @@ export const getServerSideProps: GetServerSideProps<ToolpadAppProps> = async (co
   //   context.res.setHeader('X-Frame-Options', 'DENY');
   // }
 
-  if (!app.public) {
+  if (!app.public && !config.isDemo) {
```
Comment (Janpot) [inline packages/toolpad-app/pages/deploy/[appId]/[[...path]].tsx:31]:
```suggestion
  if (!app.public) {
```

`checkBasicAuth` below should return `true` if no basic auth is configured. The demo doesn't configure basic auth so this extra condition shouldn't be necessary

packages/toolpad-app/src/server/handleDataRequest.ts
```diff
@@ -53,7 +54,7 @@ export default async (
     return;
   }
 
-  if (!app.public) {
+  if (!app.public && !config.isDemo) {
```
Janpot, line 57: Same as above
```suggestion
  if (!app.public) {
```

packages/toolpad-app/src/toolpad/Home/index.tsx
```diff
@@ -417,22 +417,24 @@ function AppSettingsDialog({ app, open, onClose }: AppSettingsDialogProps) {
...
-                    checked={value}
-                    onChange={(e) => onChange(e.target.checked)}
-                    {...field}
-                  />
-                )}
-              />
-            }
-            label="Make application public"
-          />
+          {!config.isDemo ? (
```
Janpot, line 420: Should we disable the button with a message to avoid creating an empty dialog?

apedroferreira: Good thinking, this shouldn't show anyway for now as we don't show the apps screen in the demo, but I think it's fine to have it just in case.
Screen Shot 2022-10-26 at 20 03 51

Janpot: I think @bharatkashyap  was working on bringing the menu to the app editor as well

Issue #1127, opened 2022-10-10
Add tests for app templates


apedroferreira:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Every app template that can be picked when creating an app should be covered by a test to make sure it works properly.
The tests should perform a basic use of that app's functionality.

### Examples 🌈

Testing app template for fetching dog images:
- pick dog breed in selector
- press button to fetch image
- check that an image was rendered

### Motivation 🔦

Trying to cover the most basic/popular use cases for Toolpad users.

PR #1220, opened 2022-10-26
App templates tests

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/1127

Work in progress - dogs API template throws a lot of console errors which are making the tests fail.


test/integration/appTemplates.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+  const homeModel = new ToolpadHome(page);
+  await homeModel.goto();
+  const app = await homeModel.createApplication({ appTemplateId: 'stats' });
+
+  page.waitForNavigation();
+
+  const runtimeModel = new ToolpadRuntime(page);
+  await runtimeModel.gotoPage(app.id, 'stats');
+
+  await page.locator('h3:has-text("Statistics")').waitFor({ state: 'visible' });
```
bytasv, line 15: when we use `tag:has-text(*)` instead of `text=*` is it because we have multiple occurrences of the same text and we need more specific selector?

bytasv (approved):

test/integration/appTemplates.spec.ts
apedroferreira: Just trying to be more specific and trying to avoid issues with multiple occurrences in general, not just now but in the future.

bytasv: Should you then update other locators that currently use only `text=*`?

apedroferreira: Those are just divs, so it feels less specific...
I'll give them another look though.

Issue #827, opened 2022-08-19
Support more data types in data grid


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Our data grid (in Toolpad) currently supports 5 data types namely:
1. Boolean
2. Date
3. Date time
4. String
5. Number

But Retool supports 20 data types, Appsmith supports 9.

We could add:

- A **Link** column type. Requested in #1163
- A **Custom component** column type. #1174

### Examples 🌈

- ~created_at column [here](https://master--toolpad.mui.com/_toolpad/app/cl1on1bd80041a1pdrrfdt0rq/editor/pages/cl6w1j7ev00033b6gfwvnfl6h) is being read properly in Retool but not in Toolpad~

Janpot:
@prakhargupta1 This feature request probably needs a bit more detail. Like which column types are you missing exactly?

> created_at column [here](https://master--toolpad.mui.com/_toolpad/app/cl1on1bd80041a1pdrrfdt0rq/editor/pages/cl6w1j7ev00033b6gfwvnfl6h) is being read properly in Retool but not in Toolpad

Fixed by https://github.com/mui/mui-toolpad/pull/848

prakhargupta1:
There are fields like action button, URL, image, video, but we can look into it when the need comes.

apedroferreira:
@prakhargupta1 I understand that these columns types are not natively supported by the MUI X Data Grid, correct?

So if we implement these new column types they will be something that Toolpad is adding on top of the MUI X Data Grid, not a feature of the data grid itself.
Just clarifying that as I'm not sure if there would be any interest for the MUI X team in adding these column types natively to the data grid.

prakhargupta1:
Found the same info here: https://mui.com/x/react-data-grid/column-definition/#column-types

@joserodolfofreitas This issue is about supporting more column types like link, image and even a custom component in the data grid. Aren't these natively supported?

apedroferreira:
Just to clarify: I can add all these column types to the grid on Toolpad if they're something we want, it shouldn't be that hard.
The "custom component" column type for example will be something unique to Toolpad, obviously.
As for link and image, I'm just noting that they **could** be features that the Datagrid already has, but if the MUI X team has no plans to add them I can just create them in Toolpad.

joserodolfofreitas:
[...]

PR #1223, opened 2022-10-27
Add new column types

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/827

Add new **Link** and **Image** column types.

Link:

https://user-images.githubusercontent.com/10789765/198359861-fcab9bdb-e66d-46ed-9560-57ba67471dd0.mov

Image:

https://user-images.githubusercontent.com/10789765/198359806-93aad104-e357-4fbc-b10b-429b43fbae6f.mov

:

### DIFF
--- a/packages/toolpad-components/src/DataGrid.tsx
+++ b/packages/toolpad-components/src/DataGrid.tsx
@@ -113,7 +113,7 @@ function inferColumnType(value: unknown): string {
     case 'boolean':
       return valueType;
     case 'string':
-      if (URL_REGEX.test(value as string)) {
+      if (URL_REGEX.test(value)) {
         return 'link';
       }
       return valueType;
--- a/packages/toolpad-components/src/DataGrid.tsx
+++ b/packages/toolpad-components/src/DataGrid.tsx
@@ -157,6 +157,7 @@ const COLUMN_TYPES: Record<string, Omit<ToolpadGridColDef, 'field'>> = {
     valueGetter: dateValueGetter,
   },
   link: {
+    customType: 'link',
     renderCell: ({ value }) => (
       <Link href={value} target="_blank" rel="noopener noreferrer">
         {value}
--- a/packages/toolpad-components/src/DataGrid.tsx
+++ b/packages/toolpad-components/src/DataGrid.tsx
@@ -100,6 +100,7 @@ function SkeletonLoadingOverlay() {
   );
 }
 
+// https://stackoverflow.com/questions/161738/what-is-the-best-regular-expression-to-check-if-a-string-is-a-valid-url
 const URL_REGEX =
   /((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w-_]*)?\??(?:[-+=&;%@.\w_]*)#?(?:[\w]*))?)/;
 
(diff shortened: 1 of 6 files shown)


packages/toolpad-components/src/DataGrid.tsx
```diff
@@ -315,6 +342,11 @@ const DataGridComponent = React.forwardRef(function DataGridComponent(
     [getRowId, columns],
   );
 
+  const hasImageColumns = React.useMemo(
+    () => !!columns.find((column) => column.type === 'image'),
```
bytasv, line 346: if you don't need a value that is found, you could also use [Array.some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) method in addition could slightly shorten the expression:
```suggestion
    () => columns.some(({ type }) => type === 'image'),
```

packages/toolpad-components/src/DataGrid.tsx
```diff
@@ -108,7 +114,14 @@ function inferColumnType(value: unknown): string {
   switch (typeof value) {
     case 'number':
     case 'boolean':
+      return valueType;
     case 'string':
+      if (IMAGE_REGEX.test(value)) {
+        return 'image';
+      }
+      if (URL_REGEX.test(value)) {
```
bytasv, line 122: Just an idea - we could also probably try to use native [`URL` construct](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) - seems that when we pass invalid url it throws an error. One benefit of that is that we wouldn't need to "own" regexp in our code:
```
try {
  new URL(value)
  return 'link'
} catch (error) {}
```

Maybe it's not needed right now, but it might be worth to think if it wouldn't be useful to have these methods defined somewhere in `utils` section, I believe we might also want to reuse this logic in the future for other components 🤔

bytasv (approved):

packages/toolpad-components/src/DataGrid.tsx
Janpot: Always use `URL` to parse urls: https://github.com/mui/mui-toolpad/blob/8e6d211d2beb4a346a4dd153105f5190b8a842f7/packages/toolpad-app/src/utils/strings.ts#L56.When you read a url parsing regex on stackoverflow, it will most likely be wrong.

Same for testing the images, parse the `pathname` out and use `.endsWidth`.

packages/toolpad-components/src/DataGrid.tsx
```diff
@@ -337,6 +369,7 @@ const DataGridComponent = React.forwardRef(function DataGridComponent(
             message: typeof errorProp === 'string' ? errorProp : errorProp?.message,
           },
         }}
+        {...(hasImageColumns ? { getRowHeight: () => 'auto' } : {})}
```
Janpot, line 372: from https://mui.com/x/react-data-grid/row-height/#variable-row-height

> Always memoize the function provided to `getRowHeight`. The grid bases on the referential value of these props to cache their values and optimize the rendering.

packages/toolpad-components/src/DataGrid.tsx
```diff
@@ -315,6 +344,13 @@ const DataGridComponent = React.forwardRef(function DataGridComponent(
     [getRowId, columns],
   );
 
+  const handleGetRowHeight = React.useCallback(() => {
```
Janpot, line 347: @apedroferreira This is recalculating `hasImageColumns` for every call to `getRowHeight`. This could lead to O(nm) complexity, we can turn this O(1) if we do the following:

```tsx
  const getRowHeight = React.useMemo(() => {
    const hasImageColumns = columns.some(
      ({ customType }: ToolpadGridColDef) => customType === 'image',
    );
    return hasImageColumns ? () => 'auto' : undefined;
  }, [columns])
```

apedroferreira: Ah, I did not think of that, good point.
I'll add a fix in a separate PR.

oliviertassinari:
👍 

Screenshot 2022-11-04 at 12 52 45

Issue #226, opened 2022-03-25
Undo/Redo action


oliviertassinari:
The very first frustration I had using the tool today is not being able to undo a change. I have tried a change that didn't work, now, I'm not sure how to get back to the previous state that was better. It doesn't incentive me to experiment.

To do:

- [x] basic undo/redo based on a undo history stack of the dom state https://github.com/mui/mui-toolpad/pull/1225
- [x] Keyboard shortcut: Undo <kbd>Ctrl</kbd>+<kbd>Z</kbd> and Redo <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>Z</kbd> #1274
- [x] undo/redo button needs to be disabled when the stack is empty
- [x] If undoing/redoing change that is in a different view - don't do the undo, focus that view instead (i.e. we made changes in connection view, went to canvas editor and did undo - we should end up in connection view). For extra context, read the discussion in https://github.com/mui/mui-toolpad/pull/1225#discussion_r1007856833
- [x] make atomic changes instead of using a throttle to combine changes, currently known issues:
  - [x] Combine multiple changes when typeing an input of a string property value. What's the best way?
  - [x] Make insertion of elements atomic https://github.com/mui/mui-toolpad/pull/1374
  - [ ] ...?
- [ ] Display a history of changes
- [x] Make dom selection part of undo/redo

### How do other tools push to history when typing text?
|                                   | debounce | throttle | word delimiters |
|---------------------------------|:--------:|:--------:|:---------------:|
| google docs                             | ✅        | ✅        |                 |
| notion | ✅        |          | ✅               |
| vscode            |          |          | ✅               |
| figma                             | ✅        |          |                 |

newguy-123:
To get some idea on how Undo and Redo should work in different scenarios:
https://github.com/appsmithorg/appsmith/issues/1426#issuecomment-810842190

apedroferreira:
First prerequisite for undo/redo: https://github.com/mui/mui-toolpad/issues/845

bytasv:
As a follow up to https://github.com/mui/mui-toolpad/pull/1225 we should make following UX improvement:
- [ ] If undoing/redoing change that is in a different view - focus that view (i.e. we made changes in connection view, went to canvas editor and did undo - we should end up in connection view)

High level pseudo-code proposal:
```ts
function getViewForDom(dom: AppDom): ToolpadView
function isSameView(view1: ToolpadView, view2: ToolpadView): boolean
[...]

PR #1225, opened 2022-10-28
Add undo & redo functionality

bytasv:
- Partial of https://github.com/mui/mui-toolpad/issues/226

https://user-images.githubusercontent.com/437214/198537806-41f3a4c1-a424-4c7f-8a1f-d9b26c2a1e29.mov

FYI: if we agree on this approach, I'd like to proceed with integration tests in a separate PR as I suspect it might take a while to implement


packages/toolpad-app/src/toolpad/AppEditor/PageEditor/ComponentEditor.tsx
```diff
@@ -149,7 +149,13 @@ export default function ComponentEditor({ className }: ComponentEditorProps) {
 
   const { selection } = editor;
 
-  const selectedNode = selection ? appDom.getNode(dom, selection) : null;
+  let selectedNode;
+
+  try {
+    selectedNode = selection ? appDom.getNode(dom, selection) : null;
+  } catch (error) {
```
bytasv, line 156: I'm not really sure if this is the right way to deal with it. I bumped into this issue when I add item, it becomes selected, then upon undo it gets removed, but selection ref remains.

IMO we might want to consider to include selected state in the dom, I'm seeing it's pretty common in other apps to also switch selection state with undo/redo and I think it's fairly good UX

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/EditorCanvasHost.tsx
```diff
@@ -145,10 +148,40 @@ export default React.forwardRef<EditorCanvasHostHandle, EditorCanvasHostProps>(
...
       invariant(frameRef.current, 'Iframe ref not attached');
-      setContentWindow(frameRef.current.contentWindow);
-    }, []);
+
+      const iframeWindow = frameRef.current.contentWindow;
+      setContentWindow(iframeWindow);
+
+      const keyDownHandler = iframeKeyDownHandler(iframeWindow.document);
+
+      iframeWindow?.addEventListener('keydown', keyDownHandler);
```
bytasv, line 180: I've tried getting `useShortcut` to work with this iframe flow, but seems because it's loaded async hook just didn't pick up either I just couldn't figure out what's the proper way. I'm open to suggestions if this can be improved.

packages/toolpad-app/src/toolpad/DomLoader.tsx
```diff
@@ -68,8 +75,43 @@ export type DomAction =
       node: appDom.AppDomNode;
     };
 
+const undoStack: appDom.AppDom[] = [];
+const redoStack: appDom.AppDom[] = [];
+
+const updateUndoStack = throttle((dom: appDom.AppDom) => {
+  undoStack.push(dom);
+}, 1000);
```
bytasv, line 83: throttle duration could be reduced here if we feel that we are "skipping" certain "save" points too frequently. In general throttle guards against multiple DOM actions happening one after the other. That sequence of actions for the user doesn't really matter, and we want to save only the "final" state as undo point.

packages/toolpad-app/src/utils/useShortcut.ts
```diff
@@ -16,13 +18,15 @@ export default function useShortcut(
     }
 
     const handleKeydown = (event: KeyboardEvent) => {
-      if (event.code === code && event.metaKey === metaKey) {
+      if (event.code === code && event.metaKey === metaKey && event.shiftKey === shiftKey) {
         handler();
-        event.preventDefault();
+        if (preventDefault) {
```
bytasv, line 23: Had to add this as `useShortcut` always prevented default action when I was inside input so default browser undo inside fields did not work.

IMO it could be better to refactor this method by passing event to the handle so that we can handle such cases outside of this function

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/ComponentEditor.tsx
Janpot: use `appDom.getMaybeNode`. If the selected node should always exist, otherwise it means there's a programmer error, then use an `invariant` and just let it crash.

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/EditorCanvasHost.tsx
Janpot: Just an idea, but what if we just add the undo/redo buttons as per design and leave the shortcut as a follow up? It would allow us to concentrate on functionality first without all the shortcuts clutter.

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/RenderOverlay.tsx
```diff
@@ -163,7 +163,12 @@ export default function RenderOverlay({ canvasHostRef }: RenderOverlayProps) {
     return [pageNode, ...appDom.getDescendants(dom, pageNode)];
   }, [dom, pageNode]);
 
-  const selectedNode = selection && appDom.getNode(dom, selection);
+  let selectedNode: appDom.AppDomNode | null = null;
+  try {
+    selectedNode = selection && appDom.getNode(dom, selection);
+  } catch (error) {
+    console.warn('Selected node does not exists');
```
Janpot, line 170: `invariant`?

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/EditorCanvasHost.tsx
bytasv: I can add buttons, but why would we want to omit key shortcut? IMO it's way more comfortable to use shortcuts than to every time go back and forth to the buttons 🤔

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/ComponentEditor.tsx
bytasv: Thanks, I guess that's what I was missing here, I will use `getMaybeNode`

packages/toolpad-app/src/toolpad/DomLoader.tsx
Janpot: I'm not sure whether throttling is the right paradigm here. It feels like a workaround for not doing atomic updates to the dom. Subsequent changes to the same property can be detected by comparing two consecutive updates. 

I'm not 100% convinced that we should use a non-deterministic method like throttling to work around this problem. We could argue that this is only an intermediate step towards a better implementation, but I don't really see a way forward towards an implementation that is predictable for the user.

packages/toolpad-app/src/toolpad/DomLoader.tsx
```diff
@@ -68,8 +75,43 @@ export type DomAction =
...
+
+const SKIP_UNDO_ACTIONS = ['DOM_UNDO', 'DOM_REDO', 'DOM_SAVED', 'DOM_SAVING', 'DOM_SAVING_ERROR'];
+
 export function domReducer(dom: appDom.AppDom, action: DomAction): appDom.AppDom {
+  if (!SKIP_UNDO_ACTIONS.includes(action.type)) {
+    updateUndoStack(dom);
+  }
+
   switch (action.type) {
+    case 'DOM_UNDO': {
```
Janpot, line 93: If:
- user makes a change to a page
- user goes to custom component and makes a change
- user goes back to the page and makes another change
- user presses undo 3 times

If I understand correctly, this will undo the change to the component as well without any feedback to the user. How would we deal with that? I'm not sure I see how this implementation method can being patched to fix that use case.

packages/toolpad-app/src/toolpad/DomLoader.tsx
bytasv: As much as I tested it works as expected. If we do find a specific issue with this approach we can look into alternatives or improve existing implementation, but I don't really see the reason of discarding it all together without any alternative in place. 
If/when there is any better implementation, we can do that as a follow up

packages/toolpad-app/src/toolpad/DomLoader.tsx
bytasv: IMO what you describe is basically any sort of user change that is not visual - hence there won't be any feedback, just as there is none in other apps when you make non visual changes. I don't think we have to necessarily deal with this in any way - when I'm executing undo (in any application) I expect that every single change will be reverted up until the point that I decide to stop.

If we really have to deal with the situation that you described, then I think we should separate component definition from the dom state and keep it separately

Janpot: * If I take google sheets and make edits across different sheets, upon undo I get taken to the sheet where the change is being undone.
* on google docs, when I make changes in different places in large documents, I scroll to the relevant place in the document where the change has been undone
* in figma, when I try to undo a change that happened on a different page, I get first taken to that page and need to do a second undo command to actually make the change

bytasv: Do you see any way to do this without separating them from the main dom?

Janpot: > Do you see any way to do this without separating them from the main dom?

_To answer the exact question, yes, using the [command pattern](https://en.wikipedia.org/wiki/Command_pattern) to create an undo stack of reducer actions instead of states._

But, I don't think we need to make it that complicated, it will have its own set of edge cases. So let's scratch that. When I was just looking at these 3 applications I realized that none of them have localized undo history stacks. Their undo history is global as well, which means that if we want to provide similar behavior we can still base our undo stack on dom state. (I had some wrong assumptions about how their undo works, it's simpler than I thought)
The thing that is missing in Toolpad vs. them is the visual cue. i.e. We need to detect whether an undo concerns a change that is currently in the view. And if not, we should switch to the view first.

And when combining dom updates together for the undo stack (above in the throttle), we should also detect whether two subsequent changes happen within the same view and never colocate two changes in different views. Or at least let's think about how we will implement this because it will probably share several low-level constructs. 

High level pseudo-code I'm thinking:

```tsx
function getViewForDom(dom: AppDom): ToolpadView
function isSameView(view1: ToolpadView, view2: ToolpadView)
function navigateToView(ToolpadView)
```

But I'm just including this as illustration of where my current thinking is to help illustrate. Don't take it literally, it may be based on wrong assumptions.

bytasv: > we should also detect whether two subsequent changes happen within the same view and never colocate two changes in different views

I'm wondering if this is even possible right now, I mean to execute changes in 2 different views 🤔 (from user perspective)

> The thing that is missing in Toolpad vs. them is the visual cue. i.e. We need to detect whether an undo concerns a change that is currently in the view. And if not, we should switch to the view first.

I agree, this would be a good UX improvement and IMO we should do that, I'd just prefer that it wouldn't be a blocker for the functionality itself. I imagine this could be a fairly big change that could also require to look into performance as we would probably be doing snapshot comparison on each undo/redo action for the entire list of 2 doms 🤔

Janpot: >  I imagine this could be a fairly big change that could also require to look into performance as we would probably be doing snapshot comparison on each undo/redo action for the entire list of 2 doms 🤔

Just to point out that these doms are immutable and their properties can be compared using referential equality. Two subsequent dom states will in most cases have only one node changed. We already used to diff on every change prior to the following PR: https://github.com/mui/mui-toolpad/pull/1167/files#diff-d90295f30b6ff19e7737451a20905e502d606e97be9a94d862c1a16ef8b6017dL236-L241

bytasv: Ah right, that should simplify things 👍 nevertheless, I'd like to approach this UX improvement in a follow up

Janpot: I don't mind splitting the PRs, but then let's add a todo list to https://github.com/mui/mui-toolpad/issues/226 and not close it yet.

bytasv: Done

packages/toolpad-app/src/toolpad/DomLoader.tsx
Janpot: Changing the value of a select comes to mind, or a checkbox. These are easy to fire in rapid succession, but we want them to have separate stack entries

bytasv: I agree, that's why I mentioned that the value could be reduced if we find reasonable duration in which you can physically execute a change

packages/toolpad-app/src/utils/useShortcut.ts
apedroferreira: It should be fine to pass the event to the handler, sounds like a better solution to me.

packages/toolpad-app/src/toolpad/DomLoader.tsx
```diff
@@ -75,8 +82,43 @@ export type DomAction =
       node: appDom.AppDomNode;
     };
 
+const undoStack: appDom.AppDom[] = [];
```
apedroferreira, line 85: Could we use just one stack and have some kind of position variable to define where we are in the stack?

packages/toolpad-app/src/toolpad/DomLoader.tsx
```diff
@@ -75,8 +82,43 @@ export type DomAction =
...
+const redoStack: appDom.AppDom[] = [];
+
+const updateUndoStack = throttle((dom: appDom.AppDom) => {
```
apedroferreira, line 88: Maybe we shouldn't have throttling? Don't we have to keep track of every action the user does, even if they do it quickly?

bytasv: I think you answered your own question - once we have atomic updates, we can change implementation. Until then - it's hardly a case that user could manually execute actions faster than 500ms in between them so hopefully it will be fine until we get to atomic updates

packages/toolpad-app/src/toolpad/DomLoader.tsx
bytasv: We probably could, though I don't seem much benefit

bytasv:
@Janpot I've added undo/redo buttons as request. Are there any other changes that I should do before we can wrap up this work?

packages/toolpad-app/src/utils/useShortcut.ts
bytasv: I'll keep this out of scope for now as there are quite a few places that would require refactoring to get this working correctly (mostly because of `useCallback` usage, otherwise it should be straight forward)

packages/toolpad-app/src/toolpad/DomLoader.tsx
apedroferreira: it feels much more intuitive to me - unless it's much more difficult to implement?
That stack would kind of work like a history of the DOM.

bytasv: Ok, I've just tried to get it working. While I believe it's totally possible to do that, my initial impression is that it adds a context overhead when you have to think about "pointers" and resolve the values. Maybe it's subjective, but for me it's easier to think about 2 "buckets" - one for undo and another for redo, where redo is destroyed when new action is added to undo stack.
I'd prefer to keep this version and if you think this could be done elegantly with single stack and pointer, I'd be curious to see that version

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/ComponentEditor.tsx
```diff
@@ -149,7 +149,7 @@ export default function ComponentEditor({ className }: ComponentEditorProps) {
 
   const { selection } = editor;
 
-  const selectedNode = selection ? appDom.getNode(dom, selection) : null;
+  const selectedNode = selection ? appDom.getMaybeNode(dom, selection) : null;
```
apedroferreira, line 152: This seems similar to the issue in https://github.com/mui/mui-toolpad/pull/1233.
Is there a way we could avoid using `getMaybeNode` here and keep using `getNode` somehow?
Otherwise it might be ok if there is some edge case, just making sure.

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/RenderOverlay.tsx
```diff
@@ -163,7 +163,7 @@ export default function RenderOverlay({ canvasHostRef }: RenderOverlayProps) {
     return [pageNode, ...appDom.getDescendants(dom, pageNode)];
   }, [dom, pageNode]);
 
-  const selectedNode = selection && appDom.getNode(dom, selection);
+  const selectedNode = selection && appDom.getMaybeNode(dom, selection);
```
apedroferreira, line 166: This is reverting my change in the other PR, it also shouldn't be needed?

apedroferreira (approved):
Looks good! Just a couple of comments about `getMaybeNode`

apedroferreira:
Actually tested the UI and found some more issues, should've done that before finishing the review, sorry:
- I don't think the undo/redo buttons should be grouped together with preview and deploy as they're a different feature - they could probably be in another part of the header?
- The undo/redo buttons are gray so they seem disabled - maybe they could be blue if enabled and gray if disabled?
- I got a bug?
https://user-images.githubusercontent.com/10789765/199061264-6a209e1d-8d1d-476d-9373-eafe624690d6.mov

apedroferreira (changes requested):
Noticed some UI/UX issues.

packages/toolpad-app/src/toolpad/DomLoader.tsx
Janpot: Ah, another use-case that requires atomic changes is when we want to show the history of changes. We'd need to be able to be able to attach a label to each undo stack entry that describes it in human language (e.g. "change property 'disabled' to `true`", or "move the `mySelect` component")

packages/toolpad-app/src/toolpad/DomLoader.tsx
```diff
@@ -75,8 +82,45 @@ export type DomAction =
       node: appDom.AppDomNode;
     };
 
+const undoStack: appDom.AppDom[] = [];
```
Janpot, line 85: Perhaps we can
- add the undostack to the `DomLoader` state and keep it immutable? We're basically updating mutable global state in a reducer here. Which is sort of the opposite of what a reducer should do.
- limit the length of the stack to avoid growing indefinitely?

bytasv:
@apedroferreira 

> I don't think the undo/redo buttons should be visually grouped together with preview and deploy as they're a different feature - they could probably be in another part of the header?
> The undo/redo buttons are gray so they seem disabled - maybe they could be blue if enabled and gray if disabled?

As far as I can tell that's done as per design: https://www.figma.com/file/14OEUnwh4ahMb2empYAEc8/Toolpad-UX?node-id=248%3A13264

> I got a bug? Seems to happen pretty easily, some states just show a red error box:

Good catch! I'm looking into this why it's happening. Can't seem to reproduce yet locally, but at least can reproduce on render version

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/ComponentEditor.tsx
bytasv: Do you have any suggestions in mind? I've used `try {} catch` before with `getNode` but that isn't correct since this can actually be undefined 🤔

packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -292,8 +292,9 @@ function RenderedNodeContent({ node, childNodeGroups, Component }: RenderedNodeC
 
   const layoutElementProps = React.useMemo(() => {
     if (appDom.isElement(node) && isPageRow(node)) {
+      const children = childNodeGroups.children || [];
       return {
-        layoutColumnSizes: childNodeGroups.children.map((child) => child.layout?.columnSize?.value),
+        layoutColumnSizes: children.map((child) => child.layout?.columnSize?.value),
```
bytasv, line 297: @apedroferreira I managed to reproduce the crash locally and it seems it was caused because `children` was undefined on `childNodeGroups`. If I default to empty array here it no longer crashes but something else weird happens and I'm not sure yet why exactly:

image

Upon undo/redo I get this weird row element added with `Drop component here` text. I can actually select it and delete it but it shouldn't be here. Do you have any idea why this might be happening and whether it can be related to this change to `layoutColumnsizes` calculation?

packages/toolpad-app/src/toolpad/DomLoader.tsx
bytasv: @Janpot so I've tried to move data to the state itself (attempt can be seen here: https://github.com/mui/mui-toolpad/pull/1225/commits/ff09a8220658878e9237e515bf6f704153fd57f5 ) at first I didn't understand why it lead to unpredictable behaviour, however later I discovered the problem with storing array of refs in state and the biggest problem was that I had to mutate that array from within throttled function which basically beats the purpose of having it in state since I can't derive new state properly.

If we already had atomic updates, then I would change the implementation so we don't use array of "check points" at all and instead IMO we should store the pointer to the previous and next state which would probably be the most straight forward implementation 🤔 

As for second point - I added a limit of undo actions

Janpot: > so I've tried to move data to the state itself 

[This](https://github.com/mui/mui-toolpad/blob/ff09a8220658878e9237e515bf6f704153fd57f5/packages/toolpad-app/src/toolpad/DomLoader.tsx#L159) is still doing side-effects in a reducer. 


I would try the following:

- Add an action `UNDO_HISTORY_PUSH` which reduces by taking the current dom and pushing it on the undo stack
- create a `scheduleUndoHistoryPush ` method that dispatches this action and wrap the `dispatch` method:

   ```tsx
    const scheduleUndoHistoryPush = useThrottled(() => {
      dispatch({ type: 'UNDO_HISTORY_PUSH' })
    }, [500])

    const dispatchWithUndo = React.useCallback((action) => {
      dispatch(action);
      if (isUndoable(action)) {
        scheduleUndoHistoryPush()
      }
    }, [dispatch, isUndoable, scheduleUndoHistoryPush])
   ```
- Use this `dispatchWithUndo` wrapper in the `domApi` instead of `dispatch`

packages/toolpad-app/src/toolpad/DomLoader.tsx
```diff
@@ -75,8 +82,51 @@ export type DomAction =
...
+  undoStack.push(dom);
+  // Destroy redo stack when new action is executed
+  redoStack.length = 0;
+
+  if (undoStack.length > UNDO_LIMIT) {
+    undoStack.shift();
+  }
+}, 500);
+
+const SKIP_UNDO_ACTIONS = ['DOM_UNDO', 'DOM_REDO', 'DOM_SAVED', 'DOM_SAVING', 'DOM_SAVING_ERROR'];
```
Janpot, line 100: I'd always use a `Set` if an array is used for lookup. Remember that `[].includes` has O(n) complexity and `new Set().has` has O(1) complexity. 

The impact is probably pretty small in this case, but it's good form to use the right tool for the job.

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/ComponentEditor.tsx
apedroferreira: It depends on why you made this change, which new cases were you trying to cover?
I'm taking a look to see. Ideally if there is a selected node the node should always exist.

apedroferreira: I guess that we maybe keep this for now and create an issue to also include the selected node in undo/redo (as not sure how simple that would be to do), and then replace these `getMaybeNode` with `getNode` again?

bytasv: I assume this happens because:
1. We add component to canvas
2. We select it
3. We undo add change
4. We still hold selected node reference hence the code fails. So `getMaybeNode` to me makes sense in this case, because the node might not exist

We could do as I mentioned in some previous comments and hold selection state inside dom, then this shouldn't be an issue

apedroferreira: Let's probably keep it like this for now and create an issue for figuring out how to handle the selected component when doing undo/redo, we can other see what other tools are doing.
I guess it's a complex enough problem that it should be solved separately.

Janpot: Interestingly, in figma, selecting a node seems to be considered a undoable action on its own. e.g. if I add an element and then switch selection 5 times between two elements, I need to undo 5 times before the last added element is removed.

apedroferreira: Ah that makes sense, I think we can do that then and store the selection in the DOM?

apedroferreira:
Hey @gerdadesign I guess that my first 2 comments here were design related? https://github.com/mui/mui-toolpad/pull/1225#issuecomment-1297365749
Not sure if we should do something about them but let us know your thoughts!

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/ComponentEditor.tsx
bytasv: I'd say not only figma, I've seen quite a few apps behaving like that. IMO it's quite useful, at least for me working with those apps

packages/toolpad-app/src/toolpad/DomLoader.tsx
bytasv: Thanks alot! That's great tip and I think I finally got it all working 😅 @Janpot please take a look when you can.

There is one thing that I couldn't figure out - I had to use 
```
// eslint-disable-next-line react-hooks/exhaustive-deps
const api = React.useMemo(() => createDomApi(dispatchWithHistory), []);
```
If I pass dependency as eslint requires everything seems to blow up and entire app works in unpredictable way 🤔

packages/toolpad-app/src/toolpad/DomLoader.tsx
bytasv: Thanks, updated this 👍

bytasv:
> Hey @gerdadesign I guess that my first 2 comments here were design related? [#1225 (comment)](https://github.com/mui/mui-toolpad/pull/1225#issuecomment-1297365749) Not sure if we should do something about them but let us know your thoughts!

Since this is done according to the latest design 👇 I'd suggest discussion whether grouping undo/redo actions with preview/deploy is the right way or not to take place separate from this PR (in other words not be a blocking factor)
image

packages/toolpad-app/src/runtime/ToolpadApp.tsx
apedroferreira: Maybe the throttling is not enough to stop some in-between state where rows actually don't have `children` or the `children` prop.
Does changing every `appDom.createElement(dom, PAGE_ROW_COMPONENT_ID, {})` in the `RenderOverlay` file to `appDom.createElement(dom, PAGE_ROW_COMPONENT_ID, { children: [] })` maybe fix the issue?

bytasv: My guess is that it has to do with the same reason it failed for selected nodes - when you remove something, code expects that it's there, just as you said - it should never be `undefined`, but upon undo action it probably will be `undefined`

apedroferreira: Turns out throttling is just not working as intended. I proposed a fix here: https://github.com/mui/mui-toolpad/pull/1266

bytasv: Thanks a lot! Yeah that does seem to solve that issue. It's a little bit weird because even with throttle implementation it undo/redo itself worked (in terms of getting me to previous or next state) but as you pointed out it's possible that it created unwanted intermediate state which lead to error thrown 🤔 

Anyways, thanks a lot for looking into it and helping out ❤️

Janpot: I think there's a misunderstanding about what `useThrottled` does. It throttles updates to a value, it doesn't give you a function that wraps another function with throttling.

There are 4 hooks:

* `useDebounced`: debounces a value
* `useDebouncedHandler`: wraps a function with debounce behavior
* `useThrottled`: throttles a value
* `useThrottledHandler`: wraps a function with throttling behavior (doesn't exist yet in the codebase)

It seems like you were looking for `useThrottledHandler`, and you used something analog to `useDebouncedHandler`

packages/toolpad-app/src/toolpad/DomLoader.tsx
```diff
@@ -330,8 +409,30 @@ export default function DomProvider({ appId, children }: DomContextProps) {
     saveError: null,
     savedDom: dom,
     dom,
+    pendingHistoryDom: dom,
+    undo: [],
+    redo: [],
   });
-  const api = React.useMemo(() => createDomApi(dispatch), []);
+
+  const scheduleHistoryUpdate = React.useMemo(
```
Janpot, line 417: Would this work? It could be interesting to rely less on lodash.

```tsx
const scheduleHistoryUpdate = useDebouncedHandler((action: DomAction) => {
  if (action) {
    dispatch({ type: 'DOM_UPDATE_HISTORY' });
  }
}, 500)
```

packages/toolpad-app/src/toolpad/DomLoader.tsx
```diff
@@ -276,6 +343,9 @@ export interface DomLoader {
   saving: boolean;
   unsavedChanges: number;
   saveError: string | null;
+  pendingHistoryDom: appDom.AppDom;
+  undo: appDom.AppDom[];
```
Janpot, line 347: Some nits:

* Perhaps it makes sense to already anticipate that we will take selection in the undo stack (and maybe even other things) and foresee a type that can be extended with that?

    ```suggestion
    undo: { dom: appDom.AppDom }[];
    ```

* very much a nit, but in terms of naming `undo` and `redo` sound like methods, I'd consider naming these `undoStack` and `redoStack` to better convey what they are

packages/toolpad-app/src/toolpad/DomLoader.tsx
bytasv: > It could be interesting to rely less on lodash.

It's something I noticed through the codebase - we have many methods written by ourselvs that are part of lodash suite, why do we want to keep doing that rather than use what lodash provides?
There are at least a few issues with this approach:
1. We reimplement something that we can already use
2. We don't cover those with specs (which is done by library)
3. We don't have any documentation, which makes it harder to even know that those things exist in our codebase

If anything I'd think that relying more on libraries that provide common functionality is a good thing, so we can focus on feature development or other more important things 🤔 
Lodash for common utilities and something like [ahooks](https://ahooks.js.org/) for all the different hooks that we have and we could remove quite a big chunk of utilities from our codebase

> Also, in which case would action be falsy? The types don't seem to reflect that behavior

I just checked again - and it seems I no longer get dispatch without action, maybe it's related to the previous use of throttle function 🤔 for some reason `dispatch` was called without the action before.

I just cleaned this up

packages/toolpad-app/src/toolpad/DomLoader.tsx
bytasv: > Perhaps it makes sense to already anticipate that we will take selection in the undo stack (and maybe even other things) and foresee a type that can be extended with that?

Hmm, I'd then rather wait for an actual implementation. Initially I was thinking that we should do selection state part of the dom itself. With the approach you are proposing I assume selection and action would happen at the same time, where in other apps selection itself is considered a separate undoable action. If however selection was part of the dom, then we wouldn't need to do any extra handling for the undo/redo

> very much a nit, but in terms of naming undo and redo sound like methods, I'd consider naming these undoStack and redoStack to better convey what they are

Makes sense 👍 just renamed them

packages/toolpad-app/src/toolpad/DomLoader.tsx
Janpot: > It's something I noticed through the codebase - we have many methods written by ourselvs that are part of lodash suite, why do we want to keep doing that rather than use what lodash provides?

1. Because `lodash` doesn't provide what we need in an immutable environment where we rely on referential equality to optimize our code, example:

    ```tsx
    const x = { foo: 'bar' }

    x === _.omit({foo:'bar'}, 'hello')
    ```
2. Because `lodash` is full of methods that we now, in 2022 have perfectly fine native javascript constructs for. `_.map`, `_.filter`,...
3. Because `lodash` contains methods that make code hard to read, what does `_.isEmpty` do exactly? Does it error when not pased an array? Why is there an `_.isString`? These clutter the codebase, and make it harder to read.


> If anything I'd think that relying more on libraries that provide common functionality is a good thing, so we can focus on feature development or other more important things

I'm not against relying on libraries, I am against relying on large kitchensink libraries that clutter the codebase with things that are better expressed in native javascript. It is contagious, if you ever want to extract a module from the codebase, it will now have to drag along this huge library that you just need 1 method from.

packages/toolpad-app/src/runtime/ToolpadApp.tsx
bytasv: I just checked but it seems we don't even have `useThrottledHandler` 🤔
IMO this is a good illustration of the problem I mentioned in the comment below: https://github.com/mui/mui-toolpad/pull/1225#discussion_r1012682888

1. It's not really clear whether the implementation that we have is bug free?
2. It's more difficult to know that these methods even exist in our codebase
3. It's hard to know what exactly should be used as there are no docs besides the types themselves

packages/toolpad-app/src/toolpad/DomLoader.tsx
```diff
@@ -147,6 +159,55 @@ export function domLoaderReducer(state: DomLoader, action: DomAction): DomLoader
   }
 
   switch (action.type) {
+    case 'DOM_UPDATE_HISTORY': {
+      const updatedUndoStack = [...state.undoStack, state.pendingHistoryDom];
```
Janpot, line 163: Not sure I understand why we need `pendingHistoryDom`, can't we just use `state.dom` here?

packages/toolpad-app/src/toolpad/DomLoader.tsx
Janpot: >  Initially I was thinking that we should do selection state part of the dom itself. 

The dom is what we persist on disk and what we regard as the shareable artifact containing a toolpad application. The selection is local state, it shouldn't be part of that.

> With the approach you are proposing I assume selection and action would happen at the same time

Selection would add an entry in the undo stack with the same dom but a different slection.

packages/toolpad-app/src/toolpad/DomLoader.tsx
bytasv: That was my first idea, I did try using `state.dom` here, but by the time this is called `state.dom` has updated values, where what I want is values that were in the dom before the actions were called. Storing a reference to last state that hasn't been pushed to history stack resolved this issue.
Just to be 100% sure I tried passing `state.dom` here after all these changes with throttle/debounce, but it still doesn't work.
Basically if we store `state.dom` here, we loose 1 point of history, and by using reference to "pending" state it always works as expected

packages/toolpad-app/src/toolpad/DomLoader.tsx
bytasv: > Because lodash doesn't provide what we need in an immutable environment where we rely on referential equality to optimize our code, example:

Is there no alternative library that works with immutable data structure?

> Because lodash is full of methods that we now, in 2022 have perfectly fine native javascript constructs for. _.map, _.filter,...

I don't see a problem with that - we DONT have to use these methods, there is tons of methods that we can still use from lodash

> Because lodash contains methods that make code hard to read, what does _.isEmpty do exactly? Does it error when not pased an array? Why is there an _.isString? These clutter the codebase, and make it harder to read.

Honestly never had an issue with that, maybe easier to speak about concrete examples, but if something is unclear opening up documentation usually sorted it all out. I'm not sure how having our own undocumented methods makes this less of a problem?

> I'm not against relying on libraries, I am against relying on large kitchensink libraries that clutter the codebase with things that are better expressed in native javascript, it is contagious.

I'm not proposing to use lodash method for native alternatives, I'm proposing not to reinvent exact same methods that do not have native alternatives and have them sit in codebase without a tests or proper documentation

packages/toolpad-app/src/toolpad/DomLoader.tsx
Janpot: Yes, looking good. What happens if you wrap `dispatchWithHistory` with `useEvent`? (We can't call it during render then, but I don't think we do that anywhere, and we probably never should anyway.)

packages/toolpad-app/src/toolpad/DomLoader.tsx
bytasv: Fine, regardless I'd leave this out of scope for this PR then as it's already quite big

packages/toolpad-app/src/toolpad/DomLoader.tsx
bytasv: Are you talking about this `useEvent`: https://beta.reactjs.org/apis/react/useEvent ? I think it's not supported yet in the current version?

Janpot: It's a common practice to have such hook in your codebase: https://github.com/mui/mui-toolpad/blob/9c8764b689dbb7a44cc8d6833ffd393bdf402cd6/packages/toolpad-app/src/utils/useEvent.ts#L24

This version also contains an additional check that prevents us from accidentally using it during render.

bytasv: > It's a common practice to have such hook in your codebase

```
__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.ReactCurrentDispatcher
```

Is it really a common practice? I mean you can't be more explicit than that to no use these internals 🤔

Janpot: https://github.com/mui/mui-toolpad/pull/1269

packages/toolpad-app/src/toolpad/DomLoader.tsx
Janpot: oh ok I see, it just means that the top of the stack contains the current state, and that doing an undo means first popping an item, and then resetting the dom to what's in the top of the stack instead of what we just popped off.

packages/toolpad-app/src/toolpad/DomLoader.tsx
apedroferreira: I usually don't use `lodash` and use `ramda` instead (it's supposed to be immutable), but actually even when I do, the functions that I use `ramda` for could probably easily be written by myself... So I think it's not a bad approach to avoid using these libraries and using our own implementations instead to save on bundle size, a `debounce` function isn't difficult to have ourselves either.

About some of these functions not being documented - I didn't know about them either, that's why I used lodash in my proposal, but I guess that's what code reviews are for? :) I guess `useDebouncedHandler` would be the choice here.

Also debouncing/throttling seem to be slightly different things, I think throttling always runs the function immediately the first time it's called? So we probably want to use debouncing in this case.

packages/toolpad-app/src/toolpad/DomLoader.tsx
bytasv: Hmm, I'm not sure I fully understand whether you meant that this is how it's working now or how it would work if we pushed `state.dom` instead of `pendingHistoryDom` but the idea is to keep undo stack populated only with previous states 🤔 
Do you see a way that we could avoid using `pendingHistoryDom` ref? I was trying various things, like passing state to dispatch, etc... but none of them worked predictably and this (current) approach seemed to do the trick

packages/toolpad-app/src/toolpad/DomLoader.tsx
bytasv: > could probably easily be written by myself... 

IMO easily written should not be condition of whether it's worth rewriting it in our codebase or not. It's one thing to copy/paste a snippet, have to maintain it, etc.. we should also then ensure that all such code is properly tested, we pull any updates from the source (if there are any) as well as have the documented properly.

> So I think it's not a bad approach to avoid using these libraries and using our own implementations instead to save on bundle size

I'm of an opposition opinion here - we should avoid prematurely optimising things. Especially when we copy same snippets to our code base we still have to include that code in the bundle. And with a treeshaking importing parts from library shouldn't be such a big deal.

> About some of these functions not being documented - I didn't know about them either, that's why I used lodash in my proposal, but I guess that's what code reviews are for?

While it can prevent such things from happening, I'd argue that reviews should be used for that. It's also very likely that the person knowing about the util method might also miss the different implementation or won't be even part of review in some case.

> I think throttling always runs the function immediately the first time it's called? So we probably want to use debouncing in this case.

It depends on the implementation - lodash allows to control how throttle is called and whether on leading or trailing edge of the call (or could be both). The main difference between throttle and debounce is that thorttle acts as a "rate limiter" where it allows executing function at max times per specified duration, i.e. calling function every 100ms with a throttle of 500ms for 2000ms should lead to 4 function calls. Where debounce waits for specified DURATION to pass after last function call and call the function once. So in the case of every 100ms function call with 500ms duration after 2000ms it should trigger function ONCE at 2500ms point (or something close to that)

packages/toolpad-app/src/toolpad/DomLoader.tsx
bytasv: @Janpot so instead of using `useMemo` I tried using `useEvent` like so:
```diff
diff --git a/packages/toolpad-app/src/toolpad/DomLoader.tsx b/packages/toolpad-app/src/toolpad/DomLoader.tsx
index ed031d9d..370920e8 100644
--- a/packages/toolpad-app/src/toolpad/DomLoader.tsx
+++ b/packages/toolpad-app/src/toolpad/DomLoader.tsx
@@ -11,6 +11,7 @@ import { createProvidedContext } from '../utils/react';
 import { mapValues } from '../utils/collections';
 import insecureHash from '../utils/insecureHash';
 import { NodeHashes } from '../types';
+import useEvent from '../utils/useEvent';
 
 export type DomAction =
   | {
@@ -414,16 +415,14 @@ export default function DomProvider({ appId, children }: DomContextProps) {
     redoStack: [],
   });
 
-  const scheduleHistoryUpdate = React.useMemo(
-    () =>
-      throttle(
-        () => {
-          dispatch({ type: 'DOM_UPDATE_HISTORY' });
-        },
-        500,
-        { leading: false, trailing: true },
-      ),
-    [],
+  const scheduleHistoryUpdate = useEvent(() =>
+    throttle(
+      () => {
+        dispatch({ type: 'DOM_UPDATE_HISTORY' });
+      },
+      500,
+      { leading: false, trailing: true },
+    ),
   );
 
   const dispatchWithHistory = (action: DomAction) => {

```

But I either do that incorrectly or it's not working for some reason as undo just does nothing after that 🤔

packages/toolpad-app/src/toolpad/DomLoader.tsx
Janpot: > Do you see a way that we could avoid using `pendingHistoryDom` ref?

I think it can be done if we slightly alter this philosophy: "the idea is to keep undo stack populated only with previous states" to "the idea is to keep undo stack populated with full history up until the current state". And initialize the `undoStack` with `[dom]` instead of `[]`. And then when popping an item off the stack, not use that popped off item to restore the dom, but look at the top item of the trimmed stack. 

Basically it would work the same way, only instead of adding an extra property for `pendingHistoryDom`, the top item in the stack would be `pendingHistoryDom`. I think there are some small benefits to this method:

* One less property to keep in sync `pendingHistoryDom`
* We will want to give an overview of the history of changes at some point. To do that we will have to add a human readable label to each entry in the stack. In the implementation with `pendingHistoryDom` there would be an entry missing for the last change, and we'd need to special case the last change. In the implementation without `pendingHistoryDom`, the `undoStack` would reflect the whole history up until the last change, and we can just flat render this array.

Then again, I may be wrong in my thinking, it's only based on reading the implementation and thinking about the consequences.

packages/toolpad-app/src/toolpad/DomLoader.tsx
apedroferreira: I think those are valid points too, I guess we just have to pick a standard and go with it? As long as we're consistent I don't think either of these approaches would be bad right now, and we can fix issues as we run into them?

The only point I'm not sure about is what we can do to be more aware of certain `utils` we already have that could be useful. At least some comments clarifying how the utils are supposed to be used would probably be helpful though!

packages/toolpad-app/src/toolpad/DomLoader.tsx
Janpot:
```tsx
  const dispatchWithHistory = useEvent((action: DomAction) => {
    dispatch(action);

    if (!SKIP_UNDO_ACTIONS.has(action.type)) {
      scheduleHistoryUpdate();
    }
  });

  const api = React.useMemo(() => createDomApi(dispatchWithHistory), [dispatchWithHistory]);
```

bytasv: 🤦 ah, my bad... thanks! now it's working 👍

packages/toolpad-app/src/toolpad/DomLoader.tsx
bytasv: @Janpot you are correct, thanks, another great tip 💪 I changed the implementation according to your suggestion. I guess the main con of this new approach is that undo/redo methods work slightly differently, but that should be fine 🤔 

> In the implementation with pendingHistoryDom there would be an entry missing for the last change, and we'd need to special case the last change

I don't think that's true - in the case of `pendingHistoryDom` initially we would have simply empty array, which means there is no history to undo. At the time we would trigger action and push `pendingHistoryDom` we could add a label. IMO specifically right now we will have to add extra checks if we want to render history and omit top most entry, i.e. ```
if (undoStack.length < 2) {
  // means there is no history to undo
}
```

We just need to remember in the future that top most item in the undo stack represents current state 🤔

Janpot (approved):

Janpot:
This took a lot of shaping, the result is great. We can absolutely build on this.

bytasv:
@Janpot thanks! I also added to the parent issue todo for selection being part of undo/redo

@apedroferreira do you have any feedback that I should address merging this PR or do you think it's good to go?

apedroferreira (approved):
Looks good!

oliviertassinari (commented):
It's nice to have the first version of this feature 👍.

The app bar buttons work on my end in https://master--toolpad.mui.com/_toolpad/app/cl9pj1woq0001l59l5y3kztln/pages/7k195tr. However, I'm not able to use the keyboard on macOS. Is this expected?

bytasv:
> It's nice to have the first version of this feature 👍.
> 
> The app bar buttons work on my end in https://master--toolpad.mui.com/_toolpad/app/cl9pj1woq0001l59l5y3kztln/pages/7k195tr. However, I'm not able to use the keyboard on macOS. Is this expected?

Could you share exact steps? I've just tried on the link and keyboard works for me 🤔

oliviertassinari:
>  I've just tried on the link and keyboard works for me

I had a look, here is the root problem: https://github.com/mui/mui-toolpad/pull/1274#discussion_r1017032943.

oliviertassinari:
I have found a bug using the feature, I have reported it at https://github.com/mui/mui-toolpad/issues/226#issuecomment-1310706191.

PR #1227, opened 2022-10-28
Allow adding rows to DataGrid

### LINKED ISSUE #1038: Allow adding a new row to the data grid
### Duplicates

### Latest version

### Summary 💡

The new row should be editable and should write back to the data source. For the scope of this ticket we'll focus on Fetch data source.

Relates to #994

bharatkashyap:
- Closes #1038 
- To merge #1105 before this since this branches from it

- Visual demonstration 

https://user-images.githubusercontent.com/19550456/198573654-fcdb779a-23ae-4a0c-b06f-e9f6464481f2.mov


apedroferreira:
Feedback from the video:
- So the user needs to scroll down to find the temporary new record to add? They might not know it shows up there, what if the new record showed up at the top instead?

packages/toolpad-components/src/DataGrid.tsx
```diff
@@ -329,12 +382,176 @@ const DataGridComponent = React.forwardRef(function DataGridComponent(
...
+    handleSaveClick,
+  ]);
+
+  const columnsWithActions = React.useMemo(() => {
+    return actionField ? [...columns, actionField] : columns;
+  }, [columns, actionField]);
+
+  const handleCreateClick = React.useCallback(() => {
+    const newRow = Object.fromEntries(columns.map((column) => [column.field, '']));
+    setRowId(newRow, '...');
```
apedroferreira, line 524: We could use a constance for the new row id (`...` in this case)

bharatkashyap:
> Feedback from the video:
> 
> So the user needs to scroll down to find the temporary new record to add? They might not know it shows up there... What if the new (temporary) record showed up at the top instead? And once submitted it would go to its place? It's just an idea but maybe something kind of like that could work?

Makes sense, I'll follow this given that the `scrollToIndexes` function I was using from the DataGrid `apiRef` doesn't seem to work appropriately - perhaps I'm missing something in the implementation there.

apedroferreira (approved):
Sorry I forgot this PR Bharat.
Didn't get to test it but the latest code change since the review seems good.

PR #1232, opened 2022-10-28
Improve dev mode performance

Janpot:
As per last retro, removing this (`import { Error as ErrorIcon } from '@mui/icons-material';`) import improves performance a lot. For now we can just pass a component down. Hopefully we can eventually move forward with https://github.com/mui/material-ui/pull/30510

Alternatively we can explore compiling these packages to commonjs

@apedroferreira @bytasv please try out in your local and let me know if it improves things


.eslintrc.js
```diff
@@ -97,9 +97,10 @@ module.exports = {
               {
                 // Running into issues with @mui/icons-material not being an ESM package, while the
                 // toolpad-core package is. This makes Next.js try to load @mui/icons-material/* as ESM
-                // Loading from the index export works as a workaround for now
-                group: ['@mui/icons-material/*'],
-                message: "Use `import { Icon } from '@mui/icons-material'` instead.",
+                // We'll just avoid mporting icons in these packages
+                // Remove restriction after https://github.com/mui/material-ui/pull/30510 gets resolved
+                group: ['@mui/icons-material', '@mui/icons-material/*'],
+                message: "Don't use @mui/icons-material inthese packages for now.",
```
bytasv, line 103: typo
```suggestion
                message: "Don't use @mui/icons-material in these packages for now.",
```

.eslintrc.js
```diff
@@ -97,9 +97,10 @@ module.exports = {
               {
                 // Running into issues with @mui/icons-material not being an ESM package, while the
                 // toolpad-core package is. This makes Next.js try to load @mui/icons-material/* as ESM
-                // Loading from the index export works as a workaround for now
-                group: ['@mui/icons-material/*'],
-                message: "Use `import { Icon } from '@mui/icons-material'` instead.",
+                // We'll just avoid mporting icons in these packages
```
bytasv, line 100: typo
```suggestion
                // We'll just avoid importing icons in these packages
```

bytasv (approved):
Hmm, I can't really tell the difference between how these changes work and what was before 🤔 

Could it be that we are looking at different things? I'm checking by switching between pages, but it "feels" like the time it takes to load is very similar. Nevertheless approving this ✅ 

https://user-images.githubusercontent.com/437214/198985058-bdb09090-a042-401a-ae2a-0f4a6ec6c063.mov

Janpot:
I didn't measure it exactly, but I see a clear difference:
before:

https://user-images.githubusercontent.com/2109932/198986321-6b87d678-a2c9-43ac-abd0-d0f4908ff550.mov


after:


https://user-images.githubusercontent.com/2109932/198986384-caefa144-10b0-479b-b7bd-dc559664f000.mov

bytasv:
I can see the difference in your recording, but I'm unable to reproduce same on my machine 🤔 are you running `yarn dev` or something different? Curious if Pedro will be able to feel the difference and why I'm not seeing the same 🤔

Janpot:
> are you running `yarn dev` or something different?

In usually run `yarn && yarn dev` after switching branches. How are you checking out the code of this PR?

bytasv:
> How are you checking out the code of this PR?

I have checked out branch locally, just tried to run `yarn` to install/refresh deps if anything is missing, then ran `yarn dev`. Wondering could it be related to how docker is set up? I.e. maybe you have more resources allocated 🤔 

@apedroferreira could you also check? While it seems that I might be missing something on my end, it would be good to verify

Janpot:
> Wondering could it be related to how docker is set up?

I doubt it, only the database is running in docker in development mode.

apedroferreira (approved):

Janpot:
It's important to understand that Next.js optimizes for startup time in dev mode. That means it will compile sources on-demand. As soon as it is compiled once, the HMR should make incremental compilation faster. The consequence is that the first time opening a page will be a little slower. It's either this or longer startup time. Here's to hoping that the upcoming [`turbopack`](https://nextjs.org/docs/advanced-features/turbopack) update will make this a lot faster still. I don't think there's much we can do here, maybe we can [profile](https://nextjs.org/docs/advanced-features/compiler#swc-trace-profiling) some internals? All in all, I don't think it makes much sense to look at initial compile time. If there are reasons why you often have to restart the application, I'd love to hear about it.

bytasv:
I personally don't care much about initial load time as I understand it always takes longer. The problem is that page load time for me doesn't change even when I'm not changing any code, i.e. I just click back and forth between pages it takes as much time as in your video of "before". That's why I'm puzzled why I'm not seeing same improvement as you do 🤔

apedroferreira:
If we can't find any more quick wins I'm fine with things staying as they are and hoping upcoming Next.js updates improve things, it's not that bad of an issue to spend more time on it in my opinion.

Issue #1221, opened 2022-10-27
Quickly adding components leads to error


bytasv:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

Steps:

1. Start dragging any component from drawer onto canvas
2. Upon 3rd or 4th component (maybe it will need more?) you should get `node not found` error. Seems that it doesn't matter what kind of component is being added
<img width="714" alt="image" src="https://user-images.githubusercontent.com/437214/198314131-9fa2ba8a-15e3-43f8-b767-303eaaf3ce1b.png">

https://user-images.githubusercontent.com/437214/198314954-09f6fdb6-aed1-4484-82ec-e9c1bf03b440.mov

### Current behavior 😯

Prod build swallows error and we can continue to work, seems that it's only visible in the console.

Locally we get an error dialog

### Expected behavior 🤔

No error happening when we add components to the canvas

### Your environment 🌎

Local / latest version

prakhargupta1:
Another [app](https://master--toolpad.mui.com/_toolpad/app/cl1on1bd80041a1pdrrfdt0rq/pages/cl3srtf4200003f6e1j50j390) is throwing similar error message: "Node "undefined" not found". Not sure if they are related, but I was unable to make it work.

Janpot:
@prakhargupta1 That is different. Can you open a ticket?

PR #1233, opened 2022-10-28
Fix node not found error

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/1221

Explanation: 
When moving or adding a new component in the canvas, there is logic that checks for the parent of the component node. If the component is a new component, there is no parent, therefore this checks throws an error.

This error was not throwing before because `getParent` was using `getMaybeNode`, but it's not anymore after recent changes.

I made it so that `.getParent` only gets called if we're moving a component already in the canvas, instead of adding a new component.


bytasv (approved):

Janpot (approved):
👍 Oh ok, I see. The code was assuming it could call `getParent` on a non existing node. That wasn't part of the implicit contract indeed.

PR #1234, opened 2022-10-29
Add Link Component

### LINKED ISSUE #571: Add a Link component
Should at the very least support

- href
- text

In the meantime, it can be implemented with a custom component:

```tsx
import * as React from "react";
import { Link as MUILink } from "@mui/material";
import { createComponent } from "@mui/toolpad-core";

export interface LinkProps {
  value: string;
  href: string;
}

function Link({ value, href }: LinkProps) {
  return (
    <MUILink href={href} target="_blank">
      {value}
    </MUILink>
  );
}

export default createComponent(Link, {
  argTypes: {
    value: {
      typeDef: { type: "string" },
      defaultValue: "",
    },
    href: {
      typeDef: { type: "string" },
      defaultValue: "",
    },
  },
});
```

One of #78

bharatkashyap:
- Closes #571 
- Visual demonstration 

https://user-images.githubusercontent.com/19550456/198830985-ecb480a5-c544-408b-a445-a19687089e62.mov


packages/toolpad-components/src/Link.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export default createComponent(Link, {
+  argTypes: {
+    value: {
+      typeDef: { type: 'string' },
+      defaultValue: 'Link',
+    },
+    href: {
+      typeDef: { type: 'string' },
+      defaultValue: 'about:blank',
```
apedroferreira, line 26: What if it was empty? Does `about:blank` have an advantage?

apedroferreira (approved):
LGTM!
Another note, but not sure if we should do anything about it:
Today me and @Janpot took a look at Dynaboard and they seem to have a single text component with different modes: text, link, and markdown. Not sure if we should go for that kind of approach too instead?

packages/toolpad-components/src/Link.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { createComponent } from '@mui/toolpad-core';
+
+interface LinkProps extends Omit<MuiLinkProps, 'children'> {
+  href: string;
+  value: string;
+}
+
+function Link({ value, href }: LinkProps) {
+  return (
+    <MuiLink href={href} target="_blank">
```
apedroferreira, line 12: Should we add `rel="noopener noreferrer"` for external links?

apedroferreira: `nofollow` might be useful too, so that people don't use Toolpad to generate backlinks to their site.

PR #1257, opened 2022-10-31
Fix app templates tests

apedroferreira:
One of the app templates tests seems broken. This seems to make it stable as far as I've tested locally.

### DIFF
--- a/test/integration/appTemplates.spec.ts
+++ b/test/integration/appTemplates.spec.ts
@@ -1,6 +1,7 @@
 import { ToolpadHome } from '../models/ToolpadHome';
 import { ToolpadRuntime } from '../models/ToolpadRuntime';
 import { test, expect } from '../playwright/test';
+import { waitForDataResponse } from '../utils/response';
 
 test('can use statistics app template', async ({ page }) => {
   const homeModel = new ToolpadHome(page);
@@ -18,7 +19,7 @@ test('can use statistics app template', async ({ page }) => {
   );
 });
 
-test('can use images app template', async ({ page }) => {
+test('can use images app template', async ({ page, baseURL }) => {
   const homeModel = new ToolpadHome(page);
   await homeModel.goto();
   const app = await homeModel.createApplication({ appTemplateId: 'images' });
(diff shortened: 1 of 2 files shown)


test/integration/appTemplates.spec.ts
```diff
@@ -28,21 +29,17 @@ test('can use images app template', async ({ page }) => {
...
   await breedInputLocator.click();
   await page.locator('li:has-text("australian")').click();
-  await waitForDogAPIResponse();
+  await waitForDataResponse(page, baseURL!);
 
   const subBreedInputLocator = page.locator('[aria-haspopup="listbox"]').nth(1);
   await subBreedInputLocator.click();
   await page.locator('li:has-text("shepherd")').click();
-  await waitForDogAPIResponse();
+  await waitForDataResponse(page, baseURL!);
```
Janpot, line 42: These are implementation details. Why can't we just wait until the image has the expected attribute?

apedroferreira: Is there a way? I looked in the docs but couldn't find it.

apedroferreira: Not sure why I didn't try this one: `await expect(locator).toHaveAttribute('type', 'text');`
Will try it next, maybe it will work.

Janpot: > Is there a way?

By using a locator with an [attribute selector](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors)?

test/integration/appTemplates.spec.ts
```diff
@@ -33,13 +34,12 @@ test('can use images app template', async ({ page }) => {
...
   await subBreedInputLocator.click();
-  await Promise.all([
-    page.locator('li:has-text("shepherd")').click(),
-    page.waitForResponse('https://images.dog.ceo/**'),
-  ]);
+  await page.locator('li:has-text("shepherd")').click();
+  await waitForDataResponse(page, baseURL!);
 
   const imageLocator = page.locator('img');
   expect(await imageLocator.getAttribute('src')).toContain(
```
Janpot, line 45: @apedroferreira Can you try

```suggestion
  await expect(imageLocator.getAttribute('src')).toContain(
```

Janpot (approved):
👌

test/integration/appTemplates.spec.ts
```diff
@@ -36,13 +36,11 @@ test('can use images app template', async ({ page }) => {
 
   const subBreedInputLocator = page.locator('[aria-haspopup="listbox"]').nth(1);
```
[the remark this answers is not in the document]
apedroferreira: We did but the selects are divs... Is there a quick way to find the element in that case?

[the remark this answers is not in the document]
apedroferreira: Merging but might try to improve it later, I wasn't able to so far.

PR #1264, opened 2022-11-02
Provide accessible locators

Janpot:
PR got merged before I had the chance to answer https://github.com/mui/mui-toolpad/pull/1257#discussion_r1011769767

Here's my proposal after playing with it a bit locally. This makes the locators less implementation dependant. And base them purely on accessible properties

### DIFF
--- a/test/integration/appTemplates.spec.ts
+++ b/test/integration/appTemplates.spec.ts
@@ -30,13 +30,13 @@ test('can use images app template', async ({ page }) => {
 
   await page.locator('h3:has-text("Dog Images")').waitFor({ state: 'visible' });
 
-  const breedInputLocator = page.locator('[aria-haspopup="listbox"]').first();
+  const breedInputLocator = page.getByRole('button', { name: /Pick a dog breed/ });
   await breedInputLocator.click();
-  await page.locator('li:has-text("australian")').click();
+  await page.getByRole('option', { name: 'australian' }).click();
 
-  const subBreedInputLocator = page.locator('[aria-haspopup="listbox"]').nth(1);
+  const subBreedInputLocator = page.getByRole('button', { name: /Pick a sub-breed/ });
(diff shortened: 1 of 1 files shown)


apedroferreira (approved):
Cool! I didn't know we could use this accessible name: https://w3c.github.io/accname/#dfn-accessible-name
Looks great, thanks for the improvement and for this PR!

test/integration/appTemplates.spec.ts
```diff
@@ -30,13 +30,13 @@ test('can use images app template', async ({ page }) => {
 
   await page.locator('h3:has-text("Dog Images")').waitFor({ state: 'visible' });
 
-  const breedInputLocator = page.locator('[aria-haspopup="listbox"]').first();
+  const breedInputLocator = page.getByRole('button', { name: /Pick a dog breed/ });
```
apedroferreira, line 33: Why the regex here and not just a regular string btw?

Janpot: The accessible name is "Pick a dog breed akita". Not sure it's intended or not. I also had trouble locating the empty select. And neither of them I could locate with `getByLabel`.

I almost feel like it could be interesting to add some tests to core that verifies the ergonomy of using MUI components with `playwright`, as well as a few recipes in the docs. I don't have the feeling it currently is easy to figure out.


e.g. I'd expect 


```tsx
<Textfield select label='foo' />
```

to be located with 
```tsx
page.getByLabel('foo', { exact: true })
page.getByRole('button', { name: 'foo' })
```

And should the accessible name for a select contain the value?

@mui/core interested in exploring this further together?

apedroferreira: Yeah I've also had trouble sometimes finding a good way to find certain components, not sure if that means that MUI Core could improve the semantics of the components somehow.

PR #1270, opened 2022-11-03
Codify lodash usage in eslint

Janpot:
Avoid kitchensink libraries, especially unmaintained ones like `lodash`. We write a slightly more verbose, but a universally understood style of javascript.


packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -34,6 +34,8 @@ import {
   JsExpressionAttrValue,
 } from '@mui/toolpad-core';
 import { createProvidedContext, useAssertedContext } from '@mui/toolpad-utils/react';
+import { filterKeys, mapProperties, mapValues } from '@mui/toolpad-utils/collections';
+import { set as setObjectPath } from 'lodash-es';
```
apedroferreira, line 38: If we plan to eventually completely get rid of `lodash` we could maybe create utility methods for these operations, which for now would still use `lodash` but then we would know where to replace it.
Just a possibility, maybe it helps to have the `lodash` imports less spread out.

apedroferreira (approved):
Sounds good, should we create an issue to replace `groupBy` and `isEqual`?
I remember also using `setWith` for a specific case in bindings, that one might be tougher to replace but still possible.

packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -1226,7 +1228,7 @@ function QueryNode({ page, node }: QueryNodeProps) {
     Object.fromEntries(node.params ?? []),
   );
 
-  const configBindings = _.pick(node.attributes, USE_DATA_QUERY_CONFIG_KEYS);
+  const configBindings = filterKeys(node.attributes, (key) => USE_DATA_QUERY_CONFIG_KEYS.has(key));
```
apedroferreira, line 1231: We could abstract this to its our own `pick` function maybe?

Janpot: Removing it altogether

packages/toolpad-app/src/runtime/ToolpadApp.tsx
Janpot: To be honest, I'm not sure we should even have utility functions for this.

apedroferreira: Got it, yeah for this specific type of operation maybe just normal Javascript would be possible and be more type-safe.

PR #1271, opened 2022-11-03
new object explorer implementation based on react-arborist

Janpot:
Use react-arborist for an object explorer. Currently it looks like

Screenshot 2022-12-21 at 14 53 13

For the global scope explorer, to be able to size it correctly, and have virtualization, we'd need to present it as one tree, all categories as tree nodes included. This is currently stuck on [dynamic row heights](https://github.com/brimdata/react-arborist/issues/47), to make give the headers their padding.


igorgarbuz:
Hey @Janpot, thanks for this PR. Did you encounter some problems that prevent from merging it ?

Issue #1222, opened 2022-10-27
Improve Captcha UX


apedroferreira:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

The captcha we're using may block some users from creating an app and there's nothing those users can do about it.

How we can fix/improve this:
- [x] ~~Lower captcha score requirement so that most users are able to create apps~~ (let's keep it at 0.5 actually)
- [x] Fallback to a puzzle captcha if the invisible captcha fails (reCAPTCHA V2?) (https://developers.google.com/recaptcha/docs/faq#can-i-run-recaptcha-v2-and-v3-on-the-same-page)
- [x] Attach more information to the captcha error in Sentry when captcha validation fails. Possibly get rid of current captcha validation logs if they can be completely replaced by these Sentry errors.

PR #1272, opened 2022-11-03
Fallback to visible captcha if invisible captcha fails

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/1222

https://user-images.githubusercontent.com/10789765/199806778-f3207aeb-0192-4992-8ef3-8d14e020f882.mov

- Fallback to V2 captcha, which shows an "I'm not a robot" checkbox or a puzzle, if that fails
- Increase invisible (V3) captcha score as we have a fallback now
- Implement API errors with error codes
- Show response errors in API logs + remove captcha logs (brought back some of my old logic about logging response bodies for that)

Might split this up in separate PRs if it's too much.

Wasn't able to test yet if we get proper errors from the captcha validation in Sentry - Sentry isn't working from localhost right now as I'm getting a 403, will look more into it later.


packages/toolpad-app/src/apiErrors.ts
```diff
@@ -0,0 +1,17 @@
+import { ValueOf } from './utils/types';
+
+export const API_ERROR_CODES = {
```
Janpot, line 3: I'd create a separate file that exports separate consts so that we don't have to pull in this whole object (which will inevitably grow) if you need only one value.

packages/toolpad-app/src/apiErrors.ts
```diff
@@ -0,0 +1,17 @@
...
+  APP_NAME_EXISTS: 'APP_NAME_EXISTS',
```
Janpot, line 4: I'd consider namespacing them to distinguish them from errors created by node.js or other libraries
```suggestion
  APP_NAME_EXISTS: 'TOOLPAD_APP_NAME_EXISTS',
```

packages/toolpad-app/src/toolpad/Home/index.tsx
```diff
@@ -125,37 +135,67 @@ function CreateAppDialog({ onClose, ...props }: CreateAppDialogProps) {
...
+                  },
+                }
+              : {}),
+          },
+        ]);
+      } catch (error) {
+        if (config.recaptchaV2SiteKey && !hasShownRecaptchaCheckbox) {
+          if (error instanceof ApiError && error.code === API_ERROR_CODES.VALIDATE_CAPTCHA_FAILED) {
+            // Show captcha checkbox
+            const widgetId = grecaptcha.render(CAPTCHA_TARGET_ID, {
```
Janpot, line 185: We can't use a ref to a dom element here?

apedroferreira: You mean for the id? Yeah we could, it would be a bit better so will do!

packages/toolpad-app/src/apiErrors.ts
apedroferreira: It helps with types, but I'll see if I can export consts and still get good typing.

packages/toolpad-app/src/apiErrors.ts
```diff
@@ -0,0 +1,21 @@
+import { ValueOf } from './utils/types';
+
+// Do not export this object directly for easier tree-shaking
+const API_ERROR_CODES = {
```
Janpot, line 4: import * as apiErrorCodes from './apiErrorCodes';

export type ApiErrorCode = ValueOf<typeof apiErrorCodes>;

apedroferreira: i didn't know that worked, thanks! will try it out

apedroferreira:
@Janpot can I merge this one?

Janpot:
@apedroferreira I've been thinking a bit more about the `error.code` pattern and I want to generalize it so we can use it across the codebase (https://github.com/mui/mui-toolpad/pull/1286 is beginning to show need for error codes). I propose the following:

* We create a `./src/errorCodes.ts` which exports a constant for each error we want to identify in the codebase 
    ```tsx
    export const ERR_APP_EXISTS = 'TOOLPAD_ERR_APP_EXISTS'
    ```
* We don't subclass `Error`. It doesn't scale well and there is no real benefit, we just want to check the `code`. If we namespace it well there shouldn't be any conflicts. We'll could just do 

    ```tsx
    const error = new Error('BOOM!')
    error.code = ERR_APP_EXISTS
    // or
    const error = Object.assign(new Error('BOOM!'), { code: ERR_APP_EXISTS })
    ```

* We update `errorFrom` in `./src/utils/error` to return an `Error & { code?: unknown }`. And then use the follwing pattern
    ```tsx
    } catch (rawError) {
      const error = errorFrom(rawError)
      if (error.code === ERR_APP_EXISTS) {
        // ...
      }
    }

    // instead of

    } catch (error) {
      if (error instanceof SomeCustomeError && error.code === ERR_APP_EXISTS) {
        // ...
      }
    }
    ```
    This prevents us from having to subclass errors all the time. (we use `code?: unknown` because we want to discourage using this for anything else than `===` checking)

apedroferreira:
> @apedroferreira I've been thinking a bit more about the `error.code` pattern and I want to generalize it so we can use it across the codebase (#1286 is beginning to show need for error codes). I propose the following patterns:

Sounds good, I'll adjust to follow that approach. I maybe find it a bit less readable in some aspects, but it's ok as it has other good benefits like you mentioned!

As for creating an error with a code I prefer this option as it uses just one line so feels more concise?
```const error = Object.assign(new Error('BOOM!'), { code: ERR_APP_EXISTS })```

File: packages/toolpad-app/src/utils/useLocalStorageState.ts [hunk 5]
```diff
@@ -67,7 +67,7 @@ export default function useLocalStorageState<V>(
   const storedValue: V = React.useSyncExternalStore(
     (cb) => subscribe(key, cb),
     () => getSnapshot(key) ?? initialValue,
-    () => initialValue,
+    () => getSnapshot(key) ?? initialValue,
```
apedroferreira, line 70: Is there any con to using the value in storage if it already exists?
Starting the value as `null` when it already exists was giving me some problems as we're now storing the latest app when we create it.

apedroferreira:
@Janpot I've adjusted to the error codes you proposed.
@bharatkashyap I tried to use error codes like in your PR so there will be as little conflict as possible but lmk if I missed anything.

bharatkashyap:
@apedroferreira Looks great! One nit around the alignment of the captcha response in the form, should it be centred? 

Thanks for aligning this with the errors effort in #1286 as well.

apedroferreira:
Updated video:

https://user-images.githubusercontent.com/10789765/202486391-03f6f46c-3774-4d00-af25-59ed7299135d.mov

Preview: https://toolpad-pr-1272.onrender.com

apedroferreira:
I'm going to merge this if there's no more comments.
All feedback has been addressed but let me know if there's anything else and I can add later.

PR #1274, opened 2022-11-04
Fix windows and AZERTY shortcuts for undo redo

bytasv:
Follow up of https://github.com/mui/mui-toolpad/pull/1225 - add support for windows key shortcut

Partial of https://github.com/mui/mui-toolpad/issues/226


Janpot:
We're currently using Cmd+S in some places, does it have the same problem?

Any way we can abstract the platform specific behavior so that we don't have to write this everywhere we add a shortcut?

bytasv:
> We're currently using Cmd+S in some places, does it have the same problem?

> Any way we can abstract the platform specific behavior so that we don't have to write this everywhere we add a shortcut?

Those places are using `useShortcut`, so my assumption is that when we do `metaKey` we also wanna include `ctrlKey` - this change should kinda abstract this so places that use cmd + s should also work with ctrl + s now. Does that make sense?

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/EditorCanvasHost.tsx
```diff
@@ -155,8 +155,8 @@ export default React.forwardRef<EditorCanvasHostHandle, EditorCanvasHostProps>(
             return;
           }
 
-          const { code, metaKey, shiftKey } = event;
-          const undoShortcut = code === 'KeyZ' && metaKey;
+          const { code, metaKey, shiftKey, ctrlKey } = event;
+          const undoShortcut = code === 'KeyZ' && (metaKey || ctrlKey);
```
oliviertassinari, line 159: Personally, I would recommend using destructuring at least often as possible as it makes it harder to run quick regex search on the codebase, e.g. to search for the other locations where we use `event.code`. 

```suggestion
          const undoShortcut = event.code === 'KeyZ' && (event.metaKey || event.ctrlKey);
```

---

I had a look at https://github.com/mui/mui-toolpad/pull/1225#issuecomment-1302728271. I should have seen this one coming (I forgot, what happens when you are no longer day & night in the codebase). We can't use `event.code`. Per https://github.com/mui/mui-x/pull/3660/files#r793084149, our current shortcuts only work with a QWERTY layout. I have an AZERTY layout. It's also broken with the Dvorak layout.

I would recommend:
```suggestion
          const { code, metaKey, shiftKey, ctrlKey } = event;
          const undoShortcut = event.key.toLowerCase() === 'z' && (metaKey || ctrlKey);
```

or

```suggestion
          const { code, metaKey, shiftKey, ctrlKey } = event;
          const undoShortcut = String.fromCharCode(event.keyCode) === 'Z' && (metaKey || ctrlKey);
```

bytasv: Oh, interesting find! Thanks, I will fix this

bytasv:
@oliviertassinari please take a look again when you get a chance

> Personally, I would recommend using destructuring as least often as possible as it makes it harder to run quick regex search on the codebase

Regarding this - IMO that warrants a broader discussion that might be outside of scope for this PR. "as least often as possible" sounds pretty subjective, I'm not even sure how we would apply that in practice, it might either lead to inconsistent usage or simply we could go to other extreme of not using it anywhere 🤔

Janpot:
> I would be curious about the team's takes on the merit of this coding style

I don't have strong opinions about the general use of destructuring.

In this particular instance, personally, I find that it reduces the overall readability. a property named `code` has more context when attached to `event` than when free floating. It all feels a bit unnecessary here.

bytasv:
@oliviertassinari @Janpot destructuring removed, please take a look

oliviertassinari (approved):

PR #1280, opened 2022-11-04
Add improvement issue template

bytasv:
In order to differentiate between bug and features we might wanna have "improvement" type of issue


Janpot:
Can you give a few examples of "improvements" that can't be categorized under the nomer "bug" or "feature request"?

bytasv:
> In the past we've always opted for keeping these templates uniform across the organization. Should we reach consensus about this new template with the other teams first?

I wasn't aware of it, if/when we agree on this within Toolpad team, we could tag folks from core/x teams to get their feedback as well, would that make sense?

> Can you give a few examples of "improvements" that can't be categorized under the nomer "bug" or "feature request"?

At least everything under [enchancment](https://github.com/mui/mui-toolpad/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) label, even the label itself says "This is not a bug nor a new feature"

IMO most of UI/UX issues could also fall under "improvement" category as it's unlikely  a bug (especially when speaking about UX) and hardly a new feature - if we want to "improve" how our UI looks like in certain places.

Quickly looking at [design: ui](https://github.com/mui/mui-toolpad/issues?q=is%3Aissue+is%3Aopen+label%3A%22design%3A+ui%22) and [design: ux](https://github.com/mui/mui-toolpad/issues?q=is%3Aopen+is%3Aissue+label%3A%22design%3A+ux%22) labels - some are "new features" but some for me are "improvements" that could also use this category.

If these examples feel too broad, I could pick a list of exact issues that we currently have open

prakhargupta1:
I think it would be helpful. I have felt that many issues some of which Vytautas raised should fall into Enhancement category. Also, I was not aware that there was an enhancement label. But the label comes later, firstly I am asked to choose the type of the issue that I am creating.

Janpot:
We have to keep in mind that these templates are aimed towards external users. I'm not advocating against marking things as "enhancement", I'm just not sure it's a judgement that should be left to external users.

bytasv:
> We have to keep in mind that these templates are aimed towards external users.

I personally always use templates and the need for extra template comes from times when I have to decide whether I start with a bug or feature template, and for me personally even if this isn't a huge difference between existing templates it's just one less thing to think about when I'm about to create new issue. So while me or anyone else can add the extra label after creating an issue it would simply be one less action I need to care about.

> I'm just not sure it's a judgement that should be left to external users.

Is the concern that someone would choose incorrect template and say mark bug as improvement and that would add overhead for us to categorize issues later on? Could you clarify what are your concerns for introducing a new template?

bytasv:
@mnajdova @DanailH we see that neither core nor X repos have "improvement" issue template (seems that X uses [enhancement](https://github.com/mui/mui-x/issues?q=is%3Aissue+is%3Aopen+label%3Aenhancement) label) have you ever considered having a dedicated template for that? Do you see any downsides if we had that?

mnajdova:
> @mnajdova @DanailH we see that neither core nor X repos have "improvement" issue template (seems that X uses [enhancement](https://github.com/mui/mui-x/issues?q=is%3Aissue+is%3Aopen+label%3Aenhancement) label) have you ever considered having a dedicated template for that? Do you see any downsides if we had that?

The label is useful 👍 but in my opinion, I wouldn't add the new template. The main reason is, we are leaving the decision to the person opening the issue, when it is something we need to do (many times people even confuse bug/feature, for example they are trying to do something that is not supported and report it as bug, so I wouldn't add yet another option).

bytasv:
> for example they are trying to do something that is not supported and report it as bug, so I wouldn't add yet another option

@mnajdova could it also be because you can't decide properly between bug or feature if you are looking for improvement?

oliviertassinari (commented):
I would recommend keeping the issue template limited to feature requests and bug reports. I think that the point is to **force** people to pick a side. I believe that any changes can be categorized between a bug and a new feature. I feel that introduced enhancement for cases where we can't decide between, the two, but for the community, it's probably better to force them to pick like to match [VS Code](https://github.com/microsoft/vscode/issues/new/choose), [Vue](https://new-issue.vuejs.org/?repo=vuejs/core), [React](https://github.com/facebook/react/issues/new/choose), or [Next.js](https://github.com/vercel/next.js/issues/new/choose) are doing. If we really want to keep it more open-ended, renaming the new feature template to have a broader scope might be simpler (one less issue template to maintain)

bytasv:
Closing as per previous comments

PR #1281, opened 2022-11-04
Simplify menu actions

oliviertassinari:
I have found the existing labels confusing:

1. The name of the page/connection/etc. feels confusing. It adds more text to read which make it harder for me to understand what the action is doing:

Screenshot 2022-11-04 at 15 51 36

What felt better as a user:

Screenshot 2022-11-04 at 15 54 31

So I have removed these names.

2. We already know that I'm on Toolpad, I think that we can remove "MUI Toolpad"

Screenshot 2022-11-04 at 15 54 54

### DIFF
--- a/test/models/ToolpadHome.ts
+++ b/test/models/ToolpadHome.ts
@@ -36,7 +36,7 @@ export class ToolpadHome {
     this.createNewbtn = page.locator('button:has-text("create new")');
 
     this.newAppDialog = page.locator('[role="dialog"]', {
-      hasText: 'Create a App',
+      hasText: 'Create a new App',
     });
     this.newAppNameInput = this.newAppDialog.locator('label:has-text("name")');
     this.newAppTemplateSelect = this.newAppDialog.locator(
--- a/test/models/ToolpadHome.ts
+++ b/test/models/ToolpadHome.ts
@@ -36,7 +36,7 @@ export class ToolpadHome {
     this.createNewbtn = page.locator('button:has-text("create new")');
 
     this.newAppDialog = page.locator('[role="dialog"]', {
-      hasText: 'Create a new MUI Toolpad App',
+      hasText: 'Create a App',
     });
     this.newAppNameInput = this.newAppDialog.locator('label:has-text("name")');
     this.newAppTemplateSelect = this.newAppDialog.locator(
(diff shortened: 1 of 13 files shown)


Janpot (approved):
👍  Works for me

test/models/ToolpadHome.ts
```diff
@@ -36,7 +36,7 @@ export class ToolpadHome {
     this.createNewbtn = page.locator('button:has-text("create new")');
 
     this.newAppDialog = page.locator('[role="dialog"]', {
-      hasText: 'Create a new MUI Toolpad App',
+      hasText: 'Create a App',
```
Janpot, line 39:
```suggestion
      hasText: 'Create a new App',
```

apedroferreira (approved):

apedroferreira:
> Ok, I could fix the test, I had some struggles with them. A continuation of https://github.com/mui/mui-toolpad/pull/1162#issuecomment-1284167904:

1. Sounds good to me!
2. I usually run the VSCode extension first which is headless, and then if I really want to see what's going on I run the yarn script with the flag to see the browser. But I don't mind if the yarn script for tests defaults to non-headless mode, I kind of use both modes according to what I need.
3. If we seeded fixtures the tests would probably be a bit faster, especially locally, as we would have less API calls and page transitions? But if we don't want to seed, calling the API directly should still provide a good performance improvement, I would think it's worth trying out.
4. We can probably create an option for using any port, but default to 3000?

Janpot:
1. I always use the `--project chromium` flag for this
2. I always use the `--debug` flag for this
3. We don't have a rest API, we use an RPC implementation. Preferably we rely as little as possible on implementation details in the tests. Maybe we should make an exception for app creation. We will bring this option to production at some point, it's basically the "import/export application" feature, but we'll need to implement validation before we can expose that to users.
4. Does it work if you set `PLAYWRIGHT_TEST_BASE_URL=http://localhost:3002/` in your `.env` file?
5. Worth a try

prakhargupta1 (approved):

gerdadesign (approved):
agree with simplifying the text here! 👏

Janpot:
@oliviertassinari 

> The `TOOLPAD_ENABLE_CREATE_BY_DOM` flag got me curious why there is a difference with production. Looking at the source, it seems that we have a custom API call only for the test on the client side. I think that calling the API directly in the test would be simpler.

I'm updating the tests in https://github.com/mui/mui-toolpad/pull/1384

Issue #1176, opened 2022-10-18
Adding parameter with same name clears earlier one


bharatkashyap:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

https://user-images.githubusercontent.com/19550456/196426758-0269b661-bc0f-4318-94c9-df0cca513a36.mov

### Expected behavior 🤔

The expectation is that the earlier parameter would remain bound and not be replaced by the unbound new one

### Steps to reproduce 🕹

Steps:

1. Go to https://master--toolpad.mui.com/_toolpad/app/cl91ik41i0454axrr578xjjrp/pages/cl27arb1i00043g693pzhex2r
2. Click on the `randomDogImage` query
3. Add a new parameter that begins with "subBreed"
4. See the existing parameter get removed

### Context 🔦

I was trying to add a new parameter called "subBreedAll" but kept failing to do so

PR #1282, opened 2022-11-04
Fix adding parameter with duplicate name

Janpot:
Fixes https://github.com/mui/mui-toolpad/issues/1176


apedroferreira (approved):
The code changes look good - I wasn't able to run the preview though but I guess it should be fixed by using entries, looks like a good solution.

apedroferreira:
Just tested in the preview, looks good too!

PR #1285, opened 2022-11-04
Send custom GA event when new app is created

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/997

### DIFF
--- a/packages/toolpad-app/src/utils/ga.ts
+++ b/packages/toolpad-app/src/utils/ga.ts
@@ -1,5 +1,6 @@
 import type { NextWebVitalsMetric } from 'next/app';
 import config from '../config';
+import { AppTemplateId } from '../types';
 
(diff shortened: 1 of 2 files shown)


packages/toolpad-app/src/utils/ga.ts
```diff
@@ -17,3 +18,12 @@ export const reportWebVitalsToGA = ({ id, label, name, value }: NextWebVitalsMet
     });
   }
 };
+
+export const sendAppCreatedEvent = (name: string, templateId?: AppTemplateId): void => {
+  if (config.gaId) {
+    window.gtag('event', 'app_created', {
```
Janpot, line 24: Would it make sense to namespace those `toolpad_app_created`?

apedroferreira: Why not, would be a good standard for custom events, will add now.

PR #1286, opened 2022-11-05
Unify app editor options menus

### LINKED ISSUE #1214: Unify app options menus
### Duplicates

### Latest version

### Summary 💡

I would expect that the options menu associated with the app should be the same on the Home page (containing the list of apps) and the editor page. This would also involve moving the "View DOM" option into the menu, possibly renamed as "Export" (linked to https://github.com/mui/mui-toolpad/issues/1040)

### Examples 🌈

Screenshot 2022-10-26 at 12 25 48 PM

Screenshot 2022-10-26 at 12 28 34 PM

### Motivation 🔦

_No response_

bharatkashyap:
- Closes #1214 

- Visual demonstration 

Screenshot 2022-11-05 at 3 06 16 PM

Screenshot 2022-11-05 at 3 06 34 PM


packages/toolpad-app/src/server/data.ts
```diff
@@ -529,14 +529,19 @@ export async function loadRenderTree(appId: string, version: VersionOrPreview =
   return appDom.createRenderTree(await loadDom(appId, version));
 }
 
-export async function duplicateApp(id: string, name: string): Promise<AppMeta> {
+export async function duplicateApp(
+  id: string,
+  name: string,
+  suppliedName?: boolean,
```
Janpot, line 535: I don't understand this parameter.

prakhargupta1:
I have checked the UX of all menu options, and it looks good to me.

packages/toolpad-app/src/toolpad/AppEditor/PagePanel.tsx
```diff
@@ -95,16 +50,33 @@ export default function PagePanel({ appId, className, sx }: ComponentPanelProps)
...
+            app={app}
+            onRename={handleRename}
+            onDuplicate={() => setDuplicateApp(app)}
+            onDelete={onDelete}
+            dom={dom}
+          />
+        ) : null}
       </Box>
       <Divider />
+      <AppDeleteDialog app={deletedApp} onClose={() => setDeletedApp(null)} redirectOnDelete />
```
Janpot, line 74: Would it make sense to include this in the menu as well?

packages/toolpad-app/src/server/data.ts
bharatkashyap: This is essentially to allow reusing the function in two scenarios:

1. When the user doesn't supply a name when duplicating (duplicating in the list view)
2. When we have a user supplied name while duplicating the app (duplicating in the editor view)

bharatkashyap: On second thought, we can unify both these experiences and then get rid of the need of having to do this.

packages/toolpad-app/src/toolpad/AppEditor/PagePanel.tsx
bharatkashyap: Agreed, we can move both the `AppDeleteDialog` and `AppDuplicateDialog` inside the options menu and simplify the `duplicateApp` function as well

packages/toolpad-app/src/server/data.ts
```diff
@@ -531,16 +531,26 @@ export async function loadRenderTree(appId: string, version: VersionOrPreview =
...
-    duplicateCount === 0 ? `${name} (copy)` : `${name} (copy ${duplicateCount + 1})`;
-  const newApp = await createApp(duplicateName, {
-    from: {
-      kind: 'dom',
-      dom,
-    },
-  });
-  return newApp;
+  try {
+    const newApp = await createApp(name, {
```
Janpot, line 535: Let's extract and reuse the options?
```tsx
  const appInit: CreateAppOptions = {
    from: {
      kind: 'dom',
      dom,
    },
  };

    // ...
    const newApp = await createApp(name, appInit);
```

packages/toolpad-app/src/server/data.ts
```diff
@@ -531,16 +531,26 @@ export async function loadRenderTree(appId: string, version: VersionOrPreview =
...
+      from: {
+        kind: 'dom',
+        dom,
+      },
+    });
+    return newApp;
+  } catch (error) {
+    const duplicateCount = await prismaClient.app.count({
```
Janpot, line 543: This needs to only run when the error is the unique constraint of the name column. All other errors need to be rethrown. We wouldn't want this to run when something like a network error has happened

packages/toolpad-app/src/runtime/evalJsBindings.ts
```diff
@@ -2,8 +2,7 @@ import { set } from 'lodash-es';
 import evalExpression from '../utils/evalExpression';
 import { mapValues } from '../utils/collections';
 import { errorFrom } from '../utils/errors';
-
-const TOOLPAD_LOADING_MARKER = '__TOOLPAD_LOADING_MARKER__';
```
Janpot, line 6: This constant is not used outside of this file, let's keep it local?

packages/toolpad-app/src/server/data.ts
```diff
@@ -231,7 +232,9 @@ export async function createApp(name: string, opts: CreateAppOptions = {}): Prom
   }
 
   if (await prismaClient.app.findUnique({ where: { name: appName } })) {
-    throw new Error(`An app named "${name}" already exists.`);
+    throw new Error(`An app named "${name}" already exists.`, {
+      cause: TOOLPAD_UNIQUE_CONSTRAINT_VIOLATION,
```
Janpot, line 236: * We should only use `Error` instances in `cause`.
* Add a `code` property you import from `./src/errorCodes.tsx`:
    ```tsx
    export const ERR_APP_EXISTS = `TOOLPAD_ERR_APP_EXISTS`
    ```
    See also https://github.com/mui/mui-toolpad/pull/1272#issuecomment-1312401592

packages/toolpad-app/src/server/data.ts
```diff
@@ -531,16 +534,25 @@ export async function loadRenderTree(appId: string, version: VersionOrPreview =
...
       dom,
     },
-  });
-  return newApp;
+  };
+  try {
+    const newApp = await createApp(name, appFromDom);
+    return newApp;
+  } catch (error) {
+    if (error instanceof Error && error.cause === TOOLPAD_UNIQUE_CONSTRAINT_VIOLATION) {
```
Janpot, line 547: `TOOLPAD_UNIQUE_CONSTRAINT_VIOLATION` is an implementation detail. If we're adding our own error codes, let's do `ERR_APP_EXISTS`

packages/toolpad-app/src/constants.ts
```diff
@@ -10,3 +10,5 @@ export const TOOLPAD_TARGET_PRO = 'PRO';
 export const DOCUMENTATION_URL = 'https://mui.com/toolpad/getting-started/overview/';
 export const DOCUMENTATION_INSTALLATION_URL =
   'https://mui.com/toolpad/getting-started/installation/';
+
+export const TOOLPAD_UNIQUE_CONSTRAINT_VIOLATION = 'UNIQUE_CONSTRAINT_VIOLATION';
```
Janpot, line 14:
```suggestion

```

packages/toolpad-app/src/errorCodes.ts
```diff
@@ -1,2 +1,2 @@
-export const ERR_APP_NAME_EXISTS = 'TOOLPAD_APP_NAME_EXISTS';
+export const ERR_APP_EXISTS = 'TOOLPAD_APP_NAME_EXISTS';
```
Janpot, line 1:
```suggestion
export const ERR_APP_EXISTS = 'TOOLPAD_APP_EXISTS';
```

Janpot (approved):

Janpot:
One last nit in https://github.com/mui/mui-toolpad/pull/1286#discussion_r1031144001

PR #1294, opened 2022-11-06
Upgrade to next 13

Janpot:
Seems like core [wants to move fast](https://github.com/mui/material-ui/pull/35001) on this. So let's try to find the blockers for us to move to Next.js 13 as well.

- [x] Toolpad app OOM issues during build. (https://github.com/vercel/next.js/issues/40216 ?)
- [x] Toolpad app integration tests
  - [x] why are they about twice as slow now? middleware seems to have a big impact since 12.2, fixed in https://github.com/vercel/next.js/pull/42547. Confirmed on canary, 13.0.3 will contain the fix
  - [x] frameLocator is misbehaving in firefox. Tests become very flaky once middleware gets slow
  - [x] Failing duplication tests around custom components: https://github.com/mui/mui-toolpad/pull/1306
- [x] hydration mismatch when opening the docs
- [x] To do before merge:
  - [x] Upgrade `@mui/monorepo` dependency to master when https://github.com/mui/material-ui/pull/35001 is merged
  - [x] ~playwright 1.28~ not crucial for this PR
  - [x] next.13.0.3

### memory usage (`test_static`):
before (master):

Screenshot 2022-11-07 at 10 13 54

after:
Screenshot 2022-11-07 at 10 14 01

PR #1297, opened 2022-11-07
Upgrade to next 13 for docs only

Janpot:
Isolating [next 13 upgrade](https://github.com/mui/mui-toolpad/pull/1294) to docs only to unblock core from upgrading


bharatkashyap (commented):
👍

Issue #568, opened 2022-06-16
Switch component


oliviertassinari:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Make it possible to easily replace one component for another.

### Examples 🌈

In pro-code, you can simply rename your component while keeping the same props.

<img width="215" alt="Screenshot 2022-06-17 at 00 44 38" src="https://user-images.githubusercontent.com/3165635/174190471-7a4d8454-84d7-4860-98c6-bcdc594b7fdf.png">
 
I would expect a similar feature in Toolpad.

### Motivation 🔦

I have used a TextField but I really should have used a Select. I want to keep the same props.

PR #1298, opened 2022-11-07
Add Text component to merge Typography, Markdown, Link

### LINKED ISSUE #756: Support Markdown component
### Duplicates

### Latest version

### Summary 💡

The TextField component is quite limited. By default, I would expect to be able to use Toolpad's text component like I use Notion 😁

### Motivation

In https://master--toolpad.mui.com/_toolpad/app/cl4hla83p01949xoizo5uxf2a/editor/pages/cl6i430vg00003b6ghpxth1dz, I had to mix two different components, but they are not part of the same layout flow, they should:

Screenshot 2022-08-06 at 19 53 46

### Examples 🌈
https://retool.com/blog/text-v2-app-documentation/

### LINKED ISSUE #1310: Improve markdown component's prop editor
### Duplicates

### Latest version

### Summary 💡

Currently, the prop editor is a single line text box. Most of the time the content will be multiline. This issue is about how to handle this experience. Some options:

1. We allow expanding the side panel and make the editor big.
2. We create an indicator that expands the text box to a bigger dialog box (like retool)

Note: As this experience will be introduced to other components in the future, we should aim to keep it standard.

### Examples 🌈

_No response_

### Motivation 🔦

_No response_

### LINKED ISSUE #1312: Typography to be combined with markdown component
### Duplicates

### Latest version

### Summary 💡

Typography is the most basic form of adding text/headings to a Toolpad app. This issue is about if we should merge it with the markdown component. We should do it after we organize the right pane better.

### Examples 🌈

Retool's text component also supports markdown. They don't have a separate Markdown component.
Dynaboard has a single text component with different modes: text, link, and markdown.

### Motivation 🔦

_No response_

bharatkashyap:
- Closes #756 
- Closes #1310 
- Closes #1312 

Holding off on adding HTML support currently

- Visual demonstration 

#### Update: Allowing shared state between "modes" and removing the separate section for Mode:

https://user-images.githubusercontent.com/19550456/202044224-b3efde5a-3bf2-4507-aba5-ead87503343e.mov

Screenshot 2022-11-11 at 11 27 54 PM

https://user-images.githubusercontent.com/19550456/201400638-d8f60f70-169a-45df-a454-9fd7f1ab884b.mov


Janpot:
Is there a case to be made to make the typography component support markdown? e.g. suppose I added a Typography component, wrote my text, and then decide I want a link on this text. Can I just "upgrade" my Typography, or do I have to restart with a Markdown component?

bharatkashyap:
> Is there a case to be made to make the typography component support markdown? e.g. suppose I added a Typography component, wrote my text, and then decide I want a link on this text. Can I just "upgrade" my Typography, or do I have to restart with a Markdown component?

In this case, should we then merge the Typography and Markdown components, and allow turning the Typography component into markdown through a setting?

Janpot:
> and allow turning the Typography component into markdown through a setting?

Not that we necessarily need to copy it, but we noticed that's how dynaboard does it. Also mentioned by @apedroferreira before https://github.com/mui/mui-toolpad/pull/1234#pullrequestreview-1162339324.

bharatkashyap:
For me, Link should remain separate while merging Markdown and Typography into a common "Text" component makes sense.

apedroferreira:
Just tested it, I was kind of expecting that we could write Markdown on an expandable text area, maybe in it's own separate modal?

About allowing writing HTML, that would be cool but maybe should be it should be a separate feature? I guess that if we used some library to escape the HTML we should be ok in terms of security.

prakhargupta1:
After our yesterday's call, I realized that merging Typography and Markdown may make sense, but we can do this later. We can even look into other ways of reaching the same outcome. Like: https://github.com/mui/mui-toolpad/issues/568

Regarding supporting a better prop editor than a Textfield, I think we must do it, but separately.
Supporting HTML should also be a separate component. I'll create new issues and link them here.

apedroferreira (approved):
I like the new UX, looks great!

packages/toolpad-app/src/appDom/migrations/index.ts
```diff
@@ -1,9 +1,10 @@
 import invariant from 'invariant';
 import v1 from './v1';
 import v2 from './v2';
+import v3 from './v3';
 import * as appDom from '..';
 
-const versions = [v1, v2];
+const versions = [v1, v2, v3];
```
Janpot, line 7: There already is a v3, I had just forgotten to include it 🤦 
https://github.com/mui/mui-toolpad/pull/1359

packages/toolpad-app/package.json
```diff
@@ -96,12 +96,14 @@
     "react": "^18.2.0",
     "react-devtools-inline": "^4.26.1",
     "react-dom": "^18.2.0",
+    "react-draggable": "^4.4.5",
```
Janpot, line 99: I think I prefer to keep it simple until we think a bit more hollistically about the UX of the properties/binding editors. Perhaps it should be a popver, like retool. Or an input in the component panel, like dynaboard? Or inline in the page, like notion? Or in a popup with a live preview? Or something else?

packages/toolpad-components/src/Text.tsx
```diff
@@ -0,0 +1,120 @@
...
+import {
+  Skeleton,
+  Typography as MuiTypography,
+  TypographyProps as MuiTypographyProps,
+  Link as MuiLink,
+  LinkProps as MuiLinkProps,
+  styled,
+} from '@mui/material';
+import { createComponent, TOOLPAD_COMPONENT_MODE_PROPERTY } from '@mui/toolpad-core';
+import ReactMarkdown from 'react-markdown';
```
Janpot, line 11: I wonder whether we should consider lazy loading this

packages/toolpad-components/src/Text.tsx
```diff
@@ -0,0 +1,120 @@
...
+  markdown,
+  href,
+  loading,
+  [TOOLPAD_COMPONENT_MODE_PROPERTY]: mode,
+  sx,
+  ...rest
+}: TextProps) {
+  switch (mode) {
+    case 'markdown':
+      return loading ? (
```
Janpot, line 39: Will this result in layout shift? How dies the skeleton know which height to render? If we can't guarantee a loading state without layout shift, we shouldn't show a loading state imo

packages/toolpad-components/src/Text.tsx
```diff
@@ -0,0 +1,120 @@
...
+  LinkProps as MuiLinkProps,
+  styled,
+} from '@mui/material';
+import { createComponent, TOOLPAD_COMPONENT_MODE_PROPERTY } from '@mui/toolpad-core';
+import ReactMarkdown from 'react-markdown';
+import remarkGfm from 'remark-gfm';
+
+type BaseProps = MuiLinkProps | MuiTypographyProps;
+interface TextProps extends Omit<BaseProps, 'children'> {
+  [TOOLPAD_COMPONENT_MODE_PROPERTY]: string;
```
Janpot, line 16: Why not just `mode`?

packages/toolpad-core/src/types.ts
```diff
@@ -153,6 +154,8 @@ export type PropValueTypes<K extends string = string> = Partial<{
   [key in K]?: PropValueType;
 }>;
 
+export type ComponentProps = Record<string, unknown>;
+
 export interface ArgTypeDefinition<V = unknown> {
```
Janpot, line 159: Can we do
```suggestion
export interface ArgTypeDefinition<V = unknown, P = unknown> {
```
and pass `P` in
```tsx
export type ArgTypeDefinitions<P = any> = {
  [K in keyof P & string]?: ArgTypeDefinition<P[K], P>;
};
```
?

packages/toolpad-app/src/toolpad/propertyControls/Markdown.tsx
```diff
@@ -0,0 +1,66 @@
...
+  PaperProps,
+  Skeleton,
+} from '@mui/material';
+import * as React from 'react';
+import Draggable from 'react-draggable';
+import type { EditorProps } from '../../types';
+import useShortcut from '../../utils/useShortcut';
+import lazyComponent from '../../utils/lazyComponent';
+
+function DraggablePaperComponent(props: PaperProps) {
```
apedroferreira, line 18: I had missed this change last time I reviewed.
What does this add, is it to make the markdown modal draggable?
If so, and if that's a feature worth adding a new dependency for, maybe it should be a more global behavior that can be easily added to other similar dialogs?

packages/toolpad-app/package.json
bharatkashyap: Fair enough, this can be kept out of this PR

packages/toolpad-app/src/toolpad/propertyControls/Markdown.tsx
bharatkashyap: Agreed, keeping this out of here for now. See: https://github.com/mui/mui-toolpad/pull/1298#discussion_r1030314930

packages/toolpad-components/src/Text.tsx
bharatkashyap: For me, the utility of being able to show a loading state would trump the issue of it causing a layout shift. I would propose keeping the loading state, creating an issue for it and fixing it in an upcoming PR.

packages/toolpad-components/src/Text.tsx
bharatkashyap: Agreed. `[TOOLPAD_COMPONENT_MODE_PROPERTY]` was envisaged to become a property that would reflect "rendering modes" across components. We don't need it since "mode" is just a prop now.

packages/toolpad-components/src/Text.tsx
bharatkashyap: Agreed. I imagine it would create a similar layout shift problem as in https://github.com/mui/mui-toolpad/pull/1298#discussion_r1028148876 since the `fallback` element would have a fixed height.

Janpot: The runtime is already wrapped with Suspense wrapper to load components https://github.com/mui/mui-toolpad/blob/6b4a5d6cdcfd733c539980c13dcf9323a05e9508/packages/toolpad-app/src/runtime/ToolpadApp.tsx#L957
So I don't think there would be layout shift, it would just be incorporated in the loading of the components

packages/toolpad-core/src/types.ts
bharatkashyap: Initialising `P` as `unknown` creates a bunch of `P is not assignable to unknown` errors. 

Initialising `P` as `any`
```ts 
export interface ArgTypeDefinition<V = unknown, P = any>
``` 
fixes things, since `P` is initialised as `any` here: 

```ts
export type ArgTypeDefinitions<P = any> = {
  [K in keyof P & string]?: ArgTypeDefinition<P[K], P>;
};
```

Janpot: Well yes, `any` can "fix" everything, that's why avoid using it 😄

Janpot (approved):
👌

oliviertassinari:
It seems to have broken the existing apps, e.g. https://master--toolpad.mui.com/deploy/cl4hla83p01949xoizo5uxf2a/pages/f303n5n?userId=6948633456156&userEmail=olivier@mui.com

Screenshot 2022-12-14 at 15 44 18

bharatkashyap:
> https://master--toolpad.mui.com/deploy/cl4hla83p01949xoizo5uxf2a/pages/f303n5n?userId=6948633456156&userEmail=olivier@mui.com

It shouldn't have broken the existing apps, the migration should have made the previously created apps compatible

packages/toolpad-components/package.json
```diff
@@ -31,7 +31,8 @@
   "dependencies": {
     "@mui/material": "^5.10.17",
     "@mui/toolpad-core": "^0.0.30",
-    "@mui/x-data-grid-pro": "^5.17.15"
+    "@mui/x-data-grid-pro": "^5.17.15",
+    "react-markdown": "^8.0.3"
```
oliviertassinari, line 35: Not sure about this dependency: https://github.com/mui/material-ui/issues/12290#issuecomment-453930042

Issue #1213, opened 2022-10-25
https://demo.toolpad.io/ data fetching feels limited


oliviertassinari:
Related to #592

I was hoping I could rebuild most of the mini-apps that I had created on the staging environment https://master--toolpad.mui.com/ in the demo environment: https://demo.toolpad.io/ but it's not the case. This makes me wonder:

1. Why not have the default fetch query? Right now, when I click on adding a new query, the select is empty.

<img width="577" alt="Screenshot 2022-10-25 at 23 41 50" src="https://user-images.githubusercontent.com/3165635/197887573-05304608-f1cb-45aa-8933-e2696689a4d8.png">

2. Why not have a fetch function? If we are worried about the amplification network power that Toolpad can provide to bad actors, then we could limit the number of `fetch` calls that can be made inside the function. I think that we would also need to limit the number of queries that can be added to the page, because why close one door (fetch) when the other is open (multi queries)? 

<img width="574" alt="Screenshot 2022-10-25 at 23 42 32" src="https://user-images.githubusercontent.com/3165635/197887700-08788728-e39d-4ea1-ade7-a073779a73ac.png">

3. Why not have PostgreSQL? If there is a good reason for it, I think that we should allow people to pick it and then explain why it's not available. Currently, it feels like the feature doesn't exist.

<img width="373" alt="Screenshot 2022-10-25 at 23 44 52" src="https://user-images.githubusercontent.com/3165635/197888069-9fdfcd40-c65a-4f1b-b94c-06ef6273b1af.png">

4. Why not have Google Sheet? Same to 3.

<img width="381" alt="Screenshot 2022-10-25 at 23 46 16" src="https://user-images.githubusercontent.com/3165635/197888270-c52a5d1f-61b1-48ed-a78d-f1772510306a.png">

prakhargupta1:
I agree, the Functions and Fetch should be available by default like we show in the staging environment. We had purposely limited the data sources as the intent was to give a basic idea of Toolpad. Should we disable Connection button and through an info icon show that we support 2 more data sources in the self-hosted version? We can show Fetch default, Functions default and Fake movies API directly when user will go to create a query.
Alternatively, we can support all data supports if there are no major concerns.

Also, as there is no Home view, users will create 3-4 apps in the first session itself (until they realize they can't see the apps). Either we should show apps for that session or an info saying please save this URL if you want to open this again.

We can also show Self-host CTA in the header bar.

Janpot:
[...]

PR #1304, opened 2022-11-07
Allow working on latest used app from demo home page

apedroferreira:
From https://github.com/mui/mui-toolpad/issues/1213
Allow demo users to go back to the latest app they were working on from the home page.
Used `localStorage` to save the user's latest app id and name.

https://user-images.githubusercontent.com/10789765/200396072-87874881-2a76-4a2b-bb5d-4e2816b0854d.mov

### DIFF
--- a/packages/toolpad-app/src/toolpad/AppEditor/PagePanel.tsx
+++ b/packages/toolpad-app/src/toolpad/AppEditor/PagePanel.tsx
@@ -22,6 +22,8 @@ import client from '../../api';
 import { useDom } from '../DomLoader';
 import JsonView from '../../components/JsonView';
 import useMenu from '../../utils/useMenu';
+import useLocalStorageState from '../../utils/useLocalStorageState';
+import { LatestStoredAppValue, TOOLPAD_LATEST_APP_KEY } from '../../storageKeys';
 
 const PagePanelRoot = styled('div')({
   display: 'flex',
@@ -81,6 +83,20 @@ export interface ComponentPanelProps {
 
 export default function PagePanel({ appId, className, sx }: ComponentPanelProps) {
   const { data: app, isLoading } = client.useQuery('getApp', [appId]);
+
+  const [, setLatestStoredApp] = useLocalStorageState<LatestStoredAppValue>(
+    TOOLPAD_LATEST_APP_KEY,
+    null,
+  );
+
+  React.useEffect(() => {
+    if (app) {
+      setLatestStoredApp({
+        appId,
+        appName: app.name,
+      });
+    }
+  }, [app, appId, setLatestStoredApp]);
 
   return (
     <PagePanelRoot className={className} sx={sx}>
--- a/packages/toolpad-app/src/toolpad/Home/index.tsx
+++ b/packages/toolpad-app/src/toolpad/Home/index.tsx
@@ -63,6 +63,7 @@ import config from '../../config';
 import { AppTemplateId } from '../../types';
 import { errorFrom } from '../../utils/errors';
 import { sendAppCreatedEvent } from '../../utils/ga';
+import { LatestStoredAppValue, TOOLPAD_LATEST_APP_KEY } from '../../storageKeys';
 
 export const APP_TEMPLATE_OPTIONS: Map<
   AppTemplateId,
@@ -124,6 +125,11 @@ function CreateAppDialog({ onClose, ...props }: CreateAppDialogProps) {
     },
   });
 
+  const [latestStoredApp] = useLocalStorageState<LatestStoredAppValue>(
+    TOOLPAD_LATEST_APP_KEY,
+    null,
+  );
+
   const isFormValid = Boolean(name);
 
   return (
@@ -162,7 +168,7 @@ function CreateAppDialog({ onClose, ...props }: CreateAppDialogProps) {
         <DialogTitle>Create a new MUI Toolpad App</DialogTitle>
         <DialogContent>
           {config.isDemo ? (
-            <Alert severity="warning" sx={{ mb: 2 }}>
+            <Alert severity="warning" sx={{ mb: 1 }}>
               <AlertTitle>For demo purposes only!</AlertTitle>
               Your application will be ephemeral and may be deleted at any time.
             </Alert>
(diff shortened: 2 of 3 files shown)


packages/toolpad-app/src/toolpad/AppEditor/PagePanel.tsx
```diff
@@ -82,6 +84,20 @@ export interface ComponentPanelProps {
...
   const { data: app, isLoading } = client.useQuery('getApp', [appId]);
 
+  const [latestStoredApp, setLatestStoredApp] = useLocalStorageState<LatestStoredAppValue>(
+    TOOLPAD_LATEST_APP_KEY,
+    null,
+  );
+
+  React.useEffect(() => {
+    if (app && (!latestStoredApp || appId !== latestStoredApp.appId)) {
+      setLatestStoredApp({
```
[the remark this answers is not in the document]
apedroferreira: I wanted it to work also if you load an app, not just when you create it.

packages/toolpad-app/src/toolpad/Home/index.tsx
```diff
@@ -211,28 +219,47 @@ function CreateAppDialog({ onClose, ...props }: CreateAppDialogProps) {
...
-                href="https://policies.google.com/privacy"
-                underline="none"
-                target="_blank"
-                rel="noopener noreferrer"
+          {config.isDemo && latestStoredApp ? (
+            <Box mt={1} sx={{ display: 'flex', flexDirection: 'column', alignItems: 'center' }}>
+              <Typography variant="subtitle2" textAlign="center">
+                OR
+              </Typography>
+              <Button
```
bytasv, line 227: This might be subjective, but my initial impression when I saw the popup was that this action was quite "mute", like very easy to miss. If our intention is to lead people to previously created app, maybe we should use a little bit more visual style? I.e. `outlined` variant might be a bit easier to recognise as an action 🤔

apedroferreira: I didn't want to make it too strong not to look like the submit button of a form, and trying to make both creating a new app or going to the previous one have equal weight...
If we want users to do one of these 2 things more than the other though we could try different things visually.

apedroferreira: Maybe an outlined button that's a bit narrower than the form will work, I can try that.

packages/toolpad-app/src/toolpad/AppEditor/PagePanel.tsx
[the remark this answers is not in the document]
bytasv: But before loading it you would have to create it, no? If I understand correctly this would only benefit someone who hasn't yet closed the tab and received this update - like they do still have Toolpad open in their browser, otherwise if you come to Toolpad after closing the tab your app would be gone and you would have to go through creation process. If that's the only case then IMO it would be more straight forward and less complex to do that during creation

bytasv:
One thing that I tested was simulating a case when app no longer exists (just by changing local storage id manually). That could happen if we decide to prune db on some regular schedule.

In that case we get this message 🤔 
image

However this might be considered as a follow up improvement for the app in general as this would happen for any app that's not found

bytasv (approved):
Tested it works as expected ✅ 

Some non-blocking feedback left

packages/toolpad-app/src/toolpad/AppEditor/PagePanel.tsx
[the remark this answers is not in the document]
apedroferreira: The latest app isn't gone if you close the tab, it's saved in `localStorage`, so the last app that you were working on (not necessarily the last app you created) will always be there to go back to when you load the demo home page.

In the context of how the demo works right now there's no big difference, but I think that if we set this value when an app loads then we can always rely on it in the future to get the latest app the user was working on, for any future use case.
In my opinion there's no benefit in setting this when the user creates the app as opposed to when the app loads.

apedroferreira:
> However this might be considered as a follow up improvement for the app in general as this would happen for any app that's not found

Yeah, I guess we should fix that on a more global level as it would probably not be optimal to always check if the app exists.

packages/toolpad-app/src/toolpad/AppEditor/PagePanel.tsx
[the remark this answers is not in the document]
Janpot: I'm hesitant for this. Imagine having two tabs open with different apps, it  makes these two tabs fight with each other over who's the latest app. I'm also more inclined towards only storing the last created app here.

packages/toolpad-app/src/toolpad/Home/index.tsx
```diff
@@ -124,6 +125,11 @@ function CreateAppDialog({ onClose, ...props }: CreateAppDialogProps) {
     },
   });
 
+  const [latestStoredApp] = useLocalStorageState<LatestStoredAppValue>(
```
Janpot, line 128: This hook has proven to be so useful already. The only downside I see at the moment is that the type may change from one version to the next. typescript won't help here to prevent or detect that. I'm thinking of adding serialization options with `superstruct` to guard what's stored and provide the types

packages/toolpad-app/src/toolpad/AppEditor/PagePanel.tsx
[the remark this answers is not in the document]
apedroferreira: > Imagine having two tabs open with different apps

Wouldn't that be an uncommon use case though, and the only negative thing that could happen is that a recent app would be suggested instead of really the latest one that was used?

Anyway I can move it to when the app is created, it should be virtually the same for now.

packages/toolpad-app/src/toolpad/Home/index.tsx
apedroferreira: Would be really cool to be able to pass in a key and get the correct expected types, it would probably require some thinking on how to implement it though.

Janpot: I roughly had someting like this in mind
```tsx
import { object, number, string } from 'superstruct'
import useLocalStorageState from './utils/useLocalStorageState';

const LastCreatedApp = object({
  id: string(),
  lastSavedAt: number()
})

// ...

  const [lastCreatedApp, setLastCreatedApp] = useLocalStorageState('my-article', null, LastCreatedApp)
```

apedroferreira: Could be, or might be a bit heavy to pass in a schema just for typing, especially if we use the hook in multiple places...

Maybe we could just have an interface / type object with type definitions for keys and values and pass that to useLocalStorageState as a type parameter? And maybe an abstraction called `useToolpadLocalStorageState` that already included those types? Or that might be too much abstraction... Just some rough ideas anyway.

I guess we could discuss this later as a separate issue.

Janpot: It's not just for typing, it needs to actually validate in runtime that whatever came from localstorage is compatible with with that schema

PR #1307, opened 2022-11-08
Convert scripts to ESM

Janpot:
Preparing for incoming monorepo update https://github.com/mui/material-ui/pull/35036

### To do before merge:
- [x] reset monorepo dependency


docs/package.json
```diff
@@ -74,6 +74,8 @@
     "react-is": "^18.2.0",
     "react-router": "^6.3.0",
     "react-router-dom": "^6.3.0",
+    "react-runner": "^1.0.3",
```
Janpot, line 77: These were also recently introduced since the last version

michaldudak (approved):

PR #1308, opened 2022-11-08
Undo redo tests

bytasv:
Testing for undo/redo actions https://github.com/mui/mui-toolpad/issues/226


test/integration/editor/index.spec.ts
```diff
@@ -43,7 +43,7 @@ test('can move elements in page', async ({ page, browserName }) => {
   const app = await homeModel.createApplication({ dom: domInput });
   await editorModel.goto(app.id);
 
-  await editorModel.pageRoot.waitFor();
+  await editorModel.waitForOverlay();
```
bytasv, line 46: @apedroferreira I finally figured out why specs failed on my machine. It seems that my machine might be executing steps "too fast" or at least that's the main assumption. I also find some relevant discussions on playwright repo https://github.com/microsoft/playwright/issues/14752

For some reason CI and probably for the rest of us specs executed at the "right speed" and this issue wasn't visible. On my machine the overlay might not have been "ready" to be clicked on which resulted on "misclicks" and specs not working.

test/integration/playwright.config.ts
```diff
@@ -15,10 +15,12 @@ const config: PlaywrightTestConfig = {
       name: 'chromium',
       use: { ...devices['Desktop Chrome'] },
     },
-    {
-      name: 'firefox',
-      use: { ...devices['Desktop Firefox'] },
-    },
+    !process.env.SKIP_FF
```
bytasv, line 18: Added this for debugging purposes, it was a little bit annoying to have FF executed specs too when I didn't really need them, so it's a small DX optimisation that you can opt-in

bytasv:
@mui/toolpad could I get a review on this one?

test/integration/playwright.config.ts
apedroferreira: Can't you just use the flags that Playwright already provides such as `--project=chromium`? Have you tried that?

packages/toolpad-app/src/toolpad/AppEditor/AppEditorShell.tsx
```diff
@@ -153,12 +153,12 @@ export default function AppEditorShell({ appId, ...props }: ToolpadShellProps) {
     <ToolpadShell
       actions={
         <Stack direction="row" gap={1} alignItems="center">
-          <IconButton onClick={domApi.undo}>
```
apedroferreira, line 156: We try to avoid implementation details in the tests and also using things like test ids unless strictly necessary, which is also according to Playwright's recommendations of prioritizing user-facing attributes https://playwright.dev/docs/selectors#best-practices.
in this case I guess we could use the tooltip titles for the undo/redo buttons, so test ids should not be needed?

packages/toolpad-app/src/PinholeOverlay.tsx
```diff
@@ -86,7 +86,7 @@ const PinholeOverlay = React.forwardRef(function PinholeOverlay(
 
   // We key the elements so that React doesn't reuse between pinhole off or on
   return (
-    <PinholeOverlayRoot ref={ref} className={className}>
+    <PinholeOverlayRoot ref={ref} className={className} data-testid="hud-overlay">
```
apedroferreira, line 89: This isn't really the overlay the the drag & drop interacts with - the overlay is `OverlayRoot` and it already has the test id `page-overlay` in case that also works.

test/integration/editor/index.spec.ts
apedroferreira: I see, so I guess we need to be able to wait for the overlay to be interactive?
Waiting for the overlay to show and having logic related to the overlay at all is an implementation detail that ideally we should avoid - I wonder if we could wait for something more visible to show?

Maybe we can wait for the element we want to interact with to be visible instead (e.g. the button or text field that the test clicks/types in first)? Not sure if it would work though...

test/integration/undo-redo/index.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+  // Add 3rd text field
+  await editorModel.dragNewComponentToAppCanvas('Text field');
+
+  // Ensure that we added 3rd text field
+  await expect(canvasInputLocator).toHaveCount(3);
+
+  const undoButton = page.locator('[data-testid=undo-button]');
+
+  // Wait for undo stack to be updated
+  await page.waitForTimeout(600);
```
apedroferreira, line 34: Another implementation detail that we should avoid if possible, but again it's hard to think of alternatives in this case...

test/integration/undo-redo/index.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+  const app = await homeModel.createApplication({ dom: domInput });
+  await editorModel.goto(app.id);
+
+  await editorModel.waitForOverlay();
+
+  const input = editorModel.appCanvas.locator('input').first();
+
+  clickCenter(page, input);
+
+  await page.getByLabel('defaultValue').fill('some value');
```
apedroferreira, line 66: Is this interacting with the properties panel on the right side?
We have `componentEditor` in the editor model to select that panel more specifically.

test/integration/undo-redo/index.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  const redoButton = page.locator('[data-testid=redo-button]');
+
+  redoButton.click();
+
+  // Redo should bring back text field
+  await expect(canvasInputLocator).toHaveCount(3);
+});
+
+test('test batching quick actions into single undo entry', async ({ page, browserName }) => {
```
apedroferreira, line 50: Maybe we could make it explicit somehow that this test not only tests batching changes quickly but also changes from the component editor panel.

test/models/ToolpadEditor.ts
```diff
@@ -131,9 +137,11 @@ export class ToolpadEditor {
     await this.page.mouse.move(
       sourceBoundingBox!.x + sourceBoundingBox!.width / 2,
       sourceBoundingBox!.y + sourceBoundingBox!.height / 2,
-      { steps: 5 },
+      { steps: 10 },
     );
 
+    await expect(sourceLocator).toBeVisible();
```
apedroferreira, line 143: If you're dragging a component from outside the canvas this waits for the page to be visible, otherwise it waits for the canvas. 
I guess the first scenario doesn't make a lot of sense, maybe again what we want here is what `waitForOverlay` is doing?

apedroferreira (changes requested):
Looks great in general! Just added some comments and some suggestions.

test/integration/editor/index.spec.ts
bytasv: @apedroferreira I've tried waiting for an actual element, but it seems that element becomes interactive before the overlay appears. If I try to "click center" of the element it simply focuses the actual dom element (i.e. text input) without making a dom selection. If you have some ideas how we could do it otherwise without explicitly waiting for overlay, I'm glad to implement it, but so far I couldn't find a different way

test/integration/undo-redo/index.spec.ts
bytasv: It's not implementation detail. in this case it's a behaviour detail - if you execute actions faster than 500ms they will be batched (something that is technically possible to do with keyboard/mouse)

test/integration/undo-redo/index.spec.ts
bytasv: Could you please clarify what you mean?

test/integration/playwright.config.ts
bytasv: thanks, didn't know about that, it does the trick 👍

test/integration/undo-redo/index.spec.ts
apedroferreira: Yeah, I guess it's not to bad, we could maybe pass that timeout + 100 here in case we change the timeout, if possible, other than that seems ok!

test/integration/undo-redo/index.spec.ts
apedroferreira: Actually nevermind, maybe I was thinking about testing undo/redo separately for the drag & drop and the component properties editor, but as they work with the same logic I think it's okay to cover them all in one test.

test/integration/editor/index.spec.ts
apedroferreira: I see, can't really think of a better way... So I guess this is the best we can do?

test/models/ToolpadEditor.ts
apedroferreira: Sorry, nevermind this comment, I'm rereading it now and it's just wrong.

packages/toolpad-app/src/PinholeOverlay.tsx
bytasv: Thanks! that does work 👍

test/integration/editor/index.spec.ts
bytasv: I suggest we keep an eye on this and see if we can find a better way 🤔

test/integration/undo-redo/index.spec.ts
bytasv: I assume that after your atomic changes PR is merged this will become obsolete anyways, so we can tweak it together with those changes maybe as throttle will not be needed anymore 🤔

apedroferreira: I'm keeping the throttling for now as it looks like we'll still need it for undoing or redoing changing text in text inputs for example - so whether we'll keep the throttling or not it depends on our final implementation of that.
Anyway I think using a constant like `UNDO_REDO_THROTTLE = 500 // milliseconds` and then something like `UNDO_REDO_THROTTLE + 100` in the test would be cool but anyway I think this is fine.

bytasv: We shouldn't need throttle in text fields, because when we undo/redo inside text fields, default browser functionality is used 🤔

Janpot: > We shouldn't need throttle in text fields, because when we undo/redo inside text fields, default browser functionality is used

The dom is updated on every keystroke, so that it can reflect immediately in the canvas. The undo stack shouldn't update as often.

test/models/ToolpadEditor.ts
```diff
@@ -113,7 +113,7 @@ export class ToolpadEditor {
   }
 
   waitForOverlay() {
-    const hudOverlay = this.appCanvas.getByTestId('hud-overlay');
+    const hudOverlay = this.appCanvas.getByTestId('page-overlay');
```
apedroferreira, line 116: in the ToolpadEditor model there is a `pageOverlay` property, maybe we can use it here?

apedroferreira (approved):
Looks great, thanks for addressing the feedback!
Just left another small comment.

bytasv:
That's weird, it's only FF 🤔 I don't have ideas why it would be failing randomly, I could disable it on FF

Janpot:
I saw it in chrome as well.

bytasv:
Can you send chrome failing build link if you have? If not, could you do it the next time you see it?

Janpot:
https://app.circleci.com/pipelines/github/mui/mui-toolpad/4899/workflows/8af3ad67-fcb3-4b03-8b3e-96597bf646e9/jobs/19682

PR #1317, opened 2022-11-09
Show creation dialog as loading while navigating to newly created app

Janpot:
Avoid multiple requests to create a new app


apedroferreira (approved):
Can't find the preview but looks good, this wasn't great in the UI. 
I was wondering why the mutation `loading` state wasn't enough but I see the slow navigation was the reason.

Janpot:
> I was wondering why the mutation loading state wasn't enough but I see the slow navigation was the reason.

It stops loading when the fetch is done, but then the browser still takes some time to navigate while keeping the UI responsive

Issue #1273, opened 2022-11-04
Connections drop-down changes for the demo version


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

From: https://github.com/mui/mui-toolpad/issues/1213#issuecomment-1292167847

Proposal: https://www.figma.com/file/emgFiUXtGpHNqkUyEltcBR/Studio-wireframes?node-id=401%3A2
CC: @gerdadesign 

These are minimal changes that serve the purpose. We'll keep PostgreSQL and Google Sheets disabled.

gerdadesign:
Thanks for talking about this one during our check in! 

Just adjusting the style to disabled and shortening the copy. 

![Image](https://user-images.githubusercontent.com/106991024/200378830-f8824438-1caf-405d-b98c-b261eadbf60a.jpg)

prakhargupta1:
Only further change that I have to suggest is that the dialog box copy should be shortened to: **Create a new connection**

PR #1323, opened 2022-11-09
Set connection/query available options in demo

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/1273
Preview: https://toolpad-pr-1323.onrender.com/

- Disable non-demo mode options when creating a new connection
- Show predefined query option for movies mock API
- Allow only client-side fetch and functions only in demo mode, and show message that full-featured fetch is only in the self-hosted version

https://user-images.githubusercontent.com/10789765/203380721-6415fa47-7227-49a1-a7bd-07ede1584860.mov

As discussed, we're not showing query options for the COVID and Dogs APIs after all - only for the movies query which mocks an API but only runs locally.
Once we can support server-side fetch in the demo we can create connections for the COVID and Dogs API in every demo app.


packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/CreateConnectionNodeDialog.tsx
```diff
@@ -107,12 +108,24 @@ export default function CreateConnectionDialog({
...
             {(Object.entries(dataSources) as ExactEntriesOf<typeof dataSources>).map(
-              ([type, dataSourceDef]) =>
-                dataSourceDef && (
-                  <MenuItem key={type} value={type}>
-                    {dataSourceDef.displayName}
-                  </MenuItem>
-                ),
+              ([type, dataSourceDef]) => {
+                const isDisabledInDemo = config.isDemo && !dataSourceDef?.isDemoFeature;
+                const isSelectable = dataSourceDef && !dataSourceDef.isSingleQuery;
```
apedroferreira, line 113: `isSingleQuery` defines that a datasource should not show up in the connection select because it can only have one type of query (it's not configurable) - in this case the local movies "API"

apedroferreira:
@Janpot I noticed that client-side fetch was not working with `baseUrl` because the client-side fetch is not using connection params. 
I made some changes there for it to work and adjusted the types but it might not be a complete solution yet (the runtime fetch is still not using params from connection nodes, so it works but only works with the connection templates I've created), and maybe you would prefer some different/better way to do it, so feel free to provide feedback about that.

apedroferreira:
Updated video:

https://user-images.githubusercontent.com/10789765/202011566-04269c07-4189-4c33-8686-7b231507a6f6.mov

As discussed, we're not showing query options for the COVID and Dogs APIs - only for the movies query which mocks an API but only runs locally.
Once we can support server-side fetch in the demo we can create connections for the COVID and Dogs API in every demo app.

packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/CreateConnectionNodeDialog.tsx
Janpot: I wonder whether it makes sense to encode this by making the `ConnectionParamsInput` optional? I mean, that's essentially what this does, right?

packages/toolpad-app/src/toolpadDataSources/server.ts
```diff
@@ -4,19 +4,21 @@ import postgres from './postgres/server';
 import rest from './rest/server';
 import googleSheets from './googleSheets/server';
 import movies from './movies/server';
-import config from '../config';
+
+import config from '../server/config';
 
 type ServerDataSources = { [key: string]: ServerDataSource<any, any, any> | undefined };
 
 const serverDataSources: ServerDataSources = config.isDemo
```
Janpot, line 12: Would it make sense to unify this behavior, and the `isDemoFeature` flag into 1 constant under `src/constants.tsx`?
```ts
export const DEMO_DATASOURCES = new Set(['rest', 'function', 'movies'])
export const PRODUCTION_DATASOURCES = new Set(['rest', 'function', 'googleSheets', 'postgres'])
```

apedroferreira: Sounds good, it's a bit redundant right now.

bharatkashyap:
> Closes #1273 Preview: https://toolpad-pr-1323.onrender.com/
> 
> * Disable non-demo mode options when creating a new connection
> * Show predefined query option for movies mock API
> * Allow only client-side fetch and functions only in demo mode, and show message that full-featured fetch is only in the self-hosted version
> 
>  Screen.Recording.2022-11-22.at.17.23.23.mov 
> As discussed, we're not showing query options for the COVID and Dogs APIs after all - only for the movies query which mocks an API but only runs locally. Once we can support server-side fetch in the demo we can create connections for the COVID and Dogs API in every demo app.

Do we need to create a new type of connection for the demo only:

Screenshot 2022-11-24 at 7 04 14 PM

when we could do create a connection node by default for all demo apps instead?

Screenshot 2022-11-24 at 7 05 05 PM

packages/toolpad-app/src/toolpadDataSources/client.tsx
```diff
@@ -1,23 +1,24 @@
...
-    };
+const clientDataSources: ClientDataSources = _.pick(
+  {
+    rest,
+    function: functionSrc,
+    postgres,
+    googleSheets,
+    movies,
+  },
+  [...PRODUCTION_DATASOURCES, ...(config.isDemo ? DEMO_DATASOURCES : [])],
```
Janpot, line 21: Is this correct? Shouldn't this just switch between the two lists?

```suggestion
  Array.from(config.isDemo ? DEMO_DATASOURCES : PRODUCTION_DATASOURCES),
```

apedroferreira:
> Do we need to create a new type of connection for the demo only:
> when we could do create a connection node by default for all demo apps instead?

This was kind of there before so I kept it the same way - but maybe we could make it so that a fetch connection can work with our local movies API instead of having a movies API connection... I can look into it.
The issue with having predefined connections in the demo right now is that client-side queries cannot use things like base URLs or headers that could have sensitive information, so that's something we were only planning to have once we enable server-side queries in the demo.

packages/toolpad-app/src/toolpadDataSources/client.tsx
apedroferreira: I need the list on the client side to be able to show which ones are only available when self-hosting...
Maybe I can get the full list from somewhere else though instead of defining it here.

bharatkashyap:
> > Do we need to create a new type of connection for the demo only:
> 
> > when we could do create a connection node by default for all demo apps instead?
> 
> 
> 
> This was kind of there before so I kept it the same way - but maybe we could make it so that a fetch connection can work with our local movies API instead of having a movies API connection... I can look into it.
> 
> The issue with having predefined connections in the demo right now is that client-side queries cannot use things like base URLs or headers that could have sensitive information, so that's something we were only planning to have once we enable server-side queries in the demo.

Understood. We could probably do with just a query as well, right? A predefined query which uses the `Fetch<default>` connection and provides the entire path to the mock Movies json file in the query itself. Then we won't need a predefined connection.

apedroferreira:
> Understood. We could probably do with just a query as well, right? A predefined query which uses the `Fetch<default>` connection and provides the entire path to the mock Movies json file in the query itself. Then we won't need a predefined connection.

I had tried and created "query templates" like that in a previous version of this PR, but we ended up deciding not to add more complexity just for a demo-only feature. I'll give it another look though, maybe there's a simple way to do it.

bharatkashyap:
> > Understood. We could probably do with just a query as well, right? A predefined query which uses the `Fetch<default>` connection and provides the entire path to the mock Movies json file in the query itself. Then we won't need a predefined connection.
> 
> I had tried and created "query templates" like that in a previous version of this PR, but we ended up deciding not to add more complexity just for a demo-only feature. I'll give it another look though, maybe there's a simple way to do it.

Understood. I was thinking we could create an app template with a predefined query, and simply define all apps on the demo instance as being "from" that template.  Would that work?

apedroferreira:
> Understood. I was thinking we could create an app template with a predefined query, and simply define all apps on the demo instance as being "from" that template. Would that work?

Yeah actually I didn't think about it that way, not sure why, could be a good idea! I'll try it.

packages/toolpad-app/src/toolpad/AppEditor/ConnectionEditor/index.tsx
```diff
@@ -23,15 +23,15 @@ function ConnectionParamsEditor<P>({
...
+  return ConnectionParamsInput ? (
     <ConnectionParamsInput
       handlerBasePath={handlerBasePath}
       connectionId={connectionId}
       value={value}
       onChange={onChange}
       appId={appId}
     />
-  );
+  ) : null;
```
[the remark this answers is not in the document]
apedroferreira: wait, i thought you had suggested making this optional before?
so this is actually intended, unless you think i should move the condition to render `null` elsewhere?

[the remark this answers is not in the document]
apedroferreira: i'll move the condition outside the component

apedroferreira:
@bharatkashyap I looked into using a fetch movies query but in the end I think it wouldn't make a lot of sense because the default fetch query already shows an URL for the local movies JSON file as the initial value...
We do have that additional movies connection in demo though, I guess it would be a separate discussion wether we should just get rid of it - all it does is fetch from the movies API but it allows for filtering by movie genre for example.
We could make the movies JSON filterable with a param or something but I'm afraid that would be adding more complexity to things, and demo specific logic.

bharatkashyap:
> the default fetch query already shows an URL for the local movies JSON file as the initial value...

> We do have that additional movies connection in demo though, I guess it would be a separate discussion whether we should just get rid of it - all it does is fetch from the movies API, but it allows for filtering by movie genre for example, which a fetch query does not.

Understood. For me, the idea is to remove a demo-only connection and just create a template we can use in general as well. So, we could create a fetch query which can do exactly the same thing as the demo-only movies connection, place that in a  template and use that as the default in case of demos. (We could even potentially make that the default selected option in general for all apps.)

apedroferreira:
> So, we could create a fetch query which can do exactly the same thing as the demo-only movies connection

I'm not sure how we could do that with just a JSON file, we would need some kind of local API endpoint to do the filtering?
Which we could do along with getting rid of the movies connection, but maybe we could do it in a separate issue so it doesn't block the demo, as it shouldn't be too simple to implement?
If we don't even want filtering, we could just get rid of the movies connection and that would be it, but I guess that could be a part of that separate discussion.

packages/toolpad-app/src/toolpadDataSources/client.tsx
Janpot: Perhaps we shouldn't do the filtering here then, and just export all of the datasources? Then do the filtering when we create the dropdown?

apedroferreira: this is outdated, i've already made it more like you suggested, lmk if it's cool

Janpot (approved):

apedroferreira:
@bharatkashyap I've discussed with Jan and got rid of the movies connection, but in my opinion let's not create a query for movies because the movies "API" is already the default for new default queries.
Will merge this but let me know if you have any more comments!

bharatkashyap:
> @bharatkashyap I've discussed with Jan and got rid of the movies connection, but in my opinion let's not create a query for movies because the movies "API" is already the default for new default queries.
> 
> Will merge this but let me know if you have any more comments!

Makes sense!

Issue #1196, opened 2022-10-21
Client-side API requests


oliviertassinari:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Currently, all the requests are run server-side, however, there are cases where you would prefer them to run client-side. I have found one case: https://docs.github.com/en/rest/search#rate-limit. Instead of providing a key, you can not provide any, and have each client's IP set the rate limit rather than the server's IP.

I don't really need this now, but I thought it could be interesting to track the interest from the community.

Janpot:
This could be interesting for the demo mode, we won't have the DDOS attack vector

oliviertassinari:
I have found a new use case for client-side queries: firewalls. 

<img width="320" alt="Screenshot 2022-12-30 at 17 55 37" src="https://user-images.githubusercontent.com/3165635/210094610-8c4909fe-f99a-4c4b-b68f-d0a1646859ec.png">

https://stackoverflow.com/questions/52076581/wordpress-ajax-call-response-with-well-known-captcha

oliviertassinari:
I'm reopening as I believe this support was removed.

YohanSciubukgian:
Does this feature will allow the MUI Toolpad to generate ReactJS App that are exclusively client-side rendered ?

Janpot:
@YohanSciubukgian This feature would allow for Toolpad apps that can run without a server API. i.e. it could technically be statically hosted. Currently Toolpad doesn't generate code. Looking at your comment, I'm wondering whether that is a feature you're interested in. If so, please upvote the [feature request](https://github.com/mui/mui-toolpad/issues/316).

oscni:
Is it working now to do client side API requests? In do we then no longer need any server and instead can deploy the application to an SSG site? Seems like the issue is "recently completed" in the roadmap.

YohanSciubukgian:
> @YohanSciubukgian This feature would allow for Toolpad apps that can run without a server API. i.e. it could technically be statically hosted. Currently Toolpad doesn't generate code. Looking at your comment, I'm wondering whether that is a feature you're interested in. If so, please upvote the [feature request](https://github.com/mui/mui-toolpad/issues/316).

This is exactly what I was looking for, being able to have React code generated by the Toolpad with capabilities to have my own client API calls to bind my business rules to the react code generated. Thanks for the suggestion.

PR #1325, opened 2022-11-10
Run functions client side

Janpot:
Closes https://github.com/mui/mui-toolpad/issues/1196


packages/toolpad-app/src/appDom/index.ts
```diff
@@ -1063,8 +1063,10 @@ function createRenderTreeNode(node: AppDomNode): RenderTreeNode | null {
   }
 
   if (isQuery(node) || isMutation(node)) {
+    // This is hacky, should we delegate this check to the datasources?
```
apedroferreira, line 1066: It would probably be easier to maintain if there's a way to filter these out from the datasources, I guess?
But doesn't seem that hacky, or I'm not sure if creating a method to see if a node is a browser side query would feel less hacky.

apedroferreira (approved):
LGTM

PR #1328, opened 2022-11-11
Fix undo/redo race condition

bytasv:
Fixes an issue reported by @oliviertassinari when race condition happens and undo stack is messed up

Related to https://github.com/mui/mui-toolpad/issues/226


packages/toolpad-app/src/toolpad/DomLoader.tsx
```diff
@@ -239,9 +239,22 @@ export function domLoaderReducer(state: DomLoader, action: DomAction): DomLoader
...
+  },
+  500,
+  { leading: false, trailing: true },
+);
+
 function createDomApi(dispatch: React.Dispatch<DomAction>) {
   return {
     undo() {
+      if (pendingHistoryUpdate) {
+        scheduleHistoryUpdate.flush();
```
bytasv, line 255: I looked into ways to have this done inside reducer however that would mean that inside DOM_UNDO reducer I'd have to trigger dispatch and then have DOM_UNDO action itself be triggered after timeout with dispatch (or do some other workaround, maybe there is a cleaner way).
Eventually I decided to keep this logic out of reducer so once we finally have atomic updates, this should be easier to clean up

Janpot: - we don't need the `pendingHistoryUpdate` flag, `flush` is a no-op when there is no invaction pending. You can just call `.flush()` every time in `undo`
- Can we pass the flush method as a second parameter to `createDomApi` and avoid mixing global state in instances of the domApi?

Janpot (approved):

packages/toolpad-app/src/toolpad/DomLoader.tsx
bytasv: Yeah I've tried that, but eventually decided to not do that as it might be easier to remove this logic in the future when it's less "tangled" with the code of reducer 🤔

PR #1341, opened 2022-11-13
[core] Group renovate GitHub Action dependency updates

oliviertassinari:
Having to merge #1339, #1335, #1336 as different PRs feels overkill.

The solution is based on https://github.com/renovatebot/renovate/discussions/14578.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -77,6 +77,10 @@
     {
       "matchDepTypes": ["action"],
       "pinDigests": true
+    },
+    {
+      "groupName": ["github-actions"],
+      "matchManagers": ["github-actions"]
     }
   ],
   "postUpdateOptions": ["yarnDedupeHighest"],
--- a/renovate.json
+++ b/renovate.json
@@ -79,7 +79,7 @@
       "pinDigests": true
     },
     {
-      "groupName": "GitHub Action",
+      "groupName": "GitHub Actions",
       "matchManagers": ["github-actions"]
     }
   ],
--- a/renovate.json
+++ b/renovate.json
@@ -79,7 +79,7 @@
       "pinDigests": true
     },
     {
-      "groupName": ["github-actions"],
+      "groupName": "GitHub Action",
       "matchManagers": ["github-actions"]
     }
   ],


renovate.json
```diff
@@ -77,6 +77,10 @@
     {
       "matchDepTypes": ["action"],
       "pinDigests": true
+    },
+    {
+      "groupName": ["github-actions"],
```
oliviertassinari, line 82:
```suggestion
      "groupName": "GitHub Action",
```

renovate.json
```diff
@@ -77,6 +77,10 @@
     {
       "matchDepTypes": ["action"],
       "pinDigests": true
+    },
+    {
+      "groupName": "GitHub Action",
```
oliviertassinari, line 82:
```suggestion
      "groupName": "GitHub Actions",
```

Janpot:
Isn't that essentially the same group as the one above that?

Aside: If we're going to monitor PR reviews in https://github.com/mui/mui-toolpad/security/code-scanning/22, would it makes sense to always approve dependency update PRs before merging so that it doesn't distort this metric?

oliviertassinari:
> Isn't that essentially the same group as the one above that?

I don't know, I copied https://github.com/Automattic/vip-container-images/blob/6057834e5cc1041eeb5bf5296a55b14b1c1952ec/.github/renovate.json#L14-L21

> Aside: If we're going to monitor PR reviews in https://github.com/mui/mui-toolpad/security/code-scanning/22, would it makes sense to always approve dependency update PRs before merging so that it doesn't distort this metric?

I think that the idea of having a peer review is to say that 2 people approved the changes. Renovate is a bot, does it count? I'm leaning on a not but I have personally approved the PRs for this reason before 🙃

oliviertassinari:
> Isn't that essentially the same group as the one above that?

@Janpot I had a look at the source, the interesting file is https://github.com/renovatebot/renovate/blob/30eb3dda394cd5e52a6c2eacf6cd93188d176e98/lib/modules/manager/github-actions/extract.ts.

From what I can gather`matchManagers: 'github-actions'` matches the `.workflow` GitHub files. Which has 3 types of `matchDepTypes`:

- [docker containers](https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container)
- actions
- [services](https://docs.github.com/en/actions/using-containerized-services/about-service-containers)

PR #1343, opened 2022-11-14
Remove externalResolver for Sentry

Janpot:
@apedroferreira Do I understand correctly that this won't be needed anymore?


apedroferreira (approved):
Yep, shouldn't be needed anymore.

PR #1346, opened 2022-11-14
Disable baseUrl when ran in browser

Janpot:
@apedroferreira

### DIFF
--- a/packages/toolpad-app/src/toolpadDataSources/rest/client.tsx
+++ b/packages/toolpad-app/src/toolpadDataSources/rest/client.tsx
@@ -83,7 +83,7 @@ const ButtonLink = styled('button')(({ theme }) => ({
 }));
 
 interface UrlControlProps extends RenderControlParams<string> {
-  baseUrl?: string;
+  baseUrl: string | null;
 }
 
 function UrlControl({ label, disabled, baseUrl, value, onChange }: UrlControlProps) {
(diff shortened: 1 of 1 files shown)


apedroferreira (approved):
thanks!

packages/toolpad-app/src/toolpadDataSources/rest/client.tsx
```diff
@@ -271,11 +271,12 @@ const EMPTY_PARAMS: BindableAttrEntries = [];
 
 function QueryEditor({
   globalScope,
-  connectionParams,
+  connectionParams: rawConnectionParams,
   value: input,
   onChange: setInput,
 }: QueryEditorProps<RestConnectionParams, FetchQuery>) {
-  const baseUrl = connectionParams?.baseUrl;
+  const connectionParams = input.attributes.query.value.browser ? null : rawConnectionParams;
```
apedroferreira, line 278: Maybe we can store `input.attributes.query.value.browser` as a `isClientSideQuery` variable to reuse in these 2 lines?

packages/toolpad-app/src/toolpadDataSources/rest/client.tsx
```diff
@@ -271,11 +271,12 @@ const EMPTY_PARAMS: BindableAttrEntries = [];
...
+  const baseUrl = input.attributes.query.value.browser ? null : connectionParams?.baseUrl;
```
apedroferreira, line 279: or actually here we can just have `const baseUrl = connectionParams?.baseUrl; ?? null`

PR #1350, opened 2022-11-15
Show IP address in log messages

apedroferreira:
Show probable IP address from `x-forwarded-for` header in log messages, to make logs more easily distinguishable per user.

This field is not always reliable but for this usecase should be helpful.
This is how the dashboard looks in Kibana right now without this change:

Screenshot 2022-11-15 at 18 02 23


Janpot:
Perhaps we can add another field for this in the `reqSerializer`?

apedroferreira:
> Perhaps we can just add another field for this in the `reqSerializer`?

I'm already including the `x-forwarded-for` header in that serializer.
This was just so that I can take a quick look at the logs without having to open each one and know which ones come from different users (if we had authenticated users we could have their user ids in the log message instead for the same purpose)

Janpot:
* This does some parsing of the field, the header is raw. We can add this parsing in the serializer.
* In kibana you should be able to add a column in the table for that field. We're not going to change the message every tinme we want some extra information in the table. You should also be able to store the state of this table so we can create different views.

apedroferreira:
> * This does some parsing of the field, the header is raw. We can add this parsing in the serializer.
> * In kibana you should be able to add a column in the table for that field. We're not going to change the message every tinme we want some extra information in the table. You should also be able to store the state of this table so we can create different views.

Sounds good, yes you're right, we can just use more columns and keep the message simple...
I'll just change the serializer to have a more readable `ipAddress` field then.

PR #1351, opened 2022-11-15
Automatic app names in demo

apedroferreira:
From https://github.com/mui/mui-toolpad/issues/592

Set automatic app names in demo mode and remove logic for non-duplicated app names.
Here I'm setting the app name client-side with a format like `demo_app_1668543595554`.

Screenshot 2022-11-15 at 20 19 49

### DIFF
--- a/packages/toolpad-app/src/utils/crypto.ts
+++ b/packages/toolpad-app/src/utils/crypto.ts
@@ -3,7 +3,7 @@ import invariant from 'invariant';
 const crypto =
   typeof window === 'undefined'
     ? // eslint-disable-next-line global-require
-      (invariant(!global.crypto, 'Remove the crypto polyfill'), require('crypto'))
+      (invariant(!!global.crypto, 'Remove the crypto polyfill'), require('crypto'))
     : global.crypto;
 
 export default crypto;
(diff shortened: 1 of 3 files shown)


packages/toolpad-app/src/utils/crypto.ts
```diff
@@ -3,7 +3,7 @@ import invariant from 'invariant';
 const crypto =
   typeof window === 'undefined'
     ? // eslint-disable-next-line global-require
-      (invariant(!!global.crypto, 'Remove the crypto polyfill'), require('crypto'))
+      (invariant(!global.crypto, 'Remove the crypto polyfill'), require('crypto'))
```
apedroferreira, line 6: Unrelated but I guess this was wrong?

bharatkashyap:
Good idea! Do you think it makes more sense for the name to be a bit more readable - for instance a random phrase which is often seen in random usernames? Something like the examples here, where no parameters are passed: https://www.npmjs.com/package/unique-username-generator


Feel free to ignore if you don't think it's worth the effort.

apedroferreira:
> Good idea! Do you think it makes more sense for the name to be a bit more readable - for instance a random phrase which is often seen in random usernames? Something like the examples here, where no parameters are passed: https://www.npmjs.com/package/unique-username-generator
> 
> Feel free to ignore if you don't think it's worth the effort.

I thought about that first and looked at that exact library, the main concern I had was adding to the bundle size just because of a kind of temporary feature like this.

apedroferreira:
All good with the `demo_app_xxxxx` naming? I can take suggestions if anyone has a better idea or if more people think it's worth using a library to generate random names.

Janpot (approved):

PR #1353, opened 2022-11-16
Basic rate limiter implementation

Janpot:
nginx handles only simple http proxy. but we also need to be able to proxy https traffic. For that we'll need [http tunneling](https://en.wikipedia.org/wiki/HTTP_tunnel). nginx doesn't offer this out of the box. We could opt for an alibaba made module like https://www.alibabacloud.com/blog/how-to-use-nginx-as-an-https-forward-proxy-server_595799. but preferrable not. It seems nginx isn't the exact tool for the job for a forwarding proxy. Further research shows there aren't many options for a rate limiting forward proxy besides writing one.

This leaves us with following options:

- write a forwarding proxy [ourselves](https://javascript.plainenglish.io/build-your-own-forward-and-reverse-proxy-server-using-node-js-from-scratch-eaa0f8d69e1f). Add in-memory rate limiting. On Toolpad side everything remains the same. [Proxy agent](https://www.npmjs.com/package/hpagent) in the fetch and function datasources.
- Write a node.js http(s) agent that rate limits. https://www.npmjs.com/package/agent-base + https://www.npmjs.com/package/limiter. I'm starting with trying this out in this PR.

### further exploration:

- [ ] We need to rate limit per application to prevent one bad actor messing things up for everyone else
- [ ] Add a cooldown period
- [ ] Redis

### DIFF
--- a/packages/toolpad-app/src/utils/outboundAgent.ts
+++ b/packages/toolpad-app/src/utils/outboundAgent.ts
@@ -1,0 +1,71 @@
+import agentBase from 'agent-base';
+import http from 'http';
+import https from 'https';
+import dns from 'dns/promises';
+import fetch, { Request } from 'node-fetch';
+
+const SLIDING_WINDOW = 1000;
+const REQS_PER_WINDOW = 5;
+
+const rateLimiterWindows = new Map<string, number[]>();
+
+// Regular garbage collection for the rateLimiterWindows
+setInterval(() => {
+  const now = Date.now();
+  for (const [key, window] of rateLimiterWindows) {
+    if (window.every((timestamp) => timestamp < now - SLIDING_WINDOW)) {
+      rateLimiterWindows.delete(key);
+    }
+  }
+}, 5000).unref();
+
+// Basic sliding window rate limiter
+// We can build this distributed in redis:
+//   https://developer.redis.com/develop/dotnet/aspnetcore/rate-limiting/sliding-window/#sliding-window-rate-limiter-lua-script
+async function rateLimit(address: string): Promise<boolean> {
+  let window = rateLimiterWindows.get(address) ?? [];
+  const now = Date.now();
+  window = window.filter((timestamp) => timestamp > now - SLIDING_WINDOW);
+  let allowed = true;
+  if (window.length >= REQS_PER_WINDOW) {
+    allowed = false;
+  } else {
+    window.push(now);
+  }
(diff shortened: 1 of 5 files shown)


apedroferreira:
It's a shame that nginx doesn't work for https...
How does this approach differ from in-memory rate limiting? This seems like a good enough approach for now if it works.

packages/toolpad-app/src/utils/outboundAgent.ts
```diff
@@ -0,0 +1,71 @@
...
+setInterval(() => {
+  const now = Date.now();
+  for (const [key, window] of rateLimiterWindows) {
+    if (window.every((timestamp) => timestamp < now - SLIDING_WINDOW)) {
+      rateLimiterWindows.delete(key);
+    }
+  }
+}, 5000).unref();
+
+// Basic sliding window rate limiter
```
apedroferreira, line 22: The way I've used Redis for rate-limiting was with a key that stores a number that gets increment + a certain TTL for the key to expire. I hadn't seen this sliding window method but it should also work, I wonder if it has any advantages compared to the other method i mentioned?

Janpot: For our use-case we're going to want a rate limiting algorithm that can handle bursts. e.g. It's fine to do 3 requests in rapid succession, but not 300. We may also want some cooldown period, e.g. you hit the rate limit, now you're blocked for 60s.

Janpot:
> How does this approach differ from in-memory rate limiting? 

This is in-memory rate limiting, it's currently mostly to verify this method works, it doesn't synchronize across Toolpad instances. We can expand with a redis implementation of the rate limiting part to support multiple instances. The abstraction is async, so we should be able to write a drop-in replacement on top of redis.

The alternative is to write a rate limiting proxy ourselves, this would also work in-memory, and we would have only one that all Toolpad instance will tunnel through, so we don't need redis to make it work across Toolpad instances

packages/toolpad-app/src/utils/outboundAgent.ts
apedroferreira: Got it, just read about some different rate limiting algorithms, this seems like a good pick.

apedroferreira:
I think we can start with this approach.
Writing our own proxy would be nice but I don't know how time consuming or difficult it would be, could be a fun challenge though.

Issue #1065, opened 2022-09-29
Connections should be global


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Currently, our connections are inside the app. So for each app user has to create a connection. But I think connection should be app agnostic and once connected to a data source, all apps should have access to that data source. Later on in gradnular access control, we can limit its access.

If the connection isn't set it should not show inside the app when forming a query.

From the UI perspective, we could do something like:

![Screenshot 2022-09-29 at 1 13 12 PM](https://user-images.githubusercontent.com/92228082/192970957-c457280a-75b5-429d-a194-e09e81ca030b.png)

This will clean up the app building experience, and we'll have the room to condense pages to the top left and queries to the left bar.

apedroferreira:
> @prakhargupta1 I notice these connections run on server side. App query through next.js api to get data.

Hi @WangLarry - yes, that is in general how our connections and queries work.
We believe it should be possible to have them keep working like that but be globally defined on a user account level, instead of being specific to each app.

WangLarry:
yes.
Query node should be in app level. On client side, react-query cache already same request. 
Connect template is over app level or in library. Connect template: google api, facebook api  etc. App should config some params (clientId ...).

oliviertassinari:
The direction toward moving the connections to be global makes sense to me. I think that it feels a bit similar to moving toward the direction of a mono repo. You get global services that you can use from many different locations.

I think that this direction could also help in an environment where you have a staff with different technical skills that build these simple applications. A strong backend engineer might do the setup when a product manager might later on create few new pages.

Janpot:
I opened a discussion which allows to collaborate on this under threads https://github.com/mui/mui-toolpad/discussions/1376

prakhargupta1:
I am closing this issue as it is not relevant anymore. Now each Toolpad app has multiple pages and all can share a connection, so in a way in the scope of the project, the connection is global.

PR #1357, opened 2022-11-17
Refactor overview page

Janpot:
Bring overview page closer in line with the proposal in https://github.com/mui/mui-toolpad/issues/1065
It's not pixel perfect, but this way we can start working on global connections and aren't stuck on a final design for this page.

Screenshot 2022-11-17 at 14 58 53

I'm also bringing this page under `/_toolpad/apps` and redirect `/` to there

https://toolpad-pr-1357.onrender.com/


packages/toolpad-app/src/toolpad/Apps/index.tsx
```diff
@@ -138,7 +140,7 @@ function CreateAppDialog({ onClose, ...props }: CreateAppDialogProps) {
     null,
   );
 
-  const isFormValid = config.isDemo || Boolean(name);
+  const isFormValid = Boolean(name);
```
apedroferreira, line 143: Not sure if this is was accidental during resolving conflicts, but we can assume that the form is valid in demo mode because we don't have a name field there anymore (we have automatic names)

packages/toolpad-app/src/toolpad/Apps/index.tsx
```diff
@@ -160,11 +162,9 @@ function CreateAppDialog({ onClose, ...props }: CreateAppDialogProps) {
             });
           }
 
-          const appName = config.isDemo ? `demo_app_${Date.now()}` : name;
```
apedroferreira, line 163: Same here I guess this part was accidentally removed? Let me know if it's easier if I fix this myself.

packages/toolpad-app/src/toolpad/Apps/index.tsx
```diff
@@ -428,7 +426,11 @@ interface AppOpenButtonProps {
 function AppOpenButton({ app, activeDeployment }: AppOpenButtonProps) {
   const openDisabled = !app || !activeDeployment;
   let openButton = (
-    <Button disabled={!app || !activeDeployment} component="a" href={app ? `deploy/${app.id}` : ''}>
+    <Button
```
apedroferreira, line 429: Can this one be a `Link` too as it seems to be an internal link?

packages/toolpad-app/src/toolpad/ToolpadHomeShell.tsx
```diff
@@ -0,0 +1,66 @@
...
+            mt: 3,
+            ml: 2,
+            overflow: 'auto',
+          }}
+        >
+          <List>
+            <ListItem disablePadding>
+              {/* @ts-expect-error https://github.com/mui/material-ui/issues/29875 */}
+              <NavigationListItemButton component={Link} to={'/apps'}>
+                <ListItemText primary="Apps" />
```
apedroferreira, line 39: Should we have icons for the list items too? But probably not too important for now.

apedroferreira (commented):
Looks good!
Just added some nits, most important thing is the deletions in the demo I think.

Also there is a left padding in the sidebar that might not be needed?

Screenshot 2022-11-17 at 16 10 10

The white part on the left, we could maybe have the gray when we hover take up the full width?

Janpot:
> The white part on the left, we could maybe have the gray when we hover take up the full width?

🤦 It looked like that in the screenshot in https://github.com/mui/mui-toolpad/issues/1065 but it was just whitespace around the image

packages/toolpad-app/src/toolpad/Apps/index.tsx
Janpot: No, it's not under the same `react-router`.
We're mixing next router and react-router. not 100% ideal, but (at the time of writing) next router wasn't really helpful in building the editor with its deep links to subeditors

packages/toolpad-app/src/toolpad/ToolpadHomeShell.tsx
Janpot: Let's wait for the design to be finalized

packages/toolpad-app/src/toolpad/Apps/index.tsx
apedroferreira: ah ok, nevermind then

bharatkashyap (approved):
Great improvement to the home page!

PR #1360, opened 2022-11-18
Fix next regression that broke sucrase

Janpot:
This seems to be a regression in next 13. See https://github.com/vercel/next.js/issues/43128

Fixed on canary https://github.com/vercel/next.js/issues/43104#issuecomment-1321831454


test/integration/components/index.spec.ts
```diff
@@ -1,10 +1,21 @@
...
+  await page.locator('text="foo button"').waitFor({ state: 'visible' });
+  await page.locator('img[alt="foo image"]').waitFor({ state: 'attached' });
+  await page.locator('text="foo datagrid column"').waitFor({ state: 'visible' });
+  await page.locator('text="custom component 1"').waitFor({ state: 'visible' });
+  await page.locator('label:has-text("foo textfield")').waitFor({ state: 'visible' });
+  await page.locator('text="foo typography"').waitFor({ state: 'visible' });
+  await page.locator('label:has-text("foo select")').waitFor({ state: 'visible' });
+}
+
+test('components', async ({ page, browserName }) => {
```
apedroferreira, line 18: Maybe the test name could describe its intention better, could be something like "components render correctly"?

apedroferreira (approved):
LGTM!

PR #1361, opened 2022-11-19
Remove sentry debug option

Janpot:
Was this enabled by accident? I feel like it's a bit chatty to be enabled by default in development


apedroferreira (approved):
Maybe, I used it to debug something and might have kept it by mistake.
We can remove it.

PR #1374, opened 2022-11-22
Single-update undo/redo

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/845
Closes https://github.com/mui/mui-toolpad/pull/1192
Preview URL: https://toolpad-pr-1374.onrender.com/

Update the DOM only a single time for each drag & drop operation/resizing/delete, so that undo/redo works without throttling.


apedroferreira:
@Janpot I've removed some methods from `domApi`, the ones that seemed most likely to be used in updates with more than one change. Let me know if you think I should removed more or less methods.

bytasv (commented):
If dom changes are done in atomic update why do we still need throttle? Wouldn't the goal of this PR be to get rid of throttle entirely?
You mentioned in other PR that throttle might still be needed when undoing changes in a text input, though our undo/redo implementation should not be even triggered inside text inputs, so I'm wondering if it's related to that?

bytasv:
Thanks for explanation Jan (I missed your reply in another PR), now its clear to me. I agree we need to fix that and cover with test

apedroferreira:
It was difficult to make atomic updates work with select/deselect so I've added a couple more features here:
- Optional ability to make a select/deselect skip the history so it doesn't get added to the undo stack (e.g.: dragging a new component to the canvas triggers a deselect that we probably don't want in the stack)
- Ability to define if a node should be selected/deselected along with dom updates

>  I agree we need to fix that and cover with test
We can work on that next! This should be the roadmap for this feature: https://github.com/mui/mui-toolpad/issues/226

bytasv:
> (e.g.: dragging a new component to the canvas triggers a deselect that we probably don't want in the stack)

What will happen if I start dragging and I change my mind and dont drop component on the canvas? I will no longer have a selection and undo wont bring back the selection? If that's the case, maybe instead of deselecting node, instead we could hide selection state while we are dragging new component and upon drop set new selection?

Janpot:
Yep, let's not deselect in the state when drag starts, but only in the render panel while dragging.

apedroferreira:
> This way we don't need history skip flags, right?

Yeah this should work and remove the need for the skipping option, just have to find a way to do it but should be fine!

apedroferreira:
Can I get a final review here in order to merge, please?
Changing views is ready in https://github.com/mui/mui-toolpad/pull/1417, but I can't get a preview live until I merge this part.

test/integration/undo-redo/index.spec.ts
```diff
@@ -42,42 +41,3 @@ test('test basic undo and redo', async ({ page, browserName, api }) => {
   // Redo should bring back text field
   await expect(canvasInputLocator).toHaveCount(3);
 });
-
-test('test batching quick actions into single undo entry', async ({ page, browserName, api }) => {
```
bytasv, line 46: This spec should still be relevant, specifically for text field changes

bytasv (approved):
Apart from from spec comment all seems to be working great 👍

test/integration/undo-redo/index.spec.ts
apedroferreira: We can have something more specific once we have specific logic for text inputs?
Anyway I won't merge this until i have a test that does the same that this one did.

apedroferreira: This test should work fine after all, I had the impression that it was testing more general changes but I was wrong!

apedroferreira:
I've merged text input undo/redo into this PR so that it works as a whole and can be merged.

The PR with the text input changes is here https://github.com/mui/mui-toolpad/pull/1459, basically I'm just using debouncing and completely disabling the native browser undo/redo in text inputs.

Preview URL: https://toolpad-pr-1374.onrender.com/

apedroferreira:
I'm gonna merge this PR later today if it's ok!
Feel free to check these changes (just 4 files) https://github.com/mui/mui-toolpad/pull/1459, and the preview at https://toolpad-pr-1374.onrender.com/ (includes atomic updates + text input undo/redo) in case there's any more comments.
I shouldn't have requested for reviews yesterday until having added these extra changes, so sorry for that.

packages/toolpad-app/src/toolpad/AppEditor/ConnectionEditor/index.tsx
```diff
@@ -49,18 +49,21 @@ function ConnectionEditorContent<P>({
...
+        dom,
         connectionNode,
         'attributes',
         'params',
         appDom.createSecret(connectionParams),
       );
+      domApi.update(updatedDom);
     },
-    [connectionNode, domApi],
+    [connectionNode, dom, domApi],
```
Janpot, line 66: 🙂 I think this is why I suggested to use a function in `domApi.update`. This callback invalidates on every dom change, and it's also not 100% sure it will get the last `dom`. e.g. imagine a function.
```tsx
const foo = () => domApi.update(doSomethingWith(dom));
const bar = () => domApi.update(doSomethingElseWith(dom));

// somewhere else, foo called twice in a row
foo()
bar()
```
Those would get the same initial version of `dom`, while I would expect in most cases that in the call of `bar`, there should be an update on the result of `doSomethingWith(dom)` from call of `foo`

apedroferreira: Would it work to restore and use methods such as `domApi.setNodeNamespacedProp` (which I had deleted) in cases like this? 
And use `domApi.update` as I'm using it but only when a temporary DOM is needed? `domApi.update` replaces the whole DOM anyway so in those cases maybe it's fine?

apedroferreira: Still thinking of better alternatives, your proposal does seem more "atomic", it's just more difficult to use in `RenderOverlay`.

apedroferreira: I'll try supporting `domApi.update((draft) => ...` and use it as much as possible when I can, except in the few cases where it's hard to use, I think that should be possible.

Janpot: > Still thinking of better alternatives, your proposal does seem more "atomic", it's just more difficult to use in `RenderOverlay`.

We can support both methods, like 

```tsx
update: (draft: React.SetStateAction<AppDom>) => void
```

apedroferreira: Looks like I was able to use an updater function everywhere after all!
https://github.com/mui/mui-toolpad/pull/1374/commits/1f7779e2e9d097c6d5607e11915a59d180e39bc3
After the refactoring I had made I guess it wasn't so difficult.

apedroferreira: As this PR is using the updater function now I will merge this part already!

bharatkashyap (approved):

PR #1381, opened 2022-11-24
Add a Default app template

### LINKED ISSUE #841: Components should have some pre-filled dummy data when dragged on canvas

bharatkashyap:
- Closes #841 
- Visual demonstration 

https://user-images.githubusercontent.com/19550456/204729174-ac0d1bed-43f6-497e-982a-ae91fcf98c4d.mov

@apedroferreira Does something like this work for the demo as well? This is what I meant in https://github.com/mui/mui-toolpad/pull/1323#issuecomment-1326748742


prakhargupta1:
It looks good! Is this what we'll show as dummy data in the data grid?
We can definitely extend this as a proper template.

bharatkashyap:
> Is this what we'll show as dummy data in the data grid?

My idea is that we don't add dummy data to every single data grid, and make this template the default. That allows first time users to see dummy data in the data grid, but removed the need for someone to have to remove dummy data from a data grid every successive time.

test/models/ToolpadHome.ts
```diff
@@ -40,7 +40,7 @@ export class ToolpadHome {
     });
     this.newAppNameInput = this.newAppDialog.locator('label:has-text("name")');
     this.newAppTemplateSelect = this.newAppDialog.locator(
-      '[aria-haspopup="listbox"]:has-text("blank")',
+      '[aria-haspopup="listbox"]:has-text("demo")',
```
apedroferreira, line 43: I know this wasn't correct before the PR already, but we should try to select this with the `Use template` label if there's a way.

apedroferreira (approved):
LGTM if we want this template to be the new default one.
I added a comment about trying to make the test not rely on the default app template, but could be fixed separately too, I could try it.

test/models/ToolpadHome.ts
bharatkashyap: How would we select the option with the `Use template` label? (if I'm understanding what you're saying correctly)

Janpot: https://github.com/mui/mui-toolpad/blob/28a36db057e184243941db39d7a1ef45130aaa76/test/integration/appTemplates.spec.ts#L33-L35

apedroferreira: > How would we select the option with the Use template label?
It's not the label for one of the options, it's the label for the whole select input.

bharatkashyap: I've done it with the method @Janpot mentioned up there, seems to work

bharatkashyap:
> @bharatkashyap Since we're trying to rely on external datasources as little as possible to avoid issues that we can't control, are the images in this template safe to use, or should we use our own images in the `static` folder? If so, I guess that 5 or 6 small royalty-free photos for example should work?

I've used data from https://mockapi.io which has `https://cloudflare-ipfs.com` links to the static assets, I feel that is a relatively safe choice for default data?

Issue #1380, opened 2022-11-24
Client-side fetch queries are crashing the app


apedroferreira:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

Steps:

1. Create a fetch query (e.g.: "fetch | default")
2. Enable "Run in the browser"
3. Press the "Preview button"

### Current behavior 😯

Doing the steps above crashes the app.

### Expected behavior 🤔

Doing the steps above shows the query result without crashing the app.

### Context 🔦

This seems like an important bug that should be fixed before fully launching the demo.

Janpot:
Interesting, we'll need an integration test for this. Seems to not happen when I do:

```diff
--- a/packages/toolpad-app/src/toolpadDataSources/rest/runtime.tsx
+++ b/packages/toolpad-app/src/toolpadDataSources/rest/runtime.tsx
@@ -57,7 +57,7 @@ export async function clientExec(
 
     const result = await execfetch(fetchQuery, params, {
       connection: null,
-      fetchImpl: instrumentedFetch,
+      fetchImpl: window.fetch,
       evalExpression,
     });
```

apedroferreira:
> memory leak in chrome?

Not sure, I noticed that it was making the request but something is crashing afterwards, didn't investigate much further yet.

bharatkashyap:
I can reproduce as well (similar behaviour on Firefox as well)


https://user-images.githubusercontent.com/19550456/203854359-1db446c9-898a-4302-9b3a-f4bb46ed3d61.mov

Janpot:
Fix in https://github.com/mui/mui-toolpad/pull/1383

PR #1383, opened 2022-11-25
Fix time property in HAR

Janpot:
It's the duration, not the current time.

This resulted in `perf-cascade` building massive amounts of svg dom.

fixes https://github.com/mui/mui-toolpad/issues/1380

### DIFF
--- a/packages/toolpad-app/src/toolpadDataSources/rest/runtime.tsx
+++ b/packages/toolpad-app/src/toolpadDataSources/rest/runtime.tsx
@@ -14,12 +14,12 @@ export async function clientExec(
     const har = createHarLog();
 
     const instrumentedFetch = async (...args: Parameters<typeof fetch>) => {
-      const startedDateTime = new Date().toISOString();
+      const startTime = Date.now();
       const req = new Request(...args);
       const url = new URL(req.url);
       const res = await window.fetch(req);
       const entry: Entry = {
-        startedDateTime,
+        startedDateTime: new Date(startTime).toISOString(),
         request: {
           bodySize: 0,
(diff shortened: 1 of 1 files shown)


test/integration/rest-basic/index.spec.ts
```diff
@@ -30,8 +31,19 @@ test('functions basics', async ({ page }) => {
...
-  await page.locator('text="query2: query2_value"').waitFor({ state: 'visible' });
+  await expect(page.locator('text="query2: query2_value"')).toBeVisible();
+  await expect(page.locator('text="browserQuery: browserQuery_value"')).toBeVisible();
+  await expect(page.locator('text="browserQuery: browserQuery_value"')).toBeVisible();
+
+  const editorModel = new ToolpadEditor(page, browserName);
+  await editorModel.goto(app.id);
+
+  await editorModel.componentEditor.getByRole('button', { name: 'browserQuery' }).click();
+  const queryEditor = page.getByRole('dialog', { name: 'browserQuery' });
```
apedroferreira, line 45: Should we create an abstraction for testing preview queries like this?

apedroferreira (approved):
Looks good, thanks for the quick fix!
Just added one question.

test/integration/rest-basic/index.spec.ts
Janpot: We can, I was a bit hesitant to abstract it while having only one use-case so far

apedroferreira: Yeah, just a comment but probably not really necessary for now.

PR #1384, opened 2022-11-25
Create apps programmatically in integration tests

Janpot:
This speeds up the initialization, and provides faster feedback while writing tests. This seems to speed up the integration test suite in circleci from 178s to 121s


test/integration/propControls/index.spec.ts
```diff
@@ -1,18 +1,23 @@
...
 
 test('can control component prop values in properties control panel', async ({
   page,
   browserName,
+  baseURL,
 }) => {
-  const homeModel = new ToolpadHome(page);
+  invariant(baseURL, 'playwright must be run with a a baseURL');
+
+  const dom = await readJsonFile(path.resolve(__dirname, './domInput.json'));
```
apedroferreira, line 16: I think we're using `readJsonFile` in some tests but just importing the JSON file in others?
Should we do it in a single consistent way across all tests?

test/utils/toolpadApi.ts
```diff
@@ -0,0 +1,40 @@
...
+import generateId from './generateId';
+import type { CreateAppOptions } from '../../packages/toolpad-app/src/server/data';
+
+export function getApi(endpoint: string) {
+  return createRpcClient<ServerDefinition>(endpoint);
+}
+
+interface CreateApplicationParams {
+  name?: string;
+  appTemplateId?: string;
```
apedroferreira, line 13: I created an `AppTemplateId` type that maybe could be imported to use here?

test/utils/toolpadApi.ts
```diff
@@ -0,0 +1,40 @@
...
+  dom?: any;
+}
+
+export async function createApplication(
+  baseUrl: string,
+  { name = `App ${generateId()}`, appTemplateId, dom }: CreateApplicationParams = {},
+) {
+  const api = getApi(new URL('/api/rpc', baseUrl).href);
```
apedroferreira, line 21: Could this `api` variable not be global if the playwright `baseUrl` is always the same?
Just considering if it's possible in case `baseUrl` could be retrieved outside the tests, but I don't have the whole context.

apedroferreira (approved):
Looks good, I wrote some comments and questions but they're not a big deal.

test/integration/propControls/index.spec.ts
Janpot: We should use `readJsonFile` for this. There is not much benefit in having it statically typed, and it will overwhelm the type checker

test/utils/toolpadApi.ts
Janpot: I'm not sure a union type makes sense for this, it feels like data to me.

oliviertassinari (commented):
From what I understand, we change the way we seed tests. Instead of using the UI, we call the API directly. Sounds great.

test/utils/toolpadApi.ts
apedroferreira: That's fair, just mentioning it.

test/integration/propControls/index.spec.ts
apedroferreira: Ok, we can change everything to using `readJsonFile` then, no need for it to be in this PR though.

test/utils/toolpadApi.ts
Janpot: Can't make it global because the `baseURL` is not available globally unless we stop relying on the config as single source of truth. I reworked it though to make `api` available inside the fixture.

PR #1385, opened 2022-11-25
Client side templates

apedroferreira:
Use client-side queries in app templates in demo.
This would probably just be temporary until we can use server-side queries in the demo.


apedroferreira:
Merging as I've removed connections as discussed but let me know if you have any more comments.

PR #1397, opened 2022-11-30
Don't immediately open duplicated apps

Janpot:
Current implementation leaves the user out of control. They may want to 
* only duplicate, not edit
* duplicate and edit in the same tab
* duplicate and edit in a new tab

And we shouldn't make that choice for them

This PR shows a snackbar instead with a link to open the duplicated app. similar to how e.g. figma does duplication

Now the user can
* ignore the snackbar
* click open to open in the same tab
* click Cmd+open to open in a new tab

Additionally:
* tighten the types to make more ergomic API
* make sure the app is always defined in the dialogs to ensure closing animations look ok
* reset the form on dialog open
* remove some unused props
* autoselect proposed name in dialog

https://user-images.githubusercontent.com/2109932/204859712-1e1774fc-a747-486f-ab6c-47f483fe3172.mov

### DIFF
--- a/packages/toolpad-app/src/toolpad/AppEditor/PagePanel.tsx
+++ b/packages/toolpad-app/src/toolpad/AppEditor/PagePanel.tsx
@@ -49,16 +49,7 @@ export default function PagePanel({ appId, className, sx }: ComponentPanelProps)
             loading={Boolean(!app)}
           />
         )}
-        {app ? (
-          <AppOptions
-            app={app}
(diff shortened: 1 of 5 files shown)


bharatkashyap (approved):

bharatkashyap:
Much better UX! 

Only one doubt is about the discoverability of opening the duplicated app in a new tab with Control/Command + Open.

PR #1402, opened 2022-12-01
Add search to the apps overview

bharatkashyap:
- Visual demonstration 

https://user-images.githubusercontent.com/19550456/205130879-5266edaf-99b8-4f81-a61e-ab05b93a3f06.mov

### DIFF
--- a/packages/toolpad-app/src/theme.ts
+++ b/packages/toolpad-app/src/theme.ts
@@ -896,7 +896,7 @@ export function getThemedComponents(theme: Theme): { components: Theme['componen
             '& .MuiSvgIcon-root': {
               fontSize: defaultTheme.typography.pxToRem(18),
               color:
-                theme.palette.mode === 'dark' ? theme.palette.grey[300] : theme.palette.grey[500],
+                theme.palette.mode === 'dark' ? theme.palette.grey[400] : theme.palette.grey[500],
               marginRight: theme.spacing(0.2),
             },
           },
--- a/packages/toolpad-app/src/theme.ts
+++ b/packages/toolpad-app/src/theme.ts
@@ -712,18 +712,13 @@ export function getThemedComponents(theme: Theme): { components: Theme['componen
         defaultProps: {
           size: 'small',
         },
-        styleOverrides: {
-          root: {
-            backgroundColor:
-              theme.palette.mode === 'dark' ? theme.palette.primaryDark[900] : '#fff',
-          },
-        },
       },
       MuiToggleButton: {
         styleOverrides: {
           root: {
             textTransform: 'none',
             fontWeight: 500,
+            padding: theme.spacing(0.5),
             color:
               theme.palette.mode === 'dark' ? theme.palette.grey[300] : theme.palette.grey[700],
             borderColor:
@@ -891,10 +886,30 @@ export function getThemedComponents(theme: Theme): { components: Theme['componen
         defaultProps: {
           margin: 'dense',
         },
+        styleOverrides: {
+          root: {
+            fontSize: defaultTheme.typography.pxToRem(13),
+            '& .MuiInputBase-input': {
+              paddingTop: theme.spacing(0.7),
+              paddingBottom: theme.spacing(0.7),
+            },
+            '& .MuiSvgIcon-root': {
+              fontSize: defaultTheme.typography.pxToRem(18),
+              color:
+                theme.palette.mode === 'dark' ? theme.palette.grey[300] : theme.palette.grey[500],
+              marginRight: theme.spacing(0.2),
+            },
+          },
+        },
       },
       MuiInputLabel: {
         defaultProps: {
           margin: 'dense',
+        },
+        styleOverrides: {
+          root: {
+            fontSize: defaultTheme.typography.pxToRem(13),
+          },
         },
       },
       MuiListItem: {
--- a/packages/toolpad-app/src/toolpad/Apps/index.tsx
+++ b/packages/toolpad-app/src/toolpad/Apps/index.tsx
@@ -38,6 +38,7 @@ import DialogForm from '../../components/DialogForm';
 import type { Deployment } from '../../../prisma/generated/client';
 import ToolpadHomeShell from '../ToolpadHomeShell';
 import getReadableDuration from '../../utils/readableDuration';
+import useSearch from '../../utils/useSearch';
 import type { AppMeta } from '../../server/data';
 import useLocalStorageState from '../../utils/useLocalStorageState';
 import ErrorAlert from '../AppEditor/PageEditor/ErrorAlert';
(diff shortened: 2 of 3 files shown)


packages/toolpad-app/src/theme.ts
```diff
@@ -712,18 +712,13 @@ export function getThemedComponents(theme: Theme): { components: Theme['componen
         defaultProps: {
```
bharatkashyap, line 712: Some minor changes to the theme to make the search input look aligned with the rest of the UI

packages/toolpad-app/src/toolpad/Apps/index.tsx
```diff
@@ -619,6 +620,8 @@ export default function Home() {
 
   const AppsView = viewMode === 'list' ? AppsListView : AppsGridView;
 
+  const [AppsSearchField, filteredApps] = useSearch<AppMeta>(apps, 'Search apps');
```
bharatkashyap, line 623: The intention is for this hook to be reused across pages

packages/toolpad-app/src/utils/useSearch.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+function useSearch<T>(elements: T[], placeholder?: string): [JSX.Element, T[]] {
+  const [searchText, setSearchText] = React.useState('');
+
+  const handleSearchInput = React.useCallback((event: React.ChangeEvent<HTMLInputElement>) => {
+    setSearchText(event.target.value);
+  }, []);
+
+  const filteredElements = React.useMemo(() => {
+    return elements.filter((elem: any) =>
```
bytasv, line 20: I remember in some other PR it was mentioned that we should not use abbrevations and instead use full name
```suggestion
    return elements.filter((element: any) =>
      element.name.toLowerCase().includes(searchText.toLowerCase()),
```

packages/toolpad-app/src/utils/useSearch.tsx
```diff
@@ -0,0 +1,40 @@
...
+      elem.name.toLowerCase().includes(searchText.toLowerCase()),
+    );
+  }, [elements, searchText]);
+
+  return [
+    <TextField
```
bytasv, line 26: This pattern of returning both react component and filtered values as a pair looks pretty uncommon. Since the logic is so trivial, I'd personally consider keeping component itself inside the main component.

To separate filtering/searching logic, maybe we could keep this `useSearch` hook, but instead of returning component we would return only filtered results. Then the input for hook would be `allApps, searchValue`:
```
const [searchValue, setSearchValue] = useState('')
const filteredApps = useSearch(apps, searchValue)
...
<TextField key='search' value={searchValue} onChange={(event) => setSearchValue(event.target.value)}
```

Janpot: Don't create hooks until a clear pattern of reuse emerges. The logic can go straight into `toolpad/Apps/index.ts`. We will extract reusable patterns once we align implementation between apps and connections

prakhargupta1:
Q: In UX review last week, we briefly discussed showing apps in a table, in case we do that will it impact this search bar?

packages/toolpad-app/src/theme.ts
```diff
@@ -891,11 +886,31 @@ export function getThemedComponents(theme: Theme): { components: Theme['componen
...
           margin: 'dense',
         },
+        styleOverrides: {
+          root: {
+            fontSize: defaultTheme.typography.pxToRem(13),
```
Janpot, line 891: Will it be a problem that changes made for search input will affect all inputs?

bharatkashyap: Agreed, updated

Janpot: But this is just the same but applying to all TextFields. Is there a good reason that this has to be built inside of the theme overrides? I mean, there are a lot of non-search TextFields.

packages/toolpad-app/src/toolpad/Apps/index.tsx
```diff
@@ -738,7 +751,33 @@ export default function Home() {
             Apps
           </Typography>
           <FlexFill />
-          <Button onClick={() => setCreateDialogOpen(true)}>Create New</Button>
+          <TextField
```
apedroferreira, line 754: Is there really no other way to style this text field as we want to without these a bit ugly overrides? Can't we do it with just regular props? But I understand that if the intended design is imposing certain things there might be no other way to do it.

I also noticed that the icon and the text are slightly misaligned vertically, but maybe most people wouldn't even notice:
Screenshot 2023-01-04 at 10 24 54

packages/toolpad-app/src/toolpad/Apps/index.tsx
```diff
@@ -728,6 +729,18 @@ export default function Home() {
 
   const AppsView = viewMode === 'list' ? AppsListView : AppsGridView;
 
+  const [searchText, setSearchText] = React.useState('');
+
+  const handleSearchInput = React.useCallback((event: React.ChangeEvent<HTMLInputElement>) => {
+    setSearchText(event.target.value);
+  }, []);
+
+  const filteredApps = React.useMemo(() => {
```
apedroferreira, line 738: I think this works well, but as a short-term solution - long term we should have pagination when loading a list of apps, and the search should filter the results server-side.

packages/toolpad-app/src/toolpad/Apps/index.tsx
apedroferreira: At least we should extract these styles to a variable above the component, so that we don't mix them with the JSX logic.

apedroferreira (changes requested):
Looks solid!
Just left some comments about the styles, but it shouldn't be too much.

packages/toolpad-app/src/toolpad/Apps/index.tsx
bharatkashyap: > Is there really no other way to style this text field as we want to without these a bit ugly overrides? 

In the beginning I added these overrides to the default theme so as to make all text fields look more dense and match the buttons in the theme, but in the discussion [here](https://github.com/mui/mui-toolpad/pull/1402#discussion_r1047139408) we decided to go ahead with styling just this text field for now. 

> At least we should extract these styles to a variable above the component, so that we don't mix them with the JSX logic.

Agreed, I'll do this.

> I also noticed that the icon and the text are slightly misaligned vertically, but maybe most people wouldn't even notice:

Agreed, this was much less noticeable in the dark theme. I'll fix this

apedroferreira (approved):
Thanks for the changes Bharat, the test looks good too!

This input is still not aligned though: 
Screenshot 2023-01-05 at 12 26 17

I'll approve this already anyway as that's just a small detail.

bharatkashyap:
> Thanks for the changes Bharat, the test looks good too!
> 
> This input is still not aligned though: Screenshot 2023-01-05 at 12 26 17
> 
> I'll approve this already anyway as that's just a small detail.

All right, I'll merge this and fix the alignment in a future follow up

PR #1403, opened 2022-12-01
Update dependencies

Janpot:
Bisect out what works from https://github.com/mui/mui-toolpad/pull/1386

* Isolate `@sentry/next`. It messes in some way with the next.js config that breaks the build. To be investigated.
* In my opinion sentry does too much magic with our next.js build, it's not the first time I have to go forensics on them messing with webpack config.
* We're going to need to figure out dealing with the CSS in `monaco-editor`
* `chalk` can be updated now the scripts are ESM
* Disabled `swc` in `ts-node`. To re-enable when they fix the integration


apedroferreira (approved):
Looks good!

If Sentry is causing too much hassle we can maybe look into implementing it in a more manual way, without using the plugin in next.config.

PR #1414, opened 2022-12-05
Disable flaky undo spec

bytasv:
Since the spec randomly fails in CI, I've decided to disable it until I can find a way to fix it


Janpot (changes requested):
1. Why not leave it enabled, as a forcing function for us to fix?
2. Normally there is `test.skip` for this

_edit_

When I hover over `page.waitForTimeout(600);` I see:

> Note that `page.waitForTimeout()` should only be used for debugging. Tests using the timer in production are going to be flaky. Use signals such as network events, selectors becoming visible and others instead.

Could be related?

Janpot:
Superseded by https://github.com/mui/mui-toolpad/pull/1415

PR #1417, opened 2022-12-05
Undo/redo through different views

apedroferreira:
Switches to the correct view on undo/redo.
Also closes https://github.com/mui/mui-toolpad/issues/1460.

Video (undo starts at 1:36):

https://user-images.githubusercontent.com/10789765/208684464-925c5a2f-9981-40ca-83f6-f3fbfb353180.mov

Preview URL: https://toolpad-pr-1417.onrender.com/

**Summary of changes:**
- DOM state now stores current view + properties panel tab
- Adds methods to `domApi` to set current view and tab. View can also be changed with `domApi.update` if we want 1 action to represent a DOM change + view change, for example, which was needed in some situations.
- Uses DOM state as source of truth for current active view/tab
- Add test for undo between pages

There's many different approaches we could use for this logic and to extend/improve it. I think this a good starting point for further changes, even if it's not the final solution.
(Also we might still be able to use DOM diffs to get the current view automatically, just starting out with something simpler)

Continuing to follow the checklist in https://github.com/mui/mui-toolpad/issues/226
Relevant discussion reference: https://github.com/mui/mui-toolpad/pull/1225#discussion_r1007856833


.circleci/config.yml
```diff
@@ -1,6 +1,6 @@
 version: 2.1
 orbs:
-  aws-s3: circleci/aws-s3@3.0.0
```
Janpot, line 3: ok, what's going on here? This was already done in https://github.com/mui/mui-toolpad/pull/1413

apedroferreira: Bad merge of master i guess, let me check

apedroferreira: Ah, because this commit isn't based on master, I shouldn't have merged like this

apedroferreira: reverted

.gitignore
```diff
@@ -15,6 +15,9 @@ out/
 # production
 /build
 
+# typescript
+tsconfig.tsbuildinfo
```
apedroferreira, line 19: Every time I run `yarn check-types` it creates this file so I've added it here.

test/integration/undo-redo/index.spec.ts
```diff
@@ -42,42 +38,3 @@ test('test basic undo and redo', async ({ page, browserName, api }) => {
   // Redo should bring back text field
   await expect(canvasInputLocator).toHaveCount(3);
 });
-
-test('test batching quick actions into single undo entry', async ({ page, browserName, api }) => {
```
apedroferreira, line 46: is it ok to delete this test? we can add more comprehensive testing of undo/redo - perhaps one for changing views and one for the upcoming text input logic?

packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/index.tsx
```diff
@@ -245,12 +244,17 @@ export default function HierarchyExplorer({ appId, className }: HierarchyExplore
...
       const fragment = appDom.cloneFragment(dom, nodeId);
 
-      const updatedDom = appDom.addFragment(dom, fragment, node.parentId, node.parentProp);
-
       const newNode = appDom.getNode(fragment, fragment.root);
       const editorDomView = getNodeEditorDomView(newNode);
 
-      domApi.update(updatedDom, editorDomView || { kind: 'page' });
+      domApi.update((draft) => {
+        invariant(
```
Janpot, line 251: Is this still necessary if we wrap the whole function in `domAPi.update`?

apedroferreira: This invariant was outside `domApi.update` before, but if it's not inside the function it doesn't stop the type checker from complaining.
I just tried to write this function differently, but because I've made the second argument of `domApi.update` be the view we want to change to, that view needs to be set before calling `domApi.update`, so it needs to be outside of it, along with everything else that value depends on.

packages/toolpad-app/src/toolpad/DomLoader.tsx
```diff
@@ -50,6 +60,15 @@ export type DomAction =
       type: 'DOM_UPDATE';
       updater: (dom: appDom.AppDom) => appDom.AppDom;
       selectedNodeId?: NodeId | null;
+      view?: DomView;
```
Janpot, line 63: Why can't we make switching a view its own action? It will push a single entry on the stack, so...?

apedroferreira: I created an action just for switching a view, but for some actions, such as creating a new page or connection, the UX feels nicer if the switch and the creation happen in the same action. Other than a few cases like that they work well as separate actions.

apedroferreira: Right now I'm using the `currentView` in the DOM state as the source of truth for navigation routes such as connection or code components, and for the state in some modals such as the query editor...
We could make it work the other way around, and use the route or state as the source of truth for the current view, I've thought about that, I guess it should work...

apedroferreira: We need simultaneous DOM and view updates sometimes because for example when deleting a page: "delete page" and then "set default page as current view", when reversed on undo, would first set the deleted page as the current view (without it existing), and only then recreate it. I changed the way the view can be passed to the `.update` function to try to make it more readable.

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/RenderOverlay.tsx
```diff
@@ -965,83 +955,138 @@ export default function RenderOverlay({ canvasHostRef }: RenderOverlayProps) {
       const isDraggingOverLayoutSlot = dragOverSlot?.type === 'layout';
       const isDraggingOverElement = appDom.isElement(dragOverNode);
 
-      domApi.update((draft) => {
-        let parent = appDom.getParent(draft, dragOverNode);
+      domApi.update(
```
apedroferreira, line 958: This is all just formatting changes except the second argument passed to `domApi.update`.

bharatkashyap (approved):
Great effort!

apedroferreira:
This should be good to merge, will merge soon if there's no more feedback.
I was fixing some bugs before the break, it seems good from what I've tested now, if there's any more bugs we should find them later.

Issue #1268, opened 2022-11-03
Allow viewing the current deployment from Edit mode.


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

From the Edit view, there is no option to see the current deployed app.
It can be done from the Home view by clicking on the Open button. The same should also be supported from Edit mode.
This will help the user compare the current snapshot (preview) with the last deployed version.

How it will work:
When user goes to Deploy, he will get another option that will prompt them to see the current deployed app.

From Appsmith:
<img width="225" alt="Screenshot 2022-11-03 at 1 14 22 PM" src="https://user-images.githubusercontent.com/92228082/199669248-14c11ce3-afc5-4fd9-90f0-1915cdcc4a74.png">

PR #1423, opened 2022-12-06
Add latest deployment preview in editor

bytasv:
- Closes https://github.com/mui/mui-toolpad/issues/1268

When we have deployed version:
CleanShot 2022-12-08 at 17 33 19@2x

When we don't have deployed version yet:
image


packages/toolpad-app/src/toolpad/AppEditor/AppEditorShell.tsx
```diff
@@ -138,13 +140,16 @@ export interface ToolpadShellProps {
 
 export default function AppEditorShell({ appId, ...props }: ToolpadShellProps) {
   const domLoader = useDomLoader();
+  const releases = client.useQuery('getReleases', [appId]);
```
Janpot, line 143: Probably want to use `findLastRelease` to avoid overfetching

bytasv:
> How would you feel about a [split button](https://mui.com/material-ui/react-button-group/#split-button) for this?

What would we put in a button as a text so it's clear and short? I thought about text, but what I have in tooltip right now is too long, anything other than that seems too long or not clear enough (unless you got some ideas).
Im also thinking that the latest deployed version preview is not as important as a current version preview, hence giving it less visibility/importance via icon shouldn't be as bad (subjective opinion)
I honestly don't know what's the right way to present that as we don't have design, I could change to whatever if you can specify how you see it, or we can ask Gerda for design if needed

Btw, I also fixed `findLastRelease` - thanks for tip.

Janpot:
I had something like this in mind, but maybe @gerdadesign wants to chip in?

https://user-images.githubusercontent.com/2109932/206142925-a1987afd-4604-4e9d-ac56-ac88595354d8.mov

packages/toolpad-app/src/toolpad/AppEditor/AppEditorShell.tsx
```diff
@@ -159,14 +164,48 @@ export default function AppEditorShell({ appId, ...props }: ToolpadShellProps) {
...
+                    ? {
+                        // Workaround for https://github.com/mui/material-ui/issues/31210
+                        borderTopLeftRadius: 0,
+                        borderBottomLeftRadius: 0,
+                        marginLeft: '-1px',
+                      }
+                    : {}
+                }
+              >
+                <Tooltip title="Preview latest deployed version">
```
Janpot, line 203:
```suggestion
                <Tooltip title="open last deployed version">
```

What we currently call "preview" in Toolpad nomenclature is the version that you're currently editing, i.e. the button left to the deploy button. I'd not conflate the term "preview" with deployments.

bytasv:
> I had something like this in mind, but maybe @gerdadesign wants to chip in?

Oh, I actually had done this revision too, but decided against it thinking that it adds unnecessary click to expand dropdown. Another thing that I didn't like about it is that it holds just one menu item rendering the dropdown somewhat useless for the purpose 🤔
As for using just icon vs menu item as you shown here - IMO icon with tooltip is easy to learn once and makes it one-click away, where with dropdown, you would still have to learn that this function is actually behind this dropdown 🤷 

As for using icons that might not be so common - I also don't think that it should be a problem, even looking at figma's toolbar:
image

I couldn't tell what most of those icon buttons do at first glance unless I used it once or twice, after that I'd assume it settles in my memory and I know longer need to read the title and could just use visual clues (icons)

prakhargupta1:
I think Rocket icon should be easier to understand that it stands for deployment. Also, the primary action of the button is deployment so what Jan shared makes sense to me. 

Now the secondary action that we want to show is: Open the current deployed app. It has to be around this button as it is again related to deployment. The drop-down icon looks standard UX to me.

packages/toolpad-app/src/toolpad/AppEditor/AppEditorShell.tsx
```diff
@@ -159,14 +175,42 @@ export default function AppEditorShell({ appId, ...props }: ToolpadShellProps) {
...
+                </Button>
+                <Menu
+                  open={open}
+                  onClose={handleClose}
+                  anchorEl={anchorEl}
+                  anchorOrigin={{ horizontal: 'right', vertical: 'bottom' }}
+                  transformOrigin={{ horizontal: 'right', vertical: 'top' }}
+                >
+                  <MenuItem component="a" href={`/deploy/${appId}`} target="_blank">
+                    Preview latest deployed version
```
Janpot, line 208:
```suggestion
                    Open current deployed version
```

packages/toolpad-app/src/toolpad/AppEditor/AppEditorShell.tsx
```diff
@@ -159,14 +175,42 @@ export default function AppEditorShell({ appId, ...props }: ToolpadShellProps) {
...
+              color="primary"
+              onClick={handleCreateReleaseDialogOpen}
+            >
+              Deploy
+            </Button>
+            {isDeployed ? (
+              <React.Fragment>
+                <Button
+                  size="small"
+                  aria-controls={open ? 'split-button-menu' : undefined}
```
Janpot, line 192: `split-button-menu` doesn't exist

packages/toolpad-app/src/toolpad/AppEditor/AppEditorShell.tsx
```diff
@@ -138,13 +142,25 @@ export interface ToolpadShellProps {
 
 export default function AppEditorShell({ appId, ...props }: ToolpadShellProps) {
   const domLoader = useDomLoader();
+  const releases = client.useQuery('findLastRelease', [appId]);
```
Janpot, line 145: Should we display the error state?

bytasv: I'm thinking that maybe not? This is not a critical feature and if there is an error IMO it could be handled gracefully without even letting user know about that 🤔

Janpot: Let's show a message in the dropdown when `releases.error`.
I'd also rename "releases" to something else, it doesn't really contain multiple releases.

packages/toolpad-app/src/toolpad/AppEditor/AppEditorShell.tsx
```diff
@@ -159,14 +169,37 @@ export default function AppEditorShell({ appId, ...props }: ToolpadShellProps) {
...
+                  {...menuProps}
+                  anchorOrigin={{ horizontal: 'right', vertical: 'bottom' }}
+                  transformOrigin={{ horizontal: 'right', vertical: 'top' }}
+                >
+                  {!release.error ? (
+                    <MenuItem component="a" href={`/deploy/${appId}`} target="_blank">
+                      Open current deployed version
+                    </MenuItem>
+                  ) : (
+                    <MenuItem>{release.error as string}</MenuItem>
```
bytasv, line 197: @Janpot I'm not sure if that's the correct way to deal with `error` rendering - if I leave it without `as` TS complains that no overload was found matching the `unknown` which is the type for `error` by default.

Janpot: We have a utility function for this:
```tsx
errorFrom(release.error).message
```

packages/toolpad-app/src/toolpad/AppEditor/AppEditorShell.tsx
```diff
@@ -187,26 +181,21 @@ export default function AppEditorShell({ appId, ...props }: ToolpadShellProps) {
...
-                  onClose={handleClose}
-                  anchorEl={anchorEl}
+                  {...menuProps}
                   anchorOrigin={{ horizontal: 'right', vertical: 'bottom' }}
                   transformOrigin={{ horizontal: 'right', vertical: 'top' }}
                 >
-                  <MenuItem component="a" href={`/deploy/${appId}`} target="_blank">
-                    Open current deployed version
-                  </MenuItem>
+                  {!release.error ? (
```
Janpot, line 192: I'd use positive logic and switch the operands around. It's easier to reason about
```suggestion
                  {release.error ? (
```

bytasv: I've changed. But IMO positive logic in this case is showing currently deployed version even if it's not positive considering the condition itself 🤔

bytasv:
@mui/toolpad please review

Janpot (approved):

PR #1425, opened 2022-12-06
[WIP] Global connections part 1 

Janpot:
This part deals with creation and storage of global connections:

- [x] Add new connections table and API.
- [x] Rewrite google sheets authentication so that it can work in popup mode.
- [x] Connections are represented by params (unencrypted column) and secrets (encrypted column). Secrets are never passed to the frontend in plaintext. The frontend communicates changes to secrets through actions that modify the object on the server. (set, unset, noop)
- [ ] integration test


packages/toolpad-app/prisma/schema.prisma
```diff
@@ -96,3 +95,13 @@ enum DomNodeAttributeType {
   jsExpressionAction
   navigationAction
 }
+
+model Connection {
+  id         String   @id @default(cuid())
+  name       String
+  createdAt  DateTime @default(now())
+  editedAt   DateTime @default(now())
+  datasource String
```
apedroferreira, line 104: Would it be worth defining enums in the database for fields like `datasource`?

packages/toolpad-app/src/server/data.ts
```diff
@@ -570,3 +587,142 @@ export async function duplicateApp(id: string, name: string): Promise<AppMeta> {
     return newApp;
   }
 }
+
+export async function getConnections() {
+  if (config.isDemo) {
```
apedroferreira, line 592: Just a nit and I might have written this the same way in other functions but for stuff like this maybe we should just take one line with `if (config.isDemo) return [];`

packages/toolpad-app/prisma/schema.prisma
Janpot: Not if we ever plan to allow users to define datasources as plugins.

Same as for `AppTemplateId`, this is not a good candidate for an enum since it's data and subject to receive dynamic or user defined values at some point.

apedroferreira: yeah that's a good point i hadn't considered. makes sense

PR #1427, opened 2022-12-07
Add Autocomplete component

### LINKED ISSUE #1231: Support AutoComplete component
### Duplicates

### Latest version

### Summary 💡

https://mui.com/material-ui/react-autocomplete/

One of: https://github.com/mui/mui-toolpad/issues/78

### Examples 🌈

_No response_

### Motivation 🔦

_No response_

bharatkashyap:
- Closes #1231 
- Visual demonstration 

### Autocomplete component

- Adds a https://mui.com/material-ui/react-autocomplete/ component to the catalog 
- Exposes an `options` prop on the editor which is edited in the same format as the Select component
- Allows binding to any arbitrary JSON for the options, and for this purpose, accepts a `labelField` and a `valueField`

### Demo

- Creating a bound, nested set of autocomplete components with some constant data is achievable like this: 

   https://github.com/mui/mui-toolpad/assets/19550456/5e2b3430-edbc-41be-bbb3-f82343c89734
   
- A `stateData` query returns JSON with with the following shape:
    ```ts
    [
  {
    stateId: 1,
    state: "Maharashtra",
    districts: [
      {
        districtId: 101,
        district: "Mumbai",
        blocks: [
          { blockId: 1001, block: "Mumbai City" },
          { blockId: 1002, block: "Mumbai Suburban" },
        ],
      },
      ...
     ```
- The client side binding for the `Text` component above is like so:

    ```ts
    `Your selection is:  **${
      block.options.find((s) => s.blockId === block.value)?.block
    }**, **${
      district.options.find((d) => d.districtId === district.value)?.district
    }**, **${state.options.find((s) => s.stateId === state.value)?.state}**`
    ```
where `state`, `district` and `block` are the names of the three autocomplete component nodes.

### Caveats

- The options editor currently does not generalise well when using custom `labelField` and `valueField`. This would perhaps require some changes to the [ArgControlSpec ](https://github.com/mui/mui-toolpad/blob/master/packages/toolpad-core/src/types.ts#L188) and some generalisation to [SelectOptions.tsx](https://github.com/mui/mui-toolpad/blob/master/packages/toolpad-app/src/toolpad/propertyControls/SelectOptions.tsx)


packages/toolpad-components/src/Autocomplete.tsx
```diff
@@ -0,0 +1,61 @@
...
+    loading: {
+      typeDef: { type: 'boolean' },
+    },
+    disabled: {
+      typeDef: { type: 'boolean' },
+    },
+    sx: {
+      typeDef: { type: 'object' },
+      defaultValue: {
+        width: 300,
```
apedroferreira, line 57: Does this support full-width? We should if it's easy/possible.

packages/toolpad-components/src/Autocomplete.tsx
```diff
@@ -0,0 +1,61 @@
...
+    },
+    loading: {
+      typeDef: { type: 'boolean' },
+    },
+    disabled: {
+      typeDef: { type: 'boolean' },
+    },
+    sx: {
+      typeDef: { type: 'object' },
+      defaultValue: {
```
apedroferreira, line 56: We should support full-width if possible (sorry, deleted my previous comment which was the same!)

apedroferreira (approved):
Looks great! I only have that one comment about width.

packages/toolpad-app/public/schemas/AutocompleteOptions.json
```diff
@@ -0,0 +1,13 @@
+{
+  "type": "array",
+  "items": {
+    "type": ["object", "string"],
+    "additionalProperties": true,
+    "properties": {
+      "label": {
```
apedroferreira, line 7: Wait, shouldn't options be able to have an `id` or value` different than the label?

packages/toolpad-components/src/Autocomplete.tsx
```diff
@@ -0,0 +1,61 @@
...
+      {...rest}
+    />
+  );
+}
+
+export default createComponent(Autocomplete, {
+  layoutDirection: 'both',
+  argTypes: {
+    options: {
+      typeDef: { type: 'array' },
```
Janpot, line 31: Should this be the same as the select component?

packages/toolpad-components/src/Autocomplete.tsx
Janpot: I don't think `sx` should have a `defaultValue` at all. On any of the components

apedroferreira: Oh, I agree, I missed that detail.

packages/toolpad-components/src/Autocomplete.tsx
bharatkashyap: I tried both ways and realised that this works better, since the `options` prop for the Autocomplete component is more flexible about what it can take as an input. A series of strings, or an array of objects with a `label` prop. In this scenario, a simple JSON Editor works better in my opinion

packages/toolpad-components/src/Autocomplete.tsx
bharatkashyap: Agreed, updated

packages/toolpad-components/src/Autocomplete.tsx
Janpot: > A series of strings, or an array of objects with a `label` prop.

Isn't that the same as the select component?. Seems to work when I bind to it.

bharatkashyap: The Select component treats the string as `values`, not `label`s

Janpot: I can bind the following to a Select and it recognizes all of these as a valid option:
```ts
[{ value: "foo", label: "Foo" }, "bar", { value: "baz" }]
```

bharatkashyap: I was referring to this issue when using the UI of the Select options editor:

https://user-images.githubusercontent.com/19550456/207851957-f24b8b06-6891-4148-abdc-4439f70dda51.mov

The `value` prop should be the `value` attribute of the option, not the `label`

Janpot:
I still feel like the options prop of the autocomplete and select components should be identical. Conceptually, my expectation as a user is these components are interchangeable. Is there anything preventing us from aligning these options types?

packages/toolpad-components/src/Autocomplete.tsx
```diff
@@ -0,0 +1,150 @@
...
+      type: 'string',
+      onChangeProp: 'onChange',
+      default: '',
+    },
+    label: {
+      helperText: 'The label to display for the autocomplete.',
+      type: 'string',
+      default: 'Search…',
+    },
+    labelField: {
```
Janpot, line 118: Why is this needed? Can we just start with expecting from users to supply correct `AutocompleteOption`?

packages/toolpad-components/src/Autocomplete.tsx
```diff
@@ -0,0 +1,150 @@
...
+      helperText: 'The field to use as the label for each option.',
+      type: 'string',
+    },
+    valueField: {
```
Janpot, line 122: idem

packages/toolpad-components/src/Autocomplete.tsx
bharatkashyap: Without this, we wouldn't be able to support the Autocomplete docs examples; a lot of the times, the data is in a particular shape and being able to specify which field is to act as the `label` and which is to be the `value` is a very useful feature. I would do it very frequently with React Admin; in fact the example in the video is what I would spend a lot of time doing again and again on multiple admin tools through code.

Janpot: > Without this, we wouldn't be able to support the use case where the data is in a particular shape and being able to specify which field is to act as the label and which is to be the value is needed.

I believe you could bind to:
```tsx
customOptions.map(option => ({ label: option.notLabel, value: option.notvalue }))
```

> Without valueField, I would first have to transform the data I am connecting to Toolpad, and only then use the Autocomplete with it.

Yes, I propose we start with that use case

bharatkashyap: Got it; I'll remove them and add these as a proposal in the Product discussion notes

Janpot (approved):

PR #1429, opened 2022-12-07
Fix URL for Default template

bharatkashyap:
Replace the `localhost` URL in the `default.json` file with a path to the remote asset

### DIFF
--- a/packages/toolpad-app/src/server/appTemplateDoms/default.json
+++ b/packages/toolpad-app/src/server/appTemplateDoms/default.json
@@ -13,7 +13,7 @@
           "value": {
             "url": {
               "type": "const",
-              "value": "http://localhost:3000/static/employees.json"
+              "value": "https://demo.toolpad.io/static/employees.json"
             },
             "method": "GET",
             "browser": true,


packages/toolpad-app/src/server/appTemplateDoms/default.json
```diff
@@ -13,7 +13,7 @@
           "value": {
             "url": {
               "type": "const",
-              "value": "http://localhost:3000/static/employees.json"
+              "value": "https://demo.toolpad.io/static/employees.json"
```
apedroferreira, line 16: could we do it like the movies are doing it?
```
export const MOVIES_API_DEMO_URL = new URL('/static/movies.json', config.externalUrl).href;
```

apedroferreira (changes requested):

packages/toolpad-app/src/server/appTemplateDoms/default.json
apedroferreira: ah we can't, this works for now.

apedroferreira (approved):

packages/toolpad-app/src/server/appTemplateDoms/default.json
bharatkashyap: The issue is that it needs to live inside the `default.json` file, so we would need to do some sort of string-replace at build-time?

apedroferreira: Eventually we should either generate the template from `config.externalUrl`, or serve these files from somewhere that we can access from anywhere.

bharatkashyap: > or serve these files from somewhere that we can access from anywhere.

`https://demo.toolpad.io/static` serves this purpose?

apedroferreira: Once we have our own definitive cloud domain I guess that should be fine, so until then demo should work.

bharatkashyap: This fix doesn't work for local deployments because of a CORS issue for on the remote static file. The other option I could think of is to create two separate versions of the demo dom, one pointing to local and one to `https://demo.toolpad.io`, but that wouldn't work on our `master--toolpad.mui.com` deployments.

Thinking of an alternative fix

bharatkashyap: This can be done by writing a migration (`v4.ts`) that applies that string replacement on the default dom at build time. Should work, I'll also add a test to make sure this works.

apedroferreira: I guess we could just use `require` for dynamic templates instead of `readJsonFile`. Let me write a suggestion maybe.

apedroferreira: Something like this? Would probably need some cleanup and adjusting, I can write a PR too:

```
import * as path from 'path';
import * as appDom from '../../appDom';
import { migrateUp } from '../../appDom/migrations';
import { AppTemplateId } from '../../types';
import { readJsonFile } from '../../utils/fs';
import config from '../config';
import projectRoot from '../projectRoot';

const DOMS_DIR_PATH = './src/server/appTemplateDoms';

const APP_TEMPLATE_DOM_PATHS: Record<AppTemplateId, string | null> = {
  blank: {
    path: null,
    isDynamic: false,
  ,
  images: {
    path: path.resolve(
      projectRoot,
      DOMS_DIR_PATH,
      // @TODO: Remove demo template once demo supports server-side queries
      config.isDemo ? './images-demo.json' : './images.json',
    ),
    isDynamic: false,
  }
  default: {
    path: path.resolve(projectRoot, DOMS_DIR_PATH, './default.json'),
    isDynamic: true,
  }
};

export async function getAppTemplateDom(
  appTemplateId: AppTemplateId,
): Promise<appDom.AppDom | null> {
  const appTemplateDom = APP_TEMPLATE_DOM_PATHS[appTemplateId];
  if (!appTemplateDom) return null
  
  const dom = appTemplateDom.isDynamic ? require(appTemplateDom.path) : await readJsonFile(appTemplateDom.path);
  return dom ? migrateUp(dom) : null;
}

Comment (apedroferreira) [reply packages/toolpad-app/src/server/appTemplateDoms/default.json:16]: > This makes sense longer term though. For now, I think the fix lies in writing a migration since that gives us a fix for all apps that have already been created as a free side-effect

Sounds good if we need to fix the already created apps.
I'll make a PR with dynamic templates, shouldn't be too difficult and then we can use better solutions in the future.

Comment (apedroferreira) [reply packages/toolpad-app/src/server/appTemplateDoms/default.json:16]: Here's my idea, seems to work https://github.com/mui/mui-toolpad/pull/1430

File: packages/toolpad-app/src/appDom/migrations/v4.ts [hunk 2]
```diff
@@ -0,0 +1,40 @@
...
+      ...node,
+      attributes: {
+        ...node.attributes,
+        query: {
+          ...node.attributes.query,
+          value: {
+            ...node.attributes.query.value,
+            url: {
+              type: 'const',
+              value: DEFAULT_TEMPLATE_DATA_URL,
```
apedroferreira, line 20: We should probably only replace the URL if it's using `/static/employees.json` specifically - so that if someone creates a query called `getUser` (pretty common name also) it doesn't get replaced.

bharatkashyap: Yep, good catch there

Janpot: I feel like this is not what migrations are for. Can't we just replace it in the template?

Janpot: e.g. when we load the default template
```tsx
const defaultTemplate = await readJsonFile(path.resolve(__dirname, './default.json'))
defaultTemplate.nodes['1313wn3'].attributes.query.value.url.value = DEFAULT_TEMPLATE_DATA_URL
return defaultTemplate
```
and be done with it

Janpot (approved):

PR #1430, opened 2022-12-07
Support dynamic app templates

apedroferreira:
Allows dynamic app templates that we can pass Javascript variables to.
Also fixes default template for multiple environments.

Had this idea from the discussion in https://github.com/mui/mui-toolpad/pull/1429, should also be a long-term fix for these types of issues.

### DIFF
--- a/packages/toolpad-app/src/server/appTemplateDoms/doms.ts
+++ b/packages/toolpad-app/src/server/appTemplateDoms/doms.ts
@@ -1,3 +1,4 @@
+import { NodeId } from '@mui/toolpad-core';
 import * as path from 'path';
 import * as appDom from '../../appDom';
 import { migrateUp } from '../../appDom/migrations';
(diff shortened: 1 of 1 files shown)


bharatkashyap (approved):

bharatkashyap (approved):

Janpot:
This makes it impossible to update the template by importing it in toolpad, changing and exporting again.
Let's keep it simple? We could just remove the map with paths and replace it with functions that load the correct template

```tsx
const TEMPLATES = {
  defaultTemplate: async () => {
    const tpl = await readJson(path.resolve(__dirname, './default.json'))
    tpl.nodes['1313wn3'].attributes.query.value.url.value = DEFAULT_TEMPLATE_DATA_URL
    return tpl
  },
  imagesTemplate = async () => readJson(path.resolve(__dirname, './images.json'))
}

```

apedroferreira:
> This makes it impossible to update the template by importing it in toolpad, changing and exporting again. Let's keep it simple? We could just remove the map with paths and replace it with functions that load the correct template
> 
> ```tsx
> const TEMPLATES = {
>   defaultTemplate: async () => {
>     const tpl = await readJson(path.resolve(__dirname, './default.json'))
>     tpl.nodes['1313wn3'].attributes.query.value.url.value = DEFAULT_TEMPLATE_DATA_URL
>     return tpl
>   },
>   imagesTemplate = async () => readJson(path.resolve(__dirname, './images.json'))
> }
> ```

That works too and has that advantage, I'll change it now.

apedroferreira:
Updated for new approach suggested by Jan.

bharatkashyap (approved):

PR #1442, opened 2022-12-12
Document and cleanup of utility functions

Janpot:
* Add documentation to most functionality in the `./utils` folder. 
* Remove some dead code.


packages/toolpad-app/src/utils/collections.ts
```diff
@@ -31,20 +39,29 @@ export function mapProperties<U, V>(
   );
 }
 
+/**
```
apedroferreira, line 42: Doesn't hurt to have these descriptions, but i feel like a few (not all) of these utilities are almost self-explanatory and don't really need comments? So maybe we don't need to enforce that every utility method always has comments (it's fine to keep them now though), and it's ok to rely on self-explanatory method names?
Also have we decided definitively if we want to maintain our own utilities or rely in an external library? I'm fine with either (maybe slightly more in favor of using a library), but we should just find a consensus I guess.

apedroferreira (approved):
Nice work, should be pretty useful, thanks a lot!

packages/toolpad-app/src/utils/collections.ts
Janpot: > Also have we decided definitively if we want to maintain our own utilities or rely on an external library (just for the typical stuff that for example `lodash` already does)? I'm fine with either (maybe slightly more in favor of using a library), but we should just find a consensus I guess.

We will just keep making informed decisions based on the use-case at hand, the available options, our experience, the maintenance cost required, etc... and not fall in the trap of taking on dogmas around this. Making the decision of using a library that does 85% of what you need may as well be detrimental to your overall codebase if the other 15% turn into a maintenance hell. Almost all of functions that are in the utils are there because of shortcomings in available options.

PR #1443, opened 2022-12-12
Fix yarnlock

bytasv:
Seems that after recent merge master failed because of yarn lock file


apedroferreira (approved):

bytasv:
> Yes, I already corrected in #1442 as well, but fine for me to release separately

Same, I fixed that in some other branch, but just noticed that master started failing so though it would be good to fix asap

PR #1444, opened 2022-12-12
Revert react-router-dom to 6.3

Janpot:
Second stab at https://github.com/mui/mui-toolpad/pull/1426

Not sure why renovatebot reverted this in https://github.com/mui/mui-toolpad/pull/1435 but let's assume `resolutions` will prevent this in the future.

Nevertheless, I added a long overdue code components editor smoke test which covers the regression as well.


packages/toolpad-app/src/toolpad/AppEditor/CodeComponentEditor/index.tsx
```diff
@@ -246,6 +246,7 @@ function CodeComponentEditorContent({ codeComponentNode }: CodeComponentEditorCo
           <SplitPane split="vertical" allowResize size="50%">
             <TypescriptEditor
               value={input.attributes.code.value}
+              data-testid="codecomponent editor"
```
Janpot, line 249: There didn't seem to be an easy way to make the editor accessible

bytasv (approved):

PR #1457, opened 2022-12-15
Fix display of errors during loading of data

Janpot:
Our expressions evaluate to values with 3 states:
* loading
* error
* resolved

When the `rows` property of the datagrid uses any value that is in loading state, it itself should resolve as "loading" and show the grid in loading state instead of error.

I sharpened the integration test to detect the behavior

before:

https://user-images.githubusercontent.com/2109932/207881467-b344ed8b-ea00-44ae-9377-030d74a8bb84.mov

after:

https://user-images.githubusercontent.com/2109932/207881536-0baea6cd-599b-4d2b-8ec8-065f86b735f7.mov

### DIFF
--- a/test/integration/rest-basic/index.spec.ts
+++ b/test/integration/rest-basic/index.spec.ts
@@ -29,44 +29,20 @@ test('rest basics', async ({ page, browserName, api }) => {
   httpbinConnection.attributes.params.value.baseUrl = HTTPBIN_BASEURL;
 
   const runtimeModel = new ToolpadRuntime(page);
-  // eslint-disable-next-line no-console
-  console.log(0);
   await runtimeModel.gotoPage(app.id, 'page1');
-  // eslint-disable-next-line no-console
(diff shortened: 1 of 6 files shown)


test/integration/rest-basic/index.spec.ts
```diff
@@ -16,7 +16,13 @@ if (customHttbinBaseUrl) {
 
 const HTTPBIN_BASEURL = customHttbinBaseUrl || 'https://httpbin.org/';
 
-test.use({ ignoreConsoleErrors: [/.*/] });
+test.use({
+  ignoreConsoleErrors: [
+    // For some reason this shows error in chrome in CI only
+    // Needs to be investigated
+    /Invariant Violation: canvas ref not attached/,
```
apedroferreira, line 23: I've had this error in dev, haven't seen it lately though...

apedroferreira (approved):

Issue #1174, opened 2022-10-18
Support custom column type in data grid


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Inside a data grid, a custom column type would allow a user to create custom components and use them there.

### Examples 🌈

Use cases: 

- render https://mui-plus.vercel.app/components/Sparkline inside the cells.
- render bars like in Metabase: https://mui.metabaseapp.com/question/8-mui-store-biggest-orders
  <img width="106" alt="Screenshot 2022-10-20 at 13 26 12" src="https://user-images.githubusercontent.com/3165635/196936211-87c362d5-9b0b-4113-b328-d4fa56e651a9.png">
- render status order like https://www.zendesk.com/marketplace/apps/support/273938/woocommerce-by-zenplates/.
  <img width="81" alt="Screenshot 2022-10-20 at 13 26 52" src="https://user-images.githubusercontent.com/3165635/196936312-eea8081d-103f-48fd-8d3d-b894d166add9.png">

PR #1462, opened 2022-12-16
Custom column types

bytasv:
- Closes https://github.com/mui/mui-toolpad/issues/1174
- Visual demonstration

CleanShot 2023-01-04 at 17 00 11

1. User can configure column type and choose `Custom component` now
2. Upon choosing `customComponent` type, user can select which custom component shall be used for rendering

CleanShot 2022-12-16 at 16 02 15@2x

3. Current implementation expects that `value`, `row` or `field` props will be used to pass values to the component

CleanShot 2023-01-04 at 17 10 44

CleanShot 2023-01-04 at 17 02 07


packages/toolpad-app/src/toolpad/propertyControls/GridColumns.tsx
```diff
@@ -64,6 +76,15 @@ function GridColumnsPropEditor({
   const { bindings } = usePageEditorState();
   const [editColumnsDialogOpen, setEditColumnsDialogOpen] = React.useState(false);
   const [editedIndex, setEditedIndex] = React.useState<number | null>(null);
+  const { dom } = useDom();
+  const toolpadComponents = useToolpadComponents(dom);
```
apedroferreira, line 80: Would a `useCustomComponents` hook possibly be useful in the future? Just thinking it could make things a bit more readable and reusable, but if we don't have widespread use it for it yet we can also just start like this!

apedroferreira (changes requested):
Looks great Vytas, sorry for the late review!
Just one thing: when I pick a column type as `customComponent` but I haven't picked which component yet, there's an error that also shows in the UI. Maybe we could make this part of the experience smoother somehow?

Screenshot 2022-12-20 at 09 52 54

Janpot (changes requested):
I'm not sure this is the interface we should be going for. I think if we put forefront that we're about integrating with pro-code workflow, we need to present an interface that custom components can adhere to as well. It should be easy to convert your pro-code component in a Toolpad compatible one simply by wrapping it with `createComponent`. It shouldn't require any knowledge about Toolpad specifc things in its implementation (for runtime behavior).

I propose we make something like the following work first, and build the custom column types on top of that.

```tsx
import { createComponent } from '@mui/toolpad-core';

interface MyWrapperProps {
  columns: {
    CustomComponent: React.ComponentType;
  }[];
}

export default createComponent(
  function MyWrapper({ columns }: MyWrapperProps) {
    const Comp = columns[0].CustomComponent;
    return <Comp />;
  },
  {
    argTypes: {
      columns: {
        typedef: 'array',
        items: {
          type: 'object',
          properties: {
            customComponent: { type: 'component' },
          },
        },
      },
    },
  },
);

```

bytasv:
> I propose we make something like the following work first, and build the custom column types on top of that.

It's nowhere near in the original scope of making Grid component supporting custom components, so I disagree with proposal. We can do that, but it should be a separate PR and separate task in general.

I'm also not sure how the proposed approach should be used within the current (newly implemented) UI/UX of custom component picker in grid options.

bytasv:
@apedroferreira thanks for great catch!

> Just one thing: when I pick a column type as customComponent but I haven't picked which component yet, there's an error that also shows in the UI. Maybe we could make this part of the experience smoother somehow?

I've added fallback - when this happens we simply render `No component selected` in the cell instead of crashing render. It could probably be improved overall, but should do for now 🤔 

@Janpot please re-review and leave feedback relevant to the original scope of the task

apedroferreira (approved):
Looks good! 
If I wanted to nitpick: the "No component selected" maybe could be a bit greyed out compared to normal text in the table, or be different than normal text somehow, but anyway this works! :D

bytasv:
> Looks good! If I wanted to nitpick: the "No component selected" maybe could be a bit greyed out compared to normal text in the table, or be different than normal text somehow, but anyway this works! :D

Hmm, that got me thinking - it feels that this state is more like an error-state, so I did this: 
CleanShot 2022-12-20 at 14 24 11

WDYT?

apedroferreira:
> WDYT?

Seems good, I also thought about it being red so that pale red should be a good choice.

packages/toolpad-app/src/toolpad/propertyControls/GridColumns.tsx
```diff
@@ -26,14 +26,26 @@ import {
...
-const COLUMN_TYPES: string[] = ['string', 'number', 'date', 'dateTime', 'boolean', 'link', 'image'];
+const COLUMN_TYPES: string[] = [
+  'string',
+  'number',
+  'date',
+  'dateTime',
+  'boolean',
+  'link',
+  'image',
+  'customComponent',
```
Janpot, line 47:
```suggestion
  'codeComponent',
```

Nomenclature so far everywhere else has been "code component"

packages/toolpad-components/src/DataGrid.tsx
```diff
@@ -285,11 +290,35 @@ const DataGridComponent = React.forwardRef(function DataGridComponent(
...
+
+          if (!Component) {
+            return (
+              <Typography variant="overline" sx={{ color: 'error.main', fontSize: '10px' }}>
+                No component selected
+              </Typography>
+            );
+          }
+
+          return <Component {...value} />;
```
Janpot, line 316: This doesn't seem right, do you really mean to spread the value as props?

packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -322,15 +322,18 @@ function RenderedNodeContent({ node, childNodeGroups, Component }: RenderedNodeC
...
+  const components = useComponents();
+
   const props: Record<string, any> = React.useMemo(() => {
     return {
       ...boundProps,
       ...onChangeHandlers,
       ...eventHandlers,
       ...layoutElementProps,
       ...reactChildren,
+      __toolpadComponents: components,
```
Janpot, line 334: I'm in favor of adding a context

Janpot:
> It's nowhere near in the original scope of making Grid component supporting custom components, so I disagree with proposal.

So far, the idea has been that any component we offer out of the box, a user must be able to build it as a custom component. i.e. no superpowers. Essentially this PR changes the scope of Toolpad.

This PR needs an integration test

packages/toolpad-components/src/DataGrid.tsx
bytasv: yes, because the value is what column has as a value in an array we pass to data grid

packages/toolpad-app/src/runtime/ToolpadApp.tsx
bytasv: React context or comment context? Can you ellaborate?

bytasv:
> So far, the idea has been that any component we offer out of the box, a user must be able to build it as a custom component. i.e. no superpowers.

Can you explain with specific example? What does `superpower` in this PR mean? And what specifically we can no longer build as a custom component?

Janpot:
> What does superpower in this PR mean?

That the built-in components can do things (in the runtime) that custom components can never achieve. e.g. pass a custom component in a property.

> And what specifically we can no longer build as a custom component?

I never said that we suddenly can't build certain things anymore with custom components. The idea was that you could build as powerful customizations as the built-in components, if you suddenly give the built-ins superpowers, then that promise doesn't hold anymore.

bytasv:
> That the built-in components can do things (in the runtime) that custom components can never achieve. e.g. pass a custom component in a property.

Sorry I still don't understand what you mean - can you give a concrete example of what you have in mind?

Maybe I'm missing the point but is this any different than how we render `link`, `image` components for that data type? or how we do custom number formatting via configuration?

Janpot:
> Maybe I'm missing the point but is this any different than how we render link, image components for that data type? or how we do custom number formatting via configuration?

All of these can be (and some even have been) built as a custom component. With the exact same functionality.

bytasv:
> All of these can be (and some even have been) built in a custom component. With the exact same functionality.

What exactly has been built as a custom component? Sorry I still don't understand. Can you illustrate with a specific example?

Do we have a specific requirement right now that this implementation breaks or is this something that we think we could do in the future and because of that this implementation is not correct?

Janpot:
So far, the built-in component library has been built on top of the same public API interface that we expose to custom components. The built-in components don't have access to private APIs that the custom components can't access. The idea behind that is that users of Toolpad would be able to build a just as powerful component library as the built-in one.

Now this PR adds some magic private API that only the built-in components can access and use. aka, it has gained superpowers compared to what custom components can do.

> Do we have a specific requirement right now that this implementation breaks or is this something that we think we could do in the future and because of that this implementation is not correct?

I feel like often, concerns I raise are being rephrased in terms of "possible" vs. "impossible". Or "correct" vs. "incorrect". but it's not always a dichotomy. When we build features we often also implicitly make choices on the overall direction of the project. Or on the future maintainability. Maybe the way we build them closes the ticket, but doesn't really set us on a path forward. It's perfectly valid to raise concerns about that, closing tickets is important, but so is code quality.

bytasv:
> The built-in components don't have access to private APIs that the custom components can't access. The idea behind that is that users of Toolpad would be able to build a just as powerful component library as the built-in one.

I don't really see how current implementation prevents us to extend current codebase in the future so that we can access custom components from other custom components.

We do currently have a task which requires to render custom component in a grid and this PR does exactly that. ✅ 

At the moment we don't have neither task nor RFC for custom components import inside custom component code (which you indicate as a reason for this PRs rejection). If there was such requirement, I would choose a different implementation from the one you showed - if we were to offer such import functionality IMO it should work same way as we import any other component, i.e.:
```
import { CustomComponent } from '@mui/toolpad-components' 
```
I do expect that implementing this on top of my current implementation would be possible without any overhead.

If you still think current implementation warrants rejection and can not be moved forward, please do specify in detail the task for custom component import so it's clear what's the expected behaviour, interface, approach or anything else that could unblock this PR.

packages/toolpad-app/src/toolpad/propertyControls/GridColumns.tsx
bytasv: I'd prefer to wait if we see a need, it would make it easier to know if the used abstraction makes sense 🤔

Janpot:
> I don't really see how current implementation prevents us to extend current codebase in the future so that we can access custom components from other custom components.

 The goal is for us to provide a `createComponent` that users can wrap around their existing pro-code components, and configure the properties with it. The custom component will have a `React.ComponentType` property, not a string id property. This PR puts the responsibility of resolving a string id to a `React.ComponentType` inside of the custom component. I believe the component shouldn't be responsible for resolving strings to custom components, it shouldn't have knowledge of toolpad internals, because that makes it tougher to share between pro- and low-code.

> We do currently have a task which requires to render custom component in a grid and this PR does exactly that.

I think we can make shaping the feature a part of "the task". I wouldn't treat the issue tracker as "I pick a ticket, hack myself through the minimal possible solution, ship it and go home". We're not an agency.

> At the moment we don't have neither task nor RFC for custom components import inside custom component code

I think that's orthogonal to this PR, this PR is about passing a component as a property. Not about importing components inside of components.

bytasv:
> this PR is about passing a component as a property. 

no it's not, it's about getting custom component rendered inside Grid

> Which we should probably tackle at some point as well.

Then we can tackle it how you think it should be tackled, but not in the scope of this PR.

We can shape the implementation further in separate PR, but I don't see the point in changing what's already implemented.

Have you even tested the code? Do you have feedback relevant to the current implementation?

packages/toolpad-app/src/runtime/ToolpadApp.tsx
Janpot: Move the components context to `@mui/toolpad-core` and reuse instead of passing this prop

bytasv: Prop removed and replaced with context

bytasv:
> This PR needs an integration test

Test added

bytasv:
@Janpot tests have been added, props changed with context. Any other feedback as you are now blocking this to be merged?

packages/toolpad-core/src/runtime.tsx
```diff
@@ -111,7 +120,7 @@ export function NodeRuntimeWrapper({
             componentConfig,
           }}
         >
-          {children}
+          <ComponentsProvider value={components}>{children}</ComponentsProvider>
```
Janpot, line 123: There already is a global provider in the runtime, the one that is used for `useComponents`. My suggestion was to move that one to `@mui/toolpad-core`, so that we can also use `useComponents` inside of `@mui/toolpad-components`.

I don't see why we need to recreate that context inside of every `NodeRuntimeWrapper` if there already is a global contetx with the same value

Janpot:
@bytasv 
> 3. Current implementation expects that value of grid column will be passed as a spread props on a custom component:

I don't think we should do this. This will make it impossible to add more props. I think we should have a `value` prop that gets the value of the column, and a `row` prop that gets the full row, and a `field` prop. This would allow users to combine multiple data fields into one. e.g. `${row.firstName} ${row.lastName}`

bytasv:
> I don't think we should do this. This will make it impossible to add more props.

I don't understand, can you explain how come?

Please explain your suggestion in more detail as it's a bit unclear how you see it. Best would be to have data source example, component definition that you'd have in custom components followed with steps how all is bound using column editor. Thank you

Janpot:
With row data

```tsx
[
  { field1: 'foo', field2: 'bar' }
]
```

create a column for `field1` and assign code component `Field1`:

```tsx
import { createComponent } from "@mui/toolpad-core";
import * as React from "react";

export interface statusProps {
 value: string;
}

function status({ value }: statusProps) {
  return <div>{value} passed</div>;
}

export default createComponent(status, {
  argTypes: {
    value: {
      typeDef: { type: "string" },
    },
  },
});
```

The `value` passed in the component should be `'foo'`. If `field1` were an object, then that full object should be passed in the `value` prop, not spread as props.

Additionally to the `value` prop, we should add `row` and `field` props that contain the full row and the field name respectively.

test/integration/data-grid/codeComponentCell.json
```diff
@@ -0,0 +1,96 @@
...
+  "nodes": {
+    "el23z1k": {
+      "id": "el23z1k",
+      "name": "Test",
+      "type": "codeComponent",
+      "parentId": "op360414",
+      "attributes": {
+        "code": {
+          "type": "const",
+          "value": "import { createComponent } from \"@mui/toolpad-core\";\nimport * as React from \"react\";\n\nexport interface statusProps {\n  value: string;\n}\n\nfunction status({ value }: statusProps) {\n  return <div>{value} passed</div>;\n}\n\nexport default createComponent(status, {\n  argTypes: {\n    value: {\n      typeDef: { type: \"string\" },\n      defaultValue: \"Hello\",\n    },\n  },\n});\n"
```
Janpot, line 12: nit: `statusProps` and `status` should be capitalized

packages/toolpad-core/src/runtime.tsx
bytasv: Alright my bad, I'm not sure why initially I got errors complaining about missing provider, but I tried it again and all seems to just work fine now wihtout the need for extra providers 👍

bytasv:
@Janpot context moved, types fixed, description updated, test updated, seems that should be all 🤔

packages/toolpad-components/src/DataGrid.tsx
```diff
@@ -289,7 +292,35 @@ const DataGridComponent = React.forwardRef(function DataGridComponent(
...
   const nodeRuntime = useNode<ToolpadDataGridProps>();
+  const columnTypes = React.useMemo(
+    () => ({
+      ...CUSTOM_COLUMN_TYPES,
+      codeComponent: {
+        renderCell: (params: GridRenderCellParams) => {
+          const { value, colDef, row, field } = params;
+          const column = colDef as SerializableGridColumn;
+
+          invariant(components, 'Components must be defined');
```
Janpot, line 306: Why wouldn't this be defined? We use `createProvidedContext`

bytasv: You're right, I think it might be a leftover from previous implementation when TS complained about possibility of undefined components, but that's certainly incorrect. Removed 👍

bytasv:
@Janpot added boundry and handling for such cases 
CleanShot 2023-01-05 at 09 52 20@2x

Janpot (approved):
Can you open a new ticket around the poor debuggability of having react run in production mode?

bytasv:
@Janpot PTAL if that's what you meant for this one: https://github.com/mui/mui-toolpad/issues/1516

PR #1464, opened 2022-12-16
Group scope variables by function in global scope explorer

Janpot:
Working on getting meta info of page state from the runtime into the editor. This first iteration improves the global scope editor by grouping related items.

The crux of this implementation is bringing information about scope variables from the runtime into the editor.

One of the shortcomings is the object explorer which is really hitting its limits (e.g. can't remove that object [preview](https://github.com/storybookjs/react-inspector/issues/163). It's too limited to customize). Can't make it look much better than this. Will advance https://github.com/mui/mui-toolpad/pull/1271 to improve the object explorer.

Screenshot 2022-12-21 at 14 37 09


prakhargupta1:
Just came across this change, looks much better now!

Screenshot 2022-12-22 at 12 57 51 PM

PR #1468, opened 2022-12-19
Non permanent root redirection

oliviertassinari:
I saw that https://demo.toolpad.io/ is a 308 to https://demo.toolpad.io/_toolpad in an SEO crawl. I wonder if this isn't too bold, maybe a 307 would keep our option more often in the future. We could also make the argument that it doesn't matter.


Janpot (approved):
makes sense 👍

PR #1472, opened 2022-12-19
Move recaptcha script away from _app

Janpot:
- [x] Avoids loading it on every page
- [x] Initialize v2 captcha only once 
- [x] disable button until reacpatcha script loads
- [x] Should deal with https://sentry.io/organizations/mui-org/issues/3784428523/events/cc5ccc5875634fc9a6deb6ceacd814e6/?project=6763402
- [x] Should deal with https://sentry.io/organizations/mui-org/issues/3815931212/events/28be8d43644d438ca8130369ba5719bd/?project=6763402

### DIFF
--- a/packages/toolpad-app/src/toolpad/Apps/index.tsx
+++ b/packages/toolpad-app/src/toolpad/Apps/index.tsx
@@ -86,6 +86,11 @@ export const APP_TEMPLATE_OPTIONS: Map<
 
 const NO_OP = () => {};
 
+function RecaptchaV2({ onToken }) {
+  const captchaTargetRef = React.useRef<HTMLDivElement | null>(null);
+  return <div ref={captchaTargetRef} />;
+}
+
 export interface CreateAppDialogProps {
   open: boolean;
   onClose?: () => void;
@@ -100,12 +105,6 @@ function CreateAppDialog({ onClose, open, ...props }: CreateAppDialogProps) {
   const [recaptchaV2Token, setRecaptchaV2Token] = React.useState('');
 
   const captchaTargetRef = React.useRef<HTMLDivElement | null>(null);
-
-  React.useEffect(() => {
-    if (!open) {
-      setRequestRecaptchaV2(false);
-    }
-  }, [open]);
 
   const [isNavigatingToNewApp, setIsNavigatingToNewApp] = React.useState(false);
   const [isNavigatingToExistingApp, setIsNavigatingToExistingApp] = React.useState(false);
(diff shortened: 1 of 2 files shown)


packages/toolpad-app/src/toolpad/Apps/index.tsx
```diff
@@ -295,16 +327,18 @@ function CreateAppDialog({ onClose, open, ...props }: CreateAppDialogProps) {
...
-              id="captcha-target"
-              ref={captchaTargetRef}
-              mt={1}
-              mb={1}
-              sx={{
-                display: 'flex',
-                justifyContent: 'center',
-              }}
-            />
+            <Box sx={{ visibility: requestRecaptchaV2 ? undefined : 'hidden' }}>
```
apedroferreira, line 330: Would this work as a single box?

packages/toolpad-app/src/toolpad/Apps/index.tsx
```diff
@@ -199,23 +199,55 @@ function CreateAppDialog({ onClose, open, ...props }: CreateAppDialogProps) {
...
 
   return (
     <Dialog {...props} open={open} onClose={config.isDemo ? NO_OP : onClose} maxWidth="xs">
+      {config.recaptchaV3SiteKey ? (
+        <Script
+          strategy="afterInteractive"
+          src={`https://www.google.com/recaptcha/api.js?render=${encodeURIComponent(
+            config.recaptchaV3SiteKey,
+          )}`}
+          onLoad={async () => {
```
apedroferreira, line 231: So we're always rendering but we still only show it if the V3 captcha validation fails, seems good.
We could move have this function outside the JSX if you want, just to separate the HTML from the logic a bit more.

packages/toolpad-app/src/toolpad/Apps/index.tsx
```diff
@@ -199,23 +199,55 @@ function CreateAppDialog({ onClose, open, ...props }: CreateAppDialogProps) {
...
-              sitekey: config.recaptchaV2SiteKey,
-            });
-            latestRecaptchaWidgetIdRef.current = widgetId;
-          }
+        const error = errorFrom(rawError);
+        if (config.recaptchaV2SiteKey && error.code === ERR_VALIDATE_CAPTCHA_FAILED) {
+          // Show captcha checkbox
+          setRequestRecaptchaV2(true);
+        } else {
+          throw error;
```
apedroferreira, line 207: Seems good to throw here, I think we were suppressing the error before but I wasn't sure if it was intentional.

packages/toolpad-app/src/toolpad/Apps/index.tsx
apedroferreira: I noticed that if we close the "Create app" modal and open it again the space for the visual captcha isn't there anymore, and the captcha doesn't show? Maybe the `id` and `getResponse` were handling something about this? - I can take a deeper look if you want.

apedroferreira (changes requested):
Thanks for fixing this Sentry issue!
This change seems good, there's just a bug I found if you close and reopen the dialog.
I can take a deeper look to see what's causing it, let me know if that would be helpful.

packages/toolpad-app/src/toolpad/Apps/index.tsx
Janpot: Should probably use `display` instead of `visibility`

packages/toolpad-app/src/toolpad/Apps/index.tsx
Janpot: I'm not sure we should add styling to the rendered widget. I feel like it's best to hand over the element to google and handle margin/visibility in a wrapper

apedroferreira: Ok, I thought the target was the container for the widget and not the widget itself, but if it is we can keep the styles separate, even if just to be safer.

Janpot: updated to be a `div` inside of a `Box`

apedroferreira (approved):
Works great now!

PR #1473, opened 2022-12-19
Simplify RPC logs

Janpot:
Fixes issues like: https://sentry.io/organizations/mui-org/issues/3783806362/?environment=production&project=6763402&query=&referrer=issue-stream&statsPeriod=14d

We're not using JSON to transfer messages in RPC (Though it [looks like it](https://www.npmjs.com/package/superjson) 🙂 )

- [x] Deserializing the RPC message just for logging is a leaky abstraction, why not just log directly in the rpc code?
- [x] Why disable server logs? By default we log the things we want to monitor in production, if you want more/less in your local, adjust the loglevel. This serves as a forcing function for us to not overlog.
- [x] expand error serializer to have name and stacktrace


apedroferreira (approved):
This is definitely an improvement - I was trying to log the error from the response body before I think, but then the stack wasn't working with JSON.
Also getting the response body in Next.js doesn't seem very straightforward, I hadn't found any alternative to that complicated part with the chunks...

packages/toolpad-app/src/server/logs/logger.ts
```diff
@@ -21,17 +21,42 @@ if (config.ecsNodeUrl) {
...
+interface RpcReqResLogPayload {
+  key: 'rpc';
+  type: 'query' | 'mutation';
+  name: string;
+  error: Error | null;
+}
+
+type LogPayload = ReqResLogPayload | RpcReqResLogPayload;
+
+function logMethod(method: 'info' | 'trace' | 'error' | 'warn' | 'fatal') {
```
apedroferreira, line 52: nice, this works better than a function just for `logInfo`

Janpot: Yes, I'm thinking roughly:

* `fatal`: Unrecoverable errors, the node process should crash (or has crashed and this is the report).
* `error`: Unexpected, but we can degrade in a safe way. The user experience is impacted, so this must be investigated
* `warn`: Unexpected, but we can degrade in a safe way. The user experience isn't really impacted, investigate when lots of these happen.
* `info`: informational, we want to see these appear in a log aggregator for metrics


We play with logelevel on different use-cases
* production: `info`
* local: `warn`, info will most likely lead to too many logs

apedroferreira: Sounds great, one of the alerts I set up in Sentry was for frequent warnings too.

PR #1474, opened 2022-12-19
Experiment: Try out tab syncing

Janpot:
Not sure how stable this is and whether it make sense to further allocate time to this, but a basic implementation works with minimal impact on the code

@oliviertassinari 

https://user-images.githubusercontent.com/2109932/208471075-07c23064-13b9-414c-a7c4-9bfd384e081e.mov


oliviertassinari:
It looks good. It's not unusual to have many tabs open when working on something and to go back to the wrong one. I tend to do a random walk to find the tab I'm looking for.

PR #1477, opened 2022-12-20
Polish Markdown component

### LINKED ISSUE #1470: Polish markdown component
### Duplicates

### Latest version

### Steps to reproduce 🕹

Steps:

1. Open https://master--toolpad.mui.com/deploy/cl6rqzry10009arlv9sto6qja/pages/e943n9y

### Current behavior 😯

- There is a margin around the markdown.
- The `<a>` isn't rendered with the `<Link>` of Material UI.

Screenshot 2022-12-19 at 12 16 23

### Expected behavior 🤔

The opposite.

Screenshot 2022-12-19 at 12 16 52

### Context 🔦

I'm updating some of the Toolpad app we use for the support to use markdown. Related to #1298.

### Your environment 🌎

_No response_

bharatkashyap:
- Closes #1470 
- Visual demonstration 
- Replace `react-markdown` with `markdown-to-jsx`
- Uses Material UI `Link` instead of `a`
- Links open in new tabs through `target=_blank`
- No margin

https://user-images.githubusercontent.com/19550456/208647318-f51dd1bc-d329-4a57-a1c7-c904f788117b.mov


packages/toolpad-components/src/Text.tsx
```diff
@@ -34,7 +33,21 @@ function Text({ value, markdown, href, loading, mode, sx, ...rest }: TextProps)
...
           <React.Suspense>
-            <ReactMarkdown remarkPlugins={[remarkGfm]}>{value}</ReactMarkdown>
+            <ReactMarkdown
+              options={{
+                overrides: {
+                  a: {
+                    component: MuiLink,
+                    props: {
+                      target: '_blank',
+                      rel: 'noopener noreferrer nofollow',
```
oliviertassinari, line 43: I think that we could let the search engine crawler follow and crawl these links, and be a bit less opinionated https://ahrefs.com/blog/nofollow-links/. If people want to prevent it, they could write a full `<a href` element.

```suggestion
                      rel: 'noopener noreferrer',
```

packages/toolpad-components/src/Text.tsx
```diff
@@ -34,7 +33,21 @@ function Text({ value, markdown, href, loading, mode, sx, ...rest }: TextProps)
       ) : (
         <MarkdownContainer>
           <React.Suspense>
-            <ReactMarkdown remarkPlugins={[remarkGfm]}>{value}</ReactMarkdown>
+            <ReactMarkdown
```
oliviertassinari, line 36: We need the markdown component to have a min-height when it's empty so that developers can make changes.

Screenshot 2022-12-20 at 11 43 16

oliviertassinari: We might want to handle `<code><pre>` and `<code>`. 

Screenshot 2022-12-20 at 11 45 06

oliviertassinari: To see if the ids can break toolpad

Screenshot 2022-12-20 at 11 44 57

bharatkashyap:
Review feedback: 

Screenshot 2023-01-03 at 3 49 47 PM

packages/toolpad-components/src/Text.tsx
```diff
@@ -34,7 +42,25 @@ function Text({ value, markdown, href, loading, mode, sx, ...rest }: TextProps)
...
+                    props: {
+                      target: '_blank',
+                      rel: 'noopener noreferrer',
+                    },
+                  },
+                  pre: {
+                    component: CodeContainer,
+                  },
+                },
+                slugify: (text) => `toolpad-markdown-${text}`,
```
Janpot, line 59: But this would turn the links into `#toolpad-markdown-my-id`

packages/toolpad-components/src/Text.tsx
```diff
@@ -23,9 +22,18 @@ interface TextProps extends Omit<BaseProps, 'children'> {
 
 const MarkdownContainer = styled('div')({
   display: 'block',
-  [`&:empty::before`]: { content: '""', display: 'inline-block' },
+  minHeight: '25px',
 });
 
+const CodeContainer = styled('pre')(({ theme }) => ({
+  whiteSpace: 'pre-wrap',
```
Janpot, line 29: Most code doesn't look great when wrapped. Can't we handle this using overflow?

packages/toolpad-components/src/Text.tsx
```diff
@@ -23,9 +22,18 @@ interface TextProps extends Omit<BaseProps, 'children'> {
 
 const MarkdownContainer = styled('div')({
   display: 'block',
-  [`&:empty::before`]: { content: '""', display: 'inline-block' },
+  minHeight: '25px',
```
Janpot, line 25: I'd prefer this to be solved using a similar trick as the `&:empty::before` so that it follows `fontSize` rather than an absolute height

packages/toolpad-components/src/Text.tsx
```diff
@@ -8,9 +8,8 @@ import {
   styled,
 } from '@mui/material';
 import { createComponent } from '@mui/toolpad-core';
-import remarkGfm from 'remark-gfm';
 
-const ReactMarkdown = React.lazy(() => import('react-markdown'));
+const ReactMarkdown = React.lazy(() => import('markdown-to-jsx'));
```
Janpot, line 12:
```suggestion
const Markdown = React.lazy(() => import('markdown-to-jsx'));
```
Let's rename to just `Markdown` now that we don't use `react-markdown` anymore?

packages/toolpad-components/src/Text.tsx
bharatkashyap: `white-space: pre-wrap` 
> Sequences of white space are preserved. Lines are broken at newline characters, at `<br>`, and as necessary to fill line boxes.

solves the problem acceptably well in most code snippets I tried.

packages/toolpad-components/src/Text.tsx
oliviertassinari: > When is id a problem exactly?

No idea. I left the feedback in case we rely on `getElementById`. If we don't, I think that keeping the id is simpler.

packages/toolpad-components/src/Text.tsx
```diff
@@ -23,9 +22,22 @@ interface TextProps extends Omit<BaseProps, 'children'> {
 
 const MarkdownContainer = styled('div')({
   display: 'block',
-  [`&:empty::before`]: { content: '""', display: 'inline-block' },
+  maxWidth: 'fill-available',
+  '& span:empty::before': {
```
oliviertassinari, line 26: So that the min-height triggers when the lazy import is resolving?

```suggestion
  '&:empty::before, & span:empty::before': {
```

oliviertassinari: Also, we don't want to trigger `<span>` that are rendered by the markdown, so only the first direct children

```suggestion
  '&:empty::before, & > span:empty::before': {
```

packages/toolpad-components/src/Text.tsx
```diff
@@ -23,9 +22,22 @@ interface TextProps extends Omit<BaseProps, 'children'> {
...
 });
 
+const CodeContainer = styled('pre')(({ theme }) => ({
+  backgroundColor: theme.palette.grey[200],
+  marginLeft: theme.spacing(1),
+  marginRight: theme.spacing(1),
+  borderRadius: theme.shape.borderRadius,
+  padding: theme.spacing(1),
+  overflowX: 'scroll',
+}));
```
oliviertassinari, line 39: In the future, when we look into performance, we could look if styling `& pre` in the container isn't faster.

packages/toolpad-components/src/Text.tsx
```diff
@@ -23,9 +22,22 @@ interface TextProps extends Omit<BaseProps, 'children'> {
 
 const MarkdownContainer = styled('div')({
   display: 'block',
-  [`&:empty::before`]: { content: '""', display: 'inline-block' },
+  maxWidth: 'fill-available',
+  '& span:empty::before': {
+    content: '""',
+    display: 'inline-block',
+  },
 });
```
oliviertassinari, line 30: I think that it would be good to remove the margin top and bottom. It looks like Retool had the same idea:

Screenshot 2023-01-09 at 00 41 50
Screenshot 2023-01-09 at 00 41 58

The problem right now is that

```
foo

bar
```

renders like this:

Screenshot 2023-01-09 at 00 43 18

packages/toolpad-components/src/Text.tsx
```diff
@@ -23,9 +22,22 @@ interface TextProps extends Omit<BaseProps, 'children'> {
...
 
+const CodeContainer = styled('pre')(({ theme }) => ({
+  backgroundColor: theme.palette.grey[200],
+  marginLeft: theme.spacing(1),
+  marginRight: theme.spacing(1),
+  borderRadius: theme.shape.borderRadius,
+  padding: theme.spacing(1),
+  overflowX: 'scroll',
```
oliviertassinari, line 38: overflow scroll is very rarely helpful because it always shows the scrollbar, e.g.

Screenshot 2023-01-09 at 00 47 07

So:

```suggestion
  overflow: 'auto',
```

oliviertassinari (commented):
I think that we have room to do a lot of polish on the rendering (we did a lot for mui.com's docs). I have left feedback on what seems to be the most important. I think that it's good enough 👍, we have enough for the community to leave feedback on what to improve, next problem 😁

> markdown-to-jsx is very popular, but who is it losing market share to?

@Janpot I can't find anything obvious.

packages/toolpad-components/src/Text.tsx
bharatkashyap: I don't see any margin on the markdown element: 
Screenshot 2023-01-09 at 6 25 19 PM

Janpot: I think it's [collapsed](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing) from the one in the paragraphs

packages/toolpad-components/src/Text.tsx
```diff
@@ -35,14 +34,28 @@ const gutters = (marginTop: number, marginBottom: number) => ({
...
   display: 'block',
-  [`&:empty::before`]: { content: '""', display: 'inline-block' },
-  h1: { ...theme.typography.h1, ...gutters(16, 16) },
-  h2: { ...theme.typography.h2, ...gutters(12, 12) },
-  h3: { ...theme.typography.h3, ...gutters(12, 12) },
-  h4: { ...theme.typography.h4, ...gutters(12, 12) },
-  h5: { ...theme.typography.h5, ...gutters(4, 4) },
-  h6: { ...theme.typography.h6, ...gutters(4, 4) },
-  p: { margin: 0, marginBottom: 6 },
+  maxWidth: 'fill-available',
```
Janpot, line 37:
```suggestion
```

This doesn't seem to do anything for me. Which situation is this olving?

bharatkashyap: This is to introduce scrollbars in content that spans the width of the container:

Without `fill-available`:
Screenshot 2023-02-16 at 5 19 12 PM

With:
Screenshot 2023-02-16 at 5 19 29 PM

Janpot: Oh, ok, I see. Does it have advantages over `maxWidth: '100%'`?

bharatkashyap: Nothing specifically I guess, I can use `100%` as well

Janpot (approved):

packages/toolpad-components/src/Text.tsx
```diff
@@ -34,8 +34,10 @@ const gutters = (marginTop: number, marginBottom: number) => ({
 
 const MarkdownContainer = styled('div')(({ theme }) => ({
   display: 'block',
-  maxWidth: 'fill-available',
-
+  maxWidth: '100%',
+  '&  p': {
```
Janpot, line 38: There's already a rule for this

Janpot (changes requested):

Janpot (approved):

PR #1485, opened 2022-12-23
Fix: Demo reCaptcha v2 fallback does not work

bharatkashyap:
There is a fallback reCaptcha v2 verification present in the demo, which loads if the reCaptcha v3 verification fails. However, this fallback UI will not load at all (due to a bug in the way errors are being sent to the client), resulting in a user getting stuck.

Adding `code` to `SerializedError` allows the `code` to be present in the error on the client, and the fallback scenario to be handled.

- Visual demonstration 

Before

https://user-images.githubusercontent.com/19550456/209332323-fb3fb9a5-db15-46b6-b6f3-47d8690e4622.mov

After

https://user-images.githubusercontent.com/19550456/209332914-36652901-d250-4846-bd2f-3ad4cdfe99e3.mov

Next step is to add an integration test to verify that the captcha checkbox loads in case the reCaptcha v3 verification fails.

### DIFF
--- a/packages/toolpad-core/src/types.ts
+++ b/packages/toolpad-core/src/types.ts
@@ -306,6 +306,7 @@ export interface SerializedError {
   message: string;
   name: string;
   stack?: string;
+  code?: unknown;
 }
 
 export type ExecFetchResult<T = any> = {
--- a/packages/toolpad-app/src/toolpad/Apps/index.tsx
+++ b/packages/toolpad-app/src/toolpad/Apps/index.tsx
@@ -151,7 +151,7 @@ function CreateAppDialog({
   const [name, setName] = React.useState('');
   const [appTemplateId, setAppTemplateId] = React.useState<AppTemplateId>('default');
   const [dom, setDom] = React.useState('');
-  const [recaptachaApiLoaded, setRecaptachaApiLoaded] = React.useState(false);
+  const [recaptchaApiLoaded, setRecaptchaApiLoaded] = React.useState(false);
   const [requestRecaptchaV2, setRequestRecaptchaV2] = React.useState(false);
   const [recaptchaV2Token, setRecaptchaV2Token] = React.useState('');
 
@@ -214,7 +214,7 @@ function CreateAppDialog({
 
   const recaptchaSubmitEnabled =
     config.recaptchaV2SiteKey || config.recaptchaV3SiteKey
-      ? recaptachaApiLoaded && ((requestRecaptchaV2 && recaptchaV2Token) || !requestRecaptchaV2)
+      ? recaptchaApiLoaded && ((requestRecaptchaV2 && recaptchaV2Token) || !requestRecaptchaV2)
       : true;
 
   const handleSubmit = React.useCallback(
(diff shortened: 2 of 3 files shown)


packages/toolpad-core/src/types.ts
```diff
@@ -306,6 +306,7 @@ export interface SerializedError {
   message: string;
   name: string;
   stack?: string;
+  code?: unknown;
```
oliviertassinari, line 309: So it's for https://nodejs.org/api/errors.html#errorcode?

oliviertassinari (commented):
Are we sure that SerializedError should have a `code` property? Should it be part of the message or name?

packages/toolpad-core/src/types.ts
bharatkashyap: It's a custom type, so perhaps we could rename it to better represent this (`SerializedToolpadError` ?)

bytasv:
I'm just curious - how come this PR fixes a problem? 🤔 I mean it seems that we are only renaming variables and also adding a type, none of which should change how code behaves, or am I missing something?

bharatkashyap:
> I'm just curious - how come this PR fixes a problem? 🤔 I mean it seems that we are only renaming variables and also adding a type, none of which should change how code behaves, or am I missing something?

This line makes sure that `code` is not dropped
https://github.com/mui/mui-toolpad/pull/1485/files#diff-4c1bd35b8be1fd9d892b298ad9b8b1b2e9aadfd0ed7470b4bf2e3cb1819eb476L6

when we pass errors to the client here
https://github.com/mui/mui-toolpad/blob/master/packages/toolpad-app/pages/api/rpc.ts#L95

and the `code` is what triggers the fallback UI, here
https://github.com/mui/mui-toolpad/blob/master/packages/toolpad-app/src/toolpad/Apps/index.tsx#L265

bytasv:
> > I'm just curious - how come this PR fixes a problem? 🤔 I mean it seems that we are only renaming variables and also adding a type, none of which should change how code behaves, or am I missing something?
> 
> This line makes sure that `code` is not dropped https://github.com/mui/mui-toolpad/pull/1485/files#diff-4c1bd35b8be1fd9d892b298ad9b8b1b2e9aadfd0ed7470b4bf2e3cb1819eb476L6
> 
> when we pass errors to the client here https://github.com/mui/mui-toolpad/blob/master/packages/toolpad-app/pages/api/rpc.ts#L95
> 
> and the `code` is what triggers the fallback UI, here https://github.com/mui/mui-toolpad/blob/master/packages/toolpad-app/src/toolpad/Apps/index.tsx#L265

Oh you are so fast, I just posted and realised that there was that one change that does the trick and deleted comment 😅  but thanks for explaining anyways 🙇

packages/toolpad-core/src/types.ts
Janpot: We standardize on regular JavaScript `Error`s with a `code` property similar to how node.js treats them. we frequently pass errors between server and client and this format is the most straightforward to do that with.

in general we should avoid using exceptions for control flow. that paradigm doesn't mesh well with a distributed system anyway.

Janpot (approved):

PR #1504, opened 2023-01-03
Datepicker format handling changes

bytasv:
- Addresses feedback from https://github.com/mui/mui-toolpad/pull/1499/files

https://user-images.githubusercontent.com/437214/210354930-8f4be681-1831-4054-8ec8-81b2393a617b.mp4

### DIFF
--- a/packages/toolpad-components/src/DatePicker.tsx
+++ b/packages/toolpad-components/src/DatePicker.tsx
@@ -49,6 +49,7 @@ export default createComponent(DatePicker, {
       typeDef: { type: 'string' },
       onChangeProp: 'onChange',
       onChangeHandler: (newValue: Dayjs) => {
+        // date-only form of ISO8601. See https://tc39.es/ecma262/#sec-date-time-string-format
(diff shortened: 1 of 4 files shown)


packages/toolpad-components/src/DatePicker.tsx
```diff
@@ -44,25 +48,17 @@ export default createComponent(DatePicker, {
     value: {
       typeDef: { type: 'string' },
       onChangeProp: 'onChange',
-      onChangeHandler: (newValue: Dayjs, { format, separator }: Props) => {
-        return newValue.format(resolveFormat(format, separator));
+      onChangeHandler: (newValue: Dayjs) => {
+        return newValue.format('YYYY-MM-DD');
```
Janpot, line 52:
```suggestion
        // date-only form of ISO8601. See https://tc39.es/ecma262/#sec-date-time-string-format
        return newValue.format('YYYY-MM-DD');
```

Janpot (approved):
Nit: I'd add a small clarifying comment on why we choose this format

Issue #1452, opened 2022-12-14
Cmd+S doesn't work in the function query editor


Janpot:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

Steps:

1. Open a function query
2. Make changes to the code
3. Press Cmd+S

### Current behavior 😯

Nothing happens

### Expected behavior 🤔

The query gets saved

### Context 🔦

It's supposed to work according to https://github.com/mui/mui-toolpad/blob/07e29aa8633701254000e6efe2cd422e9cd66ce6/packages/toolpad-app/src/toolpadDataSources/function/client.tsx#L210
But it seems to collide with the global Cmd+S shortcut

Ran into this while working on the "bundle size viewer" use-case

We could benchmark a few 3rd party libraries

* [`react-hot-keys`](https://www.npmjs.com/package/react-hot-keys)
* [`react-hotkeys`](https://www.npmjs.com/package/react-hotkeys)
* ...

bharatkashyap:
The shortcut also fails in the query editor. +1 to audit a third-party library that could work uniformly across

bharatkashyap:
This works if we move the hook to:
https://github.com/mui/mui-toolpad/blob/master/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/QueryEditor/QueryEditorDialog.tsx, but that results in the shortcut not working inside the `BindingEditor`

Janpot:
The hook is a dead end. The solution will likely involve binding key events to focussed dom elements.

PR #1507, opened 2023-01-03
Fix function editor save

Janpot:
Closes https://github.com/mui/mui-toolpad/issues/1452

Turns out the shortcut was working after all. The problem was that the update didn't reflect in the UI afterwards

### DIFF
--- a/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/QueryEditor/QueryEditorDialog.tsx
+++ b/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/QueryEditor/QueryEditorDialog.tsx
@@ -234,10 +234,6 @@ export default function QueryNodeEditorDialog<Q>({
     onClose();
   }, [handleCommit, onClose]);
 
-  React.useEffect(() => {
-    setInput(node);
-  }, [node]);
-
   const queryEditorContext = React.useMemo(
     () => (dataSourceId ? { appId, dataSourceId, connectionId } : null),
     [appId, dataSourceId, connectionId],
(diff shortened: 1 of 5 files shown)


packages/toolpad-app/src/toolpad/AppEditor/PageEditor/QueryEditor/QueryEditorDialog.tsx
```diff
@@ -106,6 +106,9 @@ export default function QueryNodeEditorDialog<Q>({
   const node = useLatest(nodeProp);
 
   const [input, setInput] = React.useState<appDom.QueryNode<Q>>(node);
+  React.useEffect(() => {
```
apedroferreira, line 109: This is repeated, there's the same thing in line 237 (I might've added in the last PR I merged).

apedroferreira: nvm different component :D

test/integration/function-basic/index.spec.ts
```diff
@@ -31,3 +32,33 @@ test('functions basics', async ({ page, api }) => {
...
+  const app = await api.mutation.createApp(`App ${generateId()}`, {
+    from: { kind: 'dom', dom },
+  });
+
+  const editorModel = new ToolpadEditor(page, browserName);
+  await editorModel.goto(app.id);
+
+  await expect(editorModel.appCanvas.getByText('edited hello')).toBeVisible();
+
+  await editorModel.componentEditor.getByRole('button', { name: 'edited' }).click();
```
apedroferreira, line 48: When we add more tests for the query editor it might make sense for some of these to have methods we can reuse.

apedroferreira (changes requested):
Looks good, just need to remove the duplicate effect I think.

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/QueryEditor/QueryEditorDialog.tsx
Janpot: Ok, I'm removing the other one to keep it closer to where the state is created.

apedroferreira (approved):

PR #1517, opened 2023-01-05
Correct types in Datepicker

Janpot:
The types of `argTypes` is derived from the type of the component properties.


packages/toolpad-components/src/DatePicker.tsx
```diff
@@ -14,6 +14,7 @@ export interface Props extends DesktopDatePickerProps<string, Dayjs> {
   variant: 'outlined' | 'filled' | 'standard';
   size: 'small' | 'medium';
   sx: any;
+  defaultValue: string;
```
apedroferreira, line 17: Not from this PR but shouldn't this props object be called `DatePickerProps` instead of `Props`? Especially as it's being exported.

Janpot: Done

apedroferreira (approved):

PR #1521, opened 2023-01-06
Prototype direction 13

Janpot:
Adding back the old CLI and a local mode to toolpad to edit an app and store it on disk in a yaml file

https://user-images.githubusercontent.com/2109932/211366610-1abf4721-e16c-4f58-8c04-a8b05e152d26.mov

Currently not part of this effort:

- readable yaml, multiple config files,...
- export code
- use toolpad pages in pro code
- toolpad runtime (runtime and editor are in one single command)
- ...

## custom components

Editing code components:

**READ THIS FIRST:**
* This moves the current functionality to the file system in a ./toolpad-components folder. It doesn't support anything beyond that
* It doesn't support importing from other files
* It doesn't support importing node modules

https://user-images.githubusercontent.com/2109932/214776216-f206d424-b190-4170-97a6-0d88c6ee8173.mov

## Data loading:

This accepts a `./toolpad/queries.ts` file in your project folder. All async functions exported from there are available to use in queries:

https://user-images.githubusercontent.com/2109932/215149091-f7a2049a-b1b5-4890-93e0-ed97dcfa90d8.mov

These queries can be made parametrizable by wrapping them with a `createQuery` and providing the configuration.

https://user-images.githubusercontent.com/2109932/215150379-6618b966-5988-4767-bb75-13280f79bb99.mov

These queries run in node.js, they can use environment variables and use native modules:

https://user-images.githubusercontent.com/2109932/215152962-c60338af-de6d-4efb-be72-a0650d537cf9.mov

In essence this now makes the definition of queries shareable between pages. As a continuation, I can imagine making it unnecessary to define the query in the page and allow for directly hooking up the function to a table. Additionally I can imagine us creating a `createTable` helper in the same flavour as `createQuery` that automatically hooks up pagination/sorting/filtering.


bharatkashyap:
>  Screen.Recording.2023-01-27.at.18.00.38.mov 
> These queries can be made parametrizable by wrapping them with a `createQuery` and providing the configuration.

Fan of the clean `parameters` interface with the global scope being available to bind 👍

apedroferreira:
I really liked the way we can work with queries and even details like the "Open in VSCode" button.
I think this really has a lot of potential!

oliviertassinari:
@Janpot Awesome, I think that what would be great is to see next is production. How can we deploy it, I want to try moving the apps we have in https://master--toolpad.mui.com/ to this different local env DX.

prakhargupta1:
Awesome and great progress Jan! shareable queries, global scope, accessing native modules from VS code 👍

Janpot:
> I think that what would be great is to see next is production.

I will work towards publishing this on npm under a demo tag so we can test drive it. Before that I main clean up this PR and merge it as it doesn't interfere with current Toolpad, and this way we can prevent the effort from diverging too much from what's on master

Once it's on npm we can try out what a render.com/heroku deploy would look like

Janpot:
## To do

- [ ] fix `externals` for the local functions build to work correctly for other packages than 'pg'
- [ ] test whether functions still work in production build
- [ ] publish npm packages under beta tag

Janpot:
Going to merge this to concentrate on publishing

oliviertassinari:
I have played a bit with it at https://github.com/mui/mui-toolpad-examples. The DX makes sense to me, and I think will click a lot more with our existing users. Great work

Things that I noticed that I think would be better in the DX: 

- I believe that the default for backend engineers is to have one API route = one file, so we could do the same, and split the `queries.js` into a `/queries` folder.
- The `toolpad.yaml` file isn't optimized for readability. We could maybe split this file into one per page. Maybe we don't need the `id` property as it's already in the yaml dictionary key. 
- I missed the postman-like query builder.

PR #1525, opened 2023-01-06
Add components drawer open delay of 100ms

Janpot:
Refactor drawer open logic and incorporate 100ms delay on open

### DIFF
--- a/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/ComponentCatalog/ComponentCatalog.tsx
+++ b/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/ComponentCatalog/ComponentCatalog.tsx
@@ -37,6 +37,47 @@ const ComponentCatalogRoot = styled('div')({
   overflow: 'visible',
 });
 
+interface UseDebouncedBooleanValueParams {
+  waitTrue?: number;
+  waitFalse?: number;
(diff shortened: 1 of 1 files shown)


Janpot:
Personally I find the delay noticable, and I suspect it could even contribute to negative discoverability. Some users have already indicated they find it hard finding this drawer

apedroferreira:
Just tested and it feels alright to use, I didn't have a problem to how it worked without the delay though!

> Set the drawer partially open until it has been interacted with at least once (persist in localstorage)
Not sure if this wouldn't look a bit weird? But maybe it helps.

> Make it more makeswift/dynaboard like (icons for categories of components, with popovers on hover/click)
Might be a good solution if we're having the problem of not having enough space to show things in the UI.

bytasv:
It's better but only with the cases that manages to move mouse faster than in 100ms. I was thinking that if we want to improve  "intent detection" we could try to debounce timer while mouse is moving. So as long as I haven't "parked" mouse for say longer than 100ms over drawer it would not open even if I'm moving my mouse slowly.

IMO improvement in this PR is not really worth the code complexity it adds. I'm not sure if it would work but for simple delay like this maybe it would be enough to have drawer controlled with CSS transition and have same 100ms delay in the transition 🤷

Issue #1487, opened 2022-12-24
Support lists component


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

https://mui.com/material-ui/react-list/

As this will have sub-components. We should follow the same approach as we'll do in Cards component.

### Examples 🌈

<img width="676" alt="Screenshot 2022-12-24 at 12 41 59 PM" src="https://user-images.githubusercontent.com/92228082/209425541-edc082d6-2eba-446f-b728-a4c3859f350a.png">

https://marmelab.com/ra-enterprise-demo/#/

Also check DX of retool, appsmith.

bytasv:
@prakhargupta1 so you want to be able to rebuild same design that is shown in example?
- avatar
- main text
- helper text
- numbers

Should each list item be constructed manually or should the list be built from array where I specify only how single list item looks like and the rest is automatically rendered?

prakhargupta1:
Yes, this template should be fine. Instead of Avatar we can also show image. This way no new component need to added. So Image+Typography can be there by default. 

Retool has Image+Rating component as the default sub-components and deleting image from 1 deletes from other entries as well. This experience looks good to me.

It should be like an array where defining one automatically renders others. By default, we can show 3 items.

apedroferreira:
I'll start working on this one if no one's picked it up already - will ask if I have any questions.

Janpot:
There's an interesting component in retool called "List View"

It allows you to set a number of items, and allows you to drag in other components, as a template which is repeated the same number of items.

I wonder if we could improve on it:
- We allow you to template one item
- Instead of setting a number, we allow you to assign an array, with data that can be used in the template

apedroferreira:
> There's an interesting component in retool called "List View"
> 
> It allows you to set a number of items, and allows you to drag in other components, as a template which is repeated the same number of items.
> 
> I wonder if we could improve on it:
> 
> * We allow you to template one item
> * Instead of setting a number, we allow you to assign an array, with data that can be used in the template

I was probably thinking of something like that, just passing in an array of objects with all the same keys, and build the list view from that data.
Also will probably try to make our component consistent with the API for the MUI List component as much as possible.

bytasv:
I agree, I think the user flow should be similar to data grid - where one connects data source and then can somehow (ideally via drag and drop) define "template" which will be rendered

PR #1527, opened 2023-01-06
Add List component

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/1487

Allows for building a layout from the first item position in the list, which is repeated a number of times as per the `itemCount` prop. Inside each list item, the respective `i` index can be accessed in the local bindings.

Introduces templating + local scopes, and scoped props with the `isScoped` property in the arg types, which makes the prop have a unique value per scope.

https://user-images.githubusercontent.com/10789765/216412886-06026c5a-831c-46b0-8a90-92e6d12dcec2.mov

Preview: https://toolpad-pr-1527.onrender.com/


packages/toolpad-components/src/List.tsx
```diff
@@ -0,0 +1,47 @@
...
+  // eslint-disable-next-line react/no-unused-prop-types
+  items: IteratorItem[];
+  disablePadding?: boolean;
+  renderItems: IteratorRenderer;
+  sx?: SxProps;
+};
+
+function List({ renderItems, disablePadding = false, sx }: ListProps) {
+  return (
+    <MuiList disablePadding={disablePadding} sx={{ width: '100%', ...sx }}>
```
bytasv, line 16: It might be worth to consider if we shouldn't have a generic list construct where we could use any container as a list item and then pass children inside that container so that those children elements would be rendered by iterating through list array.

I.e. once we have `CardComponent` I would expect that I can use "list" generic component to render array of items as cards.

Since this isn't what was originally specified in the task feel free to ignore my comment. I just think it's worth to have this in mind for future

apedroferreira: On an unrelated note, I might add a generic `Container` component as a stripped-down version of our current `Paper` component - it follows the same kind of thinking.

bytasv: `Iterator` I think it something what I'm thinking about. As I mentioned I would expect that we should be able to take any group of component and render it in array, thinking a list of Cards and how could that be done.
The only minor issue that I see with `Iterator` is that it's very technical term and `List` might be a more natural name to what it is suppose to do

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/RenderOverlay.tsx
```diff
@@ -824,7 +824,12 @@ export default function RenderOverlay({ bridge }: RenderOverlayProps) {
         const edgeDetectionMargin = 10; // px
 
         // Detect center in layout containers
-        if (isDraggingOverElement && activeDropNodeInfo && activeDropSlot) {
+        if (
```
apedroferreira, line 827: It's all bug fixes in this file.

packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -191,13 +206,25 @@ function RenderedNodeContent({ node, childNodeGroups, Component }: RenderedNodeC
 
   const nodeId = node.id;
 
+  const { item: iteratorItem } = useIteratorItemContext();
+
+  const globalScopeNodeState = React.useMemo<GlobalScopeNodeState>(
```
apedroferreira, line 211: Use context to get active item from where the items have been mapped.

packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -353,20 +380,37 @@ function RenderedNodeContent({ node, childNodeGroups, Component }: RenderedNodeC
...
-        const value = props[propName];
-        props[propName] = (
+        wrappedValue = (
           <Placeholder prop={propName} hasLayout={argType.control?.type === 'layoutSlot'}>
             {value}
           </Placeholder>
         );
       }
+
+      props[propName] = isElementIterator
```
apedroferreira, line 402: Wrap element iterator slot children by iterating and creating a context for the active item.

packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -671,6 +719,22 @@ function parseBindings(
           parsedBindingsMap.set(bindingId, parseBinding(binding, {}));
         }
       }
+
+      const firstIteratorAncestorItems = getFirstElementIteratorAncestorItems(dom, elm, components);
```
apedroferreira, line 723: Create local bindings for each node with an element iterator as parent.

packages/toolpad-app/src/toolpad/AppEditor/GlobalScopeExplorer.tsx
```diff
@@ -39,6 +39,10 @@ function groupScopeMeta(value: Record<string, unknown>, meta: GlobalScopeMeta):
       displayName: 'Queries',
       items: [],
     },
+    hidden: {
```
apedroferreira, line 42: Introduce hidden bindings, as there are bindings created for some nodes that I didn't want to show in others.

packages/toolpad-components/src/List.tsx
```diff
@@ -0,0 +1,47 @@
...
+export default createComponent(List, {
+  layoutDirection: 'both',
+  argTypes: {
+    items: {
+      helperText: 'List items to render.',
+      typeDef: { type: 'object' },
+      defaultValue: [...Array(3).fill({})],
+    },
+    renderItems: {
+      typeDef: { type: 'elementIterator', itemsProp: 'items' },
```
apedroferreira, line 35: Using this `itemsProp` so that the iterator is aware of what the prop for the items is. Lmk if there are any ideas to define this differently.

prakhargupta1:
When a text box is dropped in a list component, just like a button should it also not appear thrice but treated as a single component so that it can handle array? In Retool it works like that.

apedroferreira:
> When a text box is dropped in a list component, just like a button should it also not appear thrice but treated as a single component so that it can handle array? In Retool it works like that.

Sorry, I didn't understand, could you show a video example of Retool to see what you mean maybe?
If it's about the Text component not showing 3 times when I place it in the video, it seems that is because the default value for the Text component has been removed? But it wasn't in this PR.
It probably happened unintentionally in some recent commit?

packages/toolpad-app/src/canvas/getPageViewState.ts
```diff
@@ -94,6 +94,11 @@ export function getNodesViewInfo(rootElm: HTMLElement): {
           const parentId: NodeId = fiber.memoizedProps.parentId as NodeId;
           const nodeSlots = nodes[parentId]?.slots;
 
+          // If there are 2 slots for the same prop name, consider only the first one
```
Janpot, line 97: Which scenario would that be?

packages/toolpad-core/src/types.ts
```diff
@@ -75,7 +75,15 @@ export type BindableAttrEntries = [string, BindableAttrValue<any>][];
...
 export interface ValueTypeBase {
-  type: 'string' | 'boolean' | 'number' | 'object' | 'array' | 'element' | 'event';
+  type:
+    | 'string'
+    | 'boolean'
+    | 'number'
+    | 'object'
+    | 'array'
+    | 'element'
+    | 'elementIterator'
```
Janpot, line 85: It wouldn't be possible to come up with a more generic construct, that allows us to expand templating to other situations? I'm thinking custom column types. Or custom autocomplete items.

bytasv (changes requested):
All in all seems to be working fine as in to render list of items, but I found a bug (I suppose?) when I delete item, but there is still one remaining it deletes both elements 👇 

https://user-images.githubusercontent.com/437214/212868101-0d95351f-dd65-4d71-8e2b-d3df02d3b9e6.mp4

packages/toolpad-app/src/canvas/getPageViewState.ts
apedroferreira: Without this exception, every item rendered by the iterator would have its own slot where you can build layout - and any changes made in any of those slots would always apply to every item. 
So this was a quick way to make it so that in cases like this it's only possible to build a layout in the first slot, using it as a template.

packages/toolpad-core/src/types.ts
apedroferreira: Do you have more specific examples of limitation this approach has that wouldn't work for those use cases?

**Custom column types:** if this would be a Datagrid feature, it might deserve its own specific approach?
**Custom autocomplete items:** what this new `elementIterator` does is allow us to render a template n number of times based on the layout of the first item, and wrap those items with whatever we want to (in the case of this `List`, a `ListItem`). So it should be possible to use an `elementIterator` for custom autocomplete items?

apedroferreira:
@bytasv I've tried to make the part you mentioned more readable + fixed the bug you found with layout slots (thanks for finding it!). Let me know if it's good now.

bytasv (approved):
Works well now, great job

packages/toolpad-core/src/types.ts
Janpot: There are two concepts here:

* The concept of "a template", which describes what this prop holds. a template that can be used to render multiple elements, based ona  common template
* The concept of "a template editing UI". This is what the "slot" control does for the `element` type. It defines how a user interacts with the property.

It seems to me that you are bundling these concepts together here in the `elementIterator` type. My expectation would be that there is a `template` type. and something like a `dragAndDrop` control. That way we separate the logic of what the prop holds from how it's being manipulated. e.g. you can have a single `number` type, that can be manipulated by several different controls.

apedroferreira: But for example the `Paper` component is already also using a `layoutSlot` for the same layout building experience, even though it does not iterate, so they're already separate concepts? Maybe it's just a question of naming things differently?
We can clear up these details later synchronously though and finish the discussion about this.

Janpot:
We had in-depth 1-on-1 around this PR. Thinking past week about the extent of this problem, I had a hunch we were underestimating the complexity. To illustrate the issues, consider following retool vs. toolpad apps:

Retool: https://mui.retool.com/editor/3cf0ba28-c1fe-11ec-95db-df6327e52470/jans%20space
Toolpad: https://toolpad-pr-1527.onrender.com/_toolpad/app/cld8ygo8g0001kh9t719qc3wk/pages/331kqzd

Main areas to note are:
1. Retool app list items create dom based on a template, but with a new binding scope, Toolpad app list items are copied dom nodes, within the global scope. Therefore the behavior differs considerably between implementation. This app is mostly to illustrate the problem very visually.
2. The input label tries to bind to the list iterator but it's not defined.


We will have to investigate around how to create nested scopes within the binding engine. Currently there is nothing foreseen for this.

apedroferreira:
> Retool app list items create dom based on a template, but with a new binding scope, Toolpad app list items are copied dom nodes, within the global scope. Therefore the behavior differs considerably between implementation. This app is mostly to illustrate the problem very visually.

Better defined local scopes would help this but there also seems to be a difference in how Retool handles input values compared to how we do... In Toolpad the input value comes directly from a prop in the Text Field component, but there's no `value` control in Retool, the value seems to be uncontrolled but still retrievable in bindings somehow. Any other controlled prop in Retool also just copies itself unless we set a binding...

> The input label tries to bind to the list iterator but it's not defined.

I think you misunderstood this part - there is no `i` in the current implementation.

Screenshot 2023-01-23 at 18 18 04

What we have now is, in this case, this `textFieldItem` in the local bindings which retrieves what you expect to be `items[i]`. The `items` prop needs to be set and there would be an object there, as also shown in my video example above. We could also easily make an index available if it would be useful.

bytasv:
Even if Toolpad differs from retool implementation and does not support "local scopes" does it makes this a blocker in this case? I'd assume this would come as a further improvement once we actually need it. Right now if I understood correctly it does what I would expect it to do - access array item according to the index and renders the data

CleanShot 2023-01-23 at 21 49 00@2x

as far as I can tell this functions as we intend in the initial scope of this ticket 🤔

bytasv:
> But iterative also means that a subsequent step can build on top of the previous step.

I don't see why it has to be built on previous step. More often than not when you learn something new you you may want to refactor or rewrite it something from the ground up. I don't think this is any different. With the information that we have right now and basic requirements that were defined in the original task this DOES satisfy it. If there are technical reasons not to merge it now (like exposed scopes) then we should look into how we can patch it and unblock the main functionality. 
Once we learn all about the different use-case and further requirements for list component we can even create list v2 and start deprecating list v1 if that's the way we choose.

Iteration is about learning something based on feedback and improving it accordingly.

I can easily see a value with a current implementation of this component to the end users even if it was some drawbacks or limitations. We should fix/improve/change it based on an actual use-case or user's feedback. Unless we are committing something that would be irreversible going forward, I don't see a reason for this to be blocked/not merged until we figure out how to do a new, improved version of such component.

apedroferreira:
> Personally, to simplify the API further, I'd also just replace items with count and only supply i in the scope. It should be enough to display lists and iterate their items.

In my latest changes I'm passing `i` and `item`. I can pass only `i` and replace `items` with `count`, sure.
I'm also using a component API like you suggested now.

> I tried adding a click handler to a button in a list (as per your screenshot), it doesn't have access to either item or i.

Seems like my latest commits didn't deploy, I need to fix the autocomplete so the CI passes... My workarounds to make the autocomplete work were causing most of the issues you perceived, but they weren't the only way to do things.

> The lack of local scope causes those listitems to show up in all the bindings. But what does it mean to bind to them?

The global scope items were there just for the autocomplete. I just removed that and they still work in the runtime. But now I have to fix the autocomplete in some way that doesn't involve using global scope.
There's actually a "local scope" now, but it's using React context. We could see how to improve it from here?
Also I think the problem with the inputs not working like in Retool is partially a different problem related to their inputs being uncontrolled, like I mentioned.

packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -145,6 +153,213 @@ function useElmToolpadComponent(elm: appDom.ElementNode): ToolpadComponent {
...
+      expression: bindable.value,
+    };
+  }
+  return {
+    scopePath,
+    result: { value: undefined },
+  };
+}
+
+function parseBindings(
```
apedroferreira, line 213: I just moved a few functions above in this file to be able to use them anywhere, and replaced the `page: appDom.PageNode` argument with `rootNode: appDom.ElementNode | appDom.PageNode`, but I actually ended up not using this so far, not sure if it's needed...

packages/toolpad-components/src/TextField.tsx
```diff
@@ -35,6 +35,7 @@ export default createComponent(TextField, {
       onChangeProp: 'onChange',
       defaultValue: '',
       defaultValueProp: 'defaultValue',
+      isScoped: true,
```
apedroferreira, line 38: I introduced the concept of scoped parameters as in Retool input values seem unique per scope but other props are not.

Janpot:
This is coming along nice. I still have this problem:

* Open https://toolpad-pr-1527.onrender.com/_toolpad/app/cle309iyl0004pt9aqmx1g7h2/pages/z239t9x
* It shows only 3 items for me
* Drag a component in the canvas
* Suddenly it shows the correct 20 items

apedroferreira:
> It shows only 3 items for me

There was an issue with prop values that changed by themselves (e.g: query results), I've added a fix.

> If it isn't too much work I'd also add a test that verifies the different binding use-cases we discussed on Slack.

I've added a more specific test for the List component that should cover those cases.

Let me know if you find anything else!

Janpot (approved):

Issue #1395, opened 2022-11-30
Support FileInput Component


sizer:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

When building admin application, a feature uploading files - like csv or etcetera - is often necessary.

### Examples 🌈

https://retool.com/components/file-button
https://retool.com/components/file-input

bytasv:
@prakhargupta1 could you please explain:
1. should this also store files in the database for future use?
2. what are we suppose to be able to do after uploading file?
3. what file formats are we suppose to allow?

It might be easier if at least one e2e scenario would be added how this feature is suppose to be used

bytasv:
@prakhargupta1 could you provide a more specific example or at least outline which part in the entire documentation refers to the actual usecase? In the link you shared I can only find explanation that the widget allows uploading files but it doesn't mention how those files can later be used:
- Should we have some UI to display uploaded files?
- Should we be able to read the content of uploaded file in the app or do we want to upload files only for linking them?

When I say "e2e scenario" I'm thinking about something like following:
1. I upload CSV
2. I can read content of that CSV file and render it in the Data grid component
OR
1. I upload image
2. I can display image using image component
etc..

IMO this is quite a large feature both in scope and functionality and we should have a bit more specific information

apedroferreira:
I guess that for now we can just allow users to add a local file to a file input, client side, and then the file object can be accessed in the bindings panel, for example?
So then the user can potentially pass the file content to some API if they want to, in order to upload it for example?

bytasv:
> I guess that for now we can just allow users to add a local file to a file input, client side, and then the file object can be accessed in the bindings panel, for example? So then the user can potentially pass the file content to some API if they want to, in order to upload it for example?

That's what I'm trying to understand because doing that client-side only vs uploading to the backend are significantly different tasks

Janpot:
as @apedroferreira suggests, intuitively I think it could make sense separating the file input component from the file uploading.

Perhaps we could start with a benchmark? 

> e.g. I noticed retool makes the files available on component state as an array of base64 strings. 

Then try to figure out the trade-offs associated with the method

> e.g. Does retool make it possible to inspect file names? How easy would it be to feed this array as `FormData` to a POST request?...

I think I'd at least be interested in how the retool/appsmith/openblocks family compares to dynaboard

prakhargupta1:
>Should we have some UI to display uploaded files?

[...]

PR #1537, opened 2023-01-09
Add FilePicker component

bytasv:
- Closes https://github.com/mui/mui-toolpad/issues/1395

https://user-images.githubusercontent.com/437214/211545392-37bc80c0-074c-419c-9098-3bd9f805a73f.mp4


packages/toolpad-components/src/FilePicker.tsx
```diff
@@ -0,0 +1,101 @@
...
+  loaded: boolean;
+}
+
+export default createComponent(FilePicker, {
+  layoutDirection: 'both',
+  argTypes: {
+    value: {
+      typeDef: { type: 'file' },
+      onChangeProp: 'onChange',
+      onChangeHandler: (event: React.ChangeEvent<HTMLInputElement>) => {
```
bytasv, line 34: @Janpot before I sink more time into this I have a few questions:

1. Do you have any alternative suggestions how we should be passing file value to the user? Right now I'm converting `FileList` object to array which should contain actual file value so we can avoid forcing users to use `FileReader` when binding values. At the moment `FileReader` is not even accessible, but also it would require handling promises in binding editor, which I'm not sure if is supported now?
2. This change handler feels a bit ugly as it's handling `onload` and asynchronously setting values on objects. Seems that up until now we didn't need to handle async value setting via change handlers, wondering if we should consider adding `Promise` handling or keep it for sync values only?
3. Without an actual usecase for file picker it's hard to tell what we need, but wondering if dataURL and array buffer would be enough for now? Or maybe I should use something entirely different here?

Janpot: Similar to what I suggested with the DatePicker, remove `onChangeHandler`, program it in the component itself.

Something along the lines of
```tsx
function FilePicker({ multiple, onFilesChange, ...props }: Props) {
  const handleChange = React.useCallback(async (event) => {
    const files = await Promise.all(Array.from(event.target.files, async (file) => {
      return {
        name: file.name,
        type: file.type,
        base64: await readAsBase64(file)
      }
    }))
    onFilesChange(files)
  }, [onFilesChange]);
  
  return <MuiTextField {...props} type="file" onChange={handleChange} inputProps={{ multiple }} />;
}
```

bytasv: Thanks a lot! I didn't got it the first time (with DatePicker), but now it makes sense 👍

packages/toolpad-components/src/FilePicker.tsx
```diff
@@ -0,0 +1,94 @@
...
+      inputProps={{ multiple }}
+      onChange={handleChange}
+    />
+  );
+}
+
+export default createComponent(FilePicker, {
+  argTypes: {
+    value: {
+      typeDef: { type: 'file' },
```
Janpot, line 66:
```suggestion
      typeDef: { type: 'object' },
      visible: false
```

packages/toolpad-core/src/types.ts
```diff
@@ -75,7 +75,7 @@ export type BindableAttrEntries = [string, BindableAttrValue<any>][];
 export type SlotType = 'single' | 'multiple' | 'layout';
 
 export interface ValueTypeBase {
-  type: 'string' | 'boolean' | 'number' | 'object' | 'array' | 'element' | 'event';
+  type: 'string' | 'boolean' | 'number' | 'object' | 'array' | 'element' | 'event' | 'file';
```
Janpot, line 78: No need for a new type, this is an object. What we need is more power in the object type to represent a shape

Janpot:
Can you also add `helperText` properties to the component and to each arg type definition. See the other components for an example.

bytasv:
@Janpot helper text added

> Also, should we put a bit of effort in making this look a bit better? It feels a bit out of placed compared to the MUI components. Perhaps start by making a benchmark of Material Design (visually) compatible Filepicker components that have been created in the wild?

Part of this PR - I don't think so. I don't see this being a blocker for the functionality. It does however inherit props from `TextField` so it isn't entirely "out of place".
However I do agree that it should look better, though question is whether we should do a special implementation for toolpad only or cooperate with core and finally have file picker component in core library?

apedroferreira:
> -edit: alternatively I could try to use a `Button` component so it looks like a button to select files and not like an input

Sounds like a good idea to use an MUI button at least!

bytasv:
@apedroferreira replaced implementation with button

CleanShot 2023-01-17 at 12 59 29

Janpot:
> Also with an Upload icon if there is one.

Some more options:
* https://mui.com/material-ui/material-icons/?query=file&selected=AttachFile
* https://mui.com/material-ui/material-icons/?query=file&selected=FileOpen
* Maybe there are better to find in https://mui.com/material-ui/material-icons/?query=file ?

And perhaps also show a list of file names that have been selected?

> However I do agree that it should look better, though question is whether we should do a special implementation for toolpad only or cooperate with core and finally have file picker component in core library?

We can always start an implementation and promote it to core later. In any case, I can't seem to find a MUI core or X feature request for a file input, and neither anything in notion (@joserodolfofreitas Would something like this make sense for X at some point?). Maybe step one is to start tracking the demand and benchmark potential solutions?

e.g. a quick google search
- https://codepen.io/Josantonius/pen/BzNwXz
- https://dev.to/j_sakamoto/how-to-implement-a-material-designed-input-type-file-field-built-on-mudblazor-640
- https://codepen.io/akhmedgaziev/pen/JdVQQm
- https://merlosy.github.io/ngx-material-file-input/
- interestingly, it doesn't seem to bring up any with a file drop zone

bytasv:
> And perhaps also show a list of file names that have been selected?

Right in canvas? I'd leave that to the end user to decide, by using other components

> Also with an Upload icon if there is one.

Is the suggestion to hardcode icon along side the `content`?

Janpot:
> Right in canvas? I'd leave that to the end user to decide, by using other components

Yes, that would be just like the native file input, so I don't think it's an unreasonable functionality. We're a low-code tool for internal applications, so our default functionality should cover the majority of use-cases.

> Is the suggestion to hardcode icon along side the content?

The user should get a good default experience. We can allow overriding behavior, but the default should look good and cover the majority of use-cases.

bytasv:
> > Right in canvas? I'd leave that to the end user to decide, by using other components
> 
> Yes, that would be just like the native file input, so I don't think it's an unreasonable functionality. We're a low-code tool for internal applications, so our default functionality should cover the majority of use-cases.
> 
> > Is the suggestion to hardcode icon along side the content?
> 
> The user should get a good default experience. We can allow overriding behavior, but the default should look good and cover the majority of use-cases.

Can you define exactly what you want to see?

I had a default file input behaviour implemented which was just changed to the button, it did show file name out of the box, it also shown if multiple files were selected.

At this point it's not really clear how it should look like.

Janpot:
> At this point it's not really clear how it should look like.

Is it possible to make it look just like the native file input, but with an MUI Button (that has an icon for extra visual feedback)?

bytasv:
Of course it's possible, my question is just, what's the value in current stage of "over optimising" this component so much? If we want default behaviour then why not stick with previous implementation of textfield?

joserodolfofreitas:
> Would something like this make sense for X at some point?

Yes. If you're willing to get started with it, it'd be great. We can incorporate it into MUI X later.

We have an [issue](https://github.com/mui/mui-x/issues/2901) on our repository. And we had a couple of requests for a FilePicker/ Dropzone component in the 2022 survey.
However, at the moment, and with the current bandwidth, it's unlikely that we can undertake this task in 2023. 
The initial plan is to expand the components portfolio with charts and maybe advanced TreeView this year.

Janpot:
> what's the value in current stage of "over optimising" this component so much?

I don't think we're over optimizing here, any PR comment can be dismissed by using the term "over optimizing". The expectation was to visually tie in this component with the rest of the components, ideally without sacrificing too much functionality.

@joserodolfofreitas Thanks, I completely missed that ticket. Been searching on "File input" and "File picker" and "Filepicker" but it doesn't seem to appear on the first few pages. (Makes me think we could introduce a keywords section in the issue template, but that's besides the question). We'll create a basic implementation of a FileInput for toolpad, and we can promote to X when there's more demand.

bytasv:
> I don't think we're over optimizing here

And I think differently. I already had a version done which shows by default file name that's selected and uses default MUI TextField styles.

I can revert changes to that or keep this implementation that uses Button style. What matters is that actual file pick/upload works, and without an actual design or benchmark I don't see the point in optimising this without any external feedback. This can be further improved/optimised if you (or we as entire team) see the benefit.

> any PR comment can be dismissed by using the term "over optimizing"

Any PR that CAN be dismissed by any reasonable argument should be dismissed, this way we can save time and focus on things that matter and that have positive impact.

So in order to stop bikeshedding - which of the 2 options you want me to keep?

Janpot:
There are two different concerns in tension here:

1. We'd like to adhere mostly to a single visual language across our components so that our users can create coherent UIs
2. A component should visually represent the state it is in. e.g. a text input or select shows the actual value, even when it's not being interacted with.

- The default file input gives us 2.
- The Button implementation gives us 1.
- My proposal would be to accompany that Button with a Typography to give use 1. and 2.

I understand that this would be some extra work, so if I really had to choose, I'd say to use the default file input. It has the better UX for the user, and is easier to improve for us afterwards.

Before we merge, if you're up for it, I'd propose to add a simple test, like we have for all the other components.

bytasv:
This is how the 1st option looked like pretty much:

https://codesandbox.io/s/red-water-uqq1po?file=/demo.tsx

CleanShot 2023-01-17 at 19 49 08@2x

So saying that "it does not adhere to single visual language" is a bit of exaggeration, don't you think? While I agree it can be further OPTIMISED by getting rid of default browser button, I don't see value at this point as per comments explained before - there is no design (which means its likely going to change anyway) and we haven't done any benchmark (which means its likely to change anyway).

> Before we merge, if you're up for it, I'd propose to add a basic test, like we have for all the other components.

Sure I will try to add tests only if playwright allows working with file inputs 🤔

Janpot:
> So saying that "it does not adhere to single visual language" is a bit of exaggeration, don't you think?

👍  Ok, I see now where our misunderstanding originates from. For me, having a browser native button nested inside of a material design textfield would be the definition of "it does not adhere to single visual language". But I understand that the threshold is interpreted differently by different people.

bytasv:
reverted to textfield version and added test, please review

Janpot:
There's one thought I had which may be a bit of a nit, but is related to updating this to a proper FilePicker component in the future. I can imagine that new component won't have MUI `TextField` specific properties like `variant`, `size` or `fullWidth`. It could be tricky to migrate if we let users build apps with these properties and have to upgrade this component to a version that doesn't have them. Would it make sense to omit those properties for now, so that the next programmer working on this code has an easier job?

bytasv:
Props removed

Janpot (approved):

PR #1539, opened 2023-01-10
Reorganize shared utils

Janpot:
Create a `useNonNullableContext` function to replace `createProvidedContext` down the line. The problem with `createProvidedContext` is that it makes it impossible for the react eslint rules to detect stable/unstable context values. In this PR I'm replacing some instances of `createProvidedContext` with this new hook and fixing the associate eslint issues

I'm also reordering a bit the utility imports from `@mui/toolpad-core` in anticipation of creating a utils package that we can reuse. `@mui/toolpad-core` is intended to share logic between the components and the runtime, we should aim at keeping it lean, utilities mentioned for broader use can go in their own package. For this package we will encourage subpath imports to maximize treeshakeability e.g.
```tsx
import { interleaveJsx } from '@mui/toolpad-utils/react'
```

I'm not making that switch in this PR, but as a first step, I'm just disallowing importing utilities from `@mui/toolpad-core`.


apedroferreira (approved):
> I'm also reordering a bit the utility imports from @mui/toolpad-core in anticipation of creating a utils package that we can reuse. @mui/toolpad-core is intended to share logic between the components and the runtime, we should aim at keeping it lean, utilities meant for broader use can go in new package. For this package we will encourage subpath imports to maximize treeshakeability e.g.

Sounds great to keep the bundle size low!

PR #1541, opened 2023-01-10
Remove obsolete dom reducer actions

Janpot:
Taking a look at the undo/redo logic. Trying to move more of the logic inside of the reducer. Saw an opportunity to simplify some of the logic (always prefer positive logic) and remove some of the actions. Having a single `DOM_UPDATE` action should hopefully make it easier to detect the nature of an update (updating the same property, or a new one)


apedroferreira (approved):
Looks good, definitely more readable and shorter!

We can also remove some methods such as `domApi.setNodeNamespace` and `domApi.setNodeName` and make them usable only with `domApi.update` if you think it's worth doing. I kept them separate because I think they're still kind of useful like that.

PR #1542, opened 2023-01-11
Fix binding to objects with more than one property

Janpot:
Currently, adding a binding `{ a:1, b:2 }` breaks the evaluator as it is interpreting this as a code block instead of an expression. This PR fixes the runtime behavior. Also opening https://github.com/mui/mui-toolpad/issues/1543 to deal with the editor


test/integration/bindings/index.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+  const runtimeModel = new ToolpadRuntime(page);
+  await runtimeModel.gotoPage(app.id, 'bindings');
+
+  const test1 = page.getByText('-test1-');
+  await expect(test1).toBeVisible();
+  const color = await test1.evaluate((elm) =>
+    window.getComputedStyle(elm).getPropertyValue('color'),
+  );
+  expect(color).toBe('rgb(25, 118, 210)');
+  await expect(page.getByText('-test2-')).toBeVisible();
```
apedroferreira, line 32: This one is testing that even if `sx` is `{` the element still shows?

apedroferreira (approved):

test/integration/bindings/index.spec.ts
Janpot: yes, and that the app doesn't crash

oliviertassinari:
Using a bisection, I have found this commit as breaking the support for

```js
function run() {
  return "Not found"
}
run()
```

and

```
const nonProductScopeLabels = ["support: commercial"]

nonProductScopeLabels
```

in the binding. This was reported by Greg because https://master--toolpad.mui.com/_toolpad/app/cl6rqzry10009arlv9sto6qja/pages/ip23ggo throws 

Screenshot 2023-01-13 at 23 07 38

(my previous link gets redirected to a wrong URL but this is a different bug)

Janpot:
🤔 hmm, this was always intended to be javascript expressions. But you should be able to

```tsx
(() => {
  if (getRepositoryDetails.data) {
    return `Repository ID: ${getRepositoryDetails.data.id}`
  } else {
    return "Not found"
  }
})()
```

and

```tsx
(() => {
  const nonProductScopeLabels = ["support: commercial"]

  return materialUI.rows
    .concat(muix.rows)
    .filter((issue) => {
      const withoutNonProductScopeLabels = issue.labels.filter(
        (label) => !nonProductScopeLabels.includes(label.name)
      )
      return withoutNonProductScopeLabels.length === 1
    })
    .sort((issueA, issueB) => {
      if (issueA.state === "open") {
        return 1
      }
      if (issueB.state === "open") {
        return 1
      }
      return issueB.number - issueA.number
    })
})()
```

Otherwise we need to take it back to the drawing board

oliviertassinari:
@Janpot `(() => {` I will do this fix for now, Greg needs the app to work until we decide what's the right solution (in case it's not this one)

Issue #747, opened 2022-08-05
Support Tabs component


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

https://mui.com/material-ui/react-tabs/

One of: https://github.com/mui/mui-toolpad/issues/78

bytasv:
@prakhargupta1 can you confirm if I understand correctly that we want something ala `Tabbed container` that retool offers? 
<img width="81" alt="CleanShot 2023-01-02 at 15 21 26@2x" src="https://user-images.githubusercontent.com/437214/210237030-2395611b-48f5-4157-ada3-7cea7ed57e67.png">

<img width="574" alt="CleanShot 2023-01-02 at 15 21 59@2x" src="https://user-images.githubusercontent.com/437214/210237052-a1e2263c-fde0-4c9e-82d3-71f29e889275.png">

There is also `Tabs` component which allows controlling other components via special configuration
<img width="659" alt="CleanShot 2023-01-02 at 15 23 34@2x" src="https://user-images.githubusercontent.com/437214/210237309-64a807b8-f427-4f61-9a73-2eded251605f.png">

Should we implement one, the other or both?

prakhargupta1:
We should do Tabbed containers.  
Example from appsmith [here](https://docs.appsmith.com/reference/widgets/tabs#:~:text=The%20tabs%20widget%20is%20a,user%20based%20on%20their%20choices.)

JerryWu1234:
@Janpot 

Today I doved into this issue. 
Tabs can only receive only one `children` property.
my suggestion is change children into a type of array.

apedroferreira:
> @Janpot
> 
> Today I doved into this issue. Tabs can only receive only one `children` property. my suggestion is change children into a type of array.

An array of `layoutSlot` props might work? But not sure if there would be any obstacles that could be complicated to deal with.

PR #1549, opened 2023-01-12
Add tabs and container components

bytasv:
- Closes https://github.com/mui/mui-toolpad/issues/747

https://user-images.githubusercontent.com/437214/212068581-f903c5ae-81be-4e6a-8270-ecca04c55ffb.mp4

@prakhargupta1 I know it's not a "tabbed container" that was discussed originally in the task, but IMO this comes closer to what we want to achieve right now.

If there won't be objections I would propose to continue polishing the approach as follow up as I realised to do something as a tabbed container we would probably need to do major work on how things are stored in our data structure.

This PR introduces 2 components:
- Tabs
- Container which has "visible" control

By using Tabs UI we can control visibility of container.

Now to actually do a "tabbed container" which would automatically swap items, I'll probably need some guidance either from @apedroferreira or from @Janpot. If any of you guys have any tips before going that road - how should we handle adding components to a container and switching active container within same component? (asking it here, as I'm not sure if we will agree on a current implementation as a intermediate solution or should that be replaced with more advanced implementation)

### DIFF
--- a/packages/toolpad-components/src/Container.tsx
+++ b/packages/toolpad-components/src/Container.tsx
@@ -1,15 +1,15 @@
 import * as React from 'react';
-import { Container as MUIContainer } from '@mui/material';
+import { Container as MUIContainer, ContainerProps } from '@mui/material';
 import { createComponent } from '@mui/toolpad-core';
+import { SX_PROP_HELPER_TEXT } from './constants';
 
-interface Props {
-  children: React.ReactNode;
+interface Props extends ContainerProps {
   visible: boolean;
 }
 
-function Container({ children, visible, ...props }: Props) {
+function Container({ children, visible, sx, ...props }: Props) {
   return visible ? (
-    <MUIContainer disableGutters sx={{ padding: 1 }} {...props}>
+    <MUIContainer disableGutters sx={sx} {...props}>
       {children}
     </MUIContainer>
   ) : null;
@@ -25,6 +25,11 @@ export default createComponent(Container, {
       typeDef: { type: 'boolean' },
       defaultValue: true,
       helperText: 'Control whether container element is visible.',
+    },
+    sx: {
+      helperText: SX_PROP_HELPER_TEXT,
+      typeDef: { type: 'object' },
+      defaultValue: { padding: 1, border: 'solid 1px' },
     },
   },
 });
--- a/packages/toolpad-components/src/Container.tsx
+++ b/packages/toolpad-components/src/Container.tsx
@@ -1,0 +1,30 @@
+import * as React from 'react';
+import { Container as MUIContainer } from '@mui/material';
+import { createComponent } from '@mui/toolpad-core';
+
+interface Props {
+  children: React.ReactNode;
+  visible: boolean;
+}
+
+function Container({ children, visible, ...props }: Props) {
+  return visible ? (
+    <MUIContainer disableGutters sx={{ padding: 1 }} {...props}>
+      {children}
+    </MUIContainer>
+  ) : null;
+}
+
+export default createComponent(Container, {
+  argTypes: {
+    children: {
+      typeDef: { type: 'element' },
+      control: { type: 'layoutSlot' },
+    },
+    visible: {
+      typeDef: { type: 'boolean' },
+      defaultValue: true,
+      helperText: 'Control whether container element is visible.',
+    },
+  },
+});
--- a/packages/toolpad-components/src/Tabs.tsx
+++ b/packages/toolpad-components/src/Tabs.tsx
@@ -1,0 +1,60 @@
+import * as React from 'react';
+import { Tabs as MUITabs, Tab } from '@mui/material';
+import { createComponent } from '@mui/toolpad-core';
+
+interface TabProps {
+  title: string;
+  name: string;
+}
+
+interface Props {
+  active: string;
+  onChange: (value: number) => void;
+  tabs: TabProps[];
+  defaultValue: string;
+}
+
+function Tabs({ active, onChange, tabs, defaultValue }: Props) {
+  return (
+    <MUITabs
+      value={active || defaultValue}
+      onChange={(event, value) => {
+        onChange(value);
+      }}
+    >
+      {tabs.map(({ title, name }) => (
+        <Tab label={title} value={name} key={name} />
+      ))}
+    </MUITabs>
+  );
+}
+
+export default createComponent(Tabs, {
+  layoutDirection: 'horizontal',
+  argTypes: {
+    active: {
+      typeDef: { type: 'string' },
+      onChangeProp: 'onChange',
+      defaultValueProp: 'defaultValue',
+      helperText: 'Currently active tab.',
+    },
+    defaultValue: {
+      label: 'Default active tab',
+      typeDef: { type: 'string' },
+      defaultValue: 'tab-one',
+      helperText: 'The tab which will be active by default.',
+    },
+    tabs: {
+      typeDef: { type: 'array' },
+      defaultValue: [
+        {
+          title: 'Tab one',
+          name: 'tab-one',
+        },
+        { title: 'Tab two', name: 'tab-two' },
+        { title: 'Tab three', name: 'tab-three' },
+      ],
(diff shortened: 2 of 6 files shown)


apedroferreira:
Thanks for this effort, seems to work well!

I'm not sure if I like the approach with the `visible` prop, even though it works.
It seems like the only reason we are introducing the `visible` prop in the new `Container` component is for this use case.
I would prefer if we were able to make it work like you mentioned, even though it's more challenging:

> to actually do a "tabbed container" which would automatically swap items

I think this alternative sounds a lot better and more intuitive.
But we would probably need a new type of `element` prop that can take a list of children, as opposed to just one. The same thing as the `children` prop in the `Paper` component, for example, but with a list instead of a single view.

Then in the `Tabs` component logic we could make it render the correct view according to the active tab.

This might require special logic in the render overlay so that when we dropped an element in a slot, this element is added to the correct tab... I'd have to look into it deeper to see if there would be many more obstacles, but this is at least where I would start.

packages/toolpad-components/src/Container.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { createComponent } from '@mui/toolpad-core';
+
+interface Props {
+  children: React.ReactNode;
+  visible: boolean;
+}
+
+function Container({ children, visible, ...props }: Props) {
+  return visible ? (
+    <MUIContainer disableGutters sx={{ padding: 1 }} {...props}>
```
apedroferreira, line 12: If we have padding here maybe it should be controllable?

packages/toolpad-components/src/Tabs.tsx
```diff
@@ -0,0 +1,60 @@
...
+      label: 'Default active tab',
+      typeDef: { type: 'string' },
+      defaultValue: 'tab-one',
+      helperText: 'The tab which will be active by default.',
+    },
+    tabs: {
+      typeDef: { type: 'array' },
+      defaultValue: [
+        {
+          title: 'Tab one',
```
apedroferreira, line 51: Why not use numbers instead of spelling them out? `Tab 1`, `Tab 2`, `Tab 3`?

bytasv:
> I'm not sure if I like the approach with the visible prop, even though it works.
It seems like the only reason we are introducing the visible prop in the new Container component is for this use case.

At the moment yes, there is nothing else that could use this prop, however I think controlling visibility is something that we currently lack in general and it should be useful. I.e. if we built a form with checkbox/radio and expected that some content would appear depending on selection, this component could also be used. Unless we decide on some different method to control visibility of components

packages/toolpad-components/src/Tabs.tsx
bytasv: No reason, just what seemed to make tab title longer, anyways this is just for default values, so nothing prevents users to use whatever they want

apedroferreira:
> At the moment yes, there is nothing else that could use this prop, however I think controlling visibility is something that we currently lack in general and it should be useful. I.e. if we built a form with checkbox/radio and expected that some content would appear depending on selection, this component could also be used. Unless we decide on some different method to control visibility of components

Yeah that's a good point. It sounds fine to me to keep the prop in the `Container` component, at least it sounds better than having this prop in every component... Unless it can be controlled with the `sx` prop?

packages/toolpad-components/src/Container.tsx
apedroferreira: Maybe we can let the user control it with the `sx` prop if they want to?

packages/toolpad-components/src/Container.tsx
```diff
@@ -0,0 +1,30 @@
...
+      {children}
+    </MUIContainer>
+  ) : null;
+}
+
+export default createComponent(Container, {
+  argTypes: {
+    children: {
```
apedroferreira, line 20: Another thing: there is a bug I've found so if this prop is not called `children` it doesn't work, but it should. I'm including the fix in https://github.com/mui/mui-toolpad/pull/1527

prakhargupta1:
Interesting! I think for now this looks good to me.
QQ: Is it possible for this container component to have sx prop? The Reason I am asking it is, I tried previewing tabs, and they appear like shown below, if there was an option to add border to a container then I could entire page in a container (like paper but without any elevation)


Screenshot 2023-01-13 at 12 08 06 AM

bytasv:
> QQ: Is it possible for this container component to have sx prop?

@prakhargupta1 yes ofc, I can add it. I just wanted to share this option quickly, but if we are fine with intermediate solution then I can improve it as much as needed and work on tabbed container improvements in follow up

Janpot:
We will need to expand on the `'object'` argtype, to include its properties. That way we can resolve it recursively in the runtime and find and render nested children and replace them in the object before we pass it as props to the component.

```tsx
export interface ObjectValueType extends ValueTypeBase {
  type: 'object';
  /** @deprecated lets remove this: */
  schema?: string;
  properties?: Record<string, PropValueType>
}

export interface ArrayValueType extends ValueTypeBase {
  type: 'array';
  /** @deprecated lets remove this: */
  schema?: string;
  items?: PropValueType
}
```

bytasv:
@apedroferreira added `sx` prop with defaults for padding and border as @prakhargupta1 requested

apedroferreira (approved):
Looks good to me as a first approach!

Like I said maybe we could consider also controlling visibility with the `sx` prop instead of having a prop just for it in the Containers, but I'll leave that up to you, as we can also improve this with the more difficult solution later.

packages/toolpad-components/src/Tabs.tsx
```diff
@@ -0,0 +1,60 @@
...
+        <Tab label={title} value={name} key={name} />
+      ))}
+    </MUITabs>
+  );
+}
+
+export default createComponent(Tabs, {
+  layoutDirection: 'horizontal',
+  argTypes: {
+    active: {
```
Janpot, line 35: This should be `value`

bytasv: Can you clarify what you mean? the name of prop? if yes, why?

bytasv:
> I'm not sure about the tabs. It feels like a Select to me with a different UI. I'm not sure how the migration path from this towards a real tabs component would work. Would it make more sense to have "tabs" as an option on the Select component? That way we keep the Tabs available for implementing real tabs, with surface one can build UI in?

retool has both `Tabs` and `TabbedContainer` which I guess would be "real tabs" as you call it. IMO having `Tabs` placed under `Select` would significantly impact it's discoverability.

> One thing to note is that the container allows users creating flows that cause vertical layout shift, which is something we have avoided up until now.

Avoided intentionally? Or is this just part of a change with new features? Also if ensuring no layout shifts happens is so important, users can use containers with fixed height via `sx` prop I assume.

packages/toolpad-components/src/Tabs.tsx
Janpot: > the name of prop? if yes, why?

yes, because:

1. Conventions: we generally use the combination `value`/`onChange` to denote the main controlled prop of a component (yes, there are specific situations where we deviate from this, e.g. dataGrid.)
2. The MUI Tabs uses `value`
3. This component as it currently is, is in essence a Select, and the MUI Select uses `value`
4. The word "active" is an adjective, and a "tab" is "a thing", so we should use a noun for naming it in our code. the word "value" is a noun.

Janpot:
Retool ended up with a lot of components that kinda do the same, but different. And I'm not sure that it's such a good thing. Why would you need `Tabs` if you have `Tabbed container`? The latter will do for 99% of the use-cases. And former has an equal status as a component for its niche use-case. IMO it would make more sense to have a set of core components that do 99% of the work and put all the niche ones further away.

But now that I think of it, we can always migrate the current `Tabs` as an option under `Select` when the time comes. So nevermind.

> Avoided intentionally?

It was avoided intentionally. The main idea was that by default, when it comes to building good UX, users should fall in a pit of success. Making it hard to build layout shifting UI was part of that. But users were always able to circumvent with the `sx` prop, so I'm not saying this should be a hard rule. Just sharing design considerations that were adhered to up until now.

PR #1565, opened 2023-01-16
Fix navigation through DOM views with pages

### LINKED ISSUE #1551: Reload of a page editor always switches to the first page
### Duplicates

### Latest version

### Steps to reproduce 🕹

Steps:

1. Open the editor on an app with multiple pages
2. Open any of the pages other than the first page
3. Hit reload

### Current behavior 😯

The url switches to the first page

### Expected behavior 🤔

The url stays on the selected page

### Context 🔦

This line is the culprit https://github.com/mui/mui-toolpad/blob/4f4f4f1e80c943098be66a9417f455006a9ae1dc/packages/toolpad-app/src/toolpad/AppEditor/index.tsx#L58

This will need an integration test

### Your environment 🌎

_No response_

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/1551

- Loads first page in the URL, even if it's not the first page or a page.
- Does not get rid of query params in URL on first page load.
- Always use DOM view state as the source of truth for the current view. Load initial view from URL. Update URL as a side-effect.

Preview: https://toolpad-pr-1565.onrender.com/


packages/toolpad-app/src/toolpad/DomLoader.tsx
```diff
@@ -362,25 +364,49 @@ const UNDOABLE_ACTIONS = new Set<DomActionType>([
   'DESELECT_NODE',
 ]);
 
+export const getCurrentPageDomView = (location: Location): DomView => {
```
apedroferreira, line 367: Should I move this to some other file? Not sure where to move it to.

oliviertassinari:
Do we know what created this regression?

Janpot:
I'm not sure about this, it feels a bit brittle. How should the undoable views interact with the browser back button?

I don't believe it would currently be possible to keep the browser history in sync with the undo stack. I feel like as soon as we're in the editor flow, the undoable view changes shouldn't create new browser history entries. And if that is the case then:

1. Should we be not using react-router for the undoable views, instead have the route defined by the editor state? (i.e. not use `<Route>`, but switch routes based on editor state)
2. Initialize the state using the current url
3. in an effect synchronize all state changes back to the url

apedroferreira:
I tried to make the source of truth be the DOM state, but yeah I did not anticipate this issue coming from the URLs and I was tried out this solution but it's not the simplest...
Picking a single source of truth and not letting the other influence anything sounds like a way to do it - so either always use URLs (could be tough) or always using the DOM state without relying on URLs.
I can probably simplify this though instead of trying to make it work so universally - will give it a try.

apedroferreira:
> I'm not sure about this, it feels a bit brittle. How should the undoable views interact with the browser back button?
> 
> I don't believe it would currently be possible to keep the browser history in sync with the undo stack. I feel like as soon as we're in the editor flow, the undoable view changes shouldn't create new browser history entries. And if that is the case then:
> 
> 1. Should we be not using react-router for the undoable views, instead have the route defined by the editor state? (i.e. not use `<Route>`, but switch routes based on editor state)
> 2. Initialize the state using the current url
> 3. in an effect synchronize all state changes back to the url
> 
> > Do we know what created this regression?
> 
> @oliviertassinari This was introduced when taking the view switching and selection into account for undo/redo. I think the initial implementation gets into a chicken/egg situation between application state (undo stack) and the browser history. I believe it would be best to make one the single source of truth (application state) and have the other simply follow suit. For that we'll have to break out of `react-router` for this part.

@Janpot how about this? I followed the approach you suggested, seems like a good solution!

packages/toolpad-app/src/toolpad/AppEditor/index.tsx
```diff
@@ -43,48 +42,73 @@ interface FileEditorProps {
...
-        navigate(`/app/${appId}/connections/${currentView.nodeId}`, { replace: true });
-        break;
-      case 'codeComponent':
-        navigate(`/app/${appId}/codeComponents/${currentView.nodeId}`);
-        break;
-      default:
+    if (currentView.kind === 'page') {
+      const newPathname = `/app/${appId}/pages/${currentView.nodeId}`;
+
+      if (newPathname !== location.pathname) {
```
Janpot, line 54: Can this be shortcircuited in the beginning of the function?

```tsx
if (newPathname === location.pathname) {
  return
}
```

packages/toolpad-app/src/toolpad/AppEditor/index.tsx
```diff
@@ -43,48 +42,73 @@ interface FileEditorProps {
...
-  const app = appDom.getApp(dom);
-  const { pages = [] } = appDom.getChildNodes(dom, app);
+  const { currentView } = useDom();
 
+  const location = useLocation();
   const navigate = useNavigate();
 
-  const firstPage = pages.length > 0 ? pages[0] : null;
-
   React.useEffect(() => {
```
Janpot, line 50: I believe this could be a bit DRYer if written as:

```tsx
function getPathnameForView (appId, view) {
  switch (view.kind) {
    case 'page': return `/app/${appId}/pages/${view.nodeId}`;
    // ...
  }
}

React.useEffect(() => {
  const desiredPath = getPathnameForView(currentView)
  if (desiredPath !== location.pathname) {
    navigate({  pathname: desiredPath }, { replace: true });
  }
}, [appId, currentView, location.pathname]
```

packages/toolpad-app/src/toolpad/AppEditor/index.tsx
```diff
@@ -43,48 +42,73 @@ interface FileEditorProps {
...
-              <Navigate to={`pages/${firstPage.id}`} replace />
-            ) : (
-              <NoPageFound appId={appId} />
-            )
-          }
-        />
-      </Route>
-    </Routes>
+    <AppEditorShell appId={appId}>
+      {currentView.kind === 'page' ? (
```
Janpot, line 99: The intent of this isn't very clear from structuring the code this way. (The intent being "only one of these needs to be rendered"). Moreover, the last element needs to always be kept in sync with the lines above. Ideally this would be written as a `switch`. Infortunately there's still no way to write switch expressions or do pattern matching in javascript, but perhaps this could be written as:

```tsx
{(() => {
  switch (currentView.kind) {
    case 'page': return <PageEditor appId={appId} nodeId={currentView.nodeId} />;
    // ...
  }
})()}
```

It's a bit ugly with all those brackets. Alternatively this could be written in a `useMemo`, or extracted as a function.

packages/toolpad-app/src/toolpad/DomLoader.tsx
```diff
@@ -362,25 +366,54 @@ const UNDOABLE_ACTIONS = new Set<DomActionType>([
   'DESELECT_NODE',
 ]);
 
+export const getInitialPageDomView = (location: Location, defaultPageNodeId?: NodeId): DomView => {
```
Janpot, line 369: This looks like the inverse of the `getPathnameFromView`. Perhaps we could put the two functions physically together and name them similarly?

`getPathnameFromView` and `getViewFromPathname`?

Janpot:
> how about this? I followed the approach you suggested, seems like a good solution!

I like this implementation a lot, if it works well in practice. Just a few stylistic comments

apedroferreira:
> I like this implementation a lot, hopefully it works well in practice. Just a few stylistic comments

Great suggestions again, there was definitely some unneeded repetition.
I've added the suggestions now.

packages/toolpad-app/src/utils/domView.ts
```diff
@@ -0,0 +1,45 @@
...
+export function getPathnameFromView(appId: string, view: DomView): string | null {
+  switch (view.kind) {
+    case 'page':
+      return `/app/${appId}/pages/${view.nodeId}`;
+    case 'connection':
+      return `/app/${appId}/connections/${view.nodeId}`;
+    case 'codeComponent':
+      return `/app/${appId}/codeComponents/${view.nodeId}`;
+    default:
+      return null;
```
Janpot, line 24: Should there be any view for which no pathame exists? Perhaps this should throw?

apedroferreira: Sounds good! I've added it in a new commit to include small type adjustments since the result can't be `null`

Janpot (approved):
👌 Very nice!

oliviertassinari:
Great, the DX is much better with this fix. Also the e2e test was much needed, great to see it 👌.

PR #1568, opened 2023-01-17
Rename Tabs main prop

bytasv:
After feedback on https://github.com/mui/mui-toolpad/pull/1549


Janpot (approved):
👍 

Just to note: For me it's not necessary to write a migration for this. Since it hasn't been merged that long ago there is probably no code using this already. For the future, be mindful that changes like this will require a migration to not break user's existing apps.

PR #1583, opened 2023-01-19
Fix: Column changes dropped on prop update

### LINKED ISSUE #1486: Resizing columns drops all previous changes to columns
### Duplicates

### Latest version

### Steps to reproduce 🕹

Steps:

1. Go to https://master--toolpad.mui.com/_toolpad/app/cl91ik41i0454axrr578xjjrp/pages/eo03kwx
2. Delete a column in the DataGrid
3. Resize any other column

### Current behavior 😯

The deleted column reappears

There is another case where all columns disappear but I can not deterministically recreate that

### Expected behavior 🤔

No other changes occur except the resizing

### Context 🔦

_No response_

### Your environment 🌎

_No response_

bharatkashyap:
- Closes #1486 

- Visual demonstration 

https://user-images.githubusercontent.com/19550456/213419327-5e90cc18-beef-4e5c-97ab-7c77e0b2235c.mov


Janpot:
Did you try adding a test? I can imagine it being tricky

apedroferreira (approved):

bharatkashyap:
> Did you try adding a test? I can imagine it being tricky

Yes, tricky specifically since this issue is only reproduced on doing a drag interaction directly with the DataGrid (changing width/order of the columns)

Looking to use this to simulate a width change operation on one of the columsn: https://playwright.dev/docs/api/class-locator#locator-drag-to

bharatkashyap:
We would also need to modify this test to use the `hr` template and potentially rename it to "can use HR template":
https://github.com/mui/mui-toolpad/blob/default-tempalte/test/integration/appTemplates.spec.ts#L8

test/integration/data-grid/index.spec.ts
```diff
@@ -24,3 +25,54 @@ test('Code component cell', async ({ page, browserName, api }) => {
...
+  await avatarColumn.dragTo(profileColumn);
+
+  // Expect the "Avatar" column to continue to be of type "boolean" instead of "link"
+
+  await expect(
+    editorModel.pageRoot
+      .getByRole('row', {
+        name: '1 Todd Breitenberg International http://spotless-octopus.name yes',
+      })
+      .getByTestId('CheckIcon'),
```
apedroferreira, line 76: How does this work - the icon already comes with a test id?

apedroferreira (approved):
The test looks good to me!

test/integration/data-grid/index.spec.ts
bharatkashyap: Yes, the icon already has the testId

bharatkashyap:
> The test looks good to me!

I'm running into trouble getting the drag/drop interactions to run properly on Firefox, do you have any experience figuring this out?

Janpot:
> They were supposed to have fixed that after I sent them a reproduction

The solution still required us to do something on our end, right? Like call `.hover()` twice or something?

apedroferreira:
> The solution still required us to do something on our end, right? Like call `.hover()` twice or something?

I didn't think so but it might be, I'll look into it again.

Issue #1535, opened 2023-01-09
Add GA events in Toolpad demo


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

1. Before clicking on create, which template did the user choose or if they clicked on Continue with my last app.
<img width="484" alt="Screenshot 2023-01-09 at 5 01 12 PM" src="https://user-images.githubusercontent.com/92228082/211298698-5a192b6d-267e-4f0e-a251-d4486e959311.png">

2. No. of clicks coming on:

- Self-host button
- roadmap
- schedule demo

Janpot:
Whenever I use this UI, it's not very clear to me. It feels like I have to click create, but a template is forced on me.

I wonder whether we should just show a grid where the options are all equal in effort to pick. e.g. show a grid of options:

- create blank application
- start from template 1
- start from template 2
- continue where you left off last time

apedroferreira:
> Whenever I use this UI, it's not very clear to me. It feels like I have to click create, but a template is forced on me.
> 
> I wonder whether we should just show a grid where the options are all equal in effort to pick. e.g. show a grid of options:
> 
> * create blank application
> * start from template 1
> * start from template 2
> * continue where you left off last time

Yeah, it probably could be better designed... Making everything seem equal in the UI sounds like a good thing to do.
Another alternative could be for all the action buttons to be centered like the template button and stacked vertically, instead of having those actions in the corner.
Not sure if it would be worthwhile to get Gerda's input on this so we can have something that looks ok?

prakhargupta1:
Agree on the thought of creating a grid like UI. There is an existing issue for this: https://github.com/mui/mui-toolpad/issues/1390

PR #1588, opened 2023-01-19
Add new custom GA events to demo

apedroferreira:
New custom Google Analytics events for the Toolpad demo.
Closes https://github.com/mui/mui-toolpad/issues/1535

The `toolpad_app_created` event already includes `app_template_id`, so it wasn't necessary to add anything new to know which template was used.

**New events:** 

- `toolpad_app_continue_latest`
Measures when user went to continue working on latest app. Includes `app_id`.

- `toolpad_self_host_click`
Measures click on self-host link. Includes `app_id` and an additional parameter `source` that tells whether the user clicked the self-host link in the demo bar or the query panel (see screenshots).

Screenshot 2023-01-19 at 19 40 29
Screenshot 2023-01-19 at 19 49 35

- `toolpad_roadmap_click`
Measures click on roadmap link in demo bar. Includes `app_id`.

- `toolpad_schedule_demo_click`
Measures click on "schedule demo" link in demo bar. Includes `app_id`.


Janpot:
Some of these wouldn't be more interesting to solve with UTM parameters on the links?

apedroferreira:
> Would it be more interesting to solve these with UTM parameters on the links? I feel like it could be more powerful.

I don't know exactly how UTM parameters work, but if they send an event automatically by clicking a link it should be a good idea. I had the idea that they only worked if the target page has Google Analytics (in which case this would still work for the self-host button anyway).
I'll try that out and look more into UTM params.

apedroferreira:
I've added UTM parameters to the links according to what I've read about how they should work, but I'm not sure how this replaces the usefulness of having custom events directly in Toolpad?

From what I've read, outbound clicks are tracked automatically with the `click` event, however in the GA dashboard it does not seem easy to view all `click` events distributed by which destination link they used.

And I guess that UTM params in links that lead to the MUI docs could be used to track where traffic came from - but having custom events in Toolpad's Analytics could be an easier way to view and analyze this data, and easier to integrate with funnel metrics and that kind of stuff if we wanted to?

apedroferreira:
I've removed UTM params from the links to roadmap and scheduling demo. 
Kept UTM params in the links to the Toolpad docs, just in case.
Will merge this tomorrow or a bit later if there's no more feedback!

Issue #749, opened 2022-08-05
Support Form component


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Examples 🌈

- https://ant.design/components/form/
- https://retool.com/components/form, https://retool.com/components/json-schema-form
- https://mantine.dev/form/use-form/
- https://docs.appsmith.com/reference/widgets/form
- https://refine.dev/docs/core/hooks/useForm/

bharatkashyap:
Sharing an idea for the form component here: For me the most useful feature would be if a form could be auto-generated from a pre-existing DataGrid on your page. The data on submission would feed into that DataGrid.

prakhargupta1:
The JSON form as demonstrated [here](https://www.youtube.com/watch?v=Zk6df9mOtQA) makes total sense. 
It can create or update a record in the data source, provide validations, read and display correct data types.
Let's discuss it further.

prakhargupta1:
I think we can have 2 types of forms:

1. A basic form that just takes some basic input and has 2 buttons; Submit and an optional Cancel/Reset. 
    Eg 1: A login form that just asks for username and password.
    Eg 2: A complaint box that asks for a user issue and shows an additional button to upload any file.

2. A JSON form, that generates a form based on a datagrid json schema. Now, it would be interesting to have this form as a slideout component. This is also a problem that I was facing, so when I was wanted to show details of a selected row, I had to make many Typography components that connected with each cell value I wanted to show.
 Example experience can be checked [here, point 1](https://www.superblocks.com/blog/alchemy-scaling-support-through-hypergrowth-with-superblocks)

bytasv:
IMO it shouldn't be too hard to to build form kinda like we build using paper component - drag and drop buttons, inputs, selects and other form elements into form and then using context or w/e else "register" them so that they values can be captured as a form input. I think we shouldn't hardcode form layout at all

apedroferreira:
> IMO it shouldn't be too hard to to build form kinda like we build using paper component - drag and drop buttons, inputs, selects...

This would also be more consistent with how the `List` are `Card` components are probably going to work.

PR #1598, opened 2023-01-24
Form component

bytasv:
Closes https://github.com/mui/mui-toolpad/issues/749

Drag-and-drop form component.
- Supports the 4 input types we have so far: TextField, Select, DatePicker and FilePicker
- Allow for submitting inputs as a group with custom field names
- Allow for resetting form values
- Supports default values
- Allows for validation according to input type
- Individual inputs outside forms can also validate and display errors
- Add prop categories (in this case, "validation" props)
- Customizable form controls out of the box. But form controls can also be hidden and the Button component can be used instead with `button`, `reset` and `submit` as options for button type

I'd like to add tests separately not to make this PR even bigger so created a separate issue here: https://github.com/mui/mui-toolpad/issues/1777 I'll keep track of it.

For automatically generating a form from external data I've created this separate issue: https://github.com/mui/mui-toolpad/issues/1770

https://user-images.githubusercontent.com/10789765/226696560-2a224385-2720-4d5a-8783-72996bcfcde4.mov

https://user-images.githubusercontent.com/10789765/226696542-87a581ba-a559-4013-bc2b-3149ac515a99.mov


packages/toolpad-app/package.json
```diff
@@ -110,7 +110,8 @@
     "superjson": "^1.12.2",
     "ts-node": "^10.9.1",
     "web-streams-polyfill": "^3.2.1",
-    "whatwg-url": "^12.0.0"
+    "whatwg-url": "^12.0.0",
+    "formik": "2.2.9"
```
apedroferreira, line 114: Isn't Formik unmaintained?
I think I even somewhere that the author himself is now recommending using other libraries such as `react-hook-form`.
I always liked Formik though and am most familiar with it compared to other libraries.

packages/toolpad-components/src/TextField.tsx
```diff
@@ -5,24 +5,29 @@ import {
...
   value: string;
   onChange: (newValue: string) => void;
   alignItems?: BoxProps['alignItems'];
   justifyContent?: BoxProps['justifyContent'];
+  name: string;
 };
 
-function TextField({ defaultValue, onChange, ...props }: TextFieldProps) {
+function TextField({ defaultValue, onChange, value, ...props }: TextFieldProps) {
+  const [field] = useField(props.name);
```
apedroferreira, line 20: So every `TextField` is using form context?
Wouldn't it be better to pass `field.onChange` to the `onChange` prop, and `field.value` to the `value` prop of this component somehow? (probably in the runtime)
Same for the other components where you followed this approach.

packages/toolpad-components/src/Form.tsx
```diff
@@ -0,0 +1,98 @@
...
+}
+
+export default createComponent(Form, {
+  argTypes: {
+    children: {
+      visible: false,
+      typeDef: { type: 'element' },
+      control: { type: 'layoutSlot' },
+    },
+    value: {
```
apedroferreira, line 71: Maybe `values` instead of `value`? But that's just personal opinion so feel free to disagree , just my idea.

packages/toolpad-components/src/Form.tsx
```diff
@@ -0,0 +1,98 @@
...
+          </form>
+        )}
+      </Formik>
+    </MUIContainer>
+  );
+}
+
+export default createComponent(Form, {
+  argTypes: {
+    children: {
```
apedroferreira, line 66: I have a fix for the ugly yellow message that shows in the properties panel in my `Add List component` PR, I should probably extract it from it... Then we wouldn't need `visible: false` (I guess that's why you added it, to hide that control warning?).

apedroferreira (changes requested):
Works pretty much like I expected!

How about validation/errors, do we plan to tackle that later? Maybe we can have a prop to pass a `yup` schema or something?

Left many comment but it's all up for discussion of course!

packages/toolpad-components/src/TextField.tsx
bytasv: I'm not sure about that, I guess one way to do that could be adding a "global form content" where we could catch all values and still keep using them for individual components, then as soon as we add form it could capture values inside it's own context.

However the main reason why I didn't do something like that is that I wanted to limited the surface of the changes and see whether it's worth to continue building on top. So I'd say it's up for a discussion

packages/toolpad-components/src/Form.tsx
bytasv: I was thinking the same, but the idea was to keep pretty much the same API as any other form element has, i.e. `FilePicker` also provides either array of objects or object with multiple values, but it's main prop is still called `value` 🤔 I'm also not sure whether it should be `value` or `values`, both would be fine with me. Maybe @Janpot can tell the final word?

packages/toolpad-components/src/Form.tsx
bytasv: Mmm I already forgot, but it might be that 😅

packages/toolpad-app/package.json
Janpot: for the record, we're currenlty already using `react-hook-form` in some places in Toolpad.

packages/toolpad-components/src/Form.tsx
Janpot: I look at the `value`/`onChange` pattern as an abstract API for any controlled component. Simple ones, like `input`, and more complex ones, like our `MapEntriesEditor` that composes multiple controlled inputs.

The benefit lies in this being an abstract API. controlled inputs can easily be swapped by other ones, or composed into more complex ones, or wrapped by higher order APIs, or made dynamic and passed as a prop.

So from my side +1 on `value`/`onChange`

packages/toolpad-app/package.json
bytasv: Thanks, will change the implementation accordingly

Janpot:
From a high level review, (I didn't go in depth yet), I see potentially two problems coming up:

1. Ideally this works automatically with the component name, but component names are global and unique. This is another case where we need local scope for a feature to work properly. a local scope would allow us to hide those form controls from the global scope, and allow them to have a name that's unique to the form, not to the page.
2. What happens when you add a TextField to the page? Where there is no form context available? My initial hunch would be that we shouldn't touch the individual components, but rather the form component should wrap its children.

Janpot:
> What you are describe sounds more like a general problem that Toolpad has and not something that's related to form or this PR. 

What I'm describing is a limitation of the current code base that needs to be refactored if we want to implement features like List/Form in a way that is not "hacking it in to make it look like it mostly works, covering up all the nasty side-effects". Which is certainly possible, but also it's just passing on the hot potato to the person that will handle the tickets that inevitably come in for this feature.

bytasv:
> What I'm describing is a limitation of the current code base that needs to be refactored if we want to implement features like List/Form

No it doesn't require refactoring if we want to implement the base requirements that we have today

> in a way that is not "hacking it in to make it look like it mostly works, covering up all the nasty side-effects"

Working around certain limitations that are imposed on architectural decisions made in the past is not necessarily a hack. Improving code by knowing more requirements is a natural way of evolving codebase and it doesn't really matter who's the person going to be who "inevitably come in for this feature". The more important thing is to understand whether we (as a team) even want or need to solve what you call "nasty side-effects" at all, as it's also possible that nobody will even care about it until certain point of time, where it's more likely that someone (at least those who upvoted original feature) will care for not having a feature they requested.

packages/toolpad-components/src/TextField.tsx
apedroferreira: Yeah, it wouldn't be an easy change.
It can be up to you, I don't mind us doing things in steps as long as it won't cause any extra trouble.

packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -268,6 +270,30 @@ function RenderedNodeContent({ node, childNodeGroups, Component }: RenderedNodeC
     return hookResult;
   }, [isLayoutNode, liveBindings, nodeId]);
 
+  const changeClosestFormValue = React.useCallback(
```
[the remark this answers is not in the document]
apedroferreira: Not sure what the specific issue is for the missing default value - you'd probably have to look into the bindings logic in the runtime to figure it out - there's a `useEffect` there that handles default values. (`src/runtime/ToolpadApp`, around line 349?). At least this is where i would look first.

There probably are some similarities between what you need to do here and what I did for the List component in https://github.com/mui/mui-toolpad/pull/1527 (looking for a review there btw to see if my latest method sounds good? Also maybe it would be helpful if we merged the List first.).
What I did there was to create a local scope inside the list component by making it set a context, on the runtime (`src/runtime/ToolpadApp`) that wraps all of it's children. In the case of a form, this context could probably hold the form state.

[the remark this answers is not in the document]
Janpot: How I would go about it without doing low-level work to toolpad to enable defining arbitrary scopes. 

1. create shared `FormValue` and `SetFormValue` contexts to be used by both the Form and individual components. These contexts are provided by the `Form` component to provide its `value` and `onChange` prop downstream:

```tsx
<Paper>
  <FormValue.Provider value={value}>
    <SetFormValue.Provider
      value={(field, newValue) => onChange({ ...value, [field]: newValue })}
    >
      {children}
    </SetFormValue.Provider>
  </FormValue.Provider>
</Paper>
```

In the form components we use these contexts, and if they exist we override the `value` and `onChange` from the props:

```tsx
const formValue = React.useContext(FormValue);
const setFormValue = React.useContext(SetFormValue);
inveriant(!!formValue === !!setFormValue, "...");

const inputValue = formValue ? formValue[name] : valueProp;
const setInput = setFormValue
  ? (newValue) => setFormValue(name, newValue)
  : onChangeProp;

return <TextField value={input} onChange={setInput} />;
```

2. If the above works, we have the basis, but we still need to filter out the components that are bound to a form in the bindings. The second part of 1. will be shared by every form-capable component, so we can extract it and put it in the runtime. e.g. we can add some definition to `ToolpadComponentConfig` like:

```ts
argTypes: {
  /* ... */
},
formValue: ['value', 'onChange']
```

In the runtime we can check for existence of this config when rendering the component and do the logic of 1. automatically, when preparing its properties.
Then we can also read it while setting up the bindings and omit these properties from creating scope variables (so just remove `scopePath` for those)


This is a two part explanation, the end result is two, but do 1. first to verify that the reasoning works.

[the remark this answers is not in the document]
apedroferreira: @Janpot 's idea sounds good to me!

> Could you please clarify what you mean by that? Currently it's implemented in a way that child components wouldn't have to know anything about form component

In this case this `onChange` handler would probably need to be passed as a prop to each component.

I guess one of the main advantages of using context would be that you wouldn't have to "search" for the closest form manually.
As for including form logic inside components themselves, I'm fine with that if we decide it's the easiest/best way to go about it.

apedroferreira:
This form component should be ready for merging/review, let me know if you find any bugs I might have missed.
This is a complex one again so no rush on the review!

Janpot:
Can you merge/rebase master?

Janpot:
Renaming a field does strange stuff to the form value:

https://user-images.githubusercontent.com/2109932/226882563-04e9ad35-ff87-4e20-89cc-0f9ce23e0c00.mov

apedroferreira:
> Renaming a field does strange stuff to the form value

I've fixed that, should be good now!

> trouble with multiple forms

I don't think there's anything wrong there - if the input does not have a `name` attribute it defaults to the node name, which has to be unique. But if you had set the `name` prop for the input it would not have changed.
I changed the label in the control for the node name to say `Node name` so that it's distinguishable from the control for the input's `name` prop, but maybe there's more ideas/alternatives on how to improve this?

> What if I add another component in the form. e.g. a datagrid? Should we disallow that?

Right now it's allowed, but those components don't do anything in particular inside the form. I'm thinking it's okay to give users that freedom anyway in these scenarios?

> What with nested forms? e.g. would you think of forms as a way to group form controls, and can a form then itself be a form control?

Nesting forms is invalid HTML, it even shows a console error, so I've disallowed placing forms inside forms now.

test/models/ToolpadEditor.ts
```diff
@@ -108,6 +108,13 @@ export class ToolpadEditor {
     await this.createComponentDialog.createButton.click();
   }
 
+  async waitForOverlay() {
+    await this.pageOverlay.waitFor({ state: 'visible' });
+    // Some tests seem to be flaky around this waitFor and perform better with a short timeout
```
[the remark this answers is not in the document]
apedroferreira: I was removing it just to try to fix the flakiness in the tests it was causing, and also because you've said before that we should avoid test logic related to the overlay (or anything else specific to implementation) as much as possible. So I think in that sense `editorModel.pageRoot.waitFor()` might be a bit more natural?

`waitforOverlay` was making the tests work back when I first wrote them, but it has been flaky and now from what I've tested so far replacing it with `editorModel.pageRoot.waitFor()` is working just fine.

I didn't look much into the root cause of the flakiness, I thought it's probably just some specific detail on how Playwright waits for things, and the fact that we're using an iframe might make that work a bit worse than in most cases.
But if you want to keep `waitForOverlay` in case it can help us fix some bigger possible underlying issue we can do that too, and I'll just remove the instances where I replaced it.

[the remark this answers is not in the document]
Janpot: Over time we can probably rename it to `waitForCanvasReady` or something? I found that adding a slight delay in there does a lot to the flakeyness, although I don't like it as a solution

Janpot:
I feel like this feature could use an integration test

apedroferreira:
> I feel like this feature could use an integration test

I was gonna do it separately but I'll add it before we merge this then, even if in a separate PR.

apedroferreira:
I'll merge this soon if there's no more feedback, will add the integration test separately right after.

apedroferreira (approved):

apedroferreira:
Text field stopped being interactive in forms in my last change.
I've reverted, will merge again when I have the test ready.

Issue #1500, opened 2023-01-01
The default isn't a blank page


oliviertassinari:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

Steps:

1. Open https://master--toolpad.mui.com/
2. Create a new app

### Current behavior 😯

<img width="457" alt="Screenshot 2023-01-01 at 18 54 16" src="https://user-images.githubusercontent.com/3165635/210180314-0d7bdafd-479a-43b4-b606-5a70c8e2c6bb.png">

### Expected behavior 🤔

The default is a blank page, like in Figma, Framer, Retool, etc.

### Your environment 🌎

v0.0.33

prakhargupta1:
Demo should also have the blank page as default.

PR #1601, opened 2023-01-27
Change default template

bytasv:
- Closes https://github.com/mui/mui-toolpad/issues/1500

https://user-images.githubusercontent.com/437214/215082226-72b01248-708d-49c8-a4f4-c27947e46717.mp4


bharatkashyap:
We would also need to modify this test to use the `hr` template and potentially rename it to "can use HR template":
https://github.com/mui/mui-toolpad/blob/default-tempalte/test/integration/appTemplates.spec.ts#L8

bharatkashyap (approved):

Issue #1536, opened 2023-01-09
Faster way to edit a deployed app or when previewing an app


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Original source where this first occurred: Point 5 [here](https://mui-org.slack.com/archives/C040M6LAZEG/p1670832003855479)

Problem: There should be a faster way to go to edit mode.

In the preview mode we show this message at the top. 
<img width="571" alt="Screenshot 2023-01-09 at 7 05 38 PM" src="https://user-images.githubusercontent.com/92228082/211320314-1b1374ce-41ba-489c-a92b-f4a455375931.png">
Maybe we can add a button at the top-right or a floating round button?

But no such message comes when viewing a deployed app. So a floating button?

Note: This can be impacted by the decision of, 1page === 1app. 

### Examples 🌈

Figma:
<img width="860" alt="Screenshot 2023-01-09 at 7 13 01 PM" src="https://user-images.githubusercontent.com/92228082/211321785-5c9795e6-ab1d-4fad-a8e1-47d73ff2c8e6.png">

Appsmith:
<img width="427" alt="Screenshot 2023-01-09 at 7 13 49 PM" src="https://user-images.githubusercontent.com/92228082/211321941-f386f179-f122-4f4c-b1ce-9e894b6f5bf0.png">

Superblocks:
<img width="852" alt="Screenshot 2023-01-09 at 7 15 56 PM" src="https://user-images.githubusercontent.com/92228082/211322414-114130d6-1a0a-4bfc-bcc5-69ec3c9bc3d4.png">

PR #1603, opened 2023-01-27
Add fast edit option from preview

bytasv:
- Closes https://github.com/mui/mui-toolpad/issues/1536

CleanShot 2023-01-27 at 15 44 04


packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -965,7 +966,17 @@ export default function ToolpadApp({
...
             <CssBaseline enableColorScheme />
             {version === 'preview' && !hidePreviewBanner ? (
-              <Alert severity="info">This is a preview version of the application.</Alert>
+              <Alert severity="info">
+                This is a preview version of the application.
+                <Button
+                  variant="contained"
+                  size="small"
+                  href={`/_toolpad/app/${appId}`}
+                  sx={{ m: '-14px 16px -8px', fontSize: '10px' }}
```
apedroferreira, line 975: Should we not make `px` explicit here and just use a number for the values? 
Seems like that's the pattern we usually follow.

Also does it really need all these negative/different margins? Just because this layout seems kind of simple at first glance.

apedroferreira (changes requested):
Left a comment but nothing major and might be a bit opinionated.

Also more opinionated design feedback: I feel like the shadow in the button and the blue color it's using don't match the Alert color too well? Could it be flatter and more consistent in color maybe? 
Also the font in the button seems a bit small.

Screenshot 2023-01-30 at 12 54 28

Another thing I might try, but not sure if it would look better, would be to place the button on the right side of the Alert bar. But it's just an idea!

prakhargupta1:
It came up in my discussion with @gerdadesign what if we continue to show the top header bar in the preview mode? We can just replace `Preview` with `Edit` and there won't be any need to show 'This is a preview version of the application'.

Would there be any concerns with this approach?

Janpot:
> Would there be any concerns with this approach?

The preview header only shows up on preview version of the application. For the edit button this could be fine. But we were talking about adding a header bar that could house the hamburger button that opens the navigation menu when it's hidden on smaller screens.

So the header we talked about during meeting would also need to show up on deployed applications.

Janpot:
Ok, I misunderstood. So you want to show this header in the deployed/preview apps as wel:

Screenshot 2023-01-30 at 19 08 07

Works for me. It will need to be hidden as well when `hideShell` is used, so the app can still be embedded

prakhargupta1:
> So you want to show this header in the deployed/preview apps as welL

Only for preview, not in the deployed app.

Janpot:
> Only for preview, not in the deployed app.

oh ok, then we can't use it to house the hamburger menu for the navigation drawer

packages/toolpad-app/src/runtime/ToolpadApp.tsx
Janpot: Perhaps use row flexbox on the parent with `align-items: center` and `gap`?

prakhargupta1:
LGTM!

apedroferreira (changes requested):
I really like this solution! 
Just noticed one more thing, in the preview the logo is vertically stretched for some reason:

Editor:
Screenshot 2023-01-31 at 16 21 49

Preview:
Screenshot 2023-01-31 at 16 21 56

apedroferreira (approved):
Looks great now. Thanks for the changes!

PR #1616, opened 2023-01-31
Add DataGrid docs

bytasv:
- Added missing docs for custom cell rendering in DataGrid

CleanShot 2023-01-31 at 14 32 03

### DIFF
--- a/packages/toolpad-app/src/theme.ts
+++ b/packages/toolpad-app/src/theme.ts
@@ -52,7 +52,7 @@ declare module '@mui/material/Chip' {
 
 const defaultTheme = createTheme();
 
-export const blue = {
+const blue = {
   50: '#F0F7FF',
   100: '#C2E0FF',
   200: '#99CCF3',
@@ -65,7 +65,7 @@ export const blue = {
   800: '#004C99',
   900: '#003A75',
 };
-export const blueDark = {
+const blueDark = {
   50: '#E2EDF8',
   100: '#CEE0F3',
   200: '#91B9E3',
@@ -78,7 +78,7 @@ export const blueDark = {
   800: '#001E3C',
   900: '#0A1929',
 };
-export const grey = {
+const grey = {
   50: '#F3F6F9',
   100: '#E7EBF0',
   200: '#E0E3E7',
--- a/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/NodeDropArea.tsx
+++ b/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/NodeDropArea.tsx
@@ -2,7 +2,6 @@ import * as React from 'react';
 import clsx from 'clsx';
 import { styled } from '@mui/material';
 import Typography from '@mui/material/Typography';
-import AddCircleOutlineIcon from '@mui/icons-material/AddCircleOutline';
 
 import { FlowDirection } from '../../../../types';
 import * as appDom from '../../../../appDom';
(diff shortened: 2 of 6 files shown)


docs/data/toolpad/building-ui/datagrid-component.md
```diff
@@ -0,0 +1,38 @@
...
+Firstly, you will have to create a custom component which should be able to process `DataGrid` data:
+
+- `value` prop will receive the value of the specific column
+- `row` prop will receive the data for the entire row
+- `field` prop will receive field configuration data
+
+<img src="/static/toolpad/docs/building-ui/cell-component.png" width="621px" alt="Custom component" />
+
+### Configuring columns
+
```
Janpot, line 22: Perhaps it could be interesting to have a section for each column type?

Screenshot 2023-02-02 at 11 40 57

docs/data/toolpad/building-ui/datagrid-component.md
```diff
@@ -0,0 +1,38 @@
+# DataGrid component
+
+<p class="description">Build powerful and feature rich data tables</p>
+
+<img src="/static/toolpad/docs/building-ui/datagrid.png" width="884px" alt="Custom component" />
+
+## Customizing cell rendering
+
+By default you can render json data using DataGrid component and it will automatically recognize most of the data types. However sometimes it might not be enough and for that reason we provide the option to render custom components inside a datagrid.
```
apedroferreira, line 9: The content seems fine to me! But for improving the grammar and punctuation we could run this by Sam maybe?

samuelsycamore:
```suggestion
By default, you can render JSON data using the Data Grid component and it will automatically recognize most of the data types. 
If this isn't sufficient, you can also render custom components inside the Data Grid.
```

docs/data/toolpad/building-ui/datagrid-component.md
```diff
@@ -0,0 +1,39 @@
...
+
+3. Choose the column
+
+<img src="/static/toolpad/docs/building-ui/choose-column.png" width="314px" alt="Choose column" />
+
+4. Change type to use codeComponent
+
+<img src="/static/toolpad/docs/building-ui/choose-column-type.png" width="318px" alt="Choose column type" />
+
+5. Choose custom component that you created
```
apedroferreira, line 37: "Choose **the** custom component that you created"

docs/data/toolpad/building-ui/datagrid-component.md
```diff
@@ -0,0 +1,39 @@
...
+- `value` prop will receive the value of the specific column
+- `row` prop will receive the data for the entire row
+- `field` prop will receive field configuration data
+
+<img src="/static/toolpad/docs/building-ui/cell-component.png" width="621px" alt="Custom component" />
+
+### Configuring columns
+
+1. Select `DataGrid` component
+2. Click on columns edit button
```
apedroferreira, line 25: `Click on the "columns" button` maybe?

Janpot:
Towards the future, should we aim to add some documentation for other components? If so, how do you imagine we could structure these?

Janpot (approved):

Janpot:
> Decide whether we want to/should document things that we think are more or less self explanatory (in other words should we assume that everything by default may be unclear?)

I believe eventually, _everything_ in Toolpad should be documented. The main trade-off we make would be the depth in which we explain things. We wouldn't want to be condescending neither by overly explaining things. I believe if the user finds documentation about some component, their expectation would be that there is documentation for every component.

> I'd like to be on a cautious side with "over documenting" things in case we decide to update UI or change functionality.

It's always a trade-off, I would want to be remain cautious though to make sure we prioritize Toolpad user UX over Toolpad contributor DX. It's a trade-off we'll have to adapt over time as the project evolves.

Janpot:
> should we do that right away

Not necessarily, the point I tried to make was that it's a trade-off, and the ratio in which we prioritize user UX should increase as the product matures.

PR #1620, opened 2023-02-01
Remove firefox specific branch in tests

Janpot:
Just to see which tests break


apedroferreira (approved):
So it works?
That's great! I think there must be something wrong with Playwright in my CLI because many tests fail in Firefox there, even excluding these ones...

Janpot:
Maybe it works differently dev vs production?

apedroferreira:
> Maybe it works differently dev vs production?

Yeah could be some difference between the CI environment and my local one.
Anyway that's just a local issue with my machine so I'll try to figure it out.

Issue #1612, opened 2023-01-30
Align the styling of the Text component variant with the Markdown component


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Make sure that the styling of `# title` when using markdown is the same as the styling of "title" when using text+variant h1

Do the same for all other corresponding tags.

PR #1627, opened 2023-02-02
Align headers and typography in the theme

Janpot:
Fixes https://github.com/mui/mui-toolpad/issues/1612

The approach here was to teach the markdown element about the typography styles of the theme. Then define the headers sizes in the theme. That way the theme is the single source of truth for bot the Markdown and the Text components.

before:

Screenshot 2023-02-02 at 17 40 52

after:

Screenshot 2023-02-02 at 17 39 48

This also removes the margins for first and last headers. Anyone with  better eye for design, feel free to suggest changes. @gerdadesign ?


bharatkashyap (approved):

gerdadesign:
Curious about the values used for H1, H2, etc. Seems different than the expectation to use the [Material typographic scale](https://m2.material.io/design/typography/the-type-system.html#type-scale).

Janpot:
> Seems different than the expectation

To me, the [Material typographic scale](https://m2.material.io/design/typography/the-type-system.html#type-scale) doesn't feel dense enough for the use-cases we intend Toolpad for. I tried to align a bit to what the MUI docs do. Happy to revert to the defaults. I'm not quite sure yet how we ideally should handle this.

prakhargupta1:
Looks good to me and I agree that Material typography scale is not ideal for Toolpad use cases. Retool and appsmith also seem to have a smaller scale.

Q) Now that H1 is bold, if I want to use that size but not the bold variant, is there anyway I can do it through sx? I searched for it but couldn't find.

Janpot:
> Q) Now that H1 is bold, if I want to use that size but not the bold variant, is there anyway I can do it through sx? I searched for it but couldn't find.

@prakhargupta1 it can be done on a case by case basis like so https://toolpad-pr-1627.onrender.com/_toolpad/app/cldvok57d0001ql9utit3cf7t/pages/ev03mab
But ideally we support this in theming and allow you to override this style once for the whole application.

> Looks good to me and I agree that Material typography scale is not ideal for Toolpad use cases.

@gerdadesign If you agree and feel like playing around with a new scale, denser, for Toolpad specifically, I'd be happy to include your suggestions. If you think we should maintain the Material design scale I'm also happy to revert. We can also bring it up for discussion next meeting.

Janpot:
Merging this, we can always tweak later

PR #1637, opened 2023-02-07
Improve behavior of image columns

Janpot:
* Don't show anything if there is no value
* Show a thumbnail with a popover, avoid auto row height

https://user-images.githubusercontent.com/2109932/217227612-a283bf18-dcb3-43f9-a84d-2a2d18e54b65.mov

### DIFF
--- a/packages/toolpad-components/src/DataGrid.tsx
+++ b/packages/toolpad-components/src/DataGrid.tsx
@@ -29,6 +29,7 @@ import {
   styled,
   Typography,
   Tooltip,
+  Popover,
 } from '@mui/material';
 import { getObjectKey } from '@mui/toolpad-core/objectKey';
 import { hasImageExtension } from '@mui/toolpad-core/path';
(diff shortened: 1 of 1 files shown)


apedroferreira:
Is there a way to have more control over the image size, or should there?
Anyway this solution works for me in that sense as not sure what the best approach would be, could depend on different use cases...

Found a bug though, on the 2nd item the popover was looking really big to me and had a scroll bar:

https://user-images.githubusercontent.com/10789765/217327117-6e7f5299-5408-4d1a-9ec6-ad622e7704c4.mov

apedroferreira (changes requested):
Just the bug I found, code seems good!

Janpot:
> Is there a way to have more control over the image size, or should there?

You can always create a custom component. I believe this is a more sensible default as it just behaves the way text behaves, i.e. just take the available space. The previous solution uses `getRowHeight: () => 'auto'`, which had the side effect that it resized the rows on other columns as well, which is not the intended behavior.

> Found a bug though, on the 2nd item the popover was looking really big to me and had a scroll bar:

I limited the size of the popover image. I also made sure cell alignment still works

apedroferreira (approved):
> You can always create a custom component. I believe this is a more sensible default as it just behaves the way text behaves, i.e. just take the available space.

Sounds good, that's better now that we have custom component columns.
This looks good now!

PR #1639, opened 2023-02-07
Fix resizing after drag & drop UI changes

apedroferreira:
Found a few issues with resizing after pushing the drag & drop UI changes, so here are some fixes/improvements:

- Only allow resizing selected elements
- Dashed hover outlines were confusing during resizing and dragging, so this hides them while resizing and dragging
- Minimum grid row gap of 1 (feedback from Prakhar)

Preview: https://toolpad-pr-1639.onrender.com


apedroferreira:
Just noticed there are console errors now too... will include the fix here.

Janpot:
> so it wasn't possible to resize from inside an unselected element

For what it's worth, I don't think we should support resizing from non-selected components.

apedroferreira:
> For what it's worth, I don't think we should support resizing from non-selected components.

Yep I was just seeing if someone agreed on that, will make it only possible to resize the selected component.

Janpot (approved):

PR #1672, opened 2023-02-14
Add in page settings option to sidebar to allow for easier access to page module settings

scouttyg:
One of the harder things to understand with the toolpad so far is how to access your page settings. Right now:

- Clicking on the page on the left side won't show the settings (if you are currently on that same page and have another node selected in the editor)
- Accessing the menu gives me the option to duplicate or delete, but I needed to realize to click "off" any components and it would finally show the page settings.

- Even clicking between pages doesn't seem to show Page Settings again (notice the right hand side still shows "Data Grid", even when I'm on a separate page):

between-pages

The only way to really access page settings is to know to click off any component on the page, which when a page is complex, might be hard to understand (you'll need to find some whitespace to click on, and that whitespace might be minimal).

This PR adds a small item to the `NavMenu` dropdown to support "Settings", which can help alleviate this.

## Visual demonstration
page-settings-menu

Screenshot 2023-02-14 at 8 49 07 AM

### DIFF
--- a/packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/index.tsx
+++ b/packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/index.tsx
@@ -256,7 +256,10 @@ export default function HierarchyExplorer({ appId, className }: HierarchyExplore
       const node = appDom.getNode(dom, nodeId);
 
       if (appDom.isPage(node)) {
-        domApi.update(dom => dom, { view: { kind: 'page', nodeId: node.id }, selectedNodeId: null });
+        domApi.update((dom) => dom, {
+          view: { kind: 'page', nodeId: node.id },
+          selectedNodeId: null,
+        });
       }
     },
     [dom, domApi],
--- a/packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/index.tsx
+++ b/packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/index.tsx
@@ -170,6 +170,7 @@ export default function HierarchyExplorer({ appId, className }: HierarchyExplore
     }
 
     if (appDom.isPage(node)) {
+      domApi.deselectNode();
       domApi.setView({ kind: 'page', nodeId: node.id });
     }
 
--- a/packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/index.tsx
+++ b/packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/index.tsx
@@ -253,12 +253,10 @@ export default function HierarchyExplorer({ appId, className }: HierarchyExplore
 
   const handlePageSettingsNode = React.useCallback(
     (nodeId: NodeId) => {
-      if (nodeId) {
-        const node = appDom.getNode(dom, nodeId);
-        if (appDom.isPage(node)) {
-          domApi.setView({ kind: 'page', nodeId: node.id });
-          domApi.deselectNode();
-        }
+      const node = appDom.getNode(dom, nodeId);
+
+      if (appDom.isPage(node)) {
+        domApi.update(dom => dom, { view: { kind: 'page', nodeId: node.id }, selectedNodeId: null });
       }
     },
     [dom, domApi],
--- a/packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/index.tsx
+++ b/packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/index.tsx
@@ -44,12 +44,14 @@ const StyledTreeItem = styled(TreeItem)({
 type StyledTreeItemProps = TreeItemProps & {
   onDeleteNode?: (nodeId: NodeId) => void;
   onDuplicateNode?: (nodeId: NodeId) => void;
+  onSettingsNode?: (nodeId: NodeId) => void;
   onCreate?: React.MouseEventHandler;
   labelIcon?: React.ReactNode;
   labelText: string;
   createLabelText?: string;
   deleteLabelText?: string;
   duplicateLabelText?: string;
+  settingsLabelText?: string;
   toolpadNodeId?: NodeId;
 };
 
@@ -60,9 +62,11 @@ function HierarchyTreeItem(props: StyledTreeItemProps) {
     onCreate,
     onDeleteNode,
     onDuplicateNode,
+    onSettingsNode,
     createLabelText,
     deleteLabelText = 'Delete',
     duplicateLabelText = 'Duplicate',
+    settingsLabelText = 'Settings',
     toolpadNodeId,
     ...other
   } = props;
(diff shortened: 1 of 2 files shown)


packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/index.tsx
```diff
@@ -164,6 +170,7 @@ export default function HierarchyExplorer({ appId, className }: HierarchyExplore
     }
 
     if (appDom.isPage(node)) {
+      domApi.deselectNode();
```
scouttyg, line 173: This solves the issue where a node would "stay" even when navigating to a page where that node didn't exist:
Screenshot 2023-02-14 at 9 02 18 AM

packages/toolpad-app/src/toolpad/AppEditor/NodeMenu.tsx
```diff
@@ -75,6 +88,14 @@ export default function NodeMenu({
...
         }}
       >
+        {onSettingsNode && (
+          <MenuItem onClick={handleSettingsClick}>
+            <ListItemIcon>
+              <SettingsIcon />
+            </ListItemIcon>
+            <ListItemText>{settingsLabelText}</ListItemText>
+          </MenuItem>
+        )}
```
scouttyg, line 98: The idea was that a `NodeMenu` could support a generic settings button in the future if `onSettingsNode` was supplied.

packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/index.tsx
Janpot: Ah yes, interesting 👍 

@apedroferreira Would it make for the future to incorporate the selection state into the `DomView`? We could even think about making selected node deep linkable through a query parameter

apedroferreira: Yeah, actually it only makes sense for the selected node state to exist inside the main view... So I should change it to be there. I'll create an issue for that and address it.

As for setting things with query parameters, that would be a separate feature but should be totally doable.

apedroferreira:
Hi @scouttyg , thanks a lot for identifying this problem and making a PR!

Part of your problem might actually be an unintended bug that I might have introduced with recent changes, as when you change the current page every node should probably be automatically unselected, and so the page settings should show... I can address that, thanks for reporting it!
Also I might make it so that it's possible to change the current view and unselect in a single action, if it's not possible yet.

As for the new option you added: seems like a good solution, but I haven't given a lot of thought to what other ways we could provide for users to be able to easily see the page settings. Would like to see what the rest of the team thinks and try to find what the best way to improve on that issue is.

packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/index.tsx
```diff
@@ -244,6 +251,19 @@ export default function HierarchyExplorer({ appId, className }: HierarchyExplore
     [dom, domApi],
   );
 
+  const handlePageSettingsNode = React.useCallback(
+    (nodeId: NodeId) => {
+      if (nodeId) {
```
apedroferreira, line 256: We might not need to check if `nodeId` is truthy here? At least I don't see it in the other functions - I think it should always be defined?

packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/index.tsx
```diff
@@ -244,6 +251,19 @@ export default function HierarchyExplorer({ appId, className }: HierarchyExplore
...
+        const node = appDom.getNode(dom, nodeId);
+        if (appDom.isPage(node)) {
+          domApi.setView({ kind: 'page', nodeId: node.id });
```
apedroferreira, line 259: This can be done in a single action with domApi.update(dom => dom, { view: { kind: 'page', nodeId: node.id }, selectedNodeId: null }). I have a PR open to make it possible to update just the view and the selected node without "updating" the app DOM, that should make this less weird.

packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/index.tsx
apedroferreira: I have a more global fix for this specific issue in https://github.com/mui/mui-toolpad/pull/1674.

packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/index.tsx
Janpot: And in case it can be `undefined` then let's express it in the types as well.

Janpot:
> As for the new option you added: seems like a good solution, but I haven't given a lot of thought to what ways we could provide for users to be able to easily see the page settings. Would like to see what the rest of the team thinks and try to find what the best way to improve on that issue is.
>
> One alternative idea: always showing a "Page" tab? Maybe we could create that tab and have 3 tabs on the right side panel, but the "Component" tab would be disabled if there's no component selected.

We're indeed not 100% happy with the discoverability of the page options. But until we get there design-wise I think this solution could be a good stopgap.

packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/index.tsx
scouttyg: @apedroferreira Updated in https://github.com/mui/mui-toolpad/pull/1672/commits/a87f8b75e4caff4fe38acbedd884eed0ba8d304f!

packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/index.tsx
scouttyg: @apedroferreira I think you are right that we don't need to check if `nodeId` is truthy here -- I've removed this in https://github.com/mui/mui-toolpad/pull/1672/commits/a87f8b75e4caff4fe38acbedd884eed0ba8d304f

apedroferreira (changes requested):
@scouttyg Looks great now! Thanks for responding to the feedback!
I think this feature works well for now, and we can think about other possible improvements in the future.

Just one more thing: there seem to be a couple of eslint errors/warnings and formatting differences that are causing the checks to fail for this PR. 
Can you see the eslint errors in your editor, and if so can you please fix them? Running `yarn eslint` should also show you the errors/warnings.
To fix the formatting you can run `yarn prettier` if you're not auto-formatting when you save a file for example.

scouttyg:
@apedroferreira Prettified things a bit more, rebased from master, and I think all tests should be passing now! Let me know what you think.

apedroferreira (approved):
Awesome! Again thanks a lot for your contribution @scouttyg , you're welcome to keep contributing as we really appreciate it!

PR #1674, opened 2023-02-14
Deselect nodes on page change

apedroferreira:
Fixes problem reported in https://github.com/mui/mui-toolpad/pull/1672 where selected node stays selected even when user changes page.
Will move `selectedNodeId` inside the state of page views only, but separately.

https://user-images.githubusercontent.com/10789765/218828026-35652d24-ff63-483b-ba7e-ac25e506f0fd.mov


apedroferreira:
Fixed instead in https://github.com/mui/mui-toolpad/pull/1679

PR #1679, opened 2023-02-15
Move selected node state to page view only

apedroferreira:
Move `selectedNodeId` and `currentTab` state to views of `kind: 'page'`, as that's the only kind of view where it makes sense to keep that state.
Also deselects node when switching to a different page or view.
As suggested in https://github.com/mui/mui-toolpad/pull/1672#discussion_r1105875122


packages/toolpad-app/src/server/appTemplateDoms/images.json
```diff
@@ -493,7 +493,7 @@
         },
         "disabled": {
           "type": "jsExpression",
-          "value": "selectSubBreed.options?.length <= 0\n"
+          "value": "selectSubBreed.options.length <= 0\n"
```
apedroferreira, line 496: Unrelated but this `?` was unnecessary here.

apedroferreira:
I moved `currentTab` inside page views only too.

packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/CreateCodeComponentNodeDialog.tsx
```diff
@@ -101,10 +101,8 @@ export default function CreateCodeComponentDialog({
           const appNode = appDom.getApp(dom);
 
           appStateApi.update((draft) => appDom.addNode(draft, newNode, appNode, 'codeComponents'), {
```
Janpot, line 103: Not introduced in this PR, but is there any technical reason why `update` updates both the dom and the view? It feels like two different things to me

Janpot (approved):

packages/toolpad-app/src/toolpad/AppEditor/HierarchyExplorer/CreateCodeComponentNodeDialog.tsx
apedroferreira: In some cases for the undo/redo history to be good the best is that we update the DOM and the view in the same action, especially when adding or removing nodes. Maybe there's a better alternative though?

PR #1690, opened 2023-02-21
Adjust margins for Markdown component

Janpot:
* 16px between paragraphs
* remove margin under trailing paragraph

### DIFF
--- a/packages/toolpad-components/src/Text.tsx
+++ b/packages/toolpad-components/src/Text.tsx
@@ -59,7 +59,6 @@ const MarkdownContainer = styled('div')(({ theme }) => ({
     marginBottom: 4,
   },
   '& p': {
-    marginTop: 16,
     marginBottom: 16,
   },
   '& *:first-child': {
--- a/packages/toolpad-components/src/Text.tsx
+++ b/packages/toolpad-components/src/Text.tsx
@@ -76,7 +76,8 @@ const MarkdownContainer = styled('div')(({ theme }) => ({
     marginBottom: 4,
   },
   '& p': {
-    marginBottom: 16,
+    marginTop: 12,
+    marginBottom: 12,
   },
   '& *:first-child': {
     marginTop: 0,
--- a/packages/toolpad-components/src/Text.tsx
+++ b/packages/toolpad-components/src/Text.tsx
@@ -21,17 +21,6 @@ interface TextProps extends Omit<BaseProps, 'children'> {
   loading?: boolean;
 }
 
-const gutters = (marginTop: number, marginBottom: number) => ({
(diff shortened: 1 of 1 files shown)


bharatkashyap (approved):

packages/toolpad-components/src/Text.tsx
```diff
@@ -21,31 +21,53 @@ interface TextProps extends Omit<BaseProps, 'children'> {
...
+    marginTop: 4,
+    marginBottom: 4,
+  },
+  '& h6': {
+    ...theme.typography.h6,
+    marginTop: 4,
+    marginBottom: 4,
+  },
+  '& p': {
+    marginTop: 16,
```
oliviertassinari, line 62: I think that it can work without margin-top. We made it work on https://mui.com/material-ui/getting-started/overview/. So maybe:

```suggestion
```

so we can more easily transfer knowledge and fixes between these two spaces.

packages/toolpad-components/src/Text.tsx
```diff
@@ -21,31 +21,53 @@ interface TextProps extends Omit<BaseProps, 'children'> {
...
+    marginBottom: 16,
```
oliviertassinari, line 63: 16 works well for documentation. To consider if we want the markdown used in a more information-dense environment.

packages/toolpad-components/src/Text.tsx
```diff
@@ -21,31 +21,53 @@ interface TextProps extends Omit<BaseProps, 'children'> {
...
-  '& p': { margin: 0, marginBottom: 6 },
+  '& h1': {
+    ...theme.typography.h1,
+    marginTop: 16,
+    marginBottom: 16,
+  },
+  '& h2': {
+    ...theme.typography.h2,
+    marginTop: 12,
+    marginBottom: 12,
```
oliviertassinari, line 39: I think that the values should be uneven, we almost always need more margin-top than margin-bottom. For example:

```suggestion
    marginTop: 16,
    marginBottom: 12,
```

(I have no idea if this is the exact value that we want, this would be more for **design** to rule this).

oliviertassinari (commented):
👍

packages/toolpad-components/src/Text.tsx
Janpot: > this would be for design to define them

Agreed, I treated these mostly as placeholders until design gets a chance to take decisions on how it should look.

packages/toolpad-components/src/Text.tsx
Janpot: Agreed, will reduce it a bit, design can further refine it later

Issue #1586, opened 2023-01-19
For PostgreSQL, preview is missing when mutating a table 


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Problem: There was no visual response when I ran an Insert query on a PGDB.

https://user-images.githubusercontent.com/92228082/213504074-0c39dd9a-ccac-40f1-83d8-f2f3de18c358.mov

Also, docs are not updated about writing to a PosgreSQL DB.

### Examples 🌈

Appsmith:
<img width="915" alt="Screenshot 2023-01-19 at 10 12 02 PM" src="https://user-images.githubusercontent.com/92228082/213504188-a0111bbe-8194-4207-a7d2-5eec97e7250a.png">

PR #1691, opened 2023-02-21
Add feedback for mutating queries

Janpot:
Closes https://github.com/mui/mui-toolpad/issues/1586

Screenshot 2023-02-21 at 12 12 44


bharatkashyap (approved):

bharatkashyap:
What would be the workflow for creating a Postgres/MySQL connection via direction 13?

PR #1618, opened 2023-01-31
Show confirmation dialog when there are unsaved changes

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/1571

The code component part of this PR can be seen/reviewed isolated here: https://github.com/mui/mui-toolpad/pull/1628
We can review and merge that part first if it makes things easier.

- [x] Confirm when leaving code component editor with unsaved changes (improvements)
- [x] Confirm when leaving query editor with unsaved changes, unless "Cancel" is pressed
- [x] Confirm when leaving page module dialog unsaved changes, unless "Cancel" is pressed
- [x] Confirm when leaving page parameters dialog with unsaved changes, unless "Close" is pressed
- [x] Confirm when leaving bindings dialog with unsaved changes, unless "Cancel" is pressed

All except the code component flow use the same `useEditorUnsavedChangesConfirm` hook.

Preview: https://toolpad-pr-1618.onrender.com/

PR #1695, opened 2023-02-21
Fix binding editor confirm on unsaved changes

apedroferreira:
Fix issue reported by @oliviertassinari in https://github.com/mui/mui-toolpad/pull/1618#issuecomment-1438885548:

Do not show confirmation popup if there have been no changes in the binding editor since it was opened, or if there were changes but the value on leaving is the same as the initial value

https://user-images.githubusercontent.com/10789765/220441075-c3887a36-6870-4f58-9530-26ddaf0c0bde.mov


oliviertassinari (commented):
Great, thanks, I fixed a bug with the text field at the same time: https://github.com/mui/material-ui/pull/36288 😁

bharatkashyap (approved):

Janpot (approved):

PR #1700, opened 2023-02-22
Add create-toolpad-app CLI

### LINKED ISSUE #1653: Getting started with Toolpad to be improved
### Duplicates

### Latest version

### Summary 💡

- [x] add `--help` command to CLI (move to `yargs`?) https://github.com/mui/mui-toolpad/issues/1744
- [x]  `toolpad init `command? For quick setup (or something like create-toolpad-app)
- [x]  automatically pick another port if 3000 is in use, only in dev mode, when no port has been specified
- [x]  automatically open the browser on start i.e, user will not have to click, only in dev mode, with an opt-out cli argument
- [x]  avoid having to install `@mui/toolpad-core`. https://github.com/mui/mui-toolpad/pull/1793. Can we have users import e.g. `createComponent` from `@mui/toolpad`?
    - something like: `import { createComponent } from '@mui/toolpad/client'`
    - `import { createQuery } from '@mui/toolpad/server'`

### Examples 🌈

_No response_

### Motivation 🔦

_No response_

bharatkashyap:
- Part of #1653 
- Visual demonstration 

https://user-images.githubusercontent.com/19550456/222448714-61d3cf19-36d1-4923-b169-b42c4e1da791.mov


packages/create-toolpad-app/index.ts
```diff
@@ -0,0 +1,113 @@
...
+const packageManager = getPackageManager();
+
+// Create a new directory and initialize a new project
+const scaffoldProject = async () => {
+  const name = await inquirer.prompt([
+    {
+      type: 'input',
+      name: 'projectName',
+      message: 'Enter the name of your project:',
+      default: 'my-app',
```
apedroferreira, line 19: How about `my-toolpad-app` as a default?

packages/create-toolpad-app/index.ts
```diff
@@ -0,0 +1,113 @@
...
+    },
+  ]);
+
+  // eslint-disable-next-line no-console
+  console.log(`Creating a new MUI Toolpad project in ${chalk.blue(name.projectName)}`);
+  try {
+    fs.mkdirSync(name.projectName);
```
apedroferreira, line 26: We can use async methods by importing `import * as fs from 'fs/promises';`
Probably wouldn't make a huge difference here but I guess it's more consistent with how we're using it in other files.

packages/create-toolpad-app/index.ts
```diff
@@ -0,0 +1,113 @@
...
+    process.chdir(name.projectName);
+
+    const packageJson = {
```
apedroferreira, line 29: Could we generate this by copying a package.json file already configured by us instead? I feel like that could be better to work with - it's what I had tried in https://github.com/mui/mui-toolpad/pull/1705. 
On the other hand this approach could be more customizable with Javascript, so anyway just bringing up the possibility - it's probably still too early to tell what option would be best at this point.

packages/create-toolpad-app/index.ts
```diff
@@ -0,0 +1,113 @@
...
+      );
+    }
+    return null;
+  }
+};
+
+// Install the dependencies
+const installDeps = async () => {
+  const installVerb = packageManager === 'yarn' ? 'add' : 'install';
+  const command = `${packageManager} ${installVerb} @mui/toolpad @mui/toolpad-core`;
```
apedroferreira, line 55: Won't this install the latest versions? We might want to install specific versions to keep things a bit more under control.
Also if we end up wanting to install more dependencies the pre-generated package.json approach I mentioned might work best.

packages/create-toolpad-app/index.ts
```diff
@@ -0,0 +1,113 @@
...
+
+// Run the CLI interaction with Inquirer.js
+const run = async () => {
+  let projectName;
+  do {
+    // eslint-disable-next-line no-await-in-loop
+    projectName = await scaffoldProject();
+  } while (!projectName);
+
+  const installDependenciesConsent = await inquirer.prompt([
```
apedroferreira, line 70: I feel like since we're creating a new app from scratch, and not changing an existing one, that we wouldn't need to ask for this consent? But maybe I'm missing something.

bharatkashyap: I thought it was a good to have in a CLI before installing packages in the user's system, to sort of give a final heads-up before going for this bandwidth heavy operation

apedroferreira: It's fine if you want to keep it, this is just my opinion anyway, I'm not super sure about it :)

Janpot: @bharatkashyap can you check with a few similar cli tools to see what the usual workflow is?

bharatkashyap: Create Next App doesn't have something like this, it just lists the dependencies and proceeds to install them. 

Screenshot 2023-03-07 at 3 24 42 PM

It makes sense since the operation may be cancelled if the user doesn't intend to instal deps

apedroferreira: Yeah, if no other CLI tools are doing I think we could definitely skip this prompt.

Janpot: Is there value in prompting if the install would happen in an existing non-empty folder?

Janpot:
We should also make sure to resolve that scary insecure `cuid` package message

packages/create-toolpad-app/index.ts
bharatkashyap: > Is there value in prompting if the install would happen in an existing non-empty folder?

That isn't possible in the current shape, since we try and create a new directory and fail if the directory name provided exists

packages/create-toolpad-app/index.ts
```diff
@@ -0,0 +1,112 @@
...
+    await fs.writeFile('package.json', JSON.stringify(packageJson, null, 2));
+    return name.projectName;
+  } catch (error) {
+    // Directory exists, verify if it is empty to continue
+    if (!isFolderEmpty(path.join(process.cwd(), name.projectName), name.projectName)) {
+      return null;
+    }
+    try {
+      await fs.access(name.projectName);
+      const installDependenciesConsent = await inquirer.prompt([
```
apedroferreira, line 63: So there is a use case for installing in existing projects that already have a `package.json` file? If so I think this is fine.

apedroferreira (approved):
All good, unless I misunderstood in my comment!

packages/create-toolpad-app/index.ts
```diff
@@ -0,0 +1,112 @@
...
+      message: 'Enter the name of your project:',
+      default: 'my-toolpad-app',
+    },
+  ]);
+
+  // eslint-disable-next-line no-console
+  console.log(`Creating a new MUI Toolpad project in ${chalk.blue(name.projectName)}`);
+  try {
+    await fs.mkdir(name.projectName);
+    process.chdir(name.projectName);
```
Janpot, line 41:
```suggestion
```
Can this be avoided? e.g. by passing `cwd` to `execa` instead. I feel like this could lead to weird issues where half the program is working in a certain working dir, and the other half in another. Moving code around could become awkward

packages/create-toolpad-app/index.ts
```diff
@@ -0,0 +1,112 @@
...
+        build: 'toolpad build',
+        start: 'toolpad start',
+      },
+    };
+
+    await fs.writeFile('package.json', JSON.stringify(packageJson, null, 2));
+    return name.projectName;
+  } catch (error) {
+    // Directory exists, verify if it is empty to continue
+    if (!isFolderEmpty(path.join(process.cwd(), name.projectName), name.projectName)) {
```
Janpot, line 58: Ideally, you read `process.cwd()` once and pass it around. Especially when you'll be doing `process.chdir` in some places.
The Purer a functions is, the easier to debug and test.

packages/create-toolpad-app/utils/isFolderEmpty.ts
```diff
@@ -0,0 +1,66 @@
...
+    .filter((file) => !/\.iml$/.test(file));
+
+  if (conflicts.length > 0) {
+    // eslint-disable-next-line no-console
+    console.log(`The directory ${chalk.green(name)} contains files that could conflict:`);
+    // eslint-disable-next-line no-console
+    console.log();
+    for (const file of conflicts) {
+      try {
+        const stats = fs.lstatSync(path.join(root, file));
```
Janpot, line 43: Avoid sync methods

packages/create-toolpad-app/utils/isFolderEmpty.ts
```diff
@@ -0,0 +1,66 @@
...
+    'mkdocs.yml',
+    'npm-debug.log',
+    'yarn-debug.log',
+    'yarn-error.log',
+    'yarnrc.yml',
+    '.yarn',
+  ];
+
+  const conflicts = fs
+    .readdirSync(root)
```
Janpot, line 31: Avoid sync methods

packages/create-toolpad-app/utils/isFolderEmpty.ts
```diff
@@ -0,0 +1,66 @@
+// From https://github.com/vercel/next.js/blob/canary/packages/create-next-app/helpers/is-folder-empty.ts
+import chalk from 'chalk';
+import fs from 'fs';
+import path from 'path';
+
+export function isFolderEmpty(root: string, name: string): boolean {
+  const validFiles = [
```
Janpot, line 7: Are we really going to curate a list with random files? Why not just be strict? empty == no files?

packages/create-toolpad-app/utils/isFolderEmpty.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  const conflicts = fs
+    .readdirSync(root)
+    .filter((file) => !validFiles.includes(file))
+    // Support IntelliJ IDEA-based editors
+    .filter((file) => !/\.iml$/.test(file));
+
+  if (conflicts.length > 0) {
+    // eslint-disable-next-line no-console
+    console.log(`The directory ${chalk.green(name)} contains files that could conflict:`);
```
Janpot, line 38: This seems to present itself as a reusable utility function `isFolderEmpty`, but then proceeds to do very specific logic for certain files, adds logging, etc... This function is not a reusable utility. I'd remove this file and colocate the code where you are using it.

packages/create-toolpad-app/index.ts
Janpot, line 1: For package structure I'd propose to use a `src` folder, just like the other projects. Also not sure whether there is real benefit to add a utils folder in a project that has 3 source files

packages/create-toolpad-app/utils/isFolderEmpty.ts
bharatkashyap: Fair enough, I imagined that we could rely on the list curated by https://github.com/vercel/next.js/blob/canary/packages/create-next-app/helpers/is-folder-empty.ts

packages/create-toolpad-app/package.json
```diff
@@ -0,0 +1,39 @@
...
+  },
+  "author": "Toolpad Team",
+  "license": "MIT",
+  "bin": {
+    "create-toolpad-app-test": "./dist/index.js"
+  },
+  "files": [
+    "dist"
+  ],
+  "type": "module",
```
Janpot, line 23:
```suggestion
```
?

packages/create-toolpad-app/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "scripts": {
+    "dev13": "tsc -w -p ./tsconfig.json --outDir ./dist/",
```
Janpot, line 25: Can `outDir` go in `tsconfig.json`?

packages/create-toolpad-app/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "prerelease": "rimraf ./dist/",
```
Janpot, line 26:
```json
    "prebuild": "rimraf ./dist/",
    "build": "tsc -p ./tsconfig.json -outDir ./dist/",
```

Let's use the same script names as in `@mui/toolpad-core` and `@mui/toolpad-components`?

packages/create-toolpad-app/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "release": "tsc -p ./tsconfig.json -outDir ./dist/",
+    "build": "yarn release"
+  },
+  "engines": {
+    "node": ">=14.6.0"
+  },
+  "dependencies": {},
+  "devDependencies": {
+    "@types/inquirer": "^9.0.3",
+    "chalk": "^5.2.0",
```
Janpot, line 36: I think these will need to go in `dependencies` when we don't use `ncc`

packages/create-toolpad-app/src/index.ts
```diff
@@ -0,0 +1,150 @@
...
+  console.log(`${chalk.green('success')} - Dependencies installed successfully!`);
+};
+
+// Create a new directory and initialize a new project
+const scaffoldProject = async (projectName: string, cwd: string): Promise<string | null> => {
+  // eslint-disable-next-line no-console
+  console.log(`Creating a new MUI Toolpad project in ${chalk.blue(projectName)}`);
+  try {
+    await fs.mkdir(projectName);
+    // process.chdir(projectName);
```
Janpot, line 70:
```suggestion
```

packages/create-toolpad-app/src/index.ts
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// From https://github.com/vercel/next.js/blob/canary/packages/create-next-app/helpers/is-folder-empty.ts
+
+async function isFolderEmpty(root: string, name: string): Promise<boolean> {
+  const files = await fs.readdir(root);
+
+  if (files.length > 0) {
+    // eslint-disable-next-line no-console
+    console.log(`The directory ${chalk.green(name)} is non-empty.`);
```
Janpot, line 35: Nit: does it make sense to move this logging to the consumer of `isFolderEmpty`? The function name doesn't suggest it will have side-effects.

packages/create-toolpad-app/utils/isFolderEmpty.ts
bharatkashyap: Should we leave it in, then?

Janpot: As you wish

packages/create-toolpad-app/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "description": "Create Toolpad apps with one command",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/mui/mui-toolpad",
+    "directory": "packages/create-toolpad-app"
+  },
+  "author": "Toolpad Team",
+  "license": "MIT",
+  "bin": {
+    "create-toolpad-app-test": "./dist/index.js"
```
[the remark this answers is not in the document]
bharatkashyap: > Also, does this file have execute permissions? 

What do you mean by this exactly/How do I test this?

packages/create-toolpad-app/package.json
bharatkashyap: Without this I get 
```
SyntaxError: Cannot use import statement outside a module
``` 
errors when running the compiled `.js` file

bharatkashyap: Fixed this by dynamically importing all the ES modules (`chalk`, `inquirer` and `execa`)

packages/create-toolpad-app/package.json
Janpot: does `types/inquirer` have to go in `dependencies` as well?

packages/create-toolpad-app/package.json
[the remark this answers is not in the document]
Janpot: which command do you use to test the cli at the moment?

packages/create-toolpad-app/package.json
bharatkashyap: Nope, `devDependencies` should be fine

packages/create-toolpad-app/package.json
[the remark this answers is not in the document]
bharatkashyap: In development, I run the built file from the dist folder in a separate folder through `node`; to test the production version, I publish the package and run `create toolpad-app` to test the flow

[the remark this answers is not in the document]
Janpot: Oh wow, it looks like npm automatically sets the correct file permissions when you use the `bin` field. I didn't know that. Maybe we should do the same for the `@mui/toolpad` cli later on.

Janpot (approved):

Janpot:
Great work!

Issue #1666, opened 2023-02-14
Add a .gitignore file to the .toolpad-generated folder


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Small task, summary not required.

apedroferreira:
To clarify: should this `.gitignore` file be in the root of the application and include the `.toolpad-generated` folder so that that folder isn't checked out?
I took a look and this and how the direction-13 codebase works, and it feels like creating this type of file should be done along with the `init` command which we don't have yet.

Janpot:
We can start by doing it in the init command. But eventually we will do this check as well every time we run `dev` on an existing project and make sure it's there

apedroferreira:
Ok, if we should regenerate the files every time the user starts the project then I can add it already!

PR #1705, opened 2023-02-22
Generate files (.gitignore, .env...) on dev command

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/1666

Generate `.gitignore` and `.env` files on running ```toolpad dev --dev```.
Not sure if this was the correct/expected way to implement this so let me know.


packages/toolpad-app/src/toolpadDataSources/local/server.ts
```diff
@@ -496,7 +497,7 @@ export default dataSource;
 
 async function startDev() {
   const builder = await createBuilder();
-  await builder.watch();
+  builder.watch();
```
apedroferreira, line 500: VSCode was saying this `await` wasn't necessary

apedroferreira:
> #1666 is about generating a `.gitignore` inside of the `.toolpad-generated` folder. I wouldn't add an env file, not until Toolpad wants to add secrets from the UI.

So I misunderstood, is this `.gitignore` in the root still helpful though? Or should it only be generated on the `init` CLI command?
I'll add a `.gitignore` file inside `.toolpad-generated` - what is the purpose of it though, which files should it ignore?

Janpot:
> So I misunderstood, is this .gitignore in the root still helpful though?

I would generate the root .gitignore during the `create-toolpad-app` (`yarn create toolpad-app`) flow. But after that I would leave it up to the user only to author this file. 

The user doesn't have business altering anything in the `.toolpad-generated` folder. The gitignore could act as sort of a safeguard for accidentally checking this folder in source control. Maybe it's a bad idea, maybe there are use cases where a user may want to check this folder in the repo? e.g. to save them a build step during deploy?

> what is the purpose of it though, which files should it ignore?

To prevent users from accidentally checking in the toolpad build output in their repository, so it would basically be
```
*
!.gitignore
```
But then again, thinking more about this, 🤔 maybe we should just add a top-level gitignore that contains `.toolpad-generated` as an entry and leave it up to the user on how to deal with this further in their dev lifecycle. Perhaps that makes more sense? If you agree, then maybe we can close the issue, we can also  bring it up in the next grooming session if you'd like to discuss further.

apedroferreira:
> I would generate the root .gitignore during the create-toolpad-app (yarn create toolpad-app) flow. But after that I would leave it up to the user only to author this file.

Agreed! Sounds like a good approach and gives some flexibility to the user.

> But then again, thinking more about this, 🤔 maybe we should just add a top-level gitignore that contains .toolpad-generated as an entry and leave it up to the user on how to deal with this further in their dev lifecycle. Perhaps that makes more sense? If you agree, then maybe we can close the issue, we can also bring it up in the next grooming session if you'd like to discuss further.

I think this sounds good! So basically the same thing I did here, but in the `create-toolpad-app` command only?
Then we can close this PR but we can also adapt it (later?) to generate the same files in that command, and it could be used to generate any pre-configured files in root?
We can finish this discussion in the next grooming session, it should be quick to solve it.

Janpot:
> So basically the same thing I did here, but in the create-toolpad-app command only?

Yes, except I wouldn't add a `.env` file from the start. @bharatkashyap is already building the `create-toolpad-app` modules

apedroferreira:
> > So basically the same thing I did here, but in the create-toolpad-app command only?
> 
> Yes, except I wouldn't add a `.env` file from the start. @bharatkashyap is already building the `create-toolpad-app` modules

Yep, I've already removed the .env file.
We can adapt this once we have that command.

bharatkashyap (approved):
👍

PR #1707, opened 2023-02-23
Open browser to app as part of of `dev` script

bharatkashyap:
- Part of #1653 
- Visual demonstration

https://user-images.githubusercontent.com/19550456/220887449-d7bcce86-d1e7-4701-a542-86896543cdc3.mov

### DIFF
--- a/packages/toolpad-app/cli/index.ts
+++ b/packages/toolpad-app/cli/index.ts
@@ -3,6 +3,8 @@ import arg from 'arg';
 import path from 'path';
 
 const DEFAULT_PORT = 3000;
+const MAX_RETRIES = 30;
+const INTERVAL = 1000;
 
 function* getNextPort(port: number = DEFAULT_PORT) {
   while (true) {
@@ -12,16 +14,21 @@ function* getNextPort(port: number = DEFAULT_PORT) {
 }
 
 interface RunCommandArgs {
-  // Whether Toolpad editor is running in dev mode (for debugging purposes only)
-  devMode?: boolean;
+  // Whether Toolpad editor is running in debug mode
+  debugMode?: boolean;
   port?: number;
+  // Whether to disable the browser automatically opening up on startup
+  browser?: boolean;
 }
 
-async function runApp(cmd: 'dev' | 'start', { devMode = false, port }: RunCommandArgs) {
+async function runApp(
+  cmd: 'dev' | 'start',
+  { debugMode = false, port, noBrowser = false }: RunCommandArgs,
+) {
   const { execa } = await import('execa');
   const { default: getPort } = await import('get-port');
   const toolpadDir = path.resolve(__dirname, '../..'); // from ./dist/server
-  const nextCommand = devMode ? 'dev' : 'start';
+  const nextCommand = debugMode ? 'dev' : 'start';
 
   if (!port) {
     port = cmd === 'dev' ? await getPort({ port: getNextPort(DEFAULT_PORT) }) : DEFAULT_PORT;
@@ -43,11 +50,34 @@ async function runApp(cmd: 'dev' | 'start', { devMode = false, port }: RunComman
   cp.stdout?.pipe(process.stdout);
   cp.stderr?.pipe(process.stdout);
 
-  setTimeout(() => {
-    // eslint-disable-next-line no-console
-    console.log("Navigating to the app's page in the browser...");
-    execa('open', [`http://localhost:${port}`], { stdio: 'inherit' });
-  }, 3000);
+  if (cp.stdout && cmd === 'dev' && !noBrowser) {
+    const { default: fetch } = await import('node-fetch');
+
+    const checkedUrl = new URL(`https://localhost:${port || DEFAULT_PORT}`);
+    for (let i = 1; i <= MAX_RETRIES; i += 1) {
+      try {
+        // eslint-disable-next-line no-console
+        console.log(`(${i}/${MAX_RETRIES}) trying reach "${checkedUrl.href}"...`);
+        // eslint-disable-next-line no-await-in-loop
+        const res = await fetch(checkedUrl.href);
+
+        if (!res.ok) {
+          throw new Error(`HTTP ${res.status}`);
+        }
+
+        // eslint-disable-next-line no-console
+        console.log(`connected!`);
+        execa('open', [`http://localhost:${port || DEFAULT_PORT}`], { stdio: 'inherit' });
+        return;
+      } catch (err: any) {
+        console.error(` > ${err.message}`);
+        // eslint-disable-next-line no-await-in-loop
+        await new Promise((resolve) => {
+          setTimeout(resolve, INTERVAL);
+        });
+      }
+    }
+  }
 
   cp.on('exit', (code) => {
     if (code) {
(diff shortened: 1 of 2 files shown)


Janpot:
* It should only open a browser in `toolpad dev` mode, not in `toolpad start`
* We should monitor `stdout` for readiness of the app instead of a timeout **
* Are all failure modes covered? e.g. what if I start dev mode with a port that's already in use?
* It should be opt-outable. I'm currently working on writing tests and I should be able to opt out of the browser opening. Perhaps we can reuse the `create-react-app` env variable `BROWSER=none`? Perhaps we can even reuse their `openBrowser` script? https://github.com/facebook/create-react-app/blob/main/packages/react-dev-utils/openBrowser.js

** One way I'm currently achieving this for the tests is by:
```tsx
import { Readable } from 'stream';

async function waitForMatch(stream: Readable, regex: RegExp): Promise<RegExpExecArray | null> {
  return new Promise((resolve, reject) => {
    const chunks: Buffer[] = [];
    const onData = (chunk: any) => {
      chunks.push(Buffer.from(chunk));
      const string = Buffer.concat(chunks).toString('utf8');
      const match = regex.exec(string);
      if (match) {
        resolve(match);
        stream.off('data', onData);
      }
    };
    stream.on('data', onData);
    stream.on('error', (err) => reject(err));
    stream.on('end', () => resolve(null));
  });
}

// ...

const match = await waitForMatch(childProcess.stdout, /http:\/\/localhost:(\d+)/);
const port = match ? Number(match[1]) : null
```

apedroferreira:
> We should monitor stdout for readiness of the app instead of a timeout **
What about polling for some health check?

Janpot:
> What about polling for a health check?

👍  Good idea! That would work as well.

bharatkashyap:
@janpot

- [x] It should only open a browser in `toolpad dev` mode, not in `toolpad start`. `toolpad start` is for production usage, on servers.

- [x] We should poll for a health check

> * Are all failure modes covered? e.g. what happens if I start dev mode with a port that's already in use?
Doesn't this https://github.com/mui/mui-toolpad/blob/direction-13/packages/toolpad-app/cli/index.ts#L27 deal with this already?

> It should be opt-outable. 
- [x] Added a `no-browser` flag which opts out from this behaviour

packages/toolpad-app/cli/index.ts
```diff
@@ -43,6 +54,34 @@ async function runApp(cmd: 'dev' | 'start', { devMode = false, port }: RunComman
   cp.stdout?.pipe(process.stdout);
   cp.stderr?.pipe(process.stdout);
 
+  if (cp.stdout && cmd === 'dev' && !noBrowser) {
```
Janpot, line 57: Perhaps we can add an invariant instead after the childprocess is created?

```tsx
invariant(cp.stdout, 'child process must be started with "stdout: 'pipe'"')
```

Janpot:
> Doesn't this https://github.com/mui/mui-toolpad/blob/direction-13/packages/toolpad-app/cli/index.ts#L27 deal with finding an available port already?

Not if the user provides a port through `--port`. In that case we assume the user explicitly wants to use this exact port and so we'll error when it's unavailable.

> We should poll for a health check

Just want to note that doing a pure healthcheck is in the worst case scenario one second slower than listening on the output of the child process stdout.

packages/toolpad-app/cli/index.ts
```diff
@@ -43,6 +54,34 @@ async function runApp(cmd: 'dev' | 'start', { devMode = false, port }: RunComman
...
+    const checkedUrl = new URL(HEALTH_CHECK_URL, DEFAULT_URL);
```
Janpot, line 58: * The port is hardcoded in `HEALTH_CHECK_URL`. This would healthcheck the wrong server if Toolpad chooses a different port.
* I'd abstract this polling logic into its own `waitForAvailable(port: string): Promise<void>` function

bharatkashyap: I've done these, and added some changes to use a `TOOLPAD_CLI_BROWSER` environment variable which supports either the names of browsers, or a "none" value to disable this behaviour.

The default browser that the script will attempt to open is Chrome.

packages/toolpad-app/cli/index.ts
```diff
@@ -26,22 +44,19 @@ interface RunCommandArgs {
...
 
 const waitForHealthCheck = async (port: number): Promise<void> => {
-  const { default: chalk } = await import('chalk');
   const TOOLPAD_BASE_URL = `http://localhost:${port}/`;
   const checkedUrl = new URL(HEALTH_CHECK_URL, TOOLPAD_BASE_URL);
   for (let i = 1; i <= MAX_RETRIES; i += 1) {
     try {
       // eslint-disable-next-line no-console
-      console.log(`(${i}/${MAX_RETRIES}) trying to reach "${checkedUrl.href}"...`);
+      console.log(`(${i}/${MAX_RETRIES}) trying to reach "${HEALTH_CHECK_URL}"...`);
```
Janpot, line 52:
```suggestion
      console.log(`(${i}/${MAX_RETRIES}) trying to reach "${HEALTH_CHECK_URL}"...`);
```

I don't think we should log here. The healthcheck is an implementation detail, and shouldn't even kick in.

packages/toolpad-app/cli/index.ts
```diff
@@ -151,12 +168,18 @@ async function runApp(cmd: 'dev' | 'start', { debugMode = false, port }: RunComm
...
 
   if (cmd === 'dev') {
-    const TOOLPAD_BASE_URL = `http://localhost:${port}/`;
+    // Poll stdout for "http://localhost:3000" first
+    const match = await waitForMatch(cp.stdout, /http:\/\/localhost:(\d+)/);
+    const detectedPort = match ? Number(match[1]) : null;
+    invariant(detectedPort, 'Could not find port in stdout');
+
+    // Then wait for health check to pass on the detected port
     await waitForHealthCheck(port);
```
Janpot, line 177: If we pass `url` here we don't have to duplicate 
```tsx
const TOOLPAD_BASE_URL = `http://localhost:${detectedPort}/`;
```

packages/toolpad-app/cli/index.ts
Janpot: > The default browser that the script will attempt to open is Chrome.

Wouldn't `open` use the default browser of the user's operating system? If so, I believe it makes sense to use that as the default behavior

packages/toolpad-app/cli/index.ts
```diff
@@ -1,8 +1,34 @@
...
+import { Readable } from 'stream';
+import * as readline from 'readline';
 
 const DEFAULT_PORT = 3000;
+// https://github.com/sindresorhus/open#app
+const DEFAULT_BROWSER = open.apps.chrome;
+const INTERVAL = 1000;
+const MAX_RETRIES = 30;
+
+const HEALTH_CHECK_URL = `/health-check`;
```
Janpot, line 15:
```suggestion
const HEALTH_CHECK_PATHNAME = `/health-check`;
```
?

packages/toolpad-app/cli/index.ts
```diff
@@ -12,19 +38,119 @@ function* getNextPort(port: number = DEFAULT_PORT) {
 }
 
 interface RunCommandArgs {
-  // Whether Toolpad editor is running in dev mode (for debugging purposes only)
-  devMode?: boolean;
+  // Whether Toolpad editor is running in debug mode
+  debugMode?: boolean;
```
Janpot, line 42:
```suggestion
  devMode?: boolean;
```

packages/toolpad-app/cli/index.ts
```diff
@@ -12,19 +37,113 @@ function* getNextPort(port: number = DEFAULT_PORT) {
...
+      // eslint-disable-next-line no-await-in-loop
+      await new Promise((resolve) => {
+        setTimeout(resolve, INTERVAL);
+      });
+    }
+  }
+  throw new Error(`Failed to connect`);
+};
+
+// From https://github.com/facebook/create-react-app/blob/main/packages/react-dev-utils/openBrowser.js
```
Janpot, line 66: Perhaps it makes sense to keep that in its own file?

Janpot (approved):

PR #1710, opened 2023-02-24
Make Toolpad work offline

Janpot:
Discovered while on the plane this afternoon. `react-query` by default [doesn't fetch](https://tanstack.com/query/v4/docs/react/guides/network-mode) when it detects being offline


apedroferreira (approved):
Wow, didn't know about that.

Issue #1663, opened 2023-02-14
Improve type definitions of the query parameters


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Infer query `parameters` typescript type from the source.. This can be done in a build step by introspecting the typescript types.

First raised in: https://github.com/mui/mui-toolpad/issues/1650

PR #1716, opened 2023-02-27
Improve createQuery types

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/1663
Infer `createQuery` parameter types from the value of the configuration argument.

### DIFF
--- a/packages/toolpad-core/src/createQuery.tsx
+++ b/packages/toolpad-core/src/createQuery.tsx
@@ -8,7 +8,7 @@ import {
   StringValueType,
 } from './types.js';
 
-export type QueryParameterValue<V extends PrimitiveValueType> = V extends NumberValueType
+type QueryParameterValue<V extends PrimitiveValueType> = V extends NumberValueType
   ? number
   : V extends StringValueType
   ? string
--- a/packages/toolpad-core/src/createQuery.tsx
+++ b/packages/toolpad-core/src/createQuery.tsx
@@ -8,7 +8,7 @@ import {
   StringValueType,
 } from './types.js';
 
-export type QueryParameterValue<V extends PrimitiveValueType['default']> = V extends NumberValueType
+export type QueryParameterValue<V extends PrimitiveValueType> = V extends NumberValueType
   ? number
   : V extends StringValueType
   ? string
--- a/packages/toolpad-core/src/createQuery.tsx
+++ b/packages/toolpad-core/src/createQuery.tsx
@@ -20,12 +20,12 @@ type QueryParameterValue<V extends PrimitiveValueType> = V extends NumberValueTy
   ? Record<string, unknown>
   : never;
 
-export interface QueryParameterConfig<P extends Record<string, unknown>, K extends keyof P> {
+export interface QueryParameterConfig<P, K extends keyof P> {
   typeDef: PrimitiveValueType;
   defaultValue?: P[K];
 }
 
-export interface CreateQueryConfig<P extends Record<string, unknown>> {
+export interface CreateQueryConfig<P> {
   parameters: {
     [K in keyof P]: QueryParameterConfig<P, K>;
   };
--- a/packages/toolpad-core/src/createQuery.tsx
+++ b/packages/toolpad-core/src/createQuery.tsx
@@ -1,30 +1,58 @@
 import { TOOLPAD_QUERY } from './constants.js';
-import { PropValueType } from './types.js';
+import {
+  PrimitiveValueType,
+  NumberValueType,
+  BooleanValueType,
+  ArrayValueType,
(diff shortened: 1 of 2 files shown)


apedroferreira:
@Janpot I tried many different approaches and this seemed like the best one.
There might be another way to do this with Typescript that I don't know of, or maybe I didn't fully understand what "This can be done in a build step by introspecting the typescript types." from the issue description meant? So let me know if so.

packages/toolpad-core/src/createQuery.tsx
```diff
@@ -1,16 +1,27 @@
...
+} from './types.js';
 
-export interface QueryResolverParams<P> {
-  parameters: P;
-}
-
-export interface QueryResolver<P> {
-  (params: QueryResolverParams<P>): Promise<any>;
-}
+type QueryParameterValue<V extends PrimitiveValueType> = V extends NumberValueType
```
Janpot, line 11: The drawback of this is that we have to keep it in sync with the `PrimitiveValueType`. Perhaps there is a way to do it based on the `default` property? Not sure if this works

```tsx
export interface ValueTypeBase<T = unknown> {
  type: 'string' | 'boolean' | 'number' | 'object' | 'array' | 'element' | 'template' | 'event';
  default?: T;
}

export interface StringValueType extends ValueTypeBase<string> {
  type: 'string';
  enum?: string[];
}

// ...

export type TsTypeOfPropValueType<T extends PropValueType> = NonNullable<T['default']>;
```

Janpot:
> I tried many different approaches and this seemed like the best one

At its root, isn't the problem the same one as for `createComponent`? Intuitively I'd expect both implementations to be very similar. Maybe the `createComponent` implementation isn't the greatest?

> or maybe I didn't fully understand what "This can be done in a build step by introspecting the typescript types" from the issue description meant?

Me neither, I removed that sentence from the issue. Maybe we should have groomed it first. 

> There might be another way to do this with Typescript that I don't know of

This seems like a good approach to me, the only think I maybe find suboptimal is that it requires us to keep this `QueryParameterValue` type in sync with the primitives. Perhaps that part could be done with some sort of [lookup table](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgCoE8AOEAyB7PAawFdNkBvAKGWRGIFsAjaALloeagG5rkBnMFFABzNgKEhhPGowIAbCHBBtZeBUp4BfSpVCRYiFAAU4UOPQBCcPhAA8AaWQQAHpBAATPskIR0eGGhYuAQkZAC83r7+gdj4RKQAfBS8hKDubPbSyO4Q8MRyYAD8bBixIaQA2vYAulo6etDwSMgAchzQJmb0Tq4QHl6d5lY2tgDkdEzQo0lUNPRwzsXsk9y89KBLE5x1uuCNhsgAyoIig90ubp7IZ8N2o+Ii08lzCzh9wmAAFpvtq9q7+iaKAs8kUIDOPUuA1MQ2sd1U6hAT3I-zAQWuMO6ETaKwhAB8jidJPjkCC1GCzjxKGjsGg+KUIAB5GBnWxGSF9K5nJIRBlxUIVIwVUapDyjaq1HQwYggBBgYB4EDIBBQRSQACKxGg6Fs6o5-WQACUIAg8FB3LYHpIADQYroJBIACl4MGUyEdmExfDY5GQVWQoEifgC6uqJXpQWZrPVVWqSU0AEpkGEkgA3PDAdzW3g5GDe5Dq7MJtjpzMUf4qtUQTXax0er1JlPl22zZDOH3eNJsca-UbITTZmjoDui9LIUYIsF9geURNcIA) though I'm not sure whether this will impose constraints on other use cases. These types are used all over the place.

Janpot (approved):

apedroferreira:
> At its root, isn't the problem the same one as for createComponent? Intuitively I'd expect both implementations to be very similar. Maybe the createComponent implementation isn't the greatest?

I'll take a quick look at `createComponent`, maybe something can be improved there too.

> Me neither, I removed that sentence from the issue. Maybe we should have groomed it first.

Yeah the description is what might have thrown me off a bit so I didn't know if there was some better approach. If not then I think this one works, I also tried keeping the parameters as the generic but seemed like there was no good way to do that.

> This seems like a good approach to me, the only think I maybe find suboptimal is that it requires us to keep this QueryParameterValue type in sync with the primitives.

Yeah, even though I don't know if `PrimitiveValueType` will ever change a lot. I'll try using the type from `default`, I think I already tried it at some point and it worked? I'll also give the approach with a lookup table an attempt too.

apedroferreira:
The current solution throws an error if the lookup is missing any type from `PrimitiveValueType`, so we should be good!
Will merge if this is green.

Issue #1724, opened 2023-03-02
Pages delete automatically


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

Create an app with 2-3 pages and abruptly the page will be deleted.

### Current behavior 😯

I am working on an app but somehow the pages keep getting deleted automatically. It has happened 5-6 time today. 
After this toolpad.yaml also gets erased.

<img width="649" alt="Screenshot 2023-03-02 at 6 21 53 PM" src="https://user-images.githubusercontent.com/92228082/222434910-c7279b19-3c2b-4123-b6bd-1fa411d89664.png">
<img width="850" alt="Screenshot 2023-03-02 at 6 11 43 PM" src="https://user-images.githubusercontent.com/92228082/222434914-a59ebfd3-30d2-4ac7-bed8-554147cefea0.png">

Janpot:
Are you by any chance renaming the file extension to `.yaml`?

I have a potential fix in https://github.com/mui/mui-toolpad/pull/1725

prakhargupta1:
> Are you by any chance renaming the file extension to .yaml?
No no. Sorry I meant, toolpad.yml

> Do you see any errors in the output of running yarn dev? So that is the output in the terminal, preferrably pasted as text, not a screenshot.
Nope, nothing reflects on the terminal when the page gets deleted.

prakhargupta1:
This seems to have been fixed. Hasn't happened so far.

Janpot:
Ok great 👍 Closing then. Feel free to reopen if it happens again

PR #1725, opened 2023-03-02
Fix config file being deleted

Janpot:
* Do not migrate the config file by extension, just read/write from which one is available: .yml or .yaml
* Pipe childprocess output before waiting on output, to see immediate feedback of the app starting
* Make sure the healthcheck works on node.js <= 16 that don't have global fetch
     > note: in dev mode it needs to wait unnecessarily long until the `/healthcheck` path has been built. I propose we don't healthcheck and just rely on the child process output to tell when it's ready

Fixes https://github.com/mui/mui-toolpad/issues/1724 ?

### DIFF
--- a/packages/toolpad-app/cli/index.ts
+++ b/packages/toolpad-app/cli/index.ts
@@ -4,6 +4,7 @@ import path from 'path';
 import invariant from 'invariant';
 import { Readable } from 'stream';
 import * as readline from 'readline';
+import fetch from 'node-fetch';
 import openBrowser from './openBrowser';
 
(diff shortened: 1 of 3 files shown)


bharatkashyap:
> * Make sure the healthcheck works on node.js <= 16 that don't have global fetch
👍 
>   > note: in dev mode it needs to wait unnecessarily long until the `/healthcheck` path has been built. I propose we don't healthcheck and just rely on the child process output to tell when it's ready
Got it

bharatkashyap (approved):

PR #1727, opened 2023-03-03
Try replacing tsc with tsup as build tool

Janpot:
Running into trouble using modules from `toolpad-app/src/**` in the cli because they depend on ESM-only modules (`execa`, `node-fetch@3`, `fractional-indexing`, `open-editor`). The current setup makes it impossible to use these modules in a commonjs context (e.g. like the new CLI). `tsc` doesn't allow for transpiling dependencies. Proposing in this PR to move to `tsup`, which allows marking certain modules as `noExternal`. This can be used to bundle ESM only modules in the CLI.

* makes it possible to bundle ESM-only modules so we can use them in node.js
* increased performance
* Found a few missing dependencies in `@mui/toolpad-components`

### Alternatives considered:
* do nothing, `const { execa } = await import('execa')` everywhere: Not always applicable, the burden is on the module consumer, creates a split-brain between different modules, extra overhead
* `microbundle`: slower, less batteries-included, less maintained, less popular
* `babel`: slow, less batteries-included
* `@vercel/ncc`: seems to be mainly focused on creating single file, node.js only focussed
* ?

https://npmtrends.com/microbundle-vs-tsdx-vs-tsup


apedroferreira:
`tsup` definitely looks like the standard now, so this looks good!

PR #1738, opened 2023-03-10
Allow falsy MySQL variable values

evankennedy:
When sending a falsy (`0`/`false`) variable to MySQL to be substituted, it will ignore the variable and you will get a prepared statement error since it can't find all the substitutions. This fixes that bug.

## Alternatives

We could also do a `params.hasOwnProperty(varName)` check or `varName in params`, which may catch even more cases. Let me know if I should update.

### DIFF
--- a/packages/toolpad-app/src/toolpadDataSources/mysql/server.ts
+++ b/packages/toolpad-app/src/toolpadDataSources/mysql/server.ts
@@ -18,7 +18,7 @@ function prepareQuery(sql: string, params: Record<string, string>) {
       const trimmedMatch = match.trim().replaceAll(/[']+/g, '');
       if (trimmedMatch[0] === '$') {
         const varName = trimmedMatch.slice(1);
-        if (params[varName]) {
+        if (typeof params[varName] !== 'undefined') {
           substitutions.push(params[varName]);
         }
         return '?';


Janpot:
@evankennedy Thanks for your contribution. As yo may have noticed, we're currently moving Toolpad into a [new direction](https://github.com/mui/mui-toolpad/discussions/1748). At the moment we haven't fitted in database connections yet, though it's planned for the future. I will however merge this PR since we intend to keep the code around for the foreseeable future.

We'd be interested in hearing about how you plan to use Toolpad. Would you be open for doing a short interview with @prakhargupta1 and me to tell us a bit more about your use-case?

Janpot (approved):

evankennedy:
> @evankennedy Thanks for your contribution. As yo may have noticed, we're currently moving Toolpad into a [new direction](https://github.com/mui/mui-toolpad/discussions/1748). At the moment we haven't fitted in database connections yet, though it's planned for the future. I will however merge this PR since we intend to keep the code around for the foreseeable future.
> 
> We'd be interested in hearing about how you plan to use Toolpad. Would you be open for doing a short interview with @prakhargupta1 and me to tell us a bit more about your use-case?

Sure I'd be happy to go over a few use cases we were planning to use it for.

PR #1753, opened 2023-03-15
Add autocomplete to sx props

Janpot:
Test out adding autocomplete to the sx prop JSON editor. This PR just adds two props to as POC

https://user-images.githubusercontent.com/2109932/225309014-c1c51210-80f9-4829-8f6a-ba5b2cc8c529.mov

To do:
- [ ] Decide on which properties and add them
- [ ] Understand the limitations of this method. We're restricted to what Monaco editor + JSON-schema offers, no customizability possible.


apedroferreira:
Nice approach, but it could be difficult to make it work for everything that can be used in an sx prop and keep it up to date, right? Unless maybe if there is some way to keep it in sync with the Typescript types.

In a way it would also be cool to make this pro-code and customizable in the editor itself, but on the other hand that would add more files and configuration problems, and this is just one JavaScript object, so I guess that it's good to keep this in the application UI and have a good autocomplete or a good UX that guides the user effectively.

PR #1771, opened 2023-03-21
Correct moduleResolution usage across workspaces

Janpot:
* Remove `mitt` and replace with our own implementation. It's incorrectly defining its typings and seems to be unmaintained https://github.com/developit/mitt/issues/178
* `@mui/toolpad-app`: uses `next` so it can have `moduleResolution` `bundler`
* `@mui/toolpad-core`, `@mui/toolpad-components`: uses `tsc`, so `nodeNext`. Needed to add a few `.js` extensions so it produces correct output. We can change this to `bundler` once we move to [`tsup`](https://github.com/mui/mui-toolpad/pull/1727)
* Make sure the workspace uses the installed `typescript` version in thew project
* We can now remove the `paths` in `tsconfig` as `moduleResolution` now understands dual packages it correctly


Janpot:
ugh, Next.js rewrites `moduleResolution` back to `node`

Issue #1742, opened 2023-03-13
Query editor in draft state closes after switching tabs


Janpot:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

This issue concerns the branch https://github.com/mui/mui-toolpad/pull/1651

1. click add query
2. select "serverside HTTP request"
3. switch tabs
4. switch back to the editor

### Current behavior 😯

The query editor closes

### Expected behavior 🤔

The query editor is still open

Janpot:
@apedroferreira I thought at first it was related to the recent router changes, but it doesn't seem to happen on master

PR #1784, opened 2023-03-22
Fix changing tabs closing query editor

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/1742

- Fix query editor closing when switching browser tabs
- Also fixes console error when opening the query editor due to `margin: auto` in the page node.

https://user-images.githubusercontent.com/10789765/227025076-6d0c7a40-0dc5-4489-b46f-17c759865706.mov


packages/toolpad-app/src/toolpad/AppEditor/PageEditor/QueryEditor/index.tsx
```diff
@@ -130,8 +130,12 @@ export default function QueryEditor() {
   const { queries = [] } = appDom.getChildNodes(dom, page) ?? [];
 
   const handleEditStateDialogClose = React.useCallback(() => {
-    appStateApi.setView({ kind: 'page', nodeId: page.id });
-  }, [appStateApi, page.id]);
+    if (!dialogState?.isDraft) {
```
Janpot, line 133: Prefer positive logic
```suggestion
    if (dialogState?.isDraft) {
      setDialogState(null);
      // ...
```

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/QueryEditor/index.tsx
```diff
@@ -184,15 +188,19 @@ export default function QueryEditor() {
...
   React.useEffect(() => {
-    setDialogState(() => {
+    setDialogState((previousState) => {
       if (currentView.kind === 'page' && currentView.view?.kind === 'query') {
         const node = appDom.getNode(dom, currentView.view?.nodeId, 'query');
         return { node, isDraft: false };
       }
 
-      return null;
+      if (!dialogState?.isDraft) {
```
Janpot, line 197: Same,

```suggestion
      if (dialogState?.isDraft) {
        return previousState;
      }
      
      return null;
```

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/QueryEditor/index.tsx
```diff
@@ -184,15 +188,19 @@ export default function QueryEditor() {
...
+        return null;
+      }
+
+      return previousState;
     });
-  }, [dom, currentView]);
+  }, [dom, currentView, dialogState?.isDraft]);
```
Janpot, line 203: Perhaps extract this in a variable and reuse?

```tsx
const isDraft = !!dialogState?.isDraft;
```

Janpot (approved):
Few stylistic nits. Should we add/amend a test for this?

apedroferreira:
> Should we add/amend a test for this?

Not sure, this was quite a specific issue since it involves switching tabs. The `dom` object changes when we switch tabs and it's a dependency in the effect that closes the query editor, so it was running unintentionally.
I can add a test, I don't think it's the most obvious case to write a test for, but if we don't mind the extra test it also offers extra coverage. I might just try to include switching tabs in an existing test for the query editor not to add a whole new test too, I'll give it a try.

apedroferreira:
> Few stylistic nits. Should we add/amend a test for this?

I've modified the `rest-basics` test since it already tested the query editor a bit.

Janpot (approved):

PR #1786, opened 2023-03-23
[website] Landing page design tweaks

danilo-leal:
While skimming the current Toolpad landing page, I realized we have some tiny yet impactful design refinement opportunities we could make. Some of these styles are novel to this page eventually we'll bring them on to the others. 

- Haven't really changed the content of anything, it's just the design/layout.
  - The biggest exception is the removal of the "Get started" button from the hero. I was feeling like the stacked layout and two buttons were competing as to what was the main action the hero section is promoting. As we already promote the docs elsewhere in the page, focusing on capturing emails there only seems like a better direction, and it's also cleaner.
- I've created some new components and left old ones commented on. There's still code clean-up/linting stuff to correct.

**👉 https://deploy-preview-1786--mui-toolpad-docs.netlify.app/toolpad/**


bharatkashyap (approved):
This is amazing @danilo-leal! Thanks for this 🙏 🥳

docs/src/components/landing/BannerV2.js
```diff
@@ -0,0 +1,75 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import Container from '@mui/material/Container';
+import Button from '@mui/material/Button';
+import Link from 'docs/src/modules/components/Link';
+import Typography from '@mui/material/Typography';
+import KeyboardArrowRightRounded from '@mui/icons-material/KeyboardArrowRightRounded';
+
+function BannerV2(props) {
+  const { title, description, href, label, category, action, docs } = props;
```
bharatkashyap, line 10: We probably don't need the `docs` prop on this component anymore since this was only to differentiate styles when re-using the same component between the landing page and the docs

danilo-leal: Ah, cool ⎯ that makes sense! I'll remove it! Let me know if keeping this component as `BannerV2` makes sense either 😬 I did it in such a way as to not delete anything that would break.

bharatkashyap: I feel like if we're not using `Banner` anymore, we can get rid of it and just name the new component `Banner` - that is perfectly fine for me

danilo-leal:
Just a note that, this last commit (https://github.com/mui/mui-toolpad/pull/1786/commits/264bb2f9947534ea0b33d9c51e9af0f26c44cc62) pulled a lot of changes that weren't necessarily meant to be here! So, let me know if removing them is needed or if it's fine to leave them. I don't really know why they were picked up!

bharatkashyap (approved):
✅ Great work!

Janpot (approved):

docs/src/components/landing/PricingTable.js
```diff
@@ -286,17 +295,37 @@ function PricingTable({
         <Box sx={gridSx}>
           <Typography />
           {plans.map((plan) => (
-            <Box key={plan} sx={{ display: 'flex', flexDirection: 'column', p: 2 }}>
+            <Box
```
danilo-leal, line 298: Could someone help me out with this particular section? I was going for making only the _recommended_ column highlighted but could figure out a way to do so.

Screen Shot 2023-04-05 at 21 26 43

bharatkashyap: Added a commit that does this!

danilo-leal: Thank you, really appreciate it!

prakhargupta1:
Thanks Danilo for these inputs! Looks great :)

gerdadesign (commented):
@danilo-leal Just noting from this PR that the alpha label is no longer consistent throughout the site. Fine to adjust it, but we should make sure it's adjusted everywhere, including the Figma components, too! 

ToolpadLP
Products
Docs

PR #1789, opened 2023-03-23
Fixes on the dev pipeline

Janpot:
* Avoid starting the up docs (use `yarn docs:dev` for this)
* Make sure `create-toolpad-app` doesn't trash the console output
* Make sure `create-toolpad-app` only builds on `src` files changes
* align `pretty` and `preserveWatchOutput` across workspaces in `tsconfig.json`


bharatkashyap (commented):
👌 ✅

PR #1793, opened 2023-03-23
Fix imports from toolpad

Janpot:
* Only install `@mui/toolpad` and not `@mui/toolpad-core` from `creat-toolpad-app`
* Provide `@mui/toolpad/browser` and `@mui/toolpad/server` and make sure they're dual mode commonjs/esm packages
*


bharatkashyap (commented):
✅

PR #1794, opened 2023-03-23
Rewrite argument handling in `@mui/toolpad` CLI with `yargs`

### LINKED ISSUE #1744: Support CLI --help command
### Summary 💡

```
npx @mui/toolpad --help
```
Should print helpful message outlining all possible commands.

```
npx @mui/toolpad dev --help
npx @mui/toolpad build --help
npx @mui/toolpad start --help
```
Should each print a helpful message of the sub command

* Let's move to `yargs` to automate this somewhat?
* Let's add a `npx @mui/toolpad help` command as well as an alias for `npx @mui/toolpad --help`?

bharatkashyap:
- Closes #1744 
- Closes https://github.com/mui/mui-toolpad/issues/1653

- Visual demonstration 
Screenshot 2023-03-23 at 2 28 47 PM

Screenshot 2023-03-23 at 2 29 13 PM

Screenshot 2023-03-23 at 2 29 53 PM

Screenshot 2023-03-23 at 2 30 22 PM

Closes https://github.com/mui/mui-toolpad/issues/1744


Janpot:
* Let's keep `--dev` instead of changing it to `--next-prod` because:
  * prod should be the default, so we need a flag to enable dev mode
  * Let's keep the name `--dev` to avoid breaking other contributor's workflow.
* `--dev` is for Toolpad contributors only, it doesn't have to be documented in the `--help` command, regular users shouldn't use this flag.

Janpot (changes requested):
* Let's keep `--dev` instead of changing it to `--next-prod` because:
  * prod should be the default, so we need a flag to enable dev mode
  * Let's keep the name `--dev` to avoid breaking other contributor's workflow.
* `--dev` is for Toolpad contributors only, it doesn't have to be documented in the `--help` command, regular users shouldn't use this flag.

bharatkashyap:
> * I propose we keep `--dev` instead of changing it to `--next-prod` because:
>   
>   * prod should be the default, so we need a flag to enable dev mode
>   * Let's keep the name `--dev` to avoid breaking other contributor's workflow.

For me, the only reason to move away from `--dev` is the confusion it creates with the command of the same name.

> * `--dev` is for Toolpad contributors only, it doesn't have to be documented in the `--help` command, regular users shouldn't use this flag.

✅ 

> * Should we do the same for `create-toolpad-app`?

Yes, added some thoughts in #1795

Janpot:
> For me, the only reason to move away from `--dev` is the confusion it creates with the command of the same name.

Ok, let's change it to an environment variable `TOOLPAD_NEXT_DEV` (we need it for the tests anyway), but not in this PR, let's leave it at `--dev` for this PR, undocumented.

Janpot (approved):

PR #1795, opened 2023-03-23
Add folder as argument to `create-toolpad-app`

bharatkashyap:
- Brings the CLI documentation instructions https://mui.com/toolpad/getting-started/installation/#create-a-toolpad-app on par with `create-toolpad-app` to have the project name be accepted as a runtime argument


Janpot:
Some questions
* Is it optional?
* Can i use `.` to initialize in the current (empty) directory?

bharatkashyap:
> Some questions that popped in my head after reading this
> 
> * Can I use `.` to initialize in the current (empty) directory?
> * Can I use any path? e.g. `yarn create toolpad-app ../../foo/bar/toolpad-app`

I think that depends on whether the option is advertised as a path or a name; in the former case I expect that `.`, `../foo/bar/` , `newDir` and `./newDir` are all valid options, whereas in the latter case only "names" would work. 

We can expand the scope of this PR to include `yargs` for `create-toolpad-app` and add `path` as an option

Janpot:
> We can expand the scope of this PR to include `yargs` for `create-toolpad-app` and add path as an option

👍 If by "option" you mean "first positional argument" and not `--path=...`

bharatkashyap:
> > We can expand the scope of this PR to include `yargs` for `create-toolpad-app` and add path as an option
> 
> 👍 If by "option" you mean "first positional argument" and not `--path=...`

Yes ✅

packages/create-toolpad-app/src/index.ts
```diff
@@ -58,13 +59,13 @@ async function isFolderEmpty(root: string, name: string): Promise<boolean> {
...
     // eslint-disable-next-line no-console
-    console.log(`The directory ${chalk.green(name)} contains files that could conflict:`);
+    console.log(`The directory at ${chalk.green(pathDir)} contains files that could conflict:`);
     // eslint-disable-next-line no-console
     console.log();
     for (const file of conflicts) {
       try {
         // eslint-disable-next-line no-await-in-loop
-        const stats = await fs.lstat(path.join(root, file));
+        const stats = await fs.lstat(path.join(pathDir, file));
```
Janpot, line 68:
```suggestion
        const stats = await fs.lstat(path.resolve(pathDir, file));
```

packages/create-toolpad-app/src/index.ts
```diff
@@ -98,49 +99,54 @@ const installDeps = async (projectName: string, cwd: string) => {
...
+    version: '0.1.0',
+    private: true,
+    scripts: {
+      dev: 'toolpad dev',
+      build: 'toolpad build',
+      start: 'toolpad start',
+    },
+  };
+
+  await fs.writeFile(path.join(absolutePath, 'package.json'), JSON.stringify(packageJson, null, 2));
```
Janpot, line 122:
```suggestion
  await fs.writeFile(path.resolve(absolutePath, 'package.json'), JSON.stringify(packageJson, null, 2));
```

packages/create-toolpad-app/src/index.ts
```diff
@@ -154,48 +160,76 @@ const scaffoldProject = async (projectName: string, cwd: string): Promise<string
...
+      }
 
+      absolutePath = path.join(process.cwd(), pathArg);
+    }
     // eslint-disable-next-line no-await-in-loop
-    projectName = await scaffoldProject(name.projectName, cwd);
-  } while (!projectName);
+    absolutePath = await scaffoldProject(absolutePath);
+    count += 1;
+  } while (!absolutePath);
```
Janpot, line 224: What do we need the loop for?

packages/create-toolpad-app/src/index.ts
```diff
@@ -87,116 +65,136 @@ async function isFolderEmpty(root: string, name: string): Promise<boolean> {
...
-// Install the dependencies
-const installDeps = async (projectName: string, cwd: string) => {
-  const { execaCommand } = await import('execa');
+const validatePath = async (relativePath: string): Promise<boolean | string> => {
+  const { default: chalk } = await import('chalk');
+
+  const absolutePath = path.join(process.cwd(), relativePath);
+
+  try {
+    await fs.mkdir(absolutePath);
```
Janpot, line 74:
```suggestion
    await fs.mkdir(absolutePath, { recursive: true });
```

packages/create-toolpad-app/src/index.ts
```diff
@@ -87,116 +65,136 @@ async function isFolderEmpty(root: string, name: string): Promise<boolean> {
...
+    // Directory exists, verify if it's empty to proceed
+    try {
+      if (await isFolderEmpty(absolutePath)) {
+        return true;
+      }
+      return `${chalk.red('error')} - The directory at ${chalk.blue(
+        absolutePath,
+      )} contains files that could conflict. Either use a new directory, or remove conflicting files.`;
+    } catch {
+      return `${chalk.red('error')} - Unable to access directory at ${chalk.red(
```
Janpot, line 87: This way there is no feedback to the user about what went wrong exactly, since it's unexpected, I'd probably just let the error bubble up and crash the program.

packages/create-toolpad-app/src/index.ts
bharatkashyap: From https://nodejs.org/api/fs.html#fspromisesmkdirpath-options 
> Calling fsPromises.mkdir() when path is a directory that exists results in a rejection only when recursive is false

I want to be able to detect when a directory exists on calling `mkdir` so that I am able to catch that error and then call `isDirectoryEmpty` on the existing directory, but setting `recursive` to `true` causes me to miss that case

Janpot: Won't you need `recursive` if you pass a nested directory to the CLI? e.g. 

```
yarn create toolpad-app some/nested/dir
```

I'd just validate the path before calling `fs.mkdir` instead of after

packages/create-toolpad-app/src/index.ts
```diff
@@ -71,23 +71,23 @@ const validatePath = async (relativePath: string): Promise<boolean | string> =>
...
-      return `${chalk.red('error')} - Unable to access directory at ${chalk.red(
-        absolutePath,
-      )}. Please provide a different path.`;
+    if (await isFolderEmpty(absolutePath)) {
+      return true;
     }
+    return `${chalk.red('error')} - The directory at ${chalk.blue(
+      absolutePath,
+    )} contains files that could conflict. Either use a new directory, or remove conflicting files.`;
+  } catch (error: any) {
```
Janpot, line 83:
```suggestion
  } catch (rawError: unkown) {
    const error = errorFrom(rawError);
```

bharatkashyap: Do you recommend importing `errorFrom` from `@mui/toolpad-core` or rewriting that function here?

bharatkashyap: Added something local to avoid adding dependencies to this package

Janpot (approved):

PR #1813, opened 2023-03-27
Use worker scoped fixtures for integration tests

Janpot:
Avoid restarting the same fixture server over and over again for different tests in the same file. This starts just one server per test file. this allows us to reuse the running server for multiple tests creating less overhead


test/models/shared.ts
```diff
@@ -10,4 +10,6 @@ export async function gotoIfNotCurrent(
   if (pathname !== expectedPath) {
     await page.goto(expectedPath, opts);
   }
+
+  await page.waitForURL((url) => url.pathname === expectedPath);
```
apedroferreira, line 14: Would it work to replace all instances of `page.waitForNavigation` with this `page.waitForURL`?
I think that's what Playwright is recommending now anyway.

apedroferreira (approved):
Looks clean!

test/models/shared.ts
Janpot: Yes, I already did a batch in https://github.com/mui/mui-toolpad/pull/1699. Thought I had all of them.
But let's keep that for a follow up indeed

Issue #1547, opened 2023-01-12
When preview is initiated show active page


bytasv:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Currently when I'm working with specific page and click `Preview` I end up with a full list of pages and have to do 1 extra click. It would be nice to land directly on the page that I'm currently working instead

Issue #813, opened 2022-08-17
In the app view, all pages should also be shown as a list on the left menu bar 


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

When previewing an app that has multiple pages, we can show a toggle asking if the pages should appear in a menu bar on the left. This is how most admin panels look.

We can also ask the developer to set one of the pages as default and show it when the app opens.

This is just one way of implementing it, we can discuss more ways like adding this menu bar as a component. 

### Examples 🌈
<img width="1435" alt="Screenshot 2022-08-17 at 9 04 43 PM" src="https://user-images.githubusercontent.com/92228082/185181366-21cafa39-ffcf-40fa-810a-e07ecd38a7f8.png">

Later on:
1. we can have a header nav bar
2. these 2 components can be prominent in assigning a theme to the app

### Motivation
https://community.retool.com/t/building-side-nav-and-multiple-pages-dashboard/6219

Reference:
https://www.notion.so/mui-org/UI-layout-for-navigation-between-pages-5e69b677da7c4a469205b133c3133fa5

### Benchmark
https://www.npmjs.com/package/react-pro-sidebar

High level tasks

- [x] Add Toolpad shell = navigation sidebar
- [x] Make Toolpad shell removable through query param `?toolpad-display=standalone`
- [x] Remove overview of apps
- [x] initial page (first one for now?)
- [x] preview button directly to the page https://github.com/mui/mui-toolpad/issues/1547

Janpot:
All apps I've seen created so far by internal dogfooding are one page only. Another idea could be to optimize for this use-case and remove the concept of pages altogether. Each app would be just one single page.

We could then later down the road provide multi-page organization with folders in the workspace.

prakhargupta1:
This looks good, much like react admin. We would soon have to conclude how we should show pages, like this way or retool way (page and folders, no concept of app)

prakhargupta1:
I suggest we do it as it will definitely make the app look more complete.  Let me know if there are any objections.
cc: @bharatkashyap

bharatkashyap:
> I suggest we do it as it will definitely make the app look more complete. Let me know if there are any objections. cc: @bharatkashyap

I realised that this creates an issue about authorisation, since now sharing only one page of an app with a user is not possible any more.

prakhargupta1:
Can we not just change this view to the one you showed while keeping everything else same?

[...]

PR #1819, opened 2023-03-27
Add app navigation sidebar

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/813
Closes https://github.com/mui/mui-toolpad/issues/1547

- Show navigation sidebar with all pages when viewing a preview or production Toolpad app
- Ability to hide navigation bar with `?toolpad-display=standalone`
- Remove app overview page and use first page as default page (it should be easy to set any default page later, so we can do it in a separate issue)
- Remove preview button in right-side panel, as the "Preview" button in the top bar now opens the preview starting in the currently open page

### Development:

https://user-images.githubusercontent.com/10789765/228324130-b19d2f85-addc-4831-aa35-e72568b860a7.mov

### Production

https://user-images.githubusercontent.com/10789765/228049624-4196296f-40ed-406d-aaba-f7fe3b5758ae.mov


packages/toolpad-app/src/toolpad/ToolpadShell/Header/index.tsx
```diff
@@ -24,80 +24,83 @@ function Header({ actions, status, enableUserFeedback = true }: HeaderProps) {
...
+              gap: 2,
+            }}
+          >
+            {status}
+            <ThemeModeMenu mode={themeMode} onChange={handleThemeModeChange} />
+            {enableUserFeedback ? <UserFeedback /> : null}
+          </Box>
+        </Toolbar>
+      </AppBar>
+      <Toolbar />
```
apedroferreira, line 102: Adds spacing because `AppBar` position was changed to `fixed` so that it's not contrained by the new sidebar. As per recommended in MUI Drawer docs https://mui.com/material-ui/react-drawer/ ("Clipped under the app bar")

packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -1135,10 +1133,58 @@ const queryClient = new QueryClient({
   },
 });
 
+export interface ToolpadAppLayoutProps {
+  dom: appDom.RenderTree;
+  hasShell?: boolean;
+  version: VersionOrPreview;
+}
+
+function ToolpadAppLayout({ dom, version, hasShell: hasShellProp = true }: ToolpadAppLayoutProps) {
```
apedroferreira, line 1142: Moved these to their own component so that we can use React Router hooks.

prakhargupta1:
Nice! Is it that both preview options (1. right panel and 2. top bar) show the same result i.e, the preview for the entire app?
If Yes, then I think we can remove the one in the right bar. 

If we want to make `toolpad-display=standalone` more discoverable, we can also put page preview available from the ellipsis menu of the page.

apedroferreira:
> Nice! Is it that both preview options (1. right panel and 2. top bar) show the same result i.e, the preview for the entire app?
If Yes, then I think we can remove the one in the right bar.

The option at the top opens the app in preview but starting in the default page, while the right-side option opens the app starting in that specific page. So I think it's fine to keep these separate options?

If we want to make toolpad-display=standalone more discoverable, we can also put page preview available from the ellipsis menu of the page.

I can add a "Preview" option from the ellipsis menu of each page, but I feel like it should open without `toolpad-display=standalone` so that the user can navigate from there if they want to? So it would be exactly the same as the option in the right-side panel.

apedroferreira:
> Intuitively what makes most sense to me is for the top button to open the current page and to remove the button on the right.

Agreed, didn't consider that but seems more intuitive, I'll do that instead.

prakhargupta1:
> Would it make sense for the "edit" button on the Toolpad application to do the reverse operation? i.e. open the page editor of the page you were viewing.

Yes, I think that would make sense. Currently, it shows the first page in edit mode.

Janpot (approved):

apedroferreira:
> I was thinking, this will break all embeds that are currently in notion already.
> Our options:

I was thinking option 1 for now but 2 or 3 also seem fine.
Let's do 3 then, I'll try to include it in this PR as I think it shouldn't be that difficult.

bharatkashyap:
> I was thinking, this will break all embeds that are currently in notion already.
> 
> Our options:
> 
> 1. Do nothing in Toolpad. Update all embeds in notion
> 2. Automatically set `toolpad-display=standalone` when an iframe is detected. use `?toolpad-display=shell` to enable the navigation
> 3. Provide a page-level setting that sets the default for `toolpad-display`
> 4. Provide an app-level setting that sets the default for `toolpad-display`
> 
> Personally my preference goes out to 3. Would it be hard to implement? I'm fine dealing with this in a separate PR, but we need to prioritize it then so that it makes it in the same release as this one.

For me, an app-level setting is useful for this since it affects all pages. I was also wondering whether this is a good way to introduce "layouts" to the app: You can start with layout option 1 (a sidebar for navigation), or 2 (full page content, and a separate page for navigation) and then in the future, potentially, option 3 (tabs for navigation?), option 4 ...

bharatkashyap (approved):

apedroferreira:
> > I was thinking, this will break all embeds that are currently in notion already.
> > Our options:
> > 
> > 1. Do nothing in Toolpad. Update all embeds in notion
> > 2. Automatically set `toolpad-display=standalone` when an iframe is detected. use `?toolpad-display=shell` to enable the navigation
> > 3. Provide a page-level setting that sets the default for `toolpad-display`
> > 4. Provide an app-level setting that sets the default for `toolpad-display`
> > 
> > Personally my preference goes out to 3. Would it be hard to implement? I'm fine dealing with this in a separate PR, but we need to prioritize it then so that it makes it in the same release as this one.
> 
> For me, an app-level setting is useful for this since it affects all pages. I was also wondering whether this is a good way to introduce "layouts" to the app: You can start with layout option 1 (a sidebar for navigation), or 2 (full page content, and a separate page for navigation) and then in the future, potentially, option 3 (tabs for navigation?), option 4 ...

I just implemented option 3 but if we see it doesn't feel that useful we can try the other ones... I guess it would also depend on people's use cases, whether anyone would like different display settings per page or most people would use the same for every page.

Janpot:
The tools-public application has pages that we need to be chromeless, as well as pages that we need to have the sidebar

apedroferreira:
> I'm not sure the app-wide setting would be the best solution for us. The tools-public application has pages that we need to be chromeless, as well as pages that we need to have the sidebar.

Ok that's good to hear, I guess option 3 should be good then.

apedroferreira:
I've implemented option 3 as discussed:
- the `toolpad-display` URL parameter can be set as `standalone` or `shell`
- Page options have a select input where `Display mode` can be set to `App shell` or `No shell` - this controls a new `defaultDisplay` property in the page node attributes
- The initial page opened in preview/production, along with the `toolpad-display` URL parameter (which takes precedence), set the display mode for the app until the current browser tab is refreshed, for example

https://user-images.githubusercontent.com/10789765/229141902-f7f4f0c4-8a6f-4ab0-8d01-dd6dd2406faa.mov

Let me know if you have any more feedback or I can merge it later today or early next week!

Issue #1745, opened 2023-03-14
Query editor shows deprecated "query" in scope explorer


Janpot:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

Try binding a HTTP request URL. You'll see the "query" available in the global scope explorer. This is deprecated, so we should hide it

<img width="901" alt="Screenshot 2023-03-14 at 12 07 27" src="https://user-images.githubusercontent.com/2109932/224983499-2468106c-75aa-402c-b652-4f62239245b1.png">

PR #1827, opened 2023-03-28
Remove deprecated scope query

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/1745

Remove `query` from query scope.

Screenshot 2023-03-28 at 19 31 51


Janpot:
My initial idea was to just hide it, but maybe it's time to remove it altogether. Are we using it anywhere in any of our applications?

Janpot (approved):

apedroferreira:
> My initial idea was to just hide it, but maybe it's time to remove it altogether. Are we using it anywhere in any of our applications?

Oh ok, we can just hide it for some time and keep todo comments to remove it after some time has passed.

PR #1832, opened 2023-03-29
Do Toolpad app migration on startup

Janpot:
Working towards setting up migrating to pages folder. The current flow just transitions the DOM file on the fly every time it reads. This becomes  a lot more complex once we start reading and writing to multiple files.
I'm changing the flow to only migrate once during startup of Toolpad dev and that will also the place where we transform the current app dom to a pages folder for https://github.com/mui/mui-toolpad/pull/1831


apedroferreira (commented):
LGTM!

PR #1834, opened 2023-03-29
Fix broken theming

### LINKED ISSUE #1814: Fix theming
### Steps to reproduce 🕹

Steps:

1. Try to apply light/dark theme or primary/secondary color to a Toolpad app in the "Theme" tab in the right-side panel

### Current behavior 😯

Changes are not reflected in the UI

### Expected behavior 🤔

Changing theme or colors should have those changes immediately reflect in the UI, as before (this is a regression)

apedroferreira:
**Status:** Still going to add a test so leaving this PR in draft until then.

Closes https://github.com/mui/mui-toolpad/issues/1814.

Themes were not working - this fix seems to work.

Screenshot 2023-03-29 at 20 05 14

The regression was introduced in https://github.com/mui/mui-toolpad/pull/1627 - it looks like deep merging the palette like that does not work, so we have to override it in the theme object? @Janpot


packages/toolpad-app/src/runtime/AppThemeProvider.tsx
```diff
@@ -23,40 +23,42 @@ export function createToolpadTheme(toolpadTheme: AppTheme = {}): ThemeOptions {
 
   const theme = createTheme();
 
-  return createTheme(theme, {
-    typography: {
-      h1: {
-        fontSize: `3.25rem`,
-        fontWeight: 800,
-      },
+  return createTheme(
```
Janpot, line 26: Would just the following work? Not sure what that `theme` is for.

```tsx
createTheme({
  palette,
  typography: {
    // ...
  }
})
```

Janpot:
> Wait I can probably add a test too.

👍 Just a smoke test of an app with a dark theme should be fine. Not that useful to add tests for the theme editor at this point I believe.

Janpot (approved):

test/models/ToolpadEditor.ts
```diff
@@ -75,6 +77,8 @@ export class ToolpadEditor {
     this.componentCatalog = page.getByTestId('component-catalog');
     this.componentEditor = page.getByTestId('component-editor');
 
+    this.themeEditor = page.getByTestId('theme-editor');
```
Janpot, line 80: I meant to not test the theme editor, but ok 😄

apedroferreira: Oh, you meant just starting with dark theme... I thought you meant testing light/dark theme was ok but not UI colors as those could change more later.
Anyway even if we change things later I guess some of these could still be useful, if not we'll eventually delete some of this and that's fine.

PR #1840, opened 2023-03-30
Fix issues around custom datagrid columns

Janpot:
* Fix regression that prevents selecting a new column
* Add an example
* Show error when no component has been selected
* Add integration test


test/integration/data-grid/fixture-custom/toolpad/components/Other.tsx
```diff
@@ -0,0 +1,27 @@
+import { createComponent } from '@mui/toolpad/browser';
+import * as React from 'react';
+
+export interface Props {
```
apedroferreira, line 4: I guess these Props and argTypes still need to be defined here even though they're not used? Just making sure.

examples/custom-datagrid-column/toolpad/components/LabelColumn.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react';
```
apedroferreira, line 1: As this is an example would it be better if it used `value`, `row` and `field` so that it shows everything that can be done and also the typings people should use?
But this could be improved later too.

test/integration/data-grid/fixture-custom/toolpad/components/Other.tsx
Janpot: Not really necessary, I just copied the docs

examples/custom-datagrid-column/toolpad/components/LabelColumn.tsx
Janpot: I feel like using `row` would be the advanced use-case. Personally I'd rather leave it out.

apedroferreira: Ok, just a thought, i think that's fine too as it probably represents the most common use case.

apedroferreira (approved):

Issue #1785, opened 2023-03-22
For an authenticated REST API, allowed editing from the fetch editor


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

I have a REST API that I want to use in Toolpad. Since it requires authentication I am bound to do it using code queries. It would be great if I could just add my secrets and API pro-code way and expose it as a connection in the fetch editor (like earlier).  Then I can just add my endpoint, params etc. and build faster. 

Snapshot of how it looks currently in master--toolpad
<img width="1401" alt="Screenshot 2023-03-22 at 10 28 26 PM" src="https://user-images.githubusercontent.com/92228082/227047426-7c9becaf-f22c-49b4-b8a3-a1b9926009ea.png">

We could allow just binding to `process.env.MY_VAR` in the binding editors for serverside code:

![Image](https://user-images.githubusercontent.com/2109932/229537106-6aaa406a-91b7-4a60-b69b-d5b281bc2fec.png)

Janpot:
Alternatively, allow binding to environment variables directly

PR #1859, opened 2023-04-05
Ability to access process in HTTP query bindings

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/1785
Work in progress, might need to take a second look at some things.

### DIFF
--- a/packages/toolpad-app/src/toolpad/Toolpad.tsx
+++ b/packages/toolpad-app/src/toolpad/Toolpad.tsx
@@ -5,6 +5,7 @@ import { BrowserRouter, Routes, Route } from 'react-router-dom';
 import NoSsr from '../components/NoSsr';
 import AppEditor from './AppEditor';
 import ErrorAlert from './AppEditor/PageEditor/ErrorAlert';
+import { ToolpadContext } from '../ToolpadContext';
 
 const Centered = styled('div')({
   height: '100%',
@@ -39,23 +40,33 @@ function ErrorFallback({ error }: FallbackProps) {
 
 export interface ToolpadProps {
   basename: string;
+  envVarNames: string[];
 }
 
-export default function Toolpad({ basename }: ToolpadProps) {
+export default function Toolpad({ basename, envVarNames }: ToolpadProps) {
+  const toolpadContextValue = React.useMemo(
+    () => ({
+      envVarNames,
+    }),
+    [envVarNames],
+  );
+
   return (
     <NoSsr>
       {/* Container that allows children to size to it with height: 100% */}
-      <Box sx={{ height: '1px', minHeight: '100vh' }}>
-        <ErrorBoundary fallbackRender={ErrorFallback}>
-          <React.Suspense fallback={<FullPageLoader />}>
-            <BrowserRouter basename={basename}>
-              <Routes>
-                <Route path="/*" element={<AppEditor />} />
-              </Routes>
-            </BrowserRouter>
-          </React.Suspense>
-        </ErrorBoundary>
-      </Box>
+      <ToolpadContext.Provider value={toolpadContextValue}>
+        <Box sx={{ height: '1px', minHeight: '100vh' }}>
+          <ErrorBoundary fallbackRender={ErrorFallback}>
+            <React.Suspense fallback={<FullPageLoader />}>
+              <BrowserRouter basename={basename}>
+                <Routes>
+                  <Route path="/*" element={<AppEditor />} />
+                </Routes>
+              </BrowserRouter>
+            </React.Suspense>
+          </ErrorBoundary>
+        </Box>
+      </ToolpadContext.Provider>
     </NoSsr>
   );
 }
--- a/packages/toolpad-app/src/toolpadDataSources/rest/client.tsx
+++ b/packages/toolpad-app/src/toolpadDataSources/rest/client.tsx
@@ -53,17 +53,20 @@ import useQueryPreview from '../useQueryPreview';
 import TransformInput from '../TranformInput';
 import Devtools from '../../components/Devtools';
 import { createHarLog, mergeHar } from '../../utils/har';
-import config from '../../config';
 import QueryInputPanel from '../QueryInputPanel';
 import useFetchPrivate from '../useFetchPrivate';
 import { clientExec } from './runtime';
 import QueryPreview from '../QueryPreview';
+import { ToolpadContext } from '../../ToolpadContext';
 
 const HTTP_METHODS = ['GET', 'POST', 'PUT', 'DELETE', 'PATCH', 'HEAD'];
 
 const QUERY_SCOPE_META: ScopeMeta = {
   parameters: {
     description: 'Parameters that can be bound to app scope variables',
+  },
+  process: {
+    description: 'Contains information about the current Node.js process.',
   },
 };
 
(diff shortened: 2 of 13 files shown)


apedroferreira:
@Janpot How does this approach look? It should all be secure, right?
I noticed that the network console in the HTTP query dialog shows the full content of the request headers, but I guess that's only accessible to the developer of the app?

packages/toolpad-app/src/toolpad/Toolpad.tsx
```diff
@@ -55,18 +59,30 @@ export default function Toolpad({ basename, envVarNames }: ToolpadProps) {
...
-              <BrowserRouter basename={basename}>
-                <Routes>
-                  <Route path="/*" element={<AppEditor />} />
-                </Routes>
-              </BrowserRouter>
-            </React.Suspense>
-          </ErrorBoundary>
-        </Box>
+        <ThemeProvider>
+          <Head>
```
Janpot, line 63: I just removed this in https://github.com/mui/mui-toolpad/pull/1888
Is this a merge gone wrong?

bharatkashyap (approved):
Looks good!

packages/toolpad-app/src/toolpadDataSources/rest/server.ts
```diff
@@ -5,33 +5,41 @@ import { withHarInstrumentation, createHarLog } from '../../server/har';
...
+  );
 
   return { ...result, har };
 }
 
 async function execPrivate(connection: Maybe<RestConnectionParams>, query: FetchPrivateQuery) {
   switch (query.kind) {
+    case 'introspection': {
+      const env = await loadEnvFile();
+      return { env };
```
Janpot, line 41: I think we should just send the keys

packages/toolpad-app/src/toolpadDataSources/rest/client.tsx
```diff
@@ -363,10 +363,10 @@ function QueryEditor({
     () => ({
       parameters: previewParams,
       process: {
-        env,
+        env: Object.fromEntries(envVarNames.map((varName) => [varName, '<SECRET>'])),
```
Janpot, line 366: Oh, we're evaluating some bindings on the client side...
Then I guess it isn't possible to only send the variable names

apedroferreira: So it's not secure this way? I thought it only uses the real environment values if it runs server-side :/ Shouldn't that be possible?

apedroferreira: Ah, yes client-side it won't use the real values in this case, yeah...

apedroferreira: I can show `<SECRET>` and use the real values in the client anyway if that's still safe.

Janpot: I think we can do either one of those:

* Send the whole env like you did before and let it be
* Rethink how to use secrets in rest datasource and instead offer a new type of binding, next to javascript expression, allow binding directly to an environment variable through a dropdown

apedroferreira: I feel like we can go with the first option "Send the whole env like you did before and let it be".
But let me know if you think the second one is better for any reason and it would not be that much extra work anyway.

Janpot: Some arguments I can think of that would be in favor for the second solution:

- It's easier for us to detect which environment variables an app expects to be run under. That allows us to warn on any missing ones, e.g. when a user checks out an app for the first time. It could somewhat be extracted from the javascript bindings by regex, or static analysis, but it wouldn't be a foolproof solution.
- We could more fine grained specify where exactly we want this to be used. e.g. only in authentication/headers
- Thinking long term, suppose we decide that whatever we came up with in this PR was a mistake, we can migrate from `$env: MY_VAR` to `$jsExpression: 'process.env.MY_var'` and guarantee no breaking change, we can't do that on the inverse.

apedroferreira: Fair points, ok I'll change it to be a new type of binding, having it as `$env` sounds consistent with the other types.

Janpot (changes requested):
Looking good. I'd change a few things:
1. avoid the layout shift when switching tabs in the binding editor
    
   https://github.com/mui/mui-toolpad/assets/2109932/52a3599c-51cf-4d11-9cb0-ee6bdc3fe5d8

2. Make it a free input with autocomplete. We can give an indication when we don't find the environment variable in the env file, but I feel like we should allow input. As a follow up we can prompt the user for a value if we don't find it in the env file and add it ourselves

3. Update the `rest-basic` test with some checks for environment variables.

packages/toolpad-app/src/toolpadDataSources/rest/client.tsx
```diff
@@ -551,7 +565,7 @@ function getInitialQueryValue(): FetchQuery {
   return {
     method: 'GET',
     headers: [],
-    browser: config.isDemo,
```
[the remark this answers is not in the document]
apedroferreira: Yeah we should remove everything from the demo, I'll see if there's anything else - might include it in that PR too.

packages/toolpad-app/src/server/schema.ts
```diff
@@ -301,6 +308,7 @@ export const META = {
   definitions: {
     Json: jsonSchema,
     JsExpressionBinding: jsExpressionBindingSchema,
+    EnvBinding: envBindingSchema,
```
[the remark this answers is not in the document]
Janpot: An app created with this schema won't work in older Toolpad versions. Older Toolpad versions don't know what `$env` means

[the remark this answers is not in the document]
Janpot: The general idea would be:

- old schema + new toolpad => we should be able to internally migrate and display the app + give option for user to migrate files to new version, or do it automatically
- new schema + old toolpad => either all features of the schema are supported in Toolpad, or otherwise we need to tell the user "This app as created in newer Toolpad, please upgrade Toolpad to version X if you want to run/edit this application"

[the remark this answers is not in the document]
apedroferreira: At first thought the 2nd option seems more sane to me and not that unusual? 
But we can try the first one if it's not that hard, also depending on how common it would be for users to try to use new schemas in old Toolpad versions.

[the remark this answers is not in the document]
Janpot: These are two different scenarios that can happen, we have to deal with both of them

> also depending on how common it would be for users to try to use new schemas in old Toolpad versions.

This scenario could happen when people copy toolpad apps from one repo to another

apedroferreira:
> avoid the layout shift when switching tabs in the binding editor

I guess I can try to keep the top of the dialog in the same place at least maybe. The dialog with JS expression / navigation action also has this issue.

Janpot:
I think the easiest would be to keep it a consistent height like we do for the query editor https://github.com/mui/mui-toolpad/blob/f091cd625b4ca0a76acf8aa57fe2938459fec412/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/QueryEditor/QueryEditorDialog.tsx#L295

Janpot (approved):
this is great!

prakhargupta1:
This should be very helpful! I'll resume work on some of the public apps that needed this feature.
cc: @oliviertassinari

oliviertassinari:
Ah, yes, awesome, we can move more apps https://github.com/mui/mui-private/issues/267.

PR #1862, opened 2023-04-07
Refactor: extract custom components loading logic

Janpot:
This PR isolates the custom components loading logic from the app runtime and makes it injectable. This allows me to work on a new runtime that compiles components from the file system.
It simplifies the logic and takes a lot less LoC to achieve the same result


packages/toolpad-app/src/server/runtimeBuild.ts
Janpot, line 1: This was an early prototype of a runtime that can load components with correct `react` and node_modules.

Issue #1800, opened 2023-03-25
Add params option for page navigation with button onClick event


RBerthier:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

It seems not possible to add params option, when using page navigation when customising onClick event handler for button.

### Examples 🌈

![image](https://user-images.githubusercontent.com/891711/227718351-75f3b720-7744-4c56-a39c-159ca820e5aa.png)

After selecting a page to navigate to when clicking on a button, I would expect to have the possibility to add n* params to the page navigation. I didn't find any to way to do so.

### Motivation 🔦

I want to navigate from one page to another using params to specify some content.

For instance, I tried the Toolpad tutorial with the dog listing, and I created another page to display more details on a specific dog. For this second page I need to use a param in the page url to specify the dog I want to display information about. To navigate to the second page, I add a button on the first page, but I can't add the needed parameter.

Janpot:
Yes 👍, makes sense. It looks like it originally was intended to be part of https://github.com/mui/mui-toolpad/issues/76 but never got implemented. We'll use this issue to keep track of it.

Issue #1579, opened 2023-01-18
From a button component, allow navigating to the current page 


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Problem: I come to a page with some URL query parameters. Now, with the click of a button, I want to basically reload the same page but without the parameters. I can't configure this because, navigating to the same page is not allowed. Explainer screenshot:
<img width="1430" alt="Screenshot 2023-01-18 at 5 54 11 PM" src="https://user-images.githubusercontent.com/92228082/213171345-3748c377-0ac9-4fab-b7b7-df1ddeed19e2.png">

Retool supports this.
<img width="1435" alt="Screenshot 2023-01-18 at 5 58 39 PM" src="https://user-images.githubusercontent.com/92228082/213171535-5fff6129-08c4-4cc6-80a9-5c20f6d896a9.png">

bytasv:
@prakhargupta1 I'm thinking would it be worth to consider extending navigation ui a bit futher and add option to specify page params?
(i.e. with retool)
![CleanShot 2023-01-18 at 14 42 45](https://user-images.githubusercontent.com/437214/213174634-56477d1f-0b53-4d32-b9f5-0765bb9feb01.png)

I feel like just navigating between pages limits a lot what can be done, I assume it should be pretty common to link to specific page with page param, i.e. from the table click on some employee link and open different page with information about that employee 🤔

PR #1876, opened 2023-04-12
Add page parameters to navigation actions

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/1800
Closes https://github.com/mui/mui-toolpad/issues/1579

Allow setting destination page parameters in navigation actions, including binding them to the current page state.

Also add documentation about navigation actions.

https://user-images.githubusercontent.com/10789765/234996002-29569385-c068-49eb-99b1-c459ad2ee92b.mov


packages/toolpad-app/src/appDom/index.ts
```diff
@@ -1114,15 +1114,15 @@ export function ref(nodeId: NodeId): NodeReference;
 export function ref(nodeId: null | undefined): null;
 export function ref(nodeId: Maybe<NodeId>): NodeReference | null;
 export function ref(nodeId: Maybe<NodeId>): NodeReference | null {
-  return nodeId ? { $$ref: nodeId } : null;
+  return nodeId ? { $ref: nodeId } : null;
```
apedroferreira, line 1117: When we inject the initial DOM state in the HTML in the `<script>` tags all the `$$ref` properties are being changed to `$ref` now with the Vite runtime... Not sure exactly why.
Can we just use `$ref` for these properties? It shouldn't even need a migration, right? I only saw it being used in a few old connections besides navigation actions.

Janpot: We currently have two formats for bindings/actions/...

One legacy one in the appDom, one new one from the new file system format. Over time we will move the whole application over to the second format

apedroferreira: Ok! How do you suggest fixing the navigation actions for now? `appDom.deRef` tries to get `$$ref` but the local DOM has the page as `$ref`.
Are the changes in this PR about that ok or do you suggest an alternative?

apedroferreira:
I've also allowed navigating to the current page to close https://github.com/mui/mui-toolpad/issues/1579 too.

Janpot (approved):

packages/toolpad-app/src/appDom/index.ts
Janpot: Since you're changing the existing format, my suggestion would be to change it to align it close to the what's in the file schema. The more we can reduce the amount of conversion needed from schema to internal representation, the better. With the north star goal to get rid of the conversion completely.
Otherwise, if it works, it works, and this can be merged

docs/data/toolpad/data-binding.md
```diff
@@ -49,3 +49,19 @@ In the example below you can see how after clicking the button, the value is cop
...
+### Navigation actions
+
+In the binding editor for event handlers it is also possible to define **navigation actions**, which can open any page in your app when triggered.
+
+You can set a navigation action in the **Navigation** tab of the binding editor for any event handler prop:
+
+<img src="/static/toolpad/docs/data-binding/bind-12.png" alt="Binding editor for navigation actions" width="800" />
+
+<!---
+@TODO: Link "any query string parameters that the destination page supports" to page parameters documentation
```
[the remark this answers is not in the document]
apedroferreira: We should document the feature of being able to set page parameters separately from navigation actions in my opinion, and then we could link to it from here.
But I wasn't sure at the time in which section the page parameters should be documented, so I added this todo. I can make sure we do this soon enough as we're working on the docs.

packages/toolpad-app/src/appDom/index.ts
apedroferreira: I forgot this comment... will take a second look and add changes if it makes sense.

PR #1880, opened 2023-04-12
Update release instructions

apedroferreira:
Small update to release instructions with some things I noticed were missing and might be useful.


Janpot (approved):

apedroferreira:
> We let github autonumber the ordered lists, but it seems it uses roman numerals for nested lists. Not the greatest, but I don't really care either way tbh.

Looks like my first commit had changed it to numbers automatically but I think roman numbers is fine.

PR #1881, opened 2023-04-12
New vite based application runtime

Janpot:
Use vite to build the Toolpad applications:

* Allows using installed node modules in Toolpad applications
* Faster production runtime that doesn't pull in every icon and every MUI component
* Allows running react in dev mode in the canvas to allow for richer error messages
* Added behind a `--viteRuntime` CLI flag to allow canary testing

Additional fixes:

* canvas overlay disappeared when the canvas iframe reloaded, added a fix for that
* store test results in CI for debuggability
* fix `/preview/pages` not redirecting to the first page
* Remove `next/image` usage
* Align some logs to the format throughout the app
* `dragNewComponentToAppCanvas` was exiting before the drag was completed, resulting in failures if the test wasn't waiting on anything afterwards
* Recreate yarn.lock

I can open a separate PR for those if you want, but most are needed for this PR as well

PR #1883, opened 2023-04-12
Update MUI monorepo

apedroferreira:
Update MUI monorepo, following approach in https://github.com/mui/mui-x/pull/8148


Janpot (approved):

apedroferreira:
Not sure what the markdown linter is complaining about, maybe it's a bug.

Janpot:
Had the same in https://github.com/mui/mui-toolpad/pull/1881, I pushed the same fix here

PR #1888, opened 2023-04-13
Fix app theming

Janpot:
Noticed this when trying to upgrade mui-public to latest toolpad

* Remove theme provider and cssbaseline from `_app` as it is shared between production app and editor


Janpot:
Merging so I can release it

test/integration/theme/index.spec.ts
```diff
@@ -22,6 +22,8 @@ test('can change between light and dark themes', async ({ page }) => {
 
   await expect(canvasBodyLocator).toHaveCSS('background-color', 'rgb(255, 255, 255)');
 
+  await editorModel.themeEditor.getByRole('button', { name: 'Add theme' }).click();
```
apedroferreira, line 25: We could also call this file `editor.spec.ts` or something if it helps tell it apart from `prod.spec.ts`

PR #1907, opened 2023-04-17
Show border on hover in interactive nodes

### LINKED ISSUE #1904: PageRow is not selectable
### Duplicates

### Latest version

### Steps to reproduce 🕹

From https://github.com/mui/mui-toolpad/issues/1875

pageRow was not visible on canvas but was there in the yaml. It made it difficult to copy and correctly map the child parent ids, indexing.

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/1904

It was not possible to see the hover border for any ancestor/parent of the selected component anymore. That is the main issue fixed by this PR. 
It was impossible to do this and keep `pointer-events: none` to be able to click through the overlay (the hover would also get disabled), so I've used a CSS clip path.

On top of this bug fix, we can also make it so that page rows and columns never show hover borders, or are selectable at all. Let me know if you think we should do that - it would not be difficult to add.

Before:

https://user-images.githubusercontent.com/10789765/232611636-a81c326a-37a6-4722-8d45-9cda3655db4b.mov

After:

https://user-images.githubusercontent.com/10789765/232611694-2c9fa7eb-b76c-4cf5-b8c4-d4fb17eb4974.mov


packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/NodeHud.tsx
```diff
@@ -174,14 +170,41 @@ export default function NodeHud({
 }: NodeHudProps) {
   const hintPosition = rect.y > HUD_HEIGHT ? HINT_POSITION_TOP : HINT_POSITION_BOTTOM;
 
+  const interactiveNodeClipPath = React.useMemo(
+    () =>
+      isInteractive && selectedNodeRect
```
apedroferreira, line 175: The clip path is a polygon around the selection rectangle, as it's not possible do define inversed clip paths (would have been useful).

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/NodeHud.tsx
```diff
@@ -44,7 +42,7 @@ const NodeHudWrapper = styled('div', {
   outline: `1px dotted ${isOutlineVisible ? theme.palette.primary[500] : 'transparent'}`,
   zIndex: 80,
   '&:hover': {
-    outline: `2px dashed ${isHoverable ? 'transparent' : theme.palette.primary[500]}`,
+    outline: `2px dashed ${isHoverable ? theme.palette.primary[500] : 'transparent'}`,
```
apedroferreira, line 45: This was wrong too, but it's unrelated.

Janpot (approved):
Not sure how hard it would be to add a test for this, but it seems fine.

This overlay is becoming quite complex. I sometimes wonder whether there's a simpler way to implement it. Instead of creating nested elements that mirror the underlying page hierarchy, read the mouse position and determine which is the closest element underneath. Then only create geometry for that rectangle.

apedroferreira:
> On top of this bug fix, we can also make it so that page rows and columns never sho

If this fails elements such as text inputs would not be interactive in the editor, so I think it's covered by the tests.

About the complexity, for this particular issue I tried a lot of solutions and this was the first one that worked - but on a larger scale when working on this logic I always kind of followed what was already there, so we might want to try to refactor it from the ground up later and make it simpler - it should be possible.

PR #1908, opened 2023-04-18
Deprecate `createQuery` and replace with `createFunction`

bharatkashyap:
- Part of #1817 

Deprecating `createQuery` in favour of `createFunction`, and favouring `function` instead of `query` in all related contexts to better follow the new nomenclature


packages/toolpad-core/src/constants.ts
```diff
@@ -1,6 +1,11 @@
 export const RUNTIME_PROP_NODE_ID = '__toolpadNodeId';
 export const RUNTIME_PROP_SLOTS = '__toolpadSlots';
 export const TOOLPAD_COMPONENT = Symbol.for('ToolpadComponent');
+export const TOOLPAD_FUNCTION = Symbol.for('ToolpadFunction');
+/**
+ * @deprecated
+ * TOOLPAD_QUERY is deprecated. Use TOOLPAD_FUNCTION instead.
```
Janpot, line 7: TOOLPAD_QUERY is niet public API, you can just rename

packages/toolpad-core/src/createQuery.tsx
```diff
@@ -23,21 +31,37 @@ export interface CreateQueryConfig<P> {
...
 export interface ToolpadQuery<C extends CreateQueryConfig<CreateQueryConfigParameters<C>>>
   extends QueryResolver<C> {
   [TOOLPAD_QUERY]: C;
 }
 
+/**
+ * @deprecated
+ * createQuery is deprecated. Use createFunction instead.
+ */
 export default function createQuery<C extends CreateQueryConfig<CreateQueryConfigParameters<C>>>(
```
Janpot, line 65: Is it possible to avoid duplication this way?

```tsx
import createFunction from './createFuction'

/**
 * @deprecated
 * createQuery is deprecated. Use createFunction instead.
 */
const createQuery = createFunction
export default createQuery
```

Janpot (approved):

Issue #1891, opened 2023-04-14
loading serverside javascript file fail


chonamkyu:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

TypeError [ERR_INVALID_MODULE_SPECIFIER]: Invalid module ".\toolpad\resources\functions.ts" is not a valid package name imported from C:\Users\nk\Documents\GitHub\toolpad016\toolpad\.generated\functions\toolpad_main.js
    at new NodeError (node:internal/errors:399:5)
    at parsePackageName (node:internal/modules/esm/resolve:817:11)
    at packageResolve (node:internal/modules/esm/resolve:840:5)
    at moduleResolve (node:internal/modules/esm/resolve:938:20)
    at defaultResolve (node:internal/modules/esm/resolve:1153:11)
    at nextResolve (node:internal/modules/esm/loader:163:28)
    at ESMLoader.resolve (node:internal/modules/esm/loader:838:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:424:18)
    at ESMLoader.import (node:internal/modules/esm/loader:525:22)
    at importModuleDynamically (node:internal/modules/cjs/loader:1186:29) {
  code: 'ERR_INVALID_MODULE_SPECIFIER'
  
 after I clicked serverside javascript, i got this error.
 
 it worked in 0.1.3 (when its name was queries.ts)
not from ^0.1.4(~0.1.5 same)

### Current behavior 😯

load list of functions I made

### Expected behavior 🤔

loaded list of functions I made

### Context 🔦

I guess the problem is file structure for windows.

you may not able to access using dot(.) (ex .\toolpad)

### Your environment 🌎

window 10

mnajdova:
I tested a new project using @mui/toolpad@0.1.6 and I can't reproduce. After upgrading existing project to 0.1.6, adding new serverside javascript is not failing with the above mentioned error, but I can't see anything in the function dropdown, although I have two functions in the `functions.ts` file. The queries (functions) was correctly migrated. cc @Janpot

Janpot:
@chonamkyu Would you be able to share your project (github or package it up in a zip file)? I'd like to take a look. Additionally I'd love to know more about your windows version and Node.js version.

apedroferreira:
I can reproduce on Windows - looking into it.

PR #1910, opened 2023-04-18
Fix Windows issues (WIP)

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/1891

Didn't find any more issues besides:
- error in typings script when running the project in development
- unable to use/import functions (reported above)


packages/toolpad-app/src/toolpadDataSources/local/server.ts
```diff
@@ -78,7 +78,7 @@ function formatCodeFrame(location: esbuild.Location): string {
 }
 
 async function createMain(): Promise<string> {
-  const relativeFunctionsFilePath = [`.`, getFunctionsFile('.')].join(path.sep);
+  const relativeFunctionsFilePath = path.resolve([`.`, getFunctionsFile('.')].join(path.sep));
```
Janpot, line 81: Why can't we use relative paths on windows? I intended it to be relative as to keep the file portable.

packages/toolpad-app/scripts/typings.mts
```diff
@@ -38,7 +38,7 @@ async function main() {
       LIBS.map(async ({ name }) => {
         const files: { filename: string; moduleId: string }[] = [];
         const resolvedPkg = require.resolve(`${name}/package.json`);
-        const pkgJson = await import(resolvedPkg, { assert: { type: 'json' } });
+        const pkgJson = await import(url.pathToFileURL(resolvedPkg).toString(), { assert: { type: 'json' } });
```
Janpot, line 41: Not sure what benefit the `import()` brings us. I'm fine with using `readJsonFile ` as well

packages/toolpad-app/src/toolpadDataSources/local/server.ts
apedroferreira: I just tried to fix the issues at first, didn't dig enough to understand them 100% yet or see if there's a better solution so I'll still do that.

packages/toolpad-app/scripts/typings.mts
apedroferreira: Okay - I found similar issues and fixes to this one in other libraries, I'll probably post them here too once the PR is ready, this seems like a weird issue.

packages/toolpad-app/src/toolpadDataSources/local/server.ts
Janpot: Oh, I think I understand now. On windows it's normalizing the path with backslashes and that makes it invalid for node.js module identifiers. I think we should do something along the lines of

```tsx
function pathToNodeImportSpecifier(importPath: string): string {
  const normalized = path.normalize(importPath).split(path.sep).join('/');
  return normalized.startsWith('/') ? normalized : `./${normalized}`;
}
```

packages/toolpad-app/scripts/typings.mts
apedroferreira: This was the issue here I think: https://github.com/nodejs/node/issues/31710

packages/toolpad-app/src/toolpadDataSources/local/server.ts
```diff
@@ -101,8 +106,8 @@ async function createMain(): Promise<string> {
     async function getResolvers() {
       if (!resolversPromise) {
         resolversPromise = (async () => {
-          const functions = await import(${JSON.stringify(
-            relativeFunctionsFilePath,
+          const functions = await import(${pathToNodeImportSpecifier(
```
Janpot, line 109: doesn't this have to be the other way around? json.stringify on the outside

apedroferreira: Maybe, it's broken on Mac now, I'm fixing it.

Janpot: json.stringify is there to add quotes around the string and correctly escape its content

apedroferreira: Yeah that was just it, I was just having some weird issues on windows VSCode.

Janpot (approved):
👌

PR #1912, opened 2023-04-19
Fix: Use `nanonid/non-secure` instead of `cuid`

### LINKED ISSUE #1737: Update `cuid` to `paralleldrive/cuid2`
Ref: Screenshot 2023-03-09 at 11 09 00 AM

and 

https://github.com/mui/mui-toolpad/pull/1700#issuecomment-1460534989

bharatkashyap:
- Closes #1737


bharatkashyap:
@Janpot https://app.circleci.com/pipelines/github/mui/mui-toolpad/6991/workflows/e354df79-c205-4539-a273-61fc4dfb4023/jobs/30387 Is it apparent to you why this test is failing - I can't see why this fails

packages/toolpad-app/src/appDom/index.ts
```diff
@@ -235,7 +237,7 @@ function assertIsType<T extends AppDomNode>(node: AppDomNode, type: T['type']):
 }
 
 function createId(): NodeId {
```
Janpot, line 239:
```tsx
import { init } from '@paralleldrive/cuid2';

const createIdInternal = init({ length: 7 })

export function createId(): NodeId {
  return createIdInternal() as NodeId;
}

File: packages/toolpad-app/src/server/localMode.ts [hunk 2]
```diff
@@ -1077,7 +1077,7 @@ async function initToolpadFolder(root: string) {
       apiVersion: 'v1',
       kind: 'page',
       spec: {
-        id: cuid.slug(),
+        id: cuid2({ length: appDom.SLUG_LENGTH })(),
```
Janpot, line 1080:
```suggestion
        id: appDom.createId(),
```

packages/toolpad-app/src/appDom/index.ts
bharatkashyap: This still fails on `yarn test` with the same error as in the CI

Janpot: yes, I wasn't proposing this as a fix for that, just as a suggestion to use the lib in a more idiomatic way.

Let's just move to `nanoid/non-secure` instead of `cuid2`. That's more than suitable for our needs and won't have these issues

Janpot (approved):

PR #1925, opened 2023-04-21
Add page name as an alias for routing in the runtime

Janpot:
* This redirects urls of the form /prod/pages/<page name> to /prod/pages/<page id>
* Additionally allow us to optimize navigations in integration tests


test/models/ToolpadRuntime.ts
```diff
@@ -13,10 +13,7 @@ export class ToolpadRuntime {
   }
 
   async gotoPage(pageName: string) {
-    await this.goto();
```
apedroferreira, line 16: I thought that this method was intentionally to test navigation using the in-app UI?
But we can also just have a test for that separately and mainly use this logic if it's more efficient.

packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -1106,18 +1106,27 @@ function RenderedPages({ pages, defaultPage }: RenderedPagesProps) {
...
+                nodeId={page.id}
+                // Make sure the page itself mounts when the route changes. This make sure all pageBindings are reinitialized
+                // during first render. Fixes https://github.com/mui/mui-toolpad/issues/1050
+                key={page.id}
+              />
+            }
+          />
+        </React.Fragment>
+      ))}
+      {pages.map((page) => (
```
apedroferreira, line 1123: We can also have this in the same mapping function as the pages for ids above if it's more efficient (I guess it's negligible though).

apedroferreira (approved):

packages/toolpad-app/src/runtime/ToolpadApp.tsx
Janpot: The idea was that id always has precedence in case there's a naming conflict between a name and an id of two different pages

apedroferreira: I guess it would still work if the route for the name comes after the respective route for the id?
Anyway it's not a big deal, just pointing out the possibility.

apedroferreira: Ah, I misunderstood it, good thinking! Thanks for clearing it up.

test/models/ToolpadRuntime.ts
Janpot: I don't think we should look at the models this way. They are not tests themselves, they are helpers that abstract common moves so that we have a central place to update when underlying UI elements change.

PR #1938, opened 2023-04-26
Make vite runtime default

Janpot:


apedroferreira (approved):
Looks good! Just a question:
When you use a fork how can I easily run your PR locally to test? Do we have any recommended way of doing it?
If we do, I can add it to the contributor docs.

mnajdova:
@apedroferreira I usually use this and run locally:
image

PR #1942, opened 2023-04-27
Added dog app arcade to quickstart section of docs

prakhargupta1:
Please check and let me know if the resolution needs to be improved. I chose the default one, when I recorded it. 

Preview: https://deploy-preview-1942--mui-toolpad-docs.netlify.app/toolpad/getting-started/quickstart/#building-your-first-application


apedroferreira:
The resolution seems fine in terms of image quality, but maybe we should use a 16:9 ratio like most screens? Otherwise in full-screen mode it had scrollbars like this on my Macbook:

Screenshot 2023-04-27 at 19 04 40

Also if I had not seen us demonstrate Arcade in the meetings before I don't think I would know that this is an interactive image that can show multiple steps... but not sure what can be improved about that?
Maybe at least some text explaining that they can change steps in the image, if nothing else can be added to the Arcade UI itself?

apedroferreira:
Oh, also would be cool to make the Arcade UI blue (the primary MUI blue we're using) instead of purple if there's an option for that.

prakhargupta1:
Thanks for these inputs!

Changed the color to blue (#0072E5). Also, found out that my Mac has aspect ratio of 16:10 hence the vertical scrollbars. My monitor is 16:9 and if that is what most screens have, I'll record on it and share again.

apedroferreira:
> Thanks for these inputs!
> 
> Changed the color to blue (#0072E5). Also, found out that my Mac has aspect ratio of 16:10 hence the vertical scrollbars. My monitor is 16:9 and if that is what most screens have, I'll record on it and share again.

Oh my bad, looks like Macbooks are 16:10 after all, not 16:9.
So not sure, I would actually optimize for Mac, so if the current images are 16:10 I think they should be good after all.

apedroferreira (approved):

Janpot:
I'm not sure about the length of the gif. It feels to me like there's a big disconnect between the video and the steps underneath. Wouldn't it be more logical to create a video per step (or a few steps that form a single task)?

prakhargupta1:
Agree, let me fix it by covering some trivial parts through video and some through interaction so that the user remains engaged.

PR #1947, opened 2023-04-28
Fix overlay grid spacing

apedroferreira:
The spacing in the themes used in the runtime VS the editor was not the same, so for example when resizing elements using the grid there was always a slight misalignment:

https://user-images.githubusercontent.com/10789765/235211223-0d13896d-38c4-4cf1-b93f-b91c6029fc5a.mov

This PR fixes the issue by using the same spacing in the runtime and editor themes:

https://user-images.githubusercontent.com/10789765/235211380-61353bdf-345d-4937-8392-b7552d2e2406.mov

### DIFF
--- a/packages/toolpad-app/src/runtime/AppThemeProvider.tsx
+++ b/packages/toolpad-app/src/runtime/AppThemeProvider.tsx
@@ -4,6 +4,8 @@ import * as colors from '@mui/material/colors';
 import * as appDom from '../appDom';
 import { AppTheme } from '../types';
 
+import { getDesignTokens } from '../theme';
+
 export function createToolpadTheme(toolpadTheme: AppTheme = {}): ThemeOptions {
   const palette: PaletteOptions = {};
   const primary = toolpadTheme['palette.primary.main'];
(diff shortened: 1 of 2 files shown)


Janpot (approved):

Janpot:
I'm thinking we could try some snapshot testing in playwright

apedroferreira:
> I'm thinking we could try some snapshot testing in playwright

Sounds good, might be a bit more effort to maintain sometimes but we can try it!
I'll merge this PR but keep a note to add a snapshot test.

packages/toolpad-app/src/runtime/AppThemeProvider.tsx
```diff
@@ -21,8 +23,11 @@ export function createToolpadTheme(toolpadTheme: AppTheme = {}): ThemeOptions {
     palette.mode = mode;
   }
 
+  const designTokens = getDesignTokens(mode || 'light');
+
   return createTheme({
     palette,
+    spacing: designTokens.spacing,
```
Janpot, line 30: @apedroferreira Doing a second review of this and I believe this is not the right approach. There are two things here:

1. The editor theme and app theme should be completely isolated from each other. 
2. If I understand correctly, shouldn't the spacing in columns be the same as the `gap` in the `PageRow` that's currently being edited? That means the spacing could potentially change based on which row is being edited and should be read from the actual application.

PR #1960, opened 2023-05-02
small improvements

Janpot:
Bringing a few things from https://github.com/mui/mui-toolpad/pull/1959

* Remove ID from component editor. It's unnecessary information as it changes very time Toolpad loads
* Make sure application name is always the same in the regression test fixtures by using a subfolder in the temporary test folder
* Move up playwright config one level


Kri83:
test

PR #1962, opened 2023-05-03
Fix preview header missing in app preview

Janpot:
The preview doesn't contain the header anymore. This PR fixes that


packages/toolpad-app/src/runtime/AppNavigation.tsx
```diff
@@ -47,7 +47,7 @@ export default function AppNavigation({ pages, isPreview = false }: AppNavigatio
         [`& .MuiDrawer-paper`]: { width: DRAWER_WIDTH, boxSizing: 'border-box' },
       }}
     >
-      {isPreview ? <Toolbar /> : null}
+      {clipped ? <Toolbar /> : null}
```
Janpot, line 50: Decoupling this component from the display mode. I don't like the drawer implementation. This spacer should be unnecessary.

apedroferreira: iirc this spacer is as recommended in https://mui.com/material-ui/react-drawer/#clipped-under-the-app-bar (Clipped under the app bar)

PR #1965, opened 2023-05-03
Add jest tests for `create-toolpad-app`

bharatkashyap:
- Closes https://github.com/mui/mui-toolpad/issues/1901

PR #1966, opened 2023-05-04
Fix console.error being called in tests

Janpot:
Tests were printing

```
  ● simple databinding

    Expected test not to call console.error().

    If the error is expected, test for it explicitly by mocking it out using jest.spyOn(console, 'error').mockImplementation() and test that the warning occurs.

    Warning: An update to Form inside a test was not wrapped in act(...).
    
    When testing, code that causes React state updates should be wrapped into act(...):
    
    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */
    
    This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
  ...
```

But still passing

Noticed while working on https://github.com/mui/mui-toolpad/pull/1965

I suspect this started happening since adding the Form component.

Related thread: https://github.com/orgs/react-hook-form/discussions/4232


packages/toolpad-app/src/runtime/ToolpadApp.spec.tsx
```diff
@@ -2,13 +2,16 @@ import * as React from 'react';
...
 import { LiveBindings, RuntimeEvents } from '@mui/toolpad-core';
+import failOnConsole from 'jest-fail-on-console';
 import ToolpadApp from './ToolpadApp';
 import * as appDom from '../appDom';
 import createRuntimeState from '../createRuntimeState';
 import { bridge } from '../canvas/ToolpadBridge';
 import { BridgeContext } from '../canvas/BridgeContext';
 import loadComponents from './loadDomComponents';
 
+failOnConsole();
```
apedroferreira, line 13: Would it be possible to set this globally for any spec file? 
This library seems to recommend using the file for `setupFilesAfterEnv` for that.

Janpot: Sure 👍

PR #1991, opened 2023-05-10
Disallow access to DOM globals in bindings

Janpot:
This makes it impossible to bind to e.g.

```tsx
JSON.stringify(document.getElementById('hello')?.textContent)
```

As a side-effect this also makes it impossible to bind to `localStorage`. This was never intended to be supported. We will provide an alternative for its use-case.


apedroferreira (approved):
Looks good, just hopefully users don't find it confusing to be able to use certain globals but not others.

Janpot:
These are not arbitrarily chosen, these are the globals defined by the ecmascript standard. We just don't support the DOM because we're not a browser.
It shouldn't be more confusing than using any other non-browser javascript runtime. Like node.js.

apedroferreira:
> These are not arbitrarily chosen, these are the globals defined by the ecmascript standard. We just don't support the DOM because we're not a browser. It shouldn't be more confusing than using any other non-browser javascript runtime. Like node.js.

Ok, even `URL` and `URLSearchParams` are available in Node, right? I think the reasoning makes sense!

Janpot:
Yes, latest node versions. I think the most important is that whatever the types say is there is supported.

Issue #1954, opened 2023-05-02
Docs to be added for following sections


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

- [x] Amend "Building UI" > "DataGrid" docs: https://mui.com/toolpad/building-ui/data-grid-component/
  - [x] deleting a row from datagrid https://github.com/mui/mui-toolpad/issues/1039#issuecomment-1520074417
  - [x] editing rows in a datagrid
  - [x] adding new rows
- [x] "Conecting to datasources" > :Queries": https://mui.com/toolpad/connecting-to-datasources/queries/
  - [x] Separate docs into two: 1 part about queries (bringing data on the page), one part about mutations (calling queries manually)
- [x] DataBinding docs: https://mui.com/toolpad/data-binding/
  - [x] Split in two, 1 part about binding state to properties, and 1 part about event handling https://github.com/mui/mui-toolpad/pull/1996
- [x] How to use theme > bring as a page under "Building UI" https://github.com/mui/mui-toolpad/pull/1993
- [x] page configuration https://github.com/mui/mui-toolpad/pull/1997
- [x] Explain Open Editor button in Toolpad, this can go inside of the documentation that explains the create custom component workflow and the function s datasource
- [x] Support importing any node modules in custom component (currently on our docs it is in wip) => update existing documentation

Don't do yet:

- [ ] How to use local storage (Do not document this for now. This was used in https://github.com/mui/mui-toolpad/pull/1933/files#diff-10d5cf0f5186831c9e402a8661dd27bdae22dd8c884c493d64c54626f12c9730R25, but this will not guarantee the DataGrid gets updated reactively. Let's come up with a Toolpad native solution to read/write state) #1988

PR #1940, opened 2023-04-26
Add Toolpad file schema reference to the docs

Janpot:
Generates documentation that describes the Toolpad configuration files

See https://deploy-preview-1940--mui-toolpad-docs.netlify.app/toolpad/schema-reference/

PR #1993, opened 2023-05-10
Add theming to docs

apedroferreira:
Part of: https://github.com/mui/mui-toolpad/issues/1954

https://deploy-preview-1993--mui-toolpad-docs.netlify.app/toolpad/building-ui/theming/


apedroferreira:
> One thing we could add is an explanation that this creates a theme file in the toolpad folder and link to the theme file schema reference of #1940

Good ideas, yeah this first version kinda just explains the UI so it's not too useful.

PR #1994, opened 2023-05-10
Deprecate typeDef

Janpot:
Simplify `createComponent` and `createFunction` by folding `typeDef` into the argument type

```json
"myParam": {
  "typedef": { "type": "string" }
}
```
becomes
```json
"myParam": {
  "type": "string"
}
```

I'm also adding a bunch of doc comments


bharatkashyap:
Do we need to write a migration for existing apps with custom components created via the old apps?

Janpot:
> Do we need to write a migration for existing apps with custom components created via the old apps?

No, the old config will be accepted for a while, users will update manually

PR #1999, opened 2023-05-11
Use vite for Toolpad editor

Janpot:
move to vite for the Toolpad editor.

This keeps the old editor around under a `TOOLPAD_LEGACY_EDITOR` flag for inline comparison. We will remove this in the next release


bharatkashyap (approved):

apedroferreira:
Just tried it, it seems way faster in development!

PR #2016, opened 2023-05-15
Make canvas column spacing the same as in the runtime

Janpot:
Isolate app theme from editor theme and use the spacing of the runtime theme for the columns overlay instead of the other way around (See point 1 of https://github.com/mui/mui-toolpad/pull/1947/files#r1193477233


apedroferreira (approved):
This should do as it's one of the options I mentioned when replying to the other PR.
https://github.com/mui/mui-toolpad/pull/1947#discussion_r1194097012

There are still more improvements I mentioned there like making grid spacing adjust to page row gaps, I'll keep them noted + create issues.

PR #2017, opened 2023-05-15
Remove hideControls and button type

Janpot:
If I understand correctly these properties allow the user to build their own reset/submit logic?

I think as an abstraction this is both too leaky and not powerful enough.

- For any button outside of a form, this property has no meaning
- The user requires knowing implementation details in order to understand how this property is to be used ("HTML type")
- The form can't be submitted from anywhere else, or on an event

I propose to remove this for the beta release and wait for user feedback

### DIFF
--- a/packages/toolpad-components/src/Button.tsx
+++ b/packages/toolpad-components/src/Button.tsx
@@ -56,12 +56,6 @@ export default createComponent(Button, {
       helperText: 'Whether the button is disabled.',
       type: 'boolean',
     },
-    type: {
-      helperText: 'Button HTML type',
-      type: 'string',
-      enum: ['button', 'submit', 'reset'],
-      default: 'button',
-    },
     sx: {
       helperText: SX_PROP_HELPER_TEXT,
       type: 'object',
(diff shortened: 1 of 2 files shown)


apedroferreira (approved):
> If I understand correctly these properties allow the user to build their own reset/submit logic?

Yes, and to be able to customize the look of their form a bit more.

I understand your points so I'm fine with the changes. Maybe soon enough we should discuss or define better what exact problems we're trying to solve with the Form component, and how it should work.

Janpot:
perhaps for this form we could expose a reset method, then you could call that from an event from any button?

apedroferreira:
> perhaps for this form we could expose a reset method, then you could call that from an event from any button?

That could work, i'll try it in a PR. I guess we would do the same for submitting the form too.

Janpot:
> I guess we would do the same for submitting the form too.

yeah, I was thinking people could just use the `form.value` in `onClick` handler for submit logic. But then what would happen with the form `onSubmit`. So yes, it may make sense

PR #2023, opened 2023-05-16
Upgrade monorepo

bharatkashyap:


bharatkashyap:
> [mui/material-ui#37223](https://github.com/mui/material-ui/pull/37223)
> 
> Looks like these started failing because the rule got disabled, not because the issues are resolved. I don't think we want this rule disabled

Do you mean keep the rules as-is and merge with a failing CI on `test:lint`?

Janpot:
> Do you mean keep the rules as-is and merge with a failing CI on `test:lint`?

No, I think we need to look into why the rule got disabled in the first place and re-enable it for our code base. It looks like it got disabled in the top level config, but re-enabled in an override for performance reasons. We're not puling in the overrides, only the top-level config. Perhaps we can add a similar override to re-enable the rule?

.eslintrc.js
```diff
@@ -14,6 +14,7 @@ module.exports = {
   rules: {
     ...baseline.rules,
     'import/prefer-default-export': ['off'],
+    'import/no-cycle': ['error', { ignoreExternal: true }],
```
Janpot, line 17: Perhaps we can use an override just like in https://github.com/mui/material-ui/pull/37223/files so that we benefit from the perf boost as well?

PR #2026, opened 2023-05-16
Fix overlay grid spacing

apedroferreira:
@Janpot I noticed that the overlay grid horizontal paddings were misaligned now since the spacing fix, because the `Container` the grid is in was using different spacing than the `Container` the runtime UI is wrapped in.
It's difficult to make the containers match because their padding changes with breakpoints and such.

What do you think about this solution? Just wrapping the grid in a theme provider.

### DIFF
--- a/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/OverlayGrid.tsx
+++ b/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/OverlayGrid.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Grid, Container, styled } from '@mui/material';
+import { Grid, Container, styled, ThemeProvider } from '@mui/material';
 import invariant from 'invariant';
 import { createToolpadAppTheme } from '../../../../runtime/AppThemeProvider';
 import { useDom } from '../../../AppState';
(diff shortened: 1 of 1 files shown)


Janpot:
I feel like it would be simpler to get rid of `Grid` and not rely on MUI theming at all. Just building up these columns using `divs` and styling.

apedroferreira:
> I feel like it would be simpler to get rid of `Grid` and not rely on MUI theming at all. Just building up these columns using `divs` and styling. Not sure

We would still need to copy the exact breakpoints that the MUI `Container` in the runtime uses for the spacing to match...
I feel like this solution is ok because even though the Grid is in the editor it should use the same theming as the UI.

We can do it manually if you think it's better but it might be more brittle.

Janpot:
It feels wrong to wrap the theme provider around editor UI, the theme affects more than just spacing. Can't we just set the spacing on the container the same way it was done for the column?

apedroferreira:
I tried a different approach, it's not as pretty but it works and doesn't mess with the overall theme.
Couldn't think of a better solution, I don't think the spacing can be set just for a `Container` component and the breakpoints it uses under the hood are also connected with theming so it would be difficult to make the edges match using pixel units.

apedroferreira:
Using absolute position instead of `margin:auto` just worked after all - I think this should be good now.

Janpot (approved):
yes!

PR #2034, opened 2023-05-17
Ignore warning for first-child pseudo selector in Markdown component

Janpot:
Fix warnings in the markdown component:

```
The pseudo class ":first-child" is potentially unsafe when doing server-side rendering. Try changing it to ":first-of-type". 
```

We're not SSRing this so we can safely disable, also adding a test to catch the warning in the future


Janpot:
Doesn't work

apedroferreira (approved):

Issue #2039, opened 2023-05-18
Tooltip is shown even when sx modal is being shown


ManoloTonto1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

Steps:

1. Add a TextField component to the surface 
2. Access the SX menu 

### Current behavior 😯

As shown in the image below, the tooltip is still showing while In the SX editor. This is very annoying because you cannot access the save button.
![image](https://github.com/mui/mui-toolpad/assets/74105210/cd1a8ad1-7906-46fe-9fae-e3e71c19149c)

### Expected behavior 🤔

once the sx button has been clicked the tooltip should not be visible. 

### Your environment 🌎

Toolpad version: 0.1.12
Browser: Firefox 113.0.1
User Agent: "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:109.0) Gecko/20100101 Firefox/113.0"

Janpot:
> once the sx button has been clicked the tooltip should not be visible.

Agreed, this was also reported in https://github.com/mui/mui-toolpad/issues/787#issuecomment-1400286601

PR #2041, opened 2023-05-18
Fix property control tooltips showing over dialogs

Janpot:
Closes https://github.com/mui/mui-toolpad/issues/2039

Put the tooltip on the inner control instead of wrapping the whole component with it to avoid it applying to the dialog as well. Not sure whether there's a better way to do this. 

MUI components themselves also don't hide the tooltip when e.g. a Select control is open


apedroferreira (approved):
LGTM!

Issue #2033, opened 2023-05-17
Anomaly in form component or canvas


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

I wanted to add another button in the form component, but realized that it is not possible.
When I was trying it, somehow the button got in a weird state from which I can't delete it or move it. When I try to do any of this, it deletes the form/page.

https://github.com/mui/mui-toolpad/assets/92228082/baf52975-0319-4fe9-aa46-d7675f35f445

The stuck app can be found here: https://github.com/mui/mui-toolpad/pull/1445
Examples > showcase > adminApp

### Your environment 🌎

v0.1.11

apedroferreira:
Damn, lots of weirdness there, thanks for providing the info, I'll take a look.

PR #2042, opened 2023-05-18
Fix orphaned node delete order

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/2033

When some nested elements were moved out of a parent or deleted, the parent 2 levels above was being adjusted before the parent 1 level above, which could result in moving the non-deleted elements to a parent that ended up being deleted. Not easy to explain.

### DIFF
--- a/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/RenderOverlay.tsx
+++ b/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/RenderOverlay.tsx
@@ -203,6 +203,32 @@ function deleteOrphanedLayoutNodes(
           moveTargetNodeId !== parentParent.id &&
           moveTargetNodeId !== lastContainerChild.id
         ) {
+          if (
+            moveTargetNodeId !== parent.id &&
+            moveTargetNodeId !== lastContainerChild.id &&
+            isPageLayoutComponent(parentParent)
+          ) {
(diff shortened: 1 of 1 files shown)


packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/RenderOverlay.tsx
```diff
@@ -203,51 +203,51 @@ function deleteOrphanedLayoutNodes(
...
+                'layout',
+                'columnSize',
+                parent.layout?.columnSize || appDom.createConst(1),
+              );
+            }
+
+            orphanedLayoutNodeIds = [...orphanedLayoutNodeIds, parent.id];
+          }
 
-        if (
```
apedroferreira, line 227: I just moved this `if` block to above, the diff makes it seems like more than that for some reason.

Janpot:
Would it be complex to add this case to a test? Can this be added to an existing test file?

apedroferreira:
> Would it be complex to add this case to a test? Can this be added to an existing test file?

This seems to be a rare case, I'm not sure how to reproduce it yet but I can try, it should be possible.
We can include some nested deletes in the test for deleting nodes, I'll give it a look.

We should probably merge this already and then I can add the tests separately just to get the fix in?

Janpot (approved):

PR #2046, opened 2023-05-19
[docs] Fix casing

oliviertassinari:


Janpot:
👍 thanks. Almost feels like we should be able to lint for this.I see X uses [vale](https://github.com/mui/mui-x/actions/runs/5043072254/jobs/9044405779). Worth looking deeper into this?

oliviertassinari:
Yeah, I think that Vale could make sense, at minimum to scale from MUI X to https://github.com/mui/material-ui.

PR #2054, opened 2023-05-22
pin toolpad packages versions again

Janpot:
Also make sure lerna won't reset it to `^` again


bharatkashyap (approved):

apedroferreira:
Oops, I hadn't noticed this, thanks for re-fixing!

PR #2060, opened 2023-05-23
Fix manual queries losing parameters when called

### LINKED ISSUE #2019: Parameter bindings are lost when manually calling a query
When defining a query that is to be called manually, one can bind the parameters via UI, like so:

However, these are required to be re-written when calling this query, for instance in a Button `onClick` binding

The expectation is that I need not have to pass in the parameters in the code, and simply write `<query-name>.call()`

bharatkashyap:
- Closes #2019 
- Visual demonstration 
### Before

https://github.com/mui/mui-toolpad/assets/19550456/33d7e121-e372-4ab5-8b01-c33096ac8e1c

### After

https://github.com/mui/mui-toolpad/assets/19550456/89d0a8ea-1331-4ff3-8b81-cb3d46528f05

### DIFF
--- a/packages/toolpad-app/src/runtime/ToolpadApp.tsx
+++ b/packages/toolpad-app/src/runtime/ToolpadApp.tsx
@@ -579,19 +579,15 @@ function flattenNestedBindables(
 
 /**
  * Returns an object with the resolved values of the bindables.
- * @param bindings
  * Example bindings:
  * {
  *  'nodeId.params.order': { error: undefined, loading: false, value: { "OrderID": "" } },
  * }
- * @param bindingId
  * Example bindingId: 'nodeId.params'
- * @param params
  * Example params:
  * {
  *  ["order", { type: 'jsExpression', value: 'form.value\n' }]
  * }
- * @returns {Record<string, unknown>}
  * Example result:
  * {
  * order: { "OrderID": "" }
--- a/packages/toolpad-app/src/runtime/ToolpadApp.tsx
+++ b/packages/toolpad-app/src/runtime/ToolpadApp.tsx
@@ -577,6 +577,31 @@ function flattenNestedBindables(
   );
 }
 
+/**
+ * Returns an object with the resolved values of the bindables.
+ * Example:
+ * { foo: { bar: { type: 'const', value:1 } }, baz: [{ type: 'jsExpression', value: 'quux' }] }
+ * =>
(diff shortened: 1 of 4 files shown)


packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -577,6 +577,27 @@ function flattenNestedBindables(
   );
 }
 
+/**
```
apedroferreira, line 580: If this method is just for internal usage do we really need these annotations?
I feel like the Typescript types could be explanatory enough.
But of course it also doesn't hurt to have this, besides it just taking a few more lines.

apedroferreira (approved):

Janpot:
Can you add a test case to either https://github.com/mui/mui-toolpad/blob/master/test/integration/backend-basic/index.spec.ts o  https://github.com/mui/mui-toolpad/blob/master/test/integration/rest-basic/index.spec.ts ?
Just a button that triggers a query with bound parameters and displaying the output

test/integration/backend-basic/fixture/toolpad/resources/functions.ts
```diff
@@ -41,3 +41,15 @@ export const withParams = createFunction(
...
 );
+
+export const manualQueryWithParams = createFunction(
+  async ({ parameters }) => {
+    return { parameters };
+  },
+  {
+    parameters: {
+      foo: { type: 'string' },
+      // __NEW_PARAMETER__
```
Janpot, line 52:
```suggestion
```

test/integration/backend-basic/fixture/toolpad/pages/page1/page.yml
```diff
@@ -93,3 +120,13 @@ spec:
...
+    - name: manualQuery
+      mode: mutation
+      query:
+        function: manualQueryWithParams
+        kind: local
+      parameters:
+        - name: foo
+          value:
+            $$jsExpression: |
+              text.value.split(" ")?.[1]
```
Janpot, line 132: Let's keep the binding simple? Any js expression will do.

```suggestion
              'checksum'
```

Janpot (approved):
Looking good! I'd scope down the fixture a bit to focus on passing the binding only. Complexer binding setups we already test in another integration test.

Approving already

PR #2062, opened 2023-05-24
Improve functions creation UX

Janpot:
Closes https://github.com/mui/mui-toolpad/issues/2063

This PR proposes an updated UX for custom functions:
- 1 function per file as the default export, this will make sure we can isolate them better. e.g. one syntax error doesn't break all functions. **This doesn't work yet**
- "open editor" is only available when an existing function is selected, the corresponding file will open
- The autocomplete is creatable, clicking the create option initializes a new function file with the desired name
- exports from `functions.ts` are still supported for now to avoid breaking change.

Demo:

https://github.com/mui/mui-toolpad/assets/2109932/f54b89ac-76f4-4f5a-b26b-79412c8e2e87

### How should it work

**primary**

- One dropdown with functions grouped by file, see https://mui.com/material-ui/react-autocomplete/#grouped
- Creatable option opens popup, popup contains text field prefilled with creatable option, it automatically creates a new file with that name and a default export with the function

**secondary**

- Category header has "+" button to add a new function into that specific file.


prakhargupta1:
Looks good!
I think we should show **Create new function:** right after the user clicks on the drop-down as currently, it is unintuitive that it will appear when the user starts writing the name of the function.

PR #2069, opened 2023-05-25
Avoid restarting vite devserver when env file changes

Janpot:
## TODO

- [ ] test

### DIFF
--- a/packages/toolpad-app/src/server/toolpadAppBuilder.ts
+++ b/packages/toolpad-app/src/server/toolpadAppBuilder.ts
@@ -218,6 +218,7 @@ export function createViteConfig({
         },
       },
     },
+    envFile: false,
     resolve: {
       alias: [
         {


Janpot:
@apedroferreira There's a minor issue in the test that I've seen before in other tests, is that queries don't reload when the env file changes. With the current functionality this is not a huge problem because to change the env file you'd have to blur the window and window focus causes a revalidation for all queries. This currently doesn't happen in playwright and they don't support hiding/showing the tab yet. We'll have to fix this issue at some point when we start altering .env files from the UI.
second to note is that the env file gets loaded from disk on every query call, this is wasteful and I'd like to centralize all env file logic for functions and rest queries so that it loads only once and notifies on changes. Will look into this after merging https://github.com/mui/mui-toolpad/pull/2062

apedroferreira:
> @apedroferreira There's a minor issue in the test that I've seen before in other tests, is that queries in the canvas don't reload when the env file changes. With the current functionality this is not a huge problem because to change the env file you'd have to open your code editor and browser window focus causes a revalidation for all queries. Playwright doesn't support an API to simulate hiding/showing the browser tab yet. We'll have to fix this issue at some point though when we start altering .env files from the UI. second to note is that the env file gets loaded from disk on every query call, this is wasteful and I'd like to centralize all env file logic for functions and rest queries so that it loads only once and notifies on changes. Will look into this after merging #2062

Ok, we should look into loading the env data from memory and also making sure it's always up to date in queries no matter what. Thanks for looking into it, but let me know if I should create an issue for this / look into it instead if it helps!

> Playwright doesn't support an API to simulate hiding/showing the browser tab yet.

I have this in the `rest-basic` tests:

```
 // Make sure switching tabs does not close query editor
  const newTab = await context.newPage();
  await newTab.bringToFront();
  await page.bringToFront();
  await expect(newQueryEditor).toBeVisible();
  ```
  
  It might not do **exactly** what a real browser does though, if I remember correctly.

PR #2073, opened 2023-05-26
Improve error handling in image and datagrid

Janpot:
Make sure we correctly display image errors:
- when loading the image fails, e.g. wrong url
- when incoming datasource fails

Also reusing the error display for datagrid to improve consistency

Screenshot 2023-05-26 at 11 38 30


packages/toolpad-components/src/components/ErrorOverlay.tsx
```diff
@@ -0,0 +1,30 @@
...
+interface ErrorOverlayProps {
+  error?: unknown;
+}
+
+export default function ErrorOverlay({ error }: ErrorOverlayProps) {
+  const errMEssage = error ? errorFrom(error).message : null;
+  return (
+    <OverlayRoot sx={{ display: error ? undefined : 'none' }}>
+      <Typography>Error</Typography>
+      <Typography variant="body2">{errMEssage}</Typography>
```
apedroferreira, line 27: What if there's no error message? Should we have a fallback generic error message like `There was an unexpected error`?

packages/toolpad-components/src/components/ErrorOverlay.tsx
```diff
@@ -0,0 +1,30 @@
...
+  borderRadius: theme.shape.borderRadius,
+  borderColor: theme.palette.divider,
+}));
+
+interface ErrorOverlayProps {
+  error?: unknown;
+}
+
+export default function ErrorOverlay({ error }: ErrorOverlayProps) {
+  const errMEssage = error ? errorFrom(error).message : null;
```
apedroferreira, line 23: `errMEssage` => `errMessage`

apedroferreira (approved):
Good change, just left some very minor comments.

Also could we add some red and an "error" icon to the overlay? Maybe a red border would work, not sure though.

packages/toolpad-components/src/components/ErrorOverlay.tsx
Janpot: The `errorFrom` function ensures a valid error

apedroferreira: Ah, from a quick look at the logic I thought `errMessage` could be `null`, but I guess that only happens if there's no error... So it's fine then.
At first thought I would make the `error` prop be required for this component and only render the component if there is an error to pass as a prop. But if you're using the approach with the CSS `display` for a different reason it could be fine too.

PR #2080, opened 2023-05-26
Allow inlining the proptype jsonschema

Janpot:
Currently these were used in the JSON editor for validation. By inlining them we could potentially use them to construct a property validator without needing to dereference a url


apedroferreira (approved):
This should be useful for charts too, and I like that it's more straightforward to just write these schemas in the component itself.

Janpot:
this was what i initially wanted to do but i didn't realize at the time that createObjectUrl would work

Issue #789, opened 2022-08-13
Support Chart component


oliviertassinari:
keywords: "Cannot read properties of null"

### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Make it possible to display charts

### Benchmark 🌈

- https://www.notion.so/mui-org/Chart-component-7b30cff4fdab4d2e8adbe0c796821f45
- https://github.com/mui/mui-x/issues/1408
- https://retool.com/components/chart (Integrated ploty.js)
- https://docs.appsmith.com/reference/widgets/chart (Integrated [fusioncharts](https://www.fusioncharts.com/))
- Superblocks (ploty.js)
- Openblocks (echarts.js)
- Tooljet (ploty.js)

### Motivation 🔦

I would like to plot this data https://api.ossinsight.io/q/analyze-issue-open-to-first-responded?repoId=23083156 which is about the median time to get feedback on issues in https://github.com/mui/material-ui. It's meant to replace https://mui-repository-health-dashboard.vercel.app/. So far, I have tried:

- **reaviz** ❌ https://master--toolpad.mui.com/_toolpad/app/cl42qt2jx01249xnr14xjce8b/editor/codeComponents/cl6rsvyg400003b6gwlzpovht. Fail with 

<img width="579" alt="Screenshot 2022-08-13 at 13 37 00" src="https://user-images.githubusercontent.com/3165635/184491202-2adce675-e20f-436f-90f5-8473c3247643.png">

- **victory** ❌ https://master--toolpad.mui.com/_toolpad/app/cl42qt2jx01249xnr14xjce8b/editor/codeComponents/cl6rsxp8q00023b6gu7dfcvg6. Fail with 

<img width="579" alt="Screenshot 2022-08-13 at 13 37 00" src="https://user-images.githubusercontent.com/3165635/184491202-2adce675-e20f-436f-90f5-8473c3247643.png">

- **react-vis** ✅ https://master--toolpad.mui.com/_toolpad/app/cl42qt2jx01249xnr14xjce8b/editor/codeComponents/cl6rtolju00003b6gcl9wrueb

<img width="314" alt="Screenshot 2022-08-13 at 13 36 32" src="https://user-images.githubusercontent.com/3165635/184491006-c7944845-d2de-48fc-9d44-2fca2da6898b.png">

<details>
  <summary>Source</summary>

```jsx
import * as React from "react";
import { createComponent } from "@mui/toolpad-core";
[...]

PR #2081, opened 2023-05-26
Chart component

apedroferreira:
New chart component. Allows for composing line chart, bar chart, area chart and scatter chart, with independent data definitions for each chart.
Uses `recharts`, can gradually be moved to the new MUI X charts library, let's create an issue for that once this is merged.

Closes https://github.com/mui/mui-toolpad/issues/789
Related discussion: https://github.com/mui/mui-toolpad/discussions/2005

https://github.com/mui/mui-toolpad/assets/10789765/2d492937-c22c-4499-8d62-68771ef5789f


packages/toolpad-app/src/runtime/constants.ts
```diff
@@ -1,3 +1,3 @@
 export const HTML_ID_EDITOR_OVERLAY = 'editor-overlay';
 
-export const NON_BINDABLE_CONTROL_TYPES = ['GridColumns'];
+export const NON_BINDABLE_CONTROL_TYPES = ['GridColumns', 'ChartData'];
```
Janpot, line 3: This feels like a leaky abstraction. Is there any reason we couldn't put a `bindable` property on the [`ArgControlSpec`](https://github.com/mui/mui-toolpad/blob/564108a247a30b6a570974308732020a6e1d479b/packages/toolpad-core/src/types.ts#L175C18-L175C32)?

_As a general remark, it's best to avoid `[].includes()` for lookup, better to use a `Set` instead for this purpose._

packages/toolpad-app/src/runtime/evalJsBindings.ts
```diff
@@ -249,12 +249,13 @@ export default function evalJsBindings(
...
-          const nestedBindingResultValue = results[nestedBindingId]?.value;
-          if (nestedBindingResultValue) {
+          const nestedBindingResult = results[nestedBindingId];
+
+          if (nestedBindingResult) {
             bindingResult = updatePath(
               bindingResult,
               `value.${nestedBindingId.replace(bindingId, '')}`,
-              nestedBindingResultValue,
+              nestedBindingResult.value,
```
Janpot, line 258: Is this fixing a bug? If so, if so it could be a good on to be covered by a test (if it isn't already).

packages/toolpad-app/src/toolpad/propertyControls/ChartData.tsx
```diff
@@ -0,0 +1,247 @@
...
+import { useBrowserJsRuntime } from '@mui/toolpad-core/jsBrowserRuntime';
+import DeleteIcon from '@mui/icons-material/Delete';
+import { BindableAttrValue } from '@mui/toolpad-core';
+import * as appDom from '../../appDom';
+import type { EditorProps } from '../../types';
+import PropertyControl from '../../components/PropertyControl';
+import { usePageEditorState } from '../AppEditor/PageEditor/PageEditorProvider';
+import { useDom, useDomApi } from '../AppState';
+// eslint-disable-next-line import/no-cycle
+import BindableEditor from '../AppEditor/PageEditor/BindableEditor';
```
Janpot, line 27: I wonder whether we can solve all these cycles by creating a react context very high up in the toolpad app that holds a registry of all available propertyControls.

packages/toolpad-components/src/Chart.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+function Chart({ data = [], height, sx }: ChartProps) {
+  const xValues = React.useMemo(
+    () =>
+      uniq(
+        data
+          .map((dataSeries) =>
+            (dataSeries.data || []).map((dataSeriesPoint) => dataSeriesPoint[dataSeries.xKey]),
+          )
+          .flat()
```
Janpot, line 50: Why does this need to be flattened? Bad user input? Maybe it's better to make sure they're aware of this instead of guessing that this is what they intended?

packages/toolpad-app/src/toolpad/propertyControls/ChartData.tsx
apedroferreira: Yeah it's becoming a bigger issue when we add controls with bindable props, I was focused on functionality first but will see if I can find a way to fix this.

packages/toolpad-app/src/runtime/constants.ts
apedroferreira: Probably no reason, it does sound better so will try it.

packages/toolpad-components/src/Chart.tsx
apedroferreira: I'm flattening to have all the unique x values from every data series into a single array, as the x-axis is shared between all charts and it helps to have it for some things such as setting the domain of the x-axis that all the charts share.
If you see this affecting the user experience in any particular scenario I can look into different approaches.

Janpot: Ah, right I misunderstood. You can also just `.flatMap` and avoid creating an array of arrays altogether

packages/toolpad-app/src/runtime/constants.ts
Janpot: Also fine in a separate PR, it just caught my eye

packages/toolpad-app/src/toolpad/propertyControls/ChartData.tsx
apedroferreira: Turns out the cyclical dependency with prop controls was simpler than the other one with the binding editors - so passing the prop controls as context does fix the issue, nice one! I had thought it was part of the same problem.

I put the list of prop controls to pass to the context in the `ComponentEditor` component - let me know if you think this should be done differently in terms of file structure.

Janpot (approved):
looking good

Issue #2099, opened 2023-05-31
Issues with environment variables


Janpot:
Trying out new environment variable support. Having a few issues:

- [x] I can't enter an environment variable in a binding unless it's defined in the `.env` file. 
- [x] Toolpad is reading the .env file to interpolate environment variables instead of reading them from the environment. This is causing issues like https://github.com/mui/mui-private/issues/267#issuecomment-1564361133. The .env file is a convenience to locally store variables. We will also use it for suggestions and linting, but by no means is it the single source of thrut. It's just one way of setting up the variables. For reading actual values use `process.env`.
- [ ] The `$$env` method allows us to up front determine some variables the application expects to have access to. We should show a warning on startup when any of the used variables are not defined.

apedroferreira:
Ah yes, environment variables aren't going to work in production this way, we need to read from `process.env`... 
I'll add a fix for 1 and 2 first separately before the others to get it live asap.

oliviertassinari:
> 1. I can't enter an environment variable in a binding unless it's defined in the .env file.

I'm not sure it's a bad outcome. I like that it gives a clear signal of what are the .env variables available. It gives the confidence that the latest env value you added was correctly loaded.

Now, I'm not sure about the "Environment variable" UX: 

<img width="349" alt="Screenshot 2023-06-01 at 10 30 45" src="https://github.com/mui/mui-toolpad/assets/3165635/9358de7b-7e1d-49c7-a32b-28ea7246c0c8">

It might be better as a normal binding value, but making it clear that it's only available when running the queries server-side.

> 2. Toolpad is reading the .env file

Ahhh, now I get why it doesn't work.

Janpot:
> There are advantages. It gives a clear signal of what are the .env variables available. It gives the confidence that the latest env value you added was correctly loaded.

The way I'm thinking about it currently is sort of two-way. 
* We make sure we can easily statically analyze a toolpad project to see which environment variables it expects. This allows us to warn in dev and prod when one is missing. 
* Additionally in dev, we could offer to UI to automatically create it when we detect it's not in the env file. for example: 
  * We could show a google oauthflow and automatically store the token in the .env file in the end
  * We could show a basic auth user+password and save the header in .env
  * We could show a db connection UI with a "test connection" button, and upon saving store the connection string in the .env file.
* In the other direction, we can offer autocomplete support for variables that are already defined in .env file.

> It might be better as a normal binding value, but making it clear that it's only available when running the logic server-side.

[...]

PR #2109, opened 2023-06-01
Allow arbitrary env variables in bindings, use process.env

apedroferreira:
- Allow for typing in any environment variable name in environment variable bindings, even if it doesn't exist
- Correctly retrieve environment variable value from `process.env`
- Refresh `process.env` when env file changes

https://github.com/mui/mui-toolpad/assets/10789765/ea9eff13-a372-4210-9d3f-fea205b0e3bc

Fixes 1 and 2 in https://github.com/mui/mui-toolpad/issues/2099.


.circleci/config.yml
```diff
@@ -166,6 +166,7 @@ jobs:
           name: Tests real browser
           environment:
             HTTPBIN_BASEURL: 'http://localhost/'
+            TEST_ENV_VAR: 'foo'
```
Janpot, line 169: 👍 but I'd keep it local like in https://github.com/mui/mui-toolpad/pull/2103/files#diff-f16a8b4fc36e8dc38f1edc79e02791f23fed2ae1be7d7f7b1b015c1a05aec6ebR23-R25

apedroferreira: Ah so there is a way, I was looking into this right now, thanks!

test/integration/rest-basic/index.spec.ts
```diff
@@ -41,14 +44,6 @@ test('rest basics', async ({ page, context, localApp }) => {
 
   await expect(page.getByText('query4 authorization: foo')).toBeVisible();
 
-  const envFilePath = path.resolve(localApp.dir, './.env');
```
Janpot, line 44: Why was this removed? This is testing whether the editor detects changes to the env file correctly

Janpot (changes requested):
We should keep the test with `.env` file and add another one with a different variable for the behavior of environemnt variables not in the `.env` file.

It's important because we need to test that changes to the file correctly propagate

test/integration/rest-basic/index.spec.ts
apedroferreira: Dotenv wasn't working with tests locally, I'll try to make it work though so we can test like this.

Janpot: Both methods need to work. The key is to add a second env var so that the old test remains as it is. Like it's done for https://github.com/mui/mui-toolpad/pull/2103/files#diff-f16a8b4fc36e8dc38f1edc79e02791f23fed2ae1be7d7f7b1b015c1a05aec6ebR23-R25

packages/toolpad-app/src/server/EnvManager.ts
```diff
@@ -46,6 +46,8 @@ export default class EnvManager {
     envFileWatcher.on('all', async () => {
       // Invalidate cache
       this.values = undefined;
+
+      dotenv.config({ override: true });
```
apedroferreira, line 50: This here refreshes `process.env` to match the env file when it changes.

Janpot (approved):
Not a huge fan of updating `process.env` at runtime. Most code in the repo and in libraries expect `process.env` to be static throughout the app lifecycle (e.g. look at our src/config.ts, those values don't update when process.env changes). It'd probably be more maintainable to construct the env explicitly when interpolating `$$env` bindings.
Happy to merge though if you disagree.

apedroferreira:
> Not a huge fan of updating process.env at runtime. Most code in the repo and in libraries expect process.env to be static throughout the app lifecycle (e.g. look at our src/config.ts, those values don't update when process.env changes). It'd probably be more maintainable to construct the env explicitly when interpolating $$env bindings.

So you'd prefer we read the `.env` file manually in development?
Or do you mean we should read the whole yaml files and list every `$$env` binding that is there?
I thought this solution would be more solid but if there's other things depending on `process.env` being static I don't think the solution is worse.

apedroferreira:
I might also merge now but then will add that change asap, just so that the feature works.

PR #2114, opened 2023-06-02
Allow nested bindings

apedroferreira:
Allow for nested bindings, which is needed for example to have the chart component work as we intend to (with nested data series that can be bound to different datasources).

Extracted from https://github.com/mui/mui-toolpad/pull/2081

Tried many different approaches, this one:
- works universally for nested bindings inside objects or arrays
- does not add anything new to the schema, we can just nest bindings in the yaml files

Open for feedback of course!
Also not sure if I might have missed some edge cases or possible bugs with the parts of the code I altered.


packages/toolpad-app/src/bindings.ts
```diff
@@ -0,0 +1,26 @@
+import { BindableAttrValue } from '@mui/toolpad-core';
+
+export function toBindable<V>(
```
Janpot, line 3: I'd like to avoid sprinkling these functions around the whole code base and essentially have two formats of bindings used together. In the appDom, can we change the interface of a binding from
```ts
{
  type: 'jsExpression'
  value: string
}
```
to the interface used by the file schema
```ts
{
  $$jsExpression: string
}
```
so that we have only one format to worry about in the whole codebase.

apedroferreira: Yeah, I wasn't sure we should make this change already, if so it sounds great!

packages/toolpad-app/src/runtime/ToolpadApp.spec.tsx
```diff
@@ -155,7 +155,7 @@ test(`Databinding errors`, async () => {
       selfReferencing = appDom.createNode(dom, 'element', {
         name: 'selfReferencing',
         attributes: { component: 'Text' },
-        props: { value: { $$jsExpressione: 'selfReferencing.value' } },
```
apedroferreira: This typo wasn't failing the test before for some reason, but these changes made me find it.

apedroferreira:
The preview and production runtime seem to turn all `$$jsExpression` DOM bindings into `$jsExpression`... 
So as I've done before for refs, it seems that using `$jsExpression` instead of `$$jsExpression` in the works better... I'm proposing that change here with a migration in pages from API v1 from v2, feel free to provide feedback.

Janpot:
Fixing the double dollar problem in https://github.com/mui/mui-toolpad/pull/2165. We do need to think about a good migration strategy going forward, but I'm not sure we should mix that in here if we don't have to, there's a lot to consider.

apedroferreira:
I've added changing bindings from
```ts
{
  type: 'jsExpression'
  value: string
}
```
to
```ts
{
  $$jsExpression: string
}
```
as proposed.

It's a lot of changes, hopefully nothing else is broken, I could not find anything else so far.

Choices made that could deserve some attention/discussion:

- Used `as` assertions a lot sometimes because I could not find a better way to infer the type of a bindable from the key name. (for example, `bindable?.$$jsExpression` will throw an error if we don't assert the `BindableAttrValue` type to `JsExpressionAttrValue`. Maybe there's a better way to do this.)
- To make it simpler and cleaner to determine what kind of bindable is being used or retrieve its value, I've created `getBindingValue` and `getBindingType` utilities.
- Added an appDom migration (v7) to address these changes, but because the new bindable types are incompatible with the old ones, made the types in old migrations work by keeping some old types in `migrations/types/v7Legacy` to be used in old migrations.

apedroferreira:
As for testing nested bindings we don't really have a use case for nested bindings yet so I think we can cover it with the chart component tests once that component is added.

test/integration/bindings/fixture-navigation/toolpad/pages/page1/page.yml
```diff
@@ -12,7 +12,8 @@ spec:
           props:
             onClick:
               $$navigationAction:
-                page: au03sef
+                page:
+                  $ref: au03sef
```
Janpot, line 16: This doesn't seem right. `page` is supposed to be a string according to the schema: https://github.com/mui/mui-toolpad/blob/ae82d58325f769644a7fb0f2f0b6b4683e75d285/packages/toolpad-app/src/server/schema.ts#L58
how is this passing the validation? I'd expect a test to break because of this

packages/toolpad-app/src/server/localMode.ts
Janpot, line 1: yass!

packages/toolpad-app/src/runtime/evalJsBindings.ts
```diff
@@ -239,11 +240,41 @@ export default function evalJsBindings(
...
+    const mergeNestedBindings = (value: unknown, parentBindingId: string) => {
+      if (value && typeof value === 'object') {
+        for (const nestedPropName of Object.keys(value)) {
+          const nestedBindingId = `${parentBindingId}${
+            Array.isArray(value) ? `[${nestedPropName}]` : `.${nestedPropName}`
+          }`;
+
+          const nestedBindingResultValue = results[nestedBindingId]?.value;
+          if (nestedBindingResultValue) {
+            bindingResult = updatePath(
```
Janpot, line 254: I'd prefer this written with nested `update` calls as we do everywhere else. It's slightly more verbose but it's more type-safe, typescript will yell at us if we update the `AppDom` but forget to update this statement.
It's probably slightly more performant as well and it doesn't create clones if it doesn't need to.

We can always start using `immer` to simplify these sort of updates ina  type safe way.

Janpot:
Looking good, just this one [comment](https://github.com/mui/mui-toolpad/pull/2114#discussion_r1229229234) seems a blocker to me

test/integration/bindings/fixture-navigation/toolpad/pages/page1/page.yml
apedroferreira: I think the code itself was expecting a ref, anyway using a string seems good - I'll make it a string and take a look at the schema validation.

packages/toolpad-app/src/runtime/evalJsBindings.ts
apedroferreira: Got it, I'll try to use `update` for now.

test/integration/bindings/fixture-navigation/toolpad/pages/page1/page.yml
apedroferreira: It's not easy to make the schema validation work here, seems like `z.record(jsonSchema)` is allowing any object to pass validation, but without it the unions also don't work as expected. So far haven't solved it, I'm looking into it.

apedroferreira: ok i finally found a way

Janpot: Aah good catch,

One thing we could do is disallow properties in the `jsonSchema` that start with `$$`

```tsx
const literalSchema = z.union([z.string(), z.number(), z.boolean(), z.null()]);
type Literal = z.infer<typeof literalSchema>;
type Json = Literal | { [key: string]: Json } | Json[];
export const jsonSchema: z.ZodType<Json> = z
  .lazy(() =>
    z.union([
      ...literalSchema.options,
      z.array(jsonSchema),
      // disallow $$ properties here
      z.record(jsonSchema).refine((obj) => !Object.keys(obj).some((key) => key.startsWith('$$')))
    ]),
  )
```

But this would ofcourse not work for nested bindings, so we need to pull back in `jsExpressionBindingSchema`. But then we should probably not call it `jsonSchema` anymore. We can probably instead remove `jsonSchema` from `bindablePropSchema` and insteadcreate that type like. (You'll have to do some trickery to make recursive types work, see how it's done in `jsonSchema`)

```tsx
export const bindablePropSchema = z.lazy(() => z.union([
  ...literalSchema.options,
  z.array(bindablePropSchema),
  z.record(bindablePropSchema).refine((obj) => !Object.keys(obj).some((key) => key.startsWith('$$'))),
  jsExpressionBindingSchema,
  envBindingSchema,
  jsExpressionActionSchema,
  navigationActionSchema,
  templateSchema,
]));

```

apedroferreira: Yep this exactly what I was already going, I was just going to try to pass all the possible `$$...` properties to exclude in the refine function, but we can also just exclude if they start with `$$`, maybe that's easier to maintain.

packages/toolpad-app/src/runtime/evalJsBindings.ts
apedroferreira: The reason I created and used an `updatePath` method is because lodash makes it easy to update an object by passing a string with the path of the property we want to change (it also needs `clone` to be immutable, though). And in this case we could have a unknown number of nested objects or arrays.

I could create an issue to write our own method for this later using our own `update` instead of lodash methods?

Janpot (approved):
Lovely

Issue #2092, opened 2023-05-29
Cannot read properties of undefined (reading 'F_OK')


SethTurin:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

Steps:

1. All I did was try the code in the quickstart : 
 `npx create-toolpad-app dog-app`

### Current behavior 😯

it exits with the error:
`Cannot read properties of undefined (reading 'F_OK')`

### Expected behavior 🤔

I mean, I assume it should create an app or something. I didn't get very far so I don't really know.

### Context 🔦

This was the very first thing I tried, and currently I'm just experimenting.

### Your environment 🌎

It looks like the npx command installed version 0.1.13

I'm using Windows 11

SethTurin:
This issue was happening because of an old version of Node.  I updated Node and it works now.

Janpot:
Thanks for reporting. There's an open [PR](https://github.com/mui/mui-toolpad/pull/193) to lower the minimum Node.js version that should solve this exact error.

But the DX could be better. Unlike `yarn` and `pnpm`, it looks like `npx` doesn't check compatibility with the `engines` field. Perhaps we can add a quick check for node.js version in `create-toolpad-app`? cc @bharatkashyap

PR #2124, opened 2023-06-05
add ts declaration for package and check node version on initial star…

### LINKED ISSUE #2094: Update `create-toolpad-app` to match `engines`
- [ ] From https://github.com/mui/mui-toolpad/issues/2092#issuecomment-1567937877, we should have a node version check before running 
- [x] ~We should also add `engines` to the `package.json` file created by `create-toolpad-app`~ This is up to the user, as long as we don't generate code that is incompatible with certain node.js versions then the restrictions of `@mui/toolpad-app` and `create-toolpad-app` should do their jobs.
- [ ] Also a `.gitignore` on the root level with `node_modules` inside it so that this step does not have to be manually done when committing the project? @Janpot

JerryWu1234:
…t and move the gitnore to rool level

fix #2094

three points 

1. add ts declaration for the package in the create-toolpad-app folder
2. Check the node version when initially running
3.Move the .gitignore file to the root level, move the code to the create-toolpad-app folder, and remove the redundant code.


JerryWu1234:
I added a unit test for this code.

packages/create-toolpad-app/src/index.ts
```diff
@@ -150,6 +165,11 @@ const scaffoldProject = async (absolutePath: string, installFlag: boolean): Prom
 
 // Run the CLI interaction with Inquirer.js
 const run = async () => {
+  // check the node version before create
+  if (!satisfies(process.version, packageJosn.engines.node)) {
+    process.exit(1);
```
bharatkashyap, line 170: We should add error information before failing here

bharatkashyap:
Thanks @JerryWu1234 for the contribution! I've added a comment around error logging.

packages/create-toolpad-app/src/index.ts
```diff
@@ -150,6 +165,11 @@ const scaffoldProject = async (absolutePath: string, installFlag: boolean): Prom
 
 // Run the CLI interaction with Inquirer.js
 const run = async () => {
+  // check the node version before create
+  if (!satisfies(process.version, packageJosn.engines.node)) {
```
bharatkashyap, line 169:
```suggestion
  if (!satisfies(process.version, packageJson.engines.node)) {
```

packages/create-toolpad-app/tests/index.spec.ts
```diff
@@ -58,6 +58,13 @@ test('create-toolpad-app can bootstrap a Toolpad app', async () => {
...
   );
 
+  // check that file exists or not in the directory
+  const gitignore = await fs.readFile(path.resolve(testDir, './.gitignore'), {
+    encoding: 'utf-8',
+  });
+
+  expect(gitignore.length).toBeGreaterThan(0);
+
   toolpadProcess = execa('yarn', ['dev'], {
```
bharatkashyap, line 68: Could we add a test for the node version check as well? Run `nvm` inside the test to change Node versions and then check to see if the CLI stops executing as expected

packages/create-toolpad-app/tsconfig.json
```diff
@@ -10,7 +10,6 @@
     "esModuleInterop": true,
     "skipLibCheck": true,
     "noEmit": false,
-    "rootDir": "src",
```
bharatkashyap, line 13: Why is this line to be removed?

packages/create-toolpad-app/src/index.ts
```diff
@@ -118,12 +121,24 @@ const scaffoldProject = async (absolutePath: string, installFlag: boolean): Prom
...
     dependencies: {
       '@mui/toolpad': 'latest',
     },
   };
 
+  const DEFAULT_GENERATED_GITIGNORE_FILE_CONTENT = '.gitignore';
+
   await fs.writeFile(path.join(absolutePath, 'package.json'), JSON.stringify(packageJson, null, 2));
+  // eslint-disable-next-line no-console
+  console.log(`${chalk.blue('info')}  - Initializing package.json file`);
```
bharatkashyap, line 134: This log should be before the call to `fs.writeFile` above

packages/create-toolpad-app/src/index.ts
```diff
@@ -118,12 +121,24 @@ const scaffoldProject = async (absolutePath: string, installFlag: boolean): Prom
...
+
+  await fs.copyFile(
+    path.resolve(__dirname, `./${DEFAULT_GENERATED_GITIGNORE_FILE_CONTENT}`),
+    path.join(absolutePath, DEFAULT_GENERATED_GITIGNORE_FILE_CONTENT),
+  );
+  // eslint-disable-next-line no-console
+  console.log(`${chalk.blue('info')}  - Initializing .gitignore file`);
```
bharatkashyap, line 141: This log should be before the call to `fs.copyFile` above

packages/create-toolpad-app/src/index.ts
```diff
@@ -118,12 +121,24 @@ const scaffoldProject = async (absolutePath: string, installFlag: boolean): Prom
       build: 'toolpad build',
       start: 'toolpad start',
     },
+
     dependencies: {
       '@mui/toolpad': 'latest',
     },
   };
 
+  const DEFAULT_GENERATED_GITIGNORE_FILE_CONTENT = '.gitignore';
```
bharatkashyap, line 130:
```suggestion
  const DEFAULT_GENERATED_GITIGNORE_FILE = '.gitignore';
```

packages/create-toolpad-app/tsconfig.json
JerryWu1234: https://github.com/JerryWu1234/mui-toolpad/blob/653d82fc1ce5c7c82471237b6e54f3bafcdaadae/packages/create-toolpad-app/src/index.ts#L12
because of this.
I need to get the upper package.json,

I’m a little confused about what the full name of DX is.

packages/create-toolpad-app/tests/index.spec.ts
JerryWu1234: Run the specified version of nvm before we check whether the computer has this version or not. If there is no this version and it needs to be downloaded, then change the node version. It spends too long time in the process when running the test command.

packages/create-toolpad-app/tsconfig.json
bharatkashyap: "DX" usually refers to developer experience, unless you're referring to it from some other specific context.

Each folder inside `packages` is a separate "project", so you must not use the `package.json` of one project inside another. Instead, we can create a constant `create-toolpad-app` which contains the minimum supported node version and compare against that.

JerryWu1234: such as const TOOLPADAPPNODEVERSION = "minimum version"?

related to an issue https://github.com/mui/mui-toolpad/issues/2092#issuecomment-1567937877
 
I sorry I'm not clear expression, actually, I only use the package.json's `engines.node` in the create-toolpad-app.

packages/create-toolpad-app/src/packageType.ts
```diff
@@ -0,0 +1,667 @@
+declare namespace PackageJsonTypes {
```
Janpot, line 1: Looks like this was copied from https://www.npmjs.com/package/type-fest. Shall we just add this as a dependency?

packages/create-toolpad-app/tests/index.spec.ts
Janpot: I'm ok with leaving this functionality untested for now if the only way to test it is to switch node versions. The overhead isn't worth it for this specific case imo.

packages/create-toolpad-app/tsconfig.json
Janpot: I think I'd also prefer to keep `rootDir` as is. It ensures a consistent dist folder layout. How would you feel about just reading the `package.json` file from the filesystem using something like [`readJsonFile`](https://github.com/mui/mui-toolpad/blob/bf9a2c17adc921ef614d606d24b1aeaea9d05a2f/packages/toolpad-app/src/utils/fs.ts#L11)?

packages/toolpad-app/src/server/localMode.ts
```diff
@@ -1165,8 +1150,6 @@ async function initToolpadFolder(root: string) {
     };
     await writeProjectFolder(root, projectFolder);
   }
-
-  await initGitignore(root);
```
Janpot, line 1169: We can keep this file for now, it makes sure the content of the `.generated` folder is ignored.

packages/create-toolpad-app/public/.gitignore
```diff
@@ -0,0 +1,130 @@
...
+
+# Stores VSCode versions used for testing VSCode extensions
+.vscode-test
+
+# yarn v2
+.yarn/cache
+.yarn/unplugged
+.yarn/build-state.yml
+.yarn/install-state.gz
+.pnp.*
```
Janpot, line 130:
```suggestion
.pnp.*

```

packages/create-toolpad-app/src/packageType.ts
JerryWu1234: I had done it to install this library before.
But there is a lot of code we didn't use it.
if we need to get more declarations from this library, we can change to that library.

JerryWu1234: in spite of it not being important, it will waste our time installing dependency when it’s the first time to contribute.

packages/create-toolpad-app/tsconfig.json
JerryWu1234: > I think I'd also prefer to keep `rootDir` as is. It ensures a consistent dist folder layout. How would you feel about just reading the `package.json` file from the filesystem using something like [`readJsonFile`](https://github.com/mui/mui-toolpad/blob/bf9a2c17adc921ef614d606d24b1aeaea9d05a2f/packages/toolpad-app/src/utils/fs.ts#L11)?
> 
> With "DX" I referred to [developer experience](https://microsoft.github.io/code-with-engineering-playbook/developer-experience/). I should probably just have said UX here as I was referring to the user installing `toolpad` (a user who happens to be a developer). I updated my comment.

thanks your suggestion ,I will do it lete

packages/create-toolpad-app/public/.gitignore
```diff
@@ -0,0 +1,131 @@
...
+# Stores VSCode versions used for testing VSCode extensions
+.vscode-test
+
+# yarn v2
+.yarn/cache
+.yarn/unplugged
+.yarn/build-state.yml
+.yarn/install-state.gz
+
+.generated
```
Janpot, line 131:
```suggestion

```

The reason for having a .gitignore in the .generated folder is that we anticipate not all Toolpad projects will be created through `create-toolpad-app`. Happy to discuss alternatives separately, but within the context of this PR I would like to leave that logic untouched.

packages/create-toolpad-app/src/index.ts
```diff
@@ -150,6 +163,13 @@ const scaffoldProject = async (absolutePath: string, installFlag: boolean): Prom
 
 // Run the CLI interaction with Inquirer.js
 const run = async () => {
+  const pj = await readJsonFile(path.resolve(__dirname, `../package.json`));
+  // check the node version before create
+  if (!satisfies(process.version, pj.engines?.node)) {
+    console.error('Please upgrade your node version. Your node version is too low');
```
bharatkashyap, line 169:
```suggestion
    console.log(`${chalk.red('error')} - Your node version ${process.version} is unsupported. Please upgrade it to at least ${pj.engines?.node}`);
```

Janpot (approved):
Thank you

JerryWu1234:
> Thank you

You welcome 
I love low code.
I joined the low code of Vue before 
I enjoy it

Issue #2126, opened 2023-06-05
Running the build command ends in RollupError


CAGLADJ:
I was trying out the toolpad and installed it through `npx create-toolpad-app`. I used the toolpad a bit (added components to a page, etc.).

When following the [deployment guide](https://mui.com/toolpad/deployment/) in the documentation, running `npm run build` ended in the following error message:

```
Could not resolve entry module "index.html".
Error [RollupError]: Could not resolve entry module "index.html".
    at error (file:///C:/Users/.../node_modules/rollup/dist/es/shared/node-entry.js:2130:30)
    at ModuleLoader.loadEntryModule (file:///C:/Users/.../node_modules/rollup/dist/es/shared/node-entry.js:24017:20)
    at async Promise.all (index 0) {
  code: 'UNRESOLVED_ENTRY'
}
```

The installed @mui/toolpad version is 0.1.14

What did I miss? I couldn't find anything about having to create an `index.html` file inside the documentation.

The full terminal message is below:
```
C:\...>npm run build

> toolpad@0.1.0 build
> toolpad build

info  - building Toolpad application...
vite v4.3.9 building for production...
✓ 0 modules transformed.
✓ built in 20ms
Could not resolve entry module "index.html".
Error [RollupError]: Could not resolve entry module "index.html".
    at error (file:///C:/Users/.../node_modules/rollup/dist/es/shared/node-entry.js:2130:30)
    at ModuleLoader.loadEntryModule (file:///C:/Users/.../node_modules/rollup/dist/es/shared/node-entry.js:24017:20)
    at async Promise.all (index 0) {
  code: 'UNRESOLVED_ENTRY'
}
toolpad build [dir]

Build Toolpad app for production

Options:
  --version  Show version number                                       [boolean]
  --help     Show help                                                 [boolean]
  --dir      Directory of the Toolpad application        [string] [default: "."]

Error: Command failed with exit code 1: C:\Program Files\nodejs\node.exe 
C:\Users\...\node_modules\@mui\toolpad-app\dist\cli\appBuilder.js
    at makeError (C:\Users\...\node_modules\@mui\toolpad-app\dist\cli\index.js:1124:17)
    at handlePromise (C:\Users\...\node_modules\@mui\toolpad-app\dist\cli\index.js:1945:29)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
[...]

PR #2131, opened 2023-06-05
Try fixing build on windows

Janpot:
Attempts to fix https://github.com/mui/mui-toolpad/issues/2126

The leading slash would mismatch on windows paths as `rollup` doesn't pass [normalized paths](https://github.com/rollup/rollup/issues/2952). The resolver not recognizing this path will result on `index.html` not being found on windows. We can just check for trailing `'.html'` instead as there will be only one being resolved.

@apedroferreira When you have some time, would you mind testing whether you can reproduce the build error and whether this PR fixes it? Bootcamp doesn't work on M1 Macs.


Janpot:
Thank you for checking. Let's look into windows tests on circleci soon

Issue #2093, opened 2023-05-30
Enable Toolpad to infer types from typescript types


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Currently, wrapping functions with createFunction is cumbersome. Toolpad should infer parameter types from typescript types (likewise for components)

### Examples 🌈

```tsx
import QRCode from 'qrcode';
import { createFunction } from '@mui/toolpad/server';

export const generateQrCode = createFunction(
  async ({ parameters }) => {
    return QRCode.toDataURL(parameters.content);
  },
  {
    helperText: 'Encodes text into a QR code.',
    parameters: {
      content: {
        helperText: 'The content to encode.',
        type: 'string',
      },
    },
  },
```

could be:

```tsx
import QRCode from 'qrcode';

/**
 * Encodes text into a QR code.
 * @param content The content to encode.
 */
export async function generateQrCode(content: string) {
  return QRCode.toDataURL(content);
}
```

Benchmarks:
https://interval.com/docs/writing-actions#type-safety-and-validation (I/O methods do the type validation)
https://docs.airplane.dev/views/component-state#typescript-for-type-safety

PR #2133, opened 2023-06-05
Extract parameters from ts types in function files

Janpot:
Closes https://github.com/mui/mui-toolpad/issues/2093

https://github.com/mui/mui-toolpad/assets/2109932/ce8e64f8-7cc4-46c5-884a-c6b42b2a3b47

## To do:

- [ ] what with parameters that change types, but already have a value assigned of a different type? (see video ☝️ )
- [x] docs
- [x] test on mui-private and mui-public before merging


apedroferreira:
This is really cool!
Maybe boolean types could have a "checkbox" control? Right now it's the same control as for objects.

prakhargupta1:
I tried it today and it is very impressive! The code is concise now. 

I was able to run the following:
```
export async function getRepositoryDetails(slug: string) {
  const res = await fetch(`https://api.ossinsight.io/gh/repo/${slug}`, {
    method: 'GET',
  });
  if (res.status !== 200) {
    throw new Error(`HTTP ${res.status}: ${(await res.text()).slice(0, 500)}`);
  }
  return res.json();
}
```

Under this PR have we also removed the need to use **createfunction** API when creating a custom function?

Janpot:
> Under this PR have we also removed the need to use createfunction API when creating a custom function?

In https://github.com/mui/mui-public/pull/111 I was able to remove all `createServer` calls. Still have to try out the same for `mui-private`.

PR #2149, opened 2023-06-08
Improve contributing guidelines for starting the application

Janpot:
Propose a simpler way of running Toolpad apps locally. Keep the advanced way around but hide it.


JerryWu1234:
thank you so much.
I think it will save a lot of time.

PR #2154, opened 2023-06-08
[core] Remove @mui/x-data-grid-generator dependency

oliviertassinari:
With @flaviendelangle we were quickly looking at why @mui/x-data-grid-generator has quite some downloads. We found https://www.npmjs.com/browse/depended/@mui/x-data-grid-generator. As far as I understand, since the new implementation of Toolpad that runs locally, this is no longer needed.


packages/toolpad-app/src/runtime/muiExports.tsx
Janpot, line 1: actually, this whole file is dead code I think

Janpot:
> As far as I understand, since the new implementation of Toolpad that runs locally, this is no longer needed.

It was there to support copy+pasting code from the MUI docs straight into Toolpad and have it immediately work, some examples use this package. Back then we didn't support installing and using node modules.

Janpot (approved):

PR #2160, opened 2023-06-12
Increase amount of playwright workers

Janpot:
Try to saturate the circleci worker more by running tests in parallel. According to [insights](https://app.circleci.com/insights/github/mui/mui-toolpad/workflows/pipeline/jobs), the `test_unit` workflow takes 17m24s on average (This is an overestimation because failing runs often take much longer due to retries). The `test_unit` workflow of this PR ran in [10m47s](https://app.circleci.com/pipelines/github/mui/mui-toolpad/8015/workflows/3cc658c8-ab23-4c35-99bf-87ebd3688487/jobs/34988). Which is about 65% of the average, a 35% reduction in duration.

I had to update the test to run on a different port for each testcase as running on the default port resulted in race conditions. Toolpad does select a different port when the default one is in use, but this isn't built in a way that can handle concurrency-safe starting on two Toolpad processes simultaneously without specifying a port.

CPU usage before:

Screenshot 2023-06-12 at 17 43 09

CPU usage after:

Screenshot 2023-06-12 at 17 42 51


apedroferreira (approved):
looks great if our current plan can take it!

Janpot:
> looks great if our current plan can take it!

We are building the same things in less time on the same resources. Which means we'll actually be using fewer credits per pipeline.

PR #2170, opened 2023-06-14
[examples] Hacker News client example

Janpot:
Also adds a loading state to the list so it doesn't display as broken when the `itemCount` is still loading


prakhargupta1 (approved):
Just added the background color.

Screenshot 2024-02-21 at 10 51 14 AM

PR #2171, opened 2023-06-14
it will delete old page while adding new page 

JerryWu1234:
fix #1984
it will delete old page while adding new page 
- [x] I've read and followed the [contributing guide](https://github.com/mui/mui-toolpad/blob/master/CONTRIBUTING.md#sending-a-pull-request) on how to create great pull requests.
- [x] I've updated the relevant documentation for any new or updated feature
- [x] I've linked relevant GitHub issue with "Closes #<issue id>"
- [x] I've added a visual demonstration under the form of a screenshot or video

PR #2177, opened 2023-06-16
optimized deleting file

### LINKED ISSUE #1872: Better handling of deleting a page
### Duplicates

### Latest version

### Summary 💡

When I delete a page, I am shown this screen:
Screenshot 2023-04-11 at 4 56 45 PM

Instead of this message, we should take the user directly to the first/default page.

Note: If the last page is deleted then the above screen doesn't come, but if any page from the middle is deleted then this screen shows.

JerryWu1234:
fix #1872 

> to move to another page when a page is deleted through the UI (seems to work for me)

this PR is mainly for this  

>to avoid reusing a tab when opening the browser on startup if it's not set to a page that exists in the application. Currently it tries to reuse any tab that is on the domain localhost:3000, but we should consider the pathname as well when finding a suitable tab

I'm not totally understand for this.
please tell me more if you are available. thanks in advance 

>The user deletes a page on the file system that is also still open in UI. Again, take the path of least surprise, do not navigate unless a user explicitly requested navigation

other pr fixes it #2171

>The user directly opens a url to a non-existing page. The user purposefully opened this url so we should tell them it doesn't exist instead of quietly opening something they didn't intend.

could you show me what kind of error to remind the user?


packages/toolpad-app/src/server/localMode.ts
```diff
@@ -232,7 +232,7 @@ export async function createComponent(name: string) {
 export async function deletePage(name: string) {
   const root = getUserProjectRoot();
   const pageFolder = getPageFolder(root, name);
-  await fs.rm(pageFolder, { force: true, recursive: true });
```
Janpot, line 235: Why was this removed?

Janpot (approved):
Thank you for looking into this!

JerryWu1234:
> Thank you for looking into this!

Do you have a contributor group that I can join?

JerryWu1234:
> > Thank you for looking into this!
> 
> Do you have a contributor group that I can join? @Janpot

If there are any conditions for that let me know.
I would like to contribute for a long time and mainly I want to know which aspect you focus on

Janpot:
> Do you have a contributor group that I can join?

We're planning to launch a discord server soon. (cc @prakhargupta1)

JerryWu1234:
@prakhargupta1 how is it going?

Issue #2166, opened 2023-06-13
Run and fix integration tests on Windows


apedroferreira:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

Steps:

1. Fix `release:build` command on Windows
2. Run integration tests on Windows
3. Fix all failing integration tests

### Current behavior 😯

`release:build` command fails with errors on Windows

Issue #2053, opened 2023-05-22
Custom component not visible after creation in the Component library sidebar


mnajdova:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

Steps:

1. Click the plus for adding a custom component
2. Optional - update the component code
3. Expand the component library panel - the component is not visible

### Current behavior 😯

The component is appearing in the component library panel after restarting the server

### Expected behavior 🤔

The component should be visible immediately. 

### Your environment 🌎

Windows 11 Pro
yarn version 1.22.5

Janpot:
I can't reproduce on Mac with that repo/PR. It could be a windows issue.

apedroferreira:
Seems to be a Windows issue, works for me on Mac but I can reproduce it in Windows. I can look into it.

PR #2181, opened 2023-06-16
Fix Windows issues

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/2166
Closes https://github.com/mui/mui-toolpad/issues/2053

Make integration tests pass on Windows:

- Use `cross-env` for `NODE_OPTIONS`
- Fix type declaration generation for `release:build` command to be able to run
- Normalize paths where needed, use `windowsPathsNoEscape` for globs
- Some more test adjustments


packages/toolpad-core/package.json
```diff
@@ -33,7 +33,7 @@
   },
   "scripts": {
     "prebuild": "rimraf dist",
-    "build": "tsup",
+    "build": "tsup && tsc --emitDeclarationOnly --declaration",
```
apedroferreira, line 36: Not sure why it worked on Mac but `tsup` doesn't generate type declaration files. https://tsup.egoist.dev/#generate-typescript-declaration-maps--d-ts-map

apedroferreira:
A few integration tests to fix:
Capture

packages/toolpad-core/package.json
Janpot: We run the type generation in the `onsuccess` handler (https://github.com/mui/mui-toolpad/blob/bebe8a36c43dc13388d19ddad841aab1975c3893/packages/toolpad-core/tsup.config.ts#L14)

Are you sure the build is finishing successfully?

apedroferreira: I hadn't checked yet but yeah, if it works on Mac I guess the types definitions were being generated there. This explains it, should have searched for `emitDeclarationOnly` :)
So I guess the issue must be something like this `spawnSync` not running on Windows. I'll take a look at that when I'm back.

Janpot: I think you can remove these `onSuccess` handlers and just turn on `dts: true`. Looks like it was implemented this way to improve performance as per https://github.com/egoist/tsup/issues/615. But we don't have `incremental` enabled (anymore?) and I don't notice a big problem in performance.

apedroferreira:
Capture

All tests are passing in Windows now, so this should be good until we add Windows tests to CI maybe.

packages/toolpad-app/src/utils/paths.ts
```diff
@@ -1,9 +1,16 @@
 import * as path from 'path';
 
+/**
+ * Normalize a file path to be platform-agnostic.
+ */
+export function normalizePath(importPath: string): string {
```
Janpot, line 6: "normalize" is a very generic term, perhaps we should call it `asPosixPath` or something?

test/integration/components/form.spec.ts
```diff
@@ -40,16 +40,14 @@ test('submits form data', async ({ page }) => {
...
-    .toStrictEqual({
+    .toMatchObject({
       name: 'Toolpad',
       date: '1990-01-01',
       option: 'option 2',
       file: [
         {
           name: 'test.txt',
           type: 'text/plain',
-          size: 6,
```
Janpot, line 51: If this varies per platform you should also be able to do

```tsx
size: expect.any(Number),
base64: expect.stringMatching(/^data:text\/plain;base64,.+$/),
```
within `toMatchObject `

test/integration/rest-basic/index.spec.ts
```diff
@@ -64,12 +64,12 @@ test('rest editor basics', async ({ page, context, localApp }) => {
 
   await expect(newQueryEditor).toBeVisible();
 
-  await expect(editorModel.appCanvas.getByText('query4 authorization: foo')).toBeVisible();
+  await expect(editorModel.appCanvas.getByText('query4 authorization: foo')).toBeDefined();
```
Janpot, line 67: A locator is always defined. This will always pass. When writing tests, however trivial the change, best to always verify it still breaks when you expect it to.

```tsx
// this passes as well
await expect(editorModel.appCanvas.getByText("I'm not on the page at all")).toBeDefined();
```

apedroferreira: Yeah, I has thought of this and was gonna give it a second look - it's just that the dialog is covering the UI so I think this was making the test fail, I'll find the right way to do this.

test/integration/components/form.spec.ts
apedroferreira: I thought the file name and type were most important anyway, but yeah for some reason the size differs in Windows and the base64 URL was different and had `==` at the end, not sure why, I'll figure it out.

packages/toolpad-app/src/utils/paths.ts
apedroferreira: Sounds good, I actually had started with that name but then changed it :)

test/integration/rest-basic/index.spec.ts
Janpot: You could try `.toBeAttached`. I didn't think an obscured element would be omitted by `.toBeVisible`. I thought that was what `. toBeInViewport` did.

Any reason why this would only fail on windows?

apedroferreira: True that it's weird if it failed on Windows only...
I'll recheck as this might not be necessary, the test was being flaky locally sometimes so it might have been misleading.

Janpot: For this test, try with the instructions above and run httpbin in a container locally. That should help with the flakeyness. I'm planning to write a mock server into the test itself in the future.

test/integration/components/test.txt
Janpot, line 1: this file can be deleted I think?

Janpot (approved):

Janpot:
One file is unnecessary, rest looks good to me. Can we follow up with running windows tests in CI?

test/integration/components/test.txt
apedroferreira: It was still used in 2 other tests but I updated it everywhere and deleted the file for consistency and to avoid this type of issue.

Janpot (approved):

Janpot:
nice work, do we have an issue for running them in windows on circleci?

apedroferreira:
> nice work, do we have an issue for running them in windows on circleci?

I'll create one.

Issue #2025, opened 2023-05-16
Support Stats component


Janpot:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

A single numerical output, capable of showing:

- title
- number
  - formatting
  - units
 - optional description
 - optional status (good/bad/average/...)
 - 

### Examples 🌈

benchmark:

retool:

![Screenshot 2023-05-16 at 14 39 40](https://github.com/mui/mui-toolpad/assets/2109932/6e03421d-0f8f-436b-b0bb-eb82ce0cfd58)

appsmith:

![Screenshot 2023-05-16 at 14 39 18](https://github.com/mui/mui-toolpad/assets/2109932/ab3d1046-0c71-486c-8f28-309bece831aa)

grafana:

Multiple stats:

![Screenshot 2023-05-16 at 14 41 38](https://github.com/mui/mui-toolpad/assets/2109932/86347225-02a2-4d00-8beb-3d1833f10e6d)

https://grafana.com/docs/grafana/latest/panels-visualizations/visualizations/stat/

### Motivation 🔦

* Commonly used custom component in internal use cases (KPI dashboards)

### Benchmark

- https://www.notion.so/mui-org/Stat-component-cbf8ec1d45a34f3fadd6b16f21114487

apedroferreira:
Could this not be done with a Card component that elements can be dragged into, and allowing more control over text?
But if other tools have a `Stats` component I guess there might be a common enough use case for something out-of-the-box.

Janpot:
Yes, this can be done with existing components. But together with chart and table it's probably the most used primitive on dashboards. You can also see that in our public tools repo it's currently implemented as a code component, and reused multiple times https://github.com/search?q=repo%3Amui%2Fmui-public%20codeComponent.HealthBadge&type=code.

oliviertassinari:
I wonder if the people who upvoted https://github.com/mui/mui-toolpad/issues/748 don't want this issue instead.

Janpot:
> I wonder if the people who upvoted #748 don't want this issue instead.

👍 I think that's a fair assumption

PR #2202, opened 2023-06-22
Add Metric component

Janpot:
Closes https://github.com/mui/mui-toolpad/issues/2025

https://github.com/mui/mui-toolpad/assets/2109932/c4a98af8-15e3-4139-a558-9d069cef5043

To do:

- [x] "colorScale" => "Conditional format"
- [x] "numberFormat" => "Number format"
- [x] "statistic" => "metric"

Future:

- sparkline
- difference with previous value


apedroferreira:
Hey, some thoughts from watching the video:
- Why is `numberFormat` in its own popover? Could it not just be a dropdown in the props editor?
- The colors stops popover feels a bit empty in it's initial state, at least on the left side of the colored circle. Maybe there could be some text there to make things more explanatory, not sure exactly what, possibly "Default color"?

docs/data/toolpad/reference/components/metric.md
```diff
@@ -0,0 +1,18 @@
+<!-- This file has been auto-generated using `yarn docs:build:api`. -->
+
+# Metric
+
+<p class="description">API docs for the Toolpad Metric component.</p>
+
+The Metric component lets you display values.
```
apedroferreira, line 7: Should we elaborate a bit more on possible typical use cases in this description?

Janpot:
> * Why is numberFormat in its own popover? Could it not just be a dropdown in the props editor?

It could have more inputs depending on the context:

Screenshot 2023-07-12 at 09 22 03

maybe there is a better way to do this

apedroferreira (approved):
The changes look great, I think they were good choices!

PR #2221, opened 2023-06-26
Add needs triage action

Janpot:
From https://docs.github.com/en/actions/managing-issues-and-pull-requests/adding-labels-to-issues

Ensure we add the needs-triage on opened issues


oliviertassinari:
If this works, 💯 to add it to all the repositories.

PR #2230, opened 2023-06-27
Use mui charts for the npm stats example

Janpot:
Makes sense? Or too soon?

Screenshot 2023-06-27 at 15 37 04


prakhargupta1:
Looks great! I think we should use it. Just that in this stage of mui charts, frequent breaking changes would have to be tackled, is there any other downside?

Janpot:
Let's migrate the example to our new chart component instead. cc @apedroferreira

apedroferreira:
> Let's migrate the example to our new chart component instead. cc @apedroferreira

Sounds good, I guess we can still reuse this issue for that?

PR #2246, opened 2023-06-30
[feat] Add page structure explorer

### LINKED ISSUE #490: Page structure explorer
Create a panel that can show the page hierarchy as a tree and allows selecting/reordering page nodes. 

Benchmark:
https://getshogun.com/help/en/articles/1821780-using-the-layout-tab-to-view-page-structure

Retool
Image

### Shaping:

Must haves:

- left sidebar, vertical split, bottom panel will be a tree view
- Will try to use MUI lab TreeView component
- All DOM elements of current page
- We indicate the selected element, and allow selecting from this view
- Do we show rows and columns?
  - Let's show them and evaluate
- upon selection, expand the subtree that contains the selection, scroll into view
- To display nested children:
  - if the prop is `children` then we just nest them as the children of the component
  - if the prop is anything else we add an item in between with the name of the prop
- Overflow of the tree results in scroll

e.g.

```
Container:
  Paper:
    Button
    Text
Container:
  List:
    *renderItems*:
      Row:
         Button
         Text
```

Secondary:

- icons: Let's match the icons from the component catalog
- delete elements
- drag to reorder
- double-click to rename

bharatkashyap:
- Closes #490 

https://github.com/mui/mui-toolpad/assets/19550456/a21773a9-4745-41b9-ba2a-6ccff05dffd9


test/integration/backend-basic/fixture/toolpad/pages/serialization/page.yml
```diff
@@ -46,3 +46,27 @@ spec:
             value:
               $$jsExpression: |
                 `Invalid error: ${invalidError.error?.message}`
+    - component: PageRow
```
Janpot, line 49: Accidentally committed?

Janpot:
Something seems to go wrong with the hover state when renaming:

https://github.com/mui/mui-toolpad/assets/2109932/b832b954-2d69-4830-ad98-e2e31addeed2

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/NodeHud.tsx
```diff
@@ -167,6 +169,7 @@ export default function NodeHud({
   onDuplicate,
   isOutlineVisible = false,
   isHoverable = true,
+  isHovered = false,
```
apedroferreira, line 172: Since the dashed border doesn't really just mean the element is "hovered" anymore we might want to give this prop a different name just to avoid confusion. Maybe `isPreviewed`? Not sure, also not a super big deal.

apedroferreira (changes requested):
Looks awesome and works great! Just noticed a couple of issues:
- When I drag-and-drop to create a column the whole row that contains it collapses in the panel
- It's not possible yet to press backspace to delete after selecting a row or column in the explorer, but I feel like it would be a really useful feature

prakhargupta1:
> * Do you think it's time we start using `TreeView` for this UI? I feel like it would create denser UI.
> * Should we add some sort of section title above that says "explorer"?

Looks good! +1 for both of these suggestions:

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/NodeHud.tsx
apedroferreira: Ok, I think it's fine to keep it.

bharatkashyap:
I've updated to address most of the feedback here:

https://github.com/mui/mui-toolpad/assets/19550456/b7fd5446-5faa-4edd-8810-bc5e62d06387

@apedroferreira I've added an implementation of deleting via the explorer which plays the most nicely with undoing, does this make sense to you are are there any other ways of achieving this?

packages/toolpad-app/src/toolpad/AppEditor/StructureExplorer/index.tsx
```diff
@@ -187,40 +213,74 @@ export default function PageStructureExplorer() {
...
+            }
+            const toRemove = appDom.getMaybeNode(dom, selectedDomNodeId);
+            if (toRemove && appDom.isElement(toRemove)) {
+              draft = appDom.removeNode(draft, toRemove.id);
+              draft = deleteOrphanedLayoutNodes(dom, draft, toRemove);
+            }
+            return draft;
+            // return normalizePageRowColumnSizes(draft);
+          },
+          currentView.kind === 'page'
```
apedroferreira, line 238: I wonder if we can get rid of these `currentView.kind === 'page'` as I think these actions can only occur in a page view? But I may have forgotten why I added these checks in the other similar deletes...

packages/toolpad-app/src/toolpad/AppEditor/StructureExplorer/index.tsx
```diff
@@ -187,40 +213,74 @@ export default function PageStructureExplorer() {
...
+      setExpandedDomNodeIds(nodeIds);
     },
     [setExpandedDomNodeIds],
   );
 
+  const handleNodeDelete = React.useCallback(
+    (event: React.KeyboardEvent<HTMLUListElement>) => {
+      // delete selected node if event.key is Backspace
+      if (event.key === 'Backspace') {
+        appStateApi.update(
```
apedroferreira, line 225: Seems like the correct logic, except we do may need to normalize the page row column sizes after deleting as done in the other places...
And now that this logic goes beyond just the `RenderOverlay` component, we should move all of this (deleteing node + oprhaned nodes + normalizing column sizes) to somewhere where we can reuse it more agnostically, not sure if that's easy or not.

packages/toolpad-app/src/toolpad/AppEditor/StructureExplorer/index.tsx
```diff
@@ -187,40 +213,74 @@ export default function PageStructureExplorer() {
...
+      <TreeView
+        aria-label="components explorer"
+        defaultCollapseIcon={<ExpandMoreIcon sx={{ fontSize: '0.9rem', opacity: 0.5 }} />}
+        defaultExpandIcon={<ChevronRightIcon sx={{ fontSize: '0.9rem', opacity: 0.5 }} />}
+        expanded={Array.from(expandedDomNodeIdSet)}
+        selected={selectedDomNodeId as string}
+        onNodeSelect={handleNodeSelect}
+        onNodeFocus={handleNodeFocus}
+        onNodeToggle={handleNodeToggle}
+        onKeyDown={handleNodeDelete}
```
apedroferreira, line 271: I would separate `onKeyDown` to something like `handleKeyDown`, and there is backspace is pressed call a reusable delete function. 
That way for example if we allow deleting in the explorer by clicking the button we just have to use the same function that is already there.

packages/toolpad-app/src/toolpad/AppEditor/StructureExplorer/index.tsx
```diff
@@ -0,0 +1,286 @@
...
+    [setExpandedDomNodeIds],
+  );
+
+  const handleNodeDelete = React.useCallback(
+    (event: React.KeyboardEvent<HTMLUListElement>) => {
+      // delete selected node if event.key is Backspace
+      if (event.key === 'Backspace') {
+        appStateApi.update(
+          (draft) => {
+            if (!selectedDomNodeId) {
```
apedroferreira, line 227: Can you check for if there is a selected node outside of `appStateApi.update`? That way we don't add a new entry to the undo stack, I think.

apedroferreira (changes requested):
Looks good, just added comments on some minor things but still everything is working great!

prakhargupta1:
Just played with it, gives so much clarity about the page hierarchy now. Awesome UX! 🙌

packages/toolpad-app/src/toolpad/AppEditor/StructureExplorer/index.tsx
bharatkashyap: I think it's for the types, in which case I can probably replace these conditionals with 
```...(currentView as Extract<DomView, { kind: 'page' }>),```

File: packages/toolpad-app/src/toolpad/AppEditor/StructureExplorer/index.tsx (hunk 4 above)
Comment (bharatkashyap) [reply packages/toolpad-app/src/toolpad/AppEditor/StructureExplorer/index.tsx]: > And now that this logic goes beyond just the RenderOverlay component, we should move all of this (deleteing node + deleting orphaned nodes + normalizing column sizes) to somewhere where we can reuse it more 
I've added some changes to try and accomplish this, have a look

File: packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/RenderOverlay.tsx [hunk 7]
```diff
@@ -268,6 +268,51 @@ export function deleteOrphanedLayoutNodes(
   return draftDom;
 }
 
+const rowColumnCounts: Record<NodeId, number> = {};
```
apedroferreira: Looks good if it works! We could probably reset this variable if the page changes to keep it from growing too big, but no need to do it yet probably.

packages/toolpad-app/src/toolpad/AppEditor/utils/deleteNode.ts
```diff
@@ -0,0 +1,31 @@
...
+export default function deleteNode(
+  nodeId: NodeId,
+  pageNode: appDom.PageNode | null,
+  currentView: DomView,
+  appStateApi: AppStateApi,
+  dom: appDom.AppDom,
+) {
+  appStateApi.update(
+    (draft) => {
+      const toRemove = appDom.getNode(draft, nodeId);
```
apedroferreira, line 17: What if we kept just the part inside `appStateApi.update` as this reusable function, and so we didn't need to pass `appStateApi` and `currentView` as arguments, it would just be a function that transforms the DOM? 
I feel like it would be clearer that way and depend less on context, even if we had to repeat the other parts when we called the function.

packages/toolpad-app/src/toolpad/AppEditor/utils/normalizePageRowColumnSizes.ts
```diff
@@ -0,0 +1,54 @@
+import { NodeId } from '@mui/toolpad-core';
+import * as appDom from '../../../appDom';
+import { isPageRow } from '../../../runtime/toolpadComponents';
+
+const rowColumnCounts: Record<NodeId, number> = {};
+
+export default function normalizePageRowColumnSizes(
+  draftDom: appDom.AppDom,
```
apedroferreira, line 8: Perhaps we can remove the word `draft` from this function as the usage is more general now, i guess it could be just `dom` instead of `draftDom` for example?

apedroferreira (approved):
Awesome work Bharat, looks really great! Thanks a lot for addressing these!
I still had some thoughts and possible suggestions but I don't think any of them should be that blocking, it's very minor stuff, so I will approve already.
Maybe a small refactor of `deleteNode` would be the most important one to me!

bharatkashyap:
@apedroferreira Resolving the comments related to the common delete function extraction since we agreed on a different approach involving refactoring the existing delete

apedroferreira:
> @apedroferreira Resolving the comments related to the common delete function extraction since we agreed on a different approach involving refactoring the existing delete

Ok, sounds good!

PR #2262, opened 2023-07-04
Add prisma example

Janpot:
https://github.com/mui/mui-toolpad/assets/2109932/68756fc1-b54f-4c98-b436-ffd5a6e25824


apedroferreira:
Looks good! - maybe there might be a better UI pattern for deleting rows, or we could maybe even have create/edit forms on separate pages, but I think it works well to show this use case.

apedroferreira (approved):

prakhargupta1:
It requires a .env file containing: DATABASE_URL="file:./dev.db" info. I guess it got skipped due to gitignore?

The app loads like this without it:
Screenshot 2023-07-07 at 2 58 25 PM

Should we add it to the readme instructions or simply add the .env in the project folder? I don't see any harm with 2, as it is just an example.

Janpot:
Adding https://github.com/mui/mui-toolpad/pull/2282

PR #2270, opened 2023-07-05
Allow loading partial theme

Janpot:
Some problems uncovered when working on the stats component color scale

- Don't crash when certain theme colors are not defined
- Fix defaults depending on mode

https://github.com/mui/mui-toolpad/assets/2109932/34fdc0ec-85d6-4d7c-ba61-02944aa98aed


apedroferreira (approved):
It looks so cool!

PR #2276, opened 2023-07-06
[docs] Fix MUI reference

oliviertassinari:


Janpot (approved):

oliviertassinari:
It's annoying to have to run `yarn release:build` for `yarn docs:build:api` to work. I think that it would be a much better DX to resolve the source when building the API.

PR #2280, opened 2023-07-07
Resolve components from source for docs:build:api

Janpot:
Deal with https://github.com/mui/mui-toolpad/pull/2276#issuecomment-1624423714

This is now much easier since we moved to `tsx` in https://github.com/mui/mui-toolpad/pull/2146


oliviertassinari:
Nice, thanks 👌

PR #2283, opened 2023-07-07
Add note about light theme

Janpot:


bharatkashyap (commented):
I feel like we should probably add context to this since read independently, this rule can seem a bit arbitrary

Janpot:
👍  Do you have any suggestions?

bharatkashyap:
> 👍 Do you have any suggestions?

I think we could probably mention that the app dark mode has very high contrast so we want screenshots to be in light mode till such time that a (linked) issue that fixes the contrast problem in Toolpad dark mode is resolved.

PR #2300, opened 2023-07-10
Add extra note about aspect ratio of screenshots

Janpot:
Maybe we should just mandate viewport width instead of width/height?


apedroferreira:
> Maybe we should just mandate viewport width instead of width/height?

I feel like it's fine and just more consistent to enforce both width and height? Unless it's causing some particular issue, then it might not be that important.

apedroferreira (approved):

Issue #2095, opened 2023-05-30
Page tab to be shown in the right panel 


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

When a component is clicked, the right panel starts showing the related component props.  The pages state goes to the background and is reachable only when component is deselected. 
We should keep it easily discoverable irrespective of component selection. 2 ways we can do it are:

1. A page tab before Component tab would help. When user clicks a component then the component tab becomes active.
2. Move queries, page params out of the right panel (part of UX revamp)

<img width="390" alt="Screenshot 2023-05-31 at 2 13 25 PM" src="https://github.com/mui/mui-toolpad/assets/92228082/daa22c3d-c2c6-46c3-89b5-95a9bbdd9d25">

### Benchmarks 🌈
Other tools don't have their queries section in the right panel, it's either left panel or at the bottom.

JerryWu1234:
>Move queries, page params out of the right panel (part of UX revamp)

The second way, I was a little confused about it. Where should I put it if we move the queries?

JerryWu1234:
@prakhargupta1

prakhargupta1:
> The second way, I was a little confused about it. Where should I put it if we move the queries?

Currently, we show the list of queries on the right panel. Instead, we can show it in a bottom panel. We'll first design the right UX for this.

JerryWu1234:
> > The second way, I was a little confused about it. Where should I put it if we move the queries?
> 
> Currently, we show the list of queries on the right panel. Instead, we can show it in a bottom panel. We'll first design the right UX for this.

because you provided two ways 
so that How do I choose one of these ways if I want to fix it? Or whatever way is up to me?

JerryWu1234:
@prakhargupta1 
When it comes to these UIs, what is your development process like?
one of your team will offer UI ?

prakhargupta1:
> because you provided two ways
> so that How do I choose one of these ways at random if I want to fix it? Or whatever way is up to me?

For now, we'll follow the first approach. Later, (when we internally agree on a UX design) we'll work as second solution.

Issue #2138, opened 2023-06-06
Move queries section to the left


prakhargupta1:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Currently, the query panel is in the right pane. Based on whether a component is selected, queries can be visible or hidden, which is a bad DX.

This task is about shifting the query panel to the left side in the Toolpad editor.

We can leave the page parameters on the right.
As there will be space, later on, we'll support a page description text box.

### Examples 🌈

Appsmith:
<img width="507" alt="Screenshot 2023-06-06 at 3 38 17 PM" src="https://github.com/mui/mui-toolpad/assets/92228082/d2657232-1882-4c77-8cf2-79befdf3436d">

Superblocks: (They have page components and APIs both under a single section: 'Navigation')
<img width="563" alt="Screenshot 2023-06-06 at 3 41 42 PM" src="https://github.com/mui/mui-toolpad/assets/92228082/7aa645d1-9da0-424d-b390-60bbfbb2712e">

PR #2311, opened 2023-07-14
Add Page tab to right-side panel

apedroferreira:
- Add Page tab
- Remove "Settings" option for pages in left-side panel

Closes https://github.com/mui/mui-toolpad/issues/2095?
(or should we not close it to separate the queries to another panel, or should we create a separate issue for that?)

https://github.com/mui/mui-toolpad/assets/10789765/0fdaf1bb-a8d9-454f-a99d-5a4fafc0e8b7


Janpot (approved):

prakhargupta1:
We can link it to be closed. Moving queries to the left is being explored in https://github.com/mui/mui-toolpad/issues/2138

Janpot:
@apedroferreira Few things I ran into when further testing

- the page tab doesn't get activated when I deselect a component (e.g. click the page margin).
- it's pretty useless to be able to select the component tab when no component is selected. perhaps we should disable it in that case?

apedroferreira:
- the page tab doesn't get activated when I deselect a component (e.g. click the page margin).
- it's pretty useless to be able to select the component tab when no component is selected. perhaps we should disable it in that case?

Sure these are fair points, maybe they make things a little bit more intuitive.
I'll try these 2 changes to see how it feels.

PR #2322, opened 2023-07-17
Refactor deleting nodes

apedroferreira:
As explained in the comments below I still think we should normalize column sizes to add up to the amount of elements in each row, but I made changes so that:
- No refs or global variables are needed
- Normalizing column sizes is a bit simpler and more efficient
- Made the functions reusable outside the `RenderOverlay` component like Bharat had, feel free to suggest where we should place these utilities

I've also added some integration tests to cover just a few common cases with nested deletes and new element column sizes.


apedroferreira:
An update about removing normalization. At first I thought it was possible to do it but now I can see why I had added it:
- Right now when we add a new element to a page row, the "columnSize" value is 1, so that's how much weight it has on the width of the row. If rows are normalized so that the sum of all column sizes is equal to the number of elements in the row, a new element in a row with 3 elements will take 1/4 of the row, and a new element in a row with 4 elements will take up 1/5.
- I thought I could remove normalization, but with the current system a new element that is added always has a `columnSize` of 1, but an element that is removed could have `columnSize` larger or smaller than 1. So the total sum of column sizes will keep changing and there's no telling how much width a new element with a column size of 1 will take. This could lead to weird behavior after columns have been resized a lot.

So I will try to find a way to solve the issues we've discussed. If it means continuing to normalize column sizes, I will try to find a way to do it without needing any global variables at least.

Janpot:
Instead of having to recallibrate the coluns on delete, would it be easier to set an initial value of "sum of flex values / number of columns"?

apedroferreira:
> Instead of having to recallibrate the coluns on delete, would it be easier to set an initial value of "sum of flex values / number of columns"?

I thought about that - will see which one is simpler and most performant.

apedroferreira:
I kept column size "normalization" as it helps with many different scenarios, but improved everything so that it should be easier to reuse (no need for refs or global variables either).
Also added some tests around this whole thing, but just for simple cases.

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/RenderOverlay.tsx
```diff
@@ -1260,7 +1091,8 @@ export default function RenderOverlay({ bridge }: RenderOverlayProps) {
             if (
               draggedNode.layout?.columnSize &&
               draggedNodeParent &&
-              draggedNodeParent.id !== parent.id
+              originalParent &&
```
apedroferreira, line 1094: bug fix

packages/toolpad-app/src/toolpad/AppEditor/pageLayout.ts
```diff
@@ -0,0 +1,180 @@
+import { NodeId } from '@mui/toolpad-core';
+import { isPageColumn, isPageLayoutComponent, isPageRow } from '../../runtime/toolpadComponents';
+import * as appDom from '../../appDom';
+
+export function deleteOrphanedLayoutNodes(
```
apedroferreira, line 5: unchanged, just moved it

Janpot (approved):
👍  great to increase the test coverage

PR #2326, opened 2023-07-18
fix background color style

JerryWu1234:
image

### DIFF
--- a/packages/toolpad-app/src/runtime/ToolpadApp.tsx
+++ b/packages/toolpad-app/src/runtime/ToolpadApp.tsx
@@ -1184,7 +1184,7 @@ interface PageRootProps {
 
 function PageRoot({ children }: PageRootProps) {
   return (
-    <Container>
+    <Container sx={{ bgcolor: '#f7f7f7' }}>
       <Stack
         data-testid="page-root"
         direction="column"


JerryWu1234:
if the inset was set, background was #f7f7f7, I think the preview mode follows the dev

Janpot:
I believe the preview is correct, it's the canvas that overlays a slight transparent grey when nothing is selected. The idea is to emphasize something is selected by slightly dimming the rest. It gives the illusion of a grey background.

We can start by disabling this overlay when nothing is selected. If this turns out to still be confusing we will disable it altogether and rely purely on the blue box to indicate selection

JerryWu1234:
>  we will disable it altogether and rely purely on the blue box to indicate selection
I chose this way

JerryWu1234:
> > we will disable it altogether and rely purely on the blue box to indicate selection
> 
> I chose this way image image

@Janpot 
It only shows the blue button active, which can help shrink the code that doesn’t need to calculate it a lot.

Janpot:
If we don't want a background at all, you may as well remove this whole component

JerryWu1234:
> If we don't want a background at all, you may as well remove this whole component



> If we don't want a background at all, you may as well remove this whole component

actually, I removed the whole component but the `[chromium] › integration/rest-basic/index.spec.ts:36:5 › rest runtime basics ` this test will occur an error.
let me dive into deeper

JerryWu1234:
image

Janpot:
Yes, you may have to move the `data-testid="page-overlay"` one comonent up

JerryWu1234:
@Janpot 
after my in-depth study, 
We can’t maybe directly delete this part of PinholeOverlay, because it is not only for emphasizing something that is selected, but also for canceling the status of active when something is selected.

JerryWu1234:
I added a test for the status of active

apedroferreira:
If the PinholeOverlay does have utility and the only issue was the darkening of the UI I think just removing that darkening should do!
Thanks for adding a test for this, I'll leave some comments on it.

test/integration/editor/index.spec.ts
```diff
@@ -115,3 +115,20 @@ test('code editor auto-complete', async ({ page }) => {
   await page.keyboard.type('textF');
   await expect(page.getByRole('option', { name: 'textField' })).toBeVisible();
 });
+
+test('cancel the status of active when click the white space', async ({ page }) => {
```
apedroferreira, line 119: I'd reword the test name to something like `must deselect selected element when clicking outside of it`

test/integration/editor/index.spec.ts
```diff
@@ -115,3 +115,20 @@ test('code editor auto-complete', async ({ page }) => {
...
+
+  await editorModel.goToPageById('K7SkzhT');
+
+  await editorModel.waitForOverlay();
+
+  const text = editorModel.appCanvas.getByText('text-foo');
+
+  await clickCenter(page, text);
+  await expect(editorModel.appCanvas.getByTestId('node-hud-tag')).toBeVisible();
+  const element = await editorModel.appCanvas.locator('.Toolpad_HudOverlay');
```
apedroferreira, line 130: Instead of having a locator for the overlay and using `clickCenter` I think that here we should write a `page.mouse.click` with coordinates outside the text element that is selected.

So we can get `text.boundingBox()` for the text element, and then click on `textBoundingBox. x + testBoundingBox.width + 50` and `textBoundingBox.y + testBoundingBox.height + 50`, for example. This way we are sure that we are clicking outside of it.

test/integration/editor/index.spec.ts
```diff
@@ -115,3 +115,20 @@ test('code editor auto-complete', async ({ page }) => {
...
+  await clickCenter(page, element);
+  const hudButton1 = await editorModel.appCanvas.getByTestId('node-hud-tag');
```
apedroferreira, line 132: No need to write this locator again, you just have to write it once above and then reuse it.
Something like ```const textNodeHudTag = editorModel.appCanvas.getByTestId('node-hud-tag')```

apedroferreira (changes requested):
Adding some suggestions for the test, other than that this approach seems to work!

JerryWu1234:
@apedroferreira 
help agree on this pr if this pr can be merged.

test/integration/editor/index.spec.ts
```diff
@@ -192,3 +192,18 @@ test('code editor auto-complete', async ({ page }) => {
...
 });
+
+test('must deselect selected element when clicking outside of it', async ({ page }) => {
+  const editorModel = new ToolpadEditor(page);
+
+  await editorModel.goToPageById('K7SkzhT');
+
+  await editorModel.waitForOverlay();
+  const nodeHudTag = editorModel.appCanvas.getByTestId('node-hud-tag');
+  const input = editorModel.appCanvas.locator('input:nth-child(1)');
```
apedroferreira, line 203: Do we need to use `nth-child`? Or if there's only one single input in the page, can we make it even more simple?

apedroferreira, line 203: The reason I said we can use a text field was to make it full-width, so that the selection and the element have the same size.
Can you make the `fullWidth` prop be `true` for this text field?

test/integration/editor/index.spec.ts
```diff
@@ -192,3 +192,18 @@ test('code editor auto-complete', async ({ page }) => {
...
   await expect(page.getByRole('option', { name: 'textField' })).toBeVisible();
 });
+
+test('must deselect selected element when clicking outside of it', async ({ page }) => {
+  const editorModel = new ToolpadEditor(page);
+
+  await editorModel.goToPageById('K7SkzhT');
+
+  await editorModel.waitForOverlay();
+  const nodeHudTag = editorModel.appCanvas.getByTestId('node-hud-tag');
```
apedroferreira, line 202: If would be even better if this node HUD selector could only match the HUD of the `TextField`, but not the HUD of the `Text`.

apedroferreira (changes requested):
Almost there, just a few more important changes in the test!
Also just a nitpick, but feel free to use more vertical space in the test to separate bits of logic, it can really help improve the readability!

test/integration/editor/index.spec.ts
JerryWu1234: I'm not very good at the `playwright`,
Could you share your idea with me?
thanks

test/integration/editor/index.spec.ts
```diff
@@ -192,3 +192,18 @@ test('code editor auto-complete', async ({ page }) => {
...
+  const input = editorModel.appCanvas.locator('input').nth(0);
```
apedroferreira, line 203: There's only one input in this page so you can just do:
```
  const input = editorModel.appCanvas.locator('input');
```

apedroferreira:
With my comments above, here's how I would vertically space the test, order the variables and change variable names to improve readability and make the test as easy to understand as possible:
```
const editorModel = new ToolpadEditor(page);

  await editorModel.goToPageById('K7SkzhT');

  await editorModel.waitForOverlay();

  const textField = editorModel.appCanvas.locator('input');
  const textFieldNodeHudTag = editorModel.appCanvas
    .getByTestId('node-hud-tag')
    .filter({ hasText: 'textField' });

  await clickCenter(page, textField);
  await expect(textFieldNodeHudTag).toBeVisible();

  const textFieldBoundingBox = await textField.boundingBox();
  await page.mouse.click(textFieldBoundingBox!.x + 150, textFieldBoundingBox!.y + 150);
  await expect(textFieldNodeHudTag).toBeHidden();
  ```

test/integration/editor/index.spec.ts
JerryWu1234: thanks.
I got it wrong, I thought I needed to modify the tsx file before

Armen692244 (commented):
J

PR #2329, opened 2023-07-19
Fix some flakyness in the visual regression testing

Janpot:
I sometimes see an empty page turning up in argos. It seems that sometimes the screenshot is being taken before it's fully rendered. Argos doesn't care whether changes were approved or not. If it's flaky on master, too bad, it will become the new baseline. I introduced `page-ready=marker` to signal that the runtime DOM actually is being rendered. Wait for this marker before taking the screenshot for visual regression testing.

Also disabling animation for VRT screenshots as we seem to be having some non-determinism on the tabs animation after selection

Screenshot 2023-07-19 at 13 27 11

Also removed `waitForNavigation`, it wasn't doing anything for the cases where it was used as the condition was already fulfulled before this function is called.

Also adding `recharts` to the `optimizeDeps`, this also improves test performance, and flakyness


apedroferreira (approved):
Looks good!

PR #2334, opened 2023-07-20
[docs] [cli] Add `createFunction` deprecation notices

bharatkashyap:
Screenshot 2023-07-20 at 5 36 42 AM
Screenshot 2023-07-20 at 7 15 01 AM


Janpot (approved):
Nice! One thing that could be added is to log the warning only once

bharatkashyap:
> Nice! One thing that could be added is to log the warning only once

Only once per "device"?

packages/toolpad-app/src/server/functionsTypesWorker.ts
```diff
@@ -475,15 +475,17 @@ export default async function extractTypes({
...
-  console.warn(
-    `${chalk.yellow('warn')} - ${chalk.bold(
-      usingCreateFunction.length,
-    )} function are using the deprecated ${chalk.red(
-      'createFunction',
-    )} API. This will be removed from Toolpad in a future release. Please see ${chalk.underline(
-      chalk.blue('https://mui.com/toolpad/reference/api/create-function/'),
-    )} for migration information and updates.`,
-  );
+  if (usingCreateFunction.length > 0) {
```
Janpot, line 478: We could call it just once during the lifetime of the process, just like [`util.deprecate'](https://nodejs.org/api/util.html#utildeprecatefn-msg-code) does.

bharatkashyap: By storing a flag to check if the warning has been sent?

Janpot: Yes, unless `util.deprecate` already does the job reasonably well

packages/toolpad-app/src/server/functionsTypesWorker.ts
```diff
@@ -413,6 +413,7 @@ export default async function extractTypes({
   const checker = program.getTypeChecker();
 
   const usingCreateFunction = [];
+  let createFunctionWarningShown = false;
```
Janpot, line 416: Shouldn't this go in module scope? Otherwise it will log every time we call this function

Janpot: Also, have you tried just wrapping `createFunction` in `toolpad-core` with `util.deprecate`?

packages/toolpad-app/src/server/functionsTypesWorker.ts
bharatkashyap: I get https://vitejs.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility when trying to use `util` to wrap `createFunction`. 

Another benefit is being able to pinpoint how many functions are using the deprecated function, which is a helpful addition to the warning in my view

packages/toolpad-app/src/server/functionsTypesWorker.ts
bharatkashyap: I get https://vitejs.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility when trying to use `util` to wrap `createFunction`. 

Another benefit is being able to pinpoint how many functions are using the deprecated function, which is a helpful addition to the warning in my view

PR #2343, opened 2023-07-21
Chart component improvements

apedroferreira:
Improving chart according to discussion with Olivier:

- [x] Add icons for chart types/kinds
- [x] Autocomplete for chart x and y keys based on data provided
- [x] Make chart try to plot data automatically once given the data

https://github.com/mui/mui-toolpad/assets/10789765/ab2d9761-39e8-496f-a9a3-ae6921b48f9c


Janpot (approved):
interested in further exploring the heuristics in how to detect which properties of data result in series data. we should maybe spend some sync meeting time on this?

apedroferreira:
> interested in further exploring the heuristics in how to detect which properties of data result in series data. we should maybe spend some sync meeting time on this?

We can - I don't have many ideas besides what I did here right now, since it seems like a completely different problem than in the DataGrid for example. All I'm mostly doing is assuming that if something only has string values it's probably a category, so probably an x-key, otherwise it's probably a y-key. And I always pick the first property that is read that can be used.
Other than that it's easy for the user to change the keys that are automatically selected when the data is first added to the chart.

PR #2353, opened 2023-07-23
Live configurator POC

Janpot:

Preview: https://deploy-preview-2353--mui-toolpad-docs.netlify.app/toolpad/live-config/


docs/data/pages.ts
```diff
@@ -172,6 +172,10 @@ const pages: MuiPage[] = [
       },
     ],
   },
+  {
+    pathname: '/toolpad/live-config',
+    title: 'Live config',
```
oliviertassinari, line 177: to match the h1

```suggestion
    title: 'Live configuration',
```

docs/data/toolpad/live-config/index.md
```diff
@@ -0,0 +1,90 @@
+# Live Configurators
```
oliviertassinari, line 1: Sentence case

```suggestion
# Live configurators
```

docs/data/toolpad/live-config/index.md
```diff
@@ -0,0 +1,90 @@
...
+
+<p class="description">Update your datagrid configuration straight in your application</p>
+
+## Live configurators
```
oliviertassinari, line 5: Looks redundant.

```suggestion
```

docs/data/toolpad/live-config/index.md
```diff
@@ -0,0 +1,90 @@
+# Live Configurators
+
+<p class="description">Update your datagrid configuration straight in your application</p>
```
oliviertassinari, line 3:
```suggestion
<p class="description">Update your data grid configuration straight in your application.</p>
```

docs/data/toolpad/live-config/index.md
```diff
@@ -0,0 +1,90 @@
...
+
+  Don't compromise on performance, our generators produce the same optimized code that you would have hand-written. If the result satisfies your use-case
+
+## Demo
+
+Click the edit button to open the live configurator. Hit save when you're done and see the page update.
+
+{{"demo": "BasicExampleGrid.js"}}
+
+## installation
```
oliviertassinari, line 27:
```suggestion
## Installation
```

oliviertassinari (commented):
Ideas:

1. If I'm here as a developer to solve the problem of quickly configuring my columns and getting the exported code, I think I would rather upload a JSON sample of my data set.
2. If I'm here as a developer to solve the problem of quickly configuring my columns and getting the exported code, I imagine I could also open a toolpad self-hosted editor, connect my data, and use the export to code.
3. If I'm here to edit one of the components in isolation to use the data source, I'm not too versed in React, I don't want to manage too much code, then I feel the backdrop bottom popup UX feels off. I see myself configuring the columns pro-code rather than using this UI. Maybe the solution is to change the UX to place the component we are editing at the center of the screen, isolated from the rest of the page and showing all the controls around it.
4. The value of 3. might be more obvious if we add access to the long tail of props the data grid can support. It might also feel more obvious if we can automatically add all the columns provided in the data source.
5. In the case of 3. I feel I would want to do more than the data grid in a low-code fashion, to stay in the same flow as long as possible when developing the page.
6. I wonder if this island of customization isn't something that would work better for the CSS layout use cases, typically more or when you want to change the layout, or customize how components look and less for logic. Like maybe if I could customize components like https://www.mirrorful.com/ right inside my app.

Janpot:
> 6. I wonder if this island of customization isn't something that would work better for the CSS layout use cases, typically more or when you want to change the layout, or customize how components look and less for logic. Like maybe if I could customize components like https://www.mirrorful.com/ right inside my app.

A mashup of https://www.mirrorful.com/ with https://csspro.com/ :)

Janpot:
> Ideas:
> 
> 1. If I'm here as a developer to solve the problem of quickly configuring my columns and getting the exported code, I think I would rather upload a JSON sample of my data set.
> 2. If I'm here as a developer to solve the problem of quickly configuring my columns and getting the exported code, I imagine I could also open a toolpad self-hosted editor, connect my data, and use the export to code.
> 3. If I'm here to edit one of the components in isolation to use the data source, I'm not too versed in React, I don't want to manage too much code, then I feel the backdrop bottom popup UX feels off. I see myself configuring the columns pro-code rather than using this UI. Maybe the solution is to change the UX to place the component we are editing at the center of the screen, isolated from the rest of the page and showing all the controls around it.
> 4. The value of 3. might be more obvious if we add access to the long tail of props the data grid can support. It might also feel more obvious if we can automatically add all the columns provided in the data source.
> 5. In the case of 3. I feel I would want to do more than the data grid in a low-code fashion, to stay in the same flow as long as possible when developing the page.
> 6. I wonder if this island of customization isn't something that would work better for the CSS layout use cases, typically more or when you want to change the layout, or customize how components look and less for logic. Like maybe if I could customize components like https://www.mirrorful.com/ right inside my app.

These comments make a lot of sense. But they seem to approach it as a playground where one can visually create a grid configuration and then export the code. a bit like all those [CSS](https://cssgrid-generator.netlify.app/) [generators](https://cssgradient.io/) [that](https://www.cssportal.com/css-flexbox-generator/) [exist](https://www.cssmatic.com/box-shadow) [all](https://border-radius.com/) [around](https://webcode.tools/generators/css/keyframe-animation). We can venture into this area though, it may be easier to explain. The main difference with what's proposed in this PR is that they are always one way only. You generate the code, close the page and you can't go back to edit it. The toolpad live configurators aim to keep the code consumable, and low-code editable at the same time. Until you purposefully eject to break the ties and make it pro-code editable.

Issue #2340, opened 2023-07-21
Memory allocation issue


oliviertassinari:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

Steps:

1. Deploy a basic toolpad app in an environment with 512 MB of memory
2. Limit the max old space of Node.js to 396 MB `NODE_OPTIONS='--max-old-space-size=396'`

### Current behavior 😯

Crash memory allocation issue

### Expected behavior 🤔

Works

### Context 🔦

Trying to deploy the example of Toolpad in render.com

apedroferreira:
So I'm guessing something like `--optimize_for_size --max_old_space_size=460 --gc_interval=100` won't help?

Janpot:
Running some tests in docker

1. `docker run --rm -it -m 512m -w /app -v $PWD/examples/qr-generator:/app node sh`
2. `yarn`
3. `yarn build`
4. this fails due to memory issues
5. `NODE_OPTIONS='--max-old-space-size=396' yarn build`
6. fails due to memory issues as well
7. `NODE_OPTIONS='--max-old-space-size=450' yarn build`
8. this passes and builds the artifact

Let's try again without limits:
1. `docker run --rm -it -w /app -v $PWD/examples/qr-generator:/app node sh`
2. `yarn build`
3. this passes and builds the artifact

Let's try to run that build
1. `docker run --rm -it -m 512m -w /app -v $PWD/examples/qr-generator:/app node sh`
2. `yarn start`
3. this runs comfortably at around 311M memory usage

Memory issues seem to only happen at build, which is sort of expected to take quite some memory. It seems possible to build it in a container of 512m, if you tune it well. 512m is really not enough though, would love to see anyone attempt at building the docs website under those limitations.

oliviertassinari:
In render.com, it fails at the `yarn start` step. I haven't seen it fail during `yarn build`. Could it be https://github.com/prisma/prisma/issues/4972?

Janpot:
> Could it be https://github.com/prisma/prisma/issues/4972?

Perhaps, if the end-user application is using prisma. Toolpad doesn't use prisma. 

> In render.com, I believe fails at the `yarn start` step. 

I know some things that can be optimized in the function runtime. Will take another look at it.

oliviertassinari:
I have narrowed down a bit more the problem:

[...]

PR #2354, opened 2023-07-23
Decrease memory usage of functions in production mode

Janpot:
Toolpad started requiring more memory after some infrastructural updates to the function runtime. https://github.com/mui/mui-toolpad/issues/2340
This PR aims at reducing the memory footprint of the function runtime by eliminating a few unnecessary componnets

## Potential improvements

- [x] Avoid file watchers in production, treat the build as static
- [x] Avoid building function files in production, prebuild them during `toolpad build`. Run them only during `toolpad start`
- [x] avoid running the type extraction during `toolpad start`

## Will follow up with (if the build starts failing)

- [ ] run type extraction in the main process during `toolpad build`

## Progress

1. Drop of 50% on `tools-public`

    Screenshot 2023-07-25 at 17 41 57


apedroferreira (approved):
Great ideas to not use file watchers in production and prebuild what we can, for sure this will reduce the memory usage a lot and helps explain what was causing the issues in the first place :)
Everything looks good to me!

PR #2355, opened 2023-07-24
take place of lodash

### LINKED ISSUE #2310: Replace Lodash completely and remove the library from the project
### Duplicates

### Latest version

### Summary 💡

In sequence of https://github.com/mui/mui-toolpad/pull/1270, we should finish replacing and removing all usage of Lodash in the project.

### Examples 🌈

_No response_

### Motivation 🔦

_No response_

JerryWu1234:
fix #2310 

I replaced lodash with something else in the form.tsx file.


JerryWu1234:
@ryancogswell @Janpot 
I replaced lodash with something else in the form.tsx file and I created a draft pr.
please tell me if  I got it wrong

packages/toolpad-components/src/Form.tsx
```diff
@@ -3,11 +3,17 @@ import { Container, ContainerProps, Box, Stack, BoxProps } from '@mui/material';
 import { LoadingButton } from '@mui/lab';
 import { ArgTypeDefinitions, createComponent, useNode } from '@mui/toolpad-core';
 import { useForm, FieldValues, ValidationMode, FieldError, Controller } from 'react-hook-form';
-// TODO: Remove lodash-es
-// eslint-disable-next-line no-restricted-imports
-import { isEqual } from 'lodash-es';
 import { SX_PROP_HELPER_TEXT } from './constants';
 
+function isEqual(props: ValidationProps, previouProps: ValidationProps) {
```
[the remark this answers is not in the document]
Janpot: a more descriptive name would be helpful as well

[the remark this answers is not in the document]
JerryWu1234: > This seems like a good start! But it only does a **shallow** comparison - it only compares the values for the keys at the **top level** of an object. What if some properties of the object have nested objects themselves, and we also want to ensure that the values in the nested object are equal? This is necessary, for example, for file input values in a form, which are an object with specific properties.
> 
> So what we need here is a function that does a **deep** comparison: it goes through every nested object as well as well as the top level, and compares every value. And you would need to account for arrays too, as a similar thing can happen with nested arrays.
> 
> The function should be as efficient as possible, so it's good that it's returning `false` as soon as the values of two properties with the same name are not equal!

I will check `array,object,set,map`.and  how do you feel the name `isReferEqual`

JerryWu1234:
@apedroferreira please check way.
I will follow this way if it's well

packages/toolpad-components/src/Form.tsx
```diff
@@ -172,24 +170,27 @@ export default createComponent(Form, {
...
+}
+
+export interface FormInputValidationProps {
   isRequired: boolean;
   minLength: number;
   maxLength: number;
   isInvalid: boolean;
 }
 
+type ValidationProps = Partial<FormInputValidationProps>;
```
apedroferreira, line 184: I would not separate this type into `ValidationProps` as it is not reused anywhere - I would just write ```validationProps: Partial<FormInputValidationProps>``` below. 
But this is not a big deal!

packages/toolpad-utils/src/check.ts
```diff
@@ -0,0 +1,20 @@
+/**
```
apedroferreira, line 1: I would not create separate utilities for these checks, I think we can remove them and their tests until we see that we need to reuse this type of logic in many different places.
We can do the checks inline for now, as needed.

packages/toolpad-utils/src/collections.ts
```diff
@@ -93,3 +95,66 @@ export function filterKeys<U>(
 ): Record<string, U> {
   return mapProperties(obj, ([key, value]) => (filter(key) ? [key, value] : null));
 }
+
+/**
+ * Check if two values are equal - that is,
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function isReferEqual(obj: any, objToCompare: any): boolean {
```
apedroferreira, line 103: Would it be possible to use `unknown` types instead?

packages/toolpad-utils/src/collections.ts
```diff
@@ -93,3 +95,66 @@ export function filterKeys<U>(
 ): Record<string, U> {
   return mapProperties(obj, ([key, value]) => (filter(key) ? [key, value] : null));
 }
+
+/**
+ * Check if two values are equal - that is,
```
apedroferreira, line 100: The description seems cut off after "that is,"?
If you copied this function or any large parts of it from somewhere, we should reference the original source.

packages/toolpad-utils/src/collections.ts
apedroferreira, line 103: I would maybe call the function `isDeepEqual` since we're checking for deep equality here, not referential equality.
https://www.syncfusion.com/blogs/post/5-different-ways-to-deep-compare-javascript-objects.aspx

packages/toolpad-utils/src/collections.ts
```diff
@@ -93,3 +95,66 @@ export function filterKeys<U>(
...
+        const keysA = Object.keys(obj);
+        const keysB = Object.keys(objToCompare);
+        return (
+          keysA.length === keysB.length &&
+          keysA.every((key: string) => {
+            return isReferEqual(obj[key], objToCompare[key]);
+          })
+        );
+      }
+      /* istanbul ignore next */
```
apedroferreira, line 149: What do these comments about istanbul mean?

packages/toolpad-utils/src/collections.spec.ts
```diff
@@ -11,3 +11,45 @@ describe('asArray', () => {
     expect(asArray(got)).toEqual(expected);
   });
 });
+describe('isReferEqual', () => {
+  test('isReferEqual array', () => {
+    const m = [1, 2, 3];
+    const d = [1, 2, 4];
+    expect(isReferEqual(m, m)).toBe(true);
```
apedroferreira, line 18: This is the same variable `m` which I think means both values will be referentially equal?
So it might be a better test to pass a different instance of the same array as each argument -  you can pass it directly like `isReferEqual([1, 2, 3], [1, 2, 3])`, for example.
Same for similar cases below.

apedroferreira (changes requested):
Getting there!
If you got the new function from somewhere, it would be nice to reference the original source.

JerryWu1234:
> 

image
maybe it can't reach it if it only checks it by  `Object.keys`

apedroferreira:
> > 
> 
> image maybe it can't reach it if it only checks it by `Object.keys`

Ok, it's fine if the solution only considers objects and arrays, but feel free to have something more complete as long as it works well :)

packages/toolpad-utils/src/collections.ts
JerryWu1234: image
I tried this way but unfortunately, these codes don’t work. And I found a lot of materials about this that all use the any

JerryWu1234: image

packages/toolpad-utils/src/collections.ts
```diff
@@ -1,3 +1,24 @@
+/**
+ * Quick object check - this is primarily used to tell
+ * objects from primitive values when we know the value
+ * is a JSON-compliant type.
+ */
+export function isObject<T>(obj: T): boolean {
```
Janpot, line 6: Please remove `isObject`, `isSet`, `isMap` and `isArray`

packages/toolpad-utils/src/collections.ts
```diff
@@ -93,3 +114,65 @@ export function filterKeys<U>(
 ): Record<string, U> {
   return mapProperties(obj, ([key, value]) => (filter(key) ? [key, value] : null));
 }
+
+/**
+ * check if two objects are deeply equal and support  Set, Map, Date, Array, Object
+ */
+// Reference code https://github.com/vuejs/vue/blob/49b6bd4264c25ea41408f066a1835f38bf6fe9f1/src/shared/util.ts#L297
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function isDeepEqual(obj: any, objToCompare: any): boolean {
```
Janpot, line 123: So, the idea behind this ticket was not to one to one recreate lodash. It's about evaluating why the algorithm needs a nested equality check, and removing that need. The fact that we use this function seems to be rather a symptom that we're using a less than ideal algorithm.

I mean, if vendoring in some lodash functions would be all that's needed for this ticket, we would have simply done it and not opened a ticket at all.

packages/toolpad-utils/src/collections.ts
```diff
@@ -93,3 +93,65 @@ export function filterKeys<U>(
...
```
Janpot, line 102:
```suggestion
export function isDeepEqual<T>(obj: T, objToCompare: T): boolean {
```

Another reason to get rid of lodash is that it tends to turn type safe code into type unsafe code. If the types don't match, there's no reason to call this function in the first place. We try to avoid the `any` as much as we can. Prefer the `unknown` type when possible.

packages/toolpad-utils/src/collections.ts
JerryWu1234: > It's about evaluating why the algorithm needs a nested equality check, and removing that need.

I apologized. I do not totally understand this sentence.

packages/toolpad-utils/src/collections.ts
JerryWu1234: I also hope that uses the type-safe.
but I didn't find any case like this.
currently, I used the unknown

packages/toolpad-utils/src/collections.ts
JerryWu1234: I fixed it

packages/toolpad-utils/src/collections.ts
```diff
@@ -93,3 +93,65 @@ export function filterKeys<U>(
 ): Record<string, U> {
   return mapProperties(obj, ([key, value]) => (filter(key) ? [key, value] : null));
 }
+
+/**
+ * check if two objects are deeply equal and support  Set, Map, Date, Array, Object
+ */
+// Reference code https://github.com/vuejs/vue/blob/49b6bd4264c25ea41408f066a1835f38bf6fe9f1/src/shared/util.ts#L297
+export function isDeepEqual(obj: unknown, objToCompare: unknown): boolean {
```
Janpot, line 101:
```suggestion
export function isDeepEqual<T = unknown>(obj: T, objToCompare: T): boolean {
```

packages/toolpad-utils/src/collections.ts
JerryWu1234: If I get it right, you want to delete some redundant code that is not related to our industry, don’t you? Such as checking the binary and string type.

Janpot: We don't want to use lodash all over our codebase because
1. 90% of it can already be achieved using standard javsacript. No need for `_.map`, `_.filter`, ... it all exists in javascript. no need for `_.isEmpty`, `_.isObject`,... because you can write all of these with simple one-liners that everybody knows and understands. 
2. it is unmaintained.
3. its api design makes programmers turn type-safe code into type-unsafe code.

For this same reason we are not going to create a library like lodash, we will just simply aim at removing lodash.

packages/toolpad-utils/src/collections.ts
```diff
@@ -93,3 +93,65 @@ export function filterKeys<U>(
 ): Record<string, U> {
   return mapProperties(obj, ([key, value]) => (filter(key) ? [key, value] : null));
 }
+
+/**
+ * check if two objects are deeply equal and support  Set, Map, Date, Array, Object
+ */
+// Reference code https://github.com/vuejs/vue/blob/49b6bd4264c25ea41408f066a1835f38bf6fe9f1/src/shared/util.ts#L297
+export function isDeepEqual<T = unknown>(obj: T, objToCompare: T): boolean {
```
apedroferreira, line 101: Originally I thought that this function would be expected to accept object types only. But because it seems that it can compare any type of values, let's probably rename `obj` and `objToCompare` to `value` and `valueToCompare`.

apedroferreira (changes requested):
Just one more very small change and I think this is good to merge!

packages/toolpad-utils/src/collections.ts
```diff
@@ -93,3 +93,65 @@ export function filterKeys<U>(
 ): Record<string, U> {
   return mapProperties(obj, ([key, value]) => (filter(key) ? [key, value] : null));
 }
+
+/**
+ * check if two objects are deeply equal and support  Set, Map, Date, Array, Object
```
apedroferreira, line 98: Also change to: `check if two values`

apedroferreira (approved):
Please ignore my previous review - these are all technically `object` types so I think it's fine to keep this naming.
Looks good and I think we can merge!

JerryWu1234:
@JerryWu1234 @Janpot 
May I ask you about this pr?
Do we split this pr into more parts? Or all features in this pr?
Because this I only implemented one function

JerryWu1234:
> The original intent of "remove `lodash.isEqual`" is not "recreate `lodash.isEqual` in our own codebase". It is "the use of `lodash.isEqual` in this instance may indicate a bad algorithm, write a better one that doesn't need deep comparison". If this reasoning is not clear, I urge you to immediately stop spending time on this.
> 
> If it turns out we can't improve the algorithm, and we do need a deep comparison, I'd much rather stay with `lodash.isEqual` or use [`react-fast-compare`](https://www.npmjs.com/package/react-fast-compare) instead.
> 
> Some of the reasons we remove lodash are:
> 
> * JavaScript equivalents are terse enough (e.g. `isString` has no benefit over `typeof x === 'string'`)
> * It introduces unsafe typing (e.g. `.set` and `.get`)
> 
> As such, we don't want to simply replace `lodash` methods with self written equivalents. It is not an improvement.



> The original intent of "remove `lodash.isEqual`" is not "recreate `lodash.isEqual` in our own codebase". It is "the use of `lodash.isEqual` in this instance may indicate a bad algorithm, write a better one that doesn't need deep comparison". If this reasoning is not clear, I urge you to immediately stop spending time on this.
> 
> If it turns out we can't improve the algorithm, and we do need a deep comparison, I'd much rather stay with `lodash.isEqual` or use [`react-fast-compare`](https://www.npmjs.com/package/react-fast-compare) instead.
> 
> Some of the reasons we remove lodash are:
> 
> * JavaScript equivalents are terse enough (e.g. `isString` has no benefit over `typeof x === 'string'`)
> * It introduces unsafe typing (e.g. `.set` and `.get`)
> 
> As such, we don't want to simply replace `lodash` methods with self written equivalents. It is not an improvement.


make me confuse 
image

your teammate's wishes need a deep comparison.
I feel like I'm missing out on something

JerryWu1234:
@Janpot please take a look

Janpot:
@JerryWu1234 

1. As I said before the intent was not to recreate `lodash`. I'm sorry for the confusion, we should've communicated this better. We will probably close https://github.com/mui/mui-toolpad/issues/2310. The intent was to stop relying more on `lodash`.
2. As you've sunk a lot of effort into it, let's restrict this PR to replacing the `isEqual` method only, let's not add/replace any other methods.
3. As said, if we can't remove the need for deep comparison, we would probably prefer a tried and tested [solution](https://www.npmjs.com/package/react-fast-compare) instead of rolling our own.

JerryWu1234:
@Janpot 

thanks for your reply.

Where can I contact you unless the GitHub?

I wanted to discuss this problem with you before
But I'm no idea where I can contact you.

 Don't we maybe need to extract a new isequal function?
the inline also probably is a good idea because the isequal is less place and the place that was used is just a simple comparison

JerryWu1234:
@Janpot, Could you have some free time so that we can discuss this problem?

JerryWu1234:
> Feel free to join the [discord server](https://mui.com/r/discord/). But bear in mind, this is a very low-priority issue, and initially mostly a reminder for the Toolpad team itself on how we want to handle `lodash` going forward. I don't have a lot of bandwidth to spend on this at the moment.

image
I can't send anything.
My nickname is `Jerry` in discord serve.

Janpot:
> I can't send anything.

You can follow the instructions in the `welcome` challenge to get access

JerryWu1234:
> Feel free to join the [discord server](https://mui.com/r/discord/). But bear in mind, this is a very low-priority issue, and initially mostly a reminder for the Toolpad team itself on how we want to handle `lodash` going forward. I don't have a lot of bandwidth to spend on this at the moment.

PR #2356, opened 2023-07-24
[docs] Fine for production

oliviertassinari:
Since it's in beta, it's fine now?

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,8 +39,6 @@ MUI Toolpad is in its beta stages of development. Feel free to run this applicat
 
 ## Quick setup locally
 
-**NOTE: MUI Toolpad is in preview, do not run this in production.**
-
 Run:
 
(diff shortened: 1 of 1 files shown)


bharatkashyap (approved):
Agreed

Janpot (approved):
Fine with removing the warning. Not sure whether beta software should be explicitly advertised as "production ready" though.

prakhargupta1 (approved):

oliviertassinari:
> Not sure whether beta software should be explicitly advertised as "production ready" though.

Agree, the goal of this PR is to keep it undefined: it can both be OK and no-OK in production, depending on your risk tolerance. I think that developers should be able to define this for themselves at this stage, based on, e.g. our version number, npm downloads, examples, etc.

PR #2363, opened 2023-07-25
Fix rest tests flakyness around env reloading

Janpot:
- [x] reset the fixture env file after changing it
- [x] reset environment variables before reloading so that we pick up removal of variables from the env file

The rest tests are flaky because of this. The second test starts with env variables left from the first test. They don't reset. The second test always fails because of this the first time around. It always passes because playwright retries and runs the second test in isolation in which case it starts with fresh env variables. This is a big drawback of using retries, it can hide legitimate bugs.

### DIFF
--- a/test/integration/rest-basic/index.spec.ts
+++ b/test/integration/rest-basic/index.spec.ts
@@ -5,6 +5,7 @@ import { test, expect } from '../../playwright/localTest';
 import { ToolpadRuntime } from '../../models/ToolpadRuntime';
 import { ToolpadEditor } from '../../models/ToolpadEditor';
 import { startTestServer } from './testServer';
+import { withTemporaryEdits } from '../../utils/fs';
 
 test.use({
   localAppConfig: {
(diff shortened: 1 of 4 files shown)


test/utils/fs.ts
```diff
@@ -0,0 +1,13 @@
+import * as fs from 'fs/promises';
+
+export async function withTemporaryEdits<T = void>(
+  filePath: string,
+  doWork: () => Promise<T>,
+): Promise<T> {
+  const envOriginal = await fs.readFile(filePath, 'utf-8');
```
apedroferreira, line 7: This is not necessarily just for env files, right? So this variable could be called `originalFile`?

test/integration/rest-basic/index.spec.ts
```diff
@@ -43,11 +44,12 @@ test('rest runtime basics', async ({ page, localApp }) => {
   await expect(page.getByText('query4 authorization: foo')).toBeVisible();
 
   const envFilePath = path.resolve(localApp.dir, './.env');
-  await fs.writeFile(envFilePath, 'TEST_VAR=bar');
-
-  await page.reload();
-
-  await expect(page.getByText('query4 authorization: bar')).toBeVisible();
+  await withTemporaryEdits(envFilePath, async () => {
```
apedroferreira, line 47: Seems good! A possible alternative could maybe be a utility function that we could pass the file path and content to, which would write to the file, and would return a function that we could call in order to undo that edit. Not sure if that would be cleaner.

apedroferreira (approved):
Looks good, thanks! I had noticed some occasional issues here already...

I added a small comment and an idea for an alternative solution, but nothing that should be blocking!

test/integration/rest-basic/index.spec.ts
Janpot: This is sort of the poor person's version of [explicit resource management](https://github.com/tc39/proposal-explicit-resource-management)

apedroferreira: Oh I didn't consider the test failing, this seems solid then!

PR #2371, opened 2023-07-25
Auto parse date strings on data grids

bharatkashyap:
Small improvement to be able to automatically detect if a string is a "date string" and set its column to "date"

| Before  | After  | 
| ------  | ------ | 
| Screenshot 2023-07-25 at 5 21 13 PM | Screenshot 2023-07-25 at 5 21 35 PM |


packages/toolpad-components/src/DataGrid.tsx
```diff
@@ -71,6 +71,19 @@ function randomBetween(seed: number, min: number, max: number): () => number {
   return () => min + (max - min) * random();
 }
 
+function isValidDate(inputString: string) {
+  // Check if the input string contains only digits (simple numeric string)
+  if (/^\d+$/.test(inputString)) {
```
Janpot, line 76: Should we parse these columns as type 'number' then?

packages/toolpad-components/src/DataGrid.tsx
```diff
@@ -215,7 +231,12 @@ function ImageCell({ field, id, value: src }: GridRenderCellParams<any, any, any
 }
 
 function dateValueGetter({ value }: GridValueGetterParams<any, any>) {
-  return typeof value === 'number' ? new Date(value) : value;
+  try {
+    const date = new Date(value);
+    return date;
+  } catch (err) {
+    return value;
```
Janpot, line 238: When does the `Date` constructor throw?

I believe we should implement this as

```tsx
if (value === null || value === undefined || value === '') {
  return value
}
// It's fine if this turns out to be an invalid date, the user wanted a date column, if the data can't parse as a date
// it should just show as such
return new Date(value)
```

packages/toolpad-components/src/DataGrid.tsx
```diff
@@ -71,17 +71,15 @@ function randomBetween(seed: number, min: number, max: number): () => number {
   return () => min + (max - min) * random();
 }
 
-function isValidDate(inputString: string) {
-  // Check if the input string contains only digits (simple numeric string)
-  if (/^\d+$/.test(inputString)) {
-    return false;
+function isNumeric(input: string) {
+  if (/^\d+$/.test(input)) {
```
Janpot, line 75: This fails to detect
* `1.23`
* `1e23`
* `1_2_3`
* ...?

How about
```tsx
return input ? !Number.isNaN(Number(input)) : false
```

Janpot (approved):

oliviertassinari:
I have tested the change with:

```diff
diff --git a/tools-private/toolpad/resources/invoice.ts b/tools-private/toolpad/resources/invoice.ts
index 89ee04b..bf416ca 100644
--- a/tools-private/toolpad/resources/invoice.ts
+++ b/tools-private/toolpad/resources/invoice.ts
@@ -74,7 +74,8 @@ export const getOverdueOrders = async function getVendors() {
   await connection.end();
   return (rows as any).map((row: any) => ({
     ...row,
-    invoice_date: dayjs(row.invoice_date).format('MMMM D, YYYY'),
+    invoice_date: row.invoice_date,
     invoice_due: dayjs(row.invoice_due).format('MMMM D, YYYY'),
   }));
 };
```

Before:

Screenshot 2023-08-30 at 13 46 31

After:

Screenshot 2023-08-30 at 13 44 36

Nice 👌

PR #2373, opened 2023-07-25
experiment with memory improvements

Janpot:
Seeing 50% improvement in memory in the runtime when running this version on tools-public. Let's see if we see the same improvement with this example and whether the memory increase on the build survives the render.com free plan


Janpot:
🤔  Why isn't this creating a PR preview like https://github.com/mui/mui-toolpad/pull/2365#issuecomment-1649465967 ?

Janpot:
Gonna try and set it to ready for review, but you can ignore this PR

PR #2376, opened 2023-07-26
Remove @mui/x-data-grid-pro

Janpot:
- Remove non-MIT dependency `@mui/x-data-grid-pro`
- Remove bundled MUI X license


bharatkashyap (approved):

oliviertassinari:
To be sure. Do we still have DataGrid Pro for the editor?

Janpot:
> Do we still have DataGrid Pro for the editor?

This PR completely removes it. The only feature we lose will be the possibility of resizing the datagrid columns by drag and drop. They can still be resized using an input field on the column definition.

If we want to keep pro grid in the editor we will have to figure out how we will license and distribute the X license key that is needed for this feature:
- Can we mark a portion of the code as non-MIT and how?
- What does it mean to depend on non-MIT packages?

oliviertassinari:
> This PR completely removes it. The only feature we lose will be the possibility of resizing the datagrid columns by drag and drop. They can still be resized using an input field on the column definition.

This is a deal-breaker for the change IMHO, we need the UX of the editor to be great.

Janpot:
> Today's he's not licensed for it, but it's not clear.

I recently added a feature in X that allows us to scope the license to our components only. This part we can solve. On a technical level we can now make it so that Toolpad components can use pro datagrid without watermark and user components will show the watermark.

The concerns that remain are more from a legal nature:

- Toolpad will need to have a dependency on `@mui/x-data-grid-pro` which is not MIT, what's the implication for the open-source license of Toolpad. Can we still call it MIT? Or de we need to change the license?
- Toolpad will need to be bundled with a valid license key. Which software license will apply to this bundled key? MIT? Does that mean that anybody can just use that license key for any purpose they want?

oliviertassinari:
> Can we still call it MIT?

The combined license isn't MIT, but Toolpad's code itself is MIT. MUI X Pro and Toolpad are license-compatible https://en.wikipedia.org/wiki/License_compatibility they have no contradictions as long as MUI issues a custom license, e.g. to avoid https://mui.com/legal/mui-x-eula/#deployment. 

> Toolpad will need to be bundled with a valid license key. Which software license will apply to this bundled key? MIT? Does that mean that anybody can just use that license key for any purpose they want?

A license key has no value on its own, it's only here to help developers follow the license terms. The license key is between MUI X Pro and Toolpad. If we don't precise anything, https://mui.com/legal/mui-x-eula/ applies to it.

Janpot:
👍  Thanks, got it. We'll keep it as it is and instead scope the license to our components only.

PR #2381, opened 2023-07-26
Avoid running the server in a separate process

Janpot:
Closes https://github.com/mui/mui-toolpad/issues/2340

This PR builds further on https://github.com/mui/mui-toolpad/pull/2354 and concludes the memory improvements for https://github.com/mui/mui-toolpad/issues/2340

This refactor is aiming at removing the "server" process. The only real reason it is still there is because a big chunk of toolpad expects the project directory location to be available in a `TOOLPAD_PROJECT_DIR` environment variable. We set this variable when starting the process. If we want to remove the process, we will need to first remove the reliance on this global variable. 

The bulk of this PR is a refactor to inject the toolpad project everywhere on the backend instead of reading it from the environment

before (local):

Screenshot 2023-07-28 at 15 36 05

3 toolpad processes totalling to roughly 250m

after: 

Screenshot 2023-07-28 at 15 34 12

2 toolpad processes totalling to roughly 190m

Which is about 25% decrease in memory


apedroferreira (approved):

oliviertassinari:
Great to see improvements in memory consumption. I had a look at the min memory requirement for heavy apps as I was curious:

- Strapi https://docs.strapi.io/dev-docs/deployment#hardware-and-software-requirements 2GB
- Metabase https://www.metabase.com/docs/latest/installation-and-operation/running-metabase-on-elastic-beanstalk 2GB

But we can't compare Toolpad to these products so much, a bit of a different category.

I tried to upgrade in https://github.com/oliviertassinari/mui-olivier-tools/pull/3 to v0.1.22. It failed. In the last commit, I created a dummy express server to better isolate the problem.

Maybe it's an issue with the initial memory allocated during the start of the process?

PR #2383, opened 2023-07-26
Remove redundant integration tests section from contribution docs

apedroferreira:
This section seems to have duplicate information, so we could condense the 2 sections about integration tests into 1? 
Or did I miss something?


Janpot (approved):
Not sure why it's there twice 😄

PR #2385, opened 2023-07-27
Update npm stats example to use new chart component

apedroferreira:
As far as I can tell it looks the same, maybe the x-axis is even an improvement as it shows the starting and end value of the whole range.

Screenshot 2023-07-27 at 18 33 39

Also added a few small features/adustments that were in the custom chart component.


prakhargupta1:
The current background grid of the chart is too bold. Can we make it a dotted line?
Something like in current charts: https://npm-stats.onrender.com/prod/pages/evZC-gp

Janpot (approved):

PR #2393, opened 2023-07-30
Add a new query panel

bharatkashyap:
Closes: https://github.com/mui/mui-toolpad/issues/2138

https://github.com/mui/mui-toolpad/assets/19550456/07fa4a29-cfd4-4495-9741-a5a273aa572c

### TODO

- https://excalidraw.com/#json=AQ-1k8DkwEVa2RUtGCQMG,TwZB3D-vbIgvCysL-mJ8mw

- [x] Mode: Add a label in the explorer to distinguish automatic and manual queries and create separate buttons for creating them

- [x] Settings: Either use tabs, or a settings cog icon which switches the display to settings, the icon switching to close. Explorer either experience and choose one.

- [x] Response tabs: The preview should be one tab in a set which could include the network response etc. in a vertically expandable section.

- [x] Scrolling: Panel should be expandable vertically.

- [x] Tests: Edit tests to adapt to query panel

- [x] Saving: Bunch changes to the queries and apply via `Ctrl/Cmd + S`.

- [x] Unsaved changes: Detect click-away from the query tab and show a dialog to manage unsaved changes; Show an unsaved changes indicator on query tab

- [x]  Order of panels: Pages, Page Hierarchy, Queries, Actions

- [x]  Make Actions/Queries separate sections

- [x]  “Save” button in the top panel, "Preview" button in the Preview/Dev Tools panel

- [x]  When creating a new query, it immediately saves the query to the project. Instead it should create it in draft and let me save or discard it when I'm done. There shouldn't be any query in the project because that makes the runtime immediately start executing it. This is how it currently works and I believe we should keep this behavior.

- [x]  Differentiate queries/actions in binding modal

- [x] Scroll to new file on “New File” click

- [x] Fix functions autocomplete anchor position?

- [ ]  Try and reduce the attention the sub-tabs draw


apedroferreira:
Wow, this looks great! 
Here's some of my UI/UX feedback from trying it out:
- Pressing "+ New file" in a custom query/action doesn't seem to do anything for me?
- The action settings have options such as "Set refetch interval:" - shouldn't these be for queries only?
- The popover where we can pick a custom function can shift its position if the number of options changes - for example if I have 20 results, but I search to filter them down to 1 result only, the popover goes below the button instead of on top. This can be a bit confusing.
- The binding modal still doesn't tell apart queries from actions in the scope explorer. Not sure if it should.
- This blue box's padding seems a bit off:
Screenshot 2023-10-31 at 17 35 51

Also not sure if the new input font is too small or I was just too used to the previous size...

The popover for picking functions looks really nice, I might try to implement some of it in the global functions explorer. Also this UI and the global functions explorer are a bit redundant right now, but I guess they might differ more when the global functions explorer also manages other global resources...

apedroferreira (changes requested):
I've added mostly UI/UX feedback, let me know if there's any part of the code in particular where you'd like me to take a closer look since there are so many changes - what I've glanced over seemed fine.

packages/toolpad-app/src/toolpad/AppState.tsx
```diff
@@ -294,6 +329,327 @@ export function appStateReducer(state: AppState, action: AppStateAction): AppSta
...
+      if (state.currentView.kind === 'page' && state.currentView.view?.kind === 'query') {
+        const queryTabs = state.currentView.queryPanel?.queryTabs || [];
+        const currentTabIndex = state.currentView.queryPanel?.currentTabIndex;
+        if (currentTabIndex !== undefined && queryTabs) {
+          return update(state, {
+            currentView: {
+              ...state.currentView,
+              queryPanel: {
+                ...state.currentView.queryPanel,
+                queryTabs: state.currentView?.queryPanel?.queryTabs?.map((tab, index) => {
```
apedroferreira, line 342: Maybe we can create some utility that makes operations like this shorter and more readable?

bharatkashyap:
> * Pressing "+ New file" in a custom query/action doesn't seem to do anything for me?

Ah, I understand - the search box doesn't scroll when the new file is created, so it's not immediately apparent that something happened. I think I can aim to make the option scroll to the one just created?

https://github.com/mui/mui-toolpad/assets/19550456/3b43fe65-1100-4b83-ace0-b75b21bcea80

> The action settings have options such as "Set refetch interval:" - shouldn't these be for queries only?

They are there but disabled - agreed, it's hard to spot that these are disabled without actually clicking to edit. I can make them more visibly disabled as well.

> The popover where we can pick a custom function can shift its position if the number of options changes - for example if I have 20 results, but I search to filter them down to 1 result only, the popover goes below the button instead of on top. This can be a bit confusing.

Although I couldn't reproduce, it might be problematic. I'll see if there's a way to fix the anchor position of the search element.

> The binding modal still doesn't tell apart queries from actions in the scope explorer. Not sure if it should.

Good catch - I can explore if we can make this happen without too much change in the internals of the nodes.

> This blue box's padding seems a bit off:

Agreed, I'll fix the alignment

> Also not sure if the new input font is too small or I was just too used to the previous size...

We can definitely bring it up a notch again if everyone thinks this is too small, but I do feel the current size is too big

> The popover for picking functions looks really nice, I might try to implement some of it in the global functions explorer. Also this UI and the global functions explorer are a bit redundant right now, but I guess they might differ more when the global functions explorer also manages other global resources...

We can explore how we can reuse this in the global functions explorer - I think it won't be redundant to the extent that the global explorer is focused on managing the functions throughout the pages, whereas this is more focused on individual queries/actions per page

apedroferreira:
> > * Pressing "+ New file" in a custom query/action doesn't seem to do anything for me?
> 
> Ah, I understand - the search box doesn't scroll when the new file is created, so it's not immediately apparent that something happened. I think I can aim to make the option scroll to the one just created?

Maybe if we scroll to the new file + highlight it in blue or something it will be more obvious what happened?


Also, if I brought up too many things feel free to just address the ones that feel easier or more important and we can improve the rest later!

bharatkashyap:
> > > * Pressing "+ New file" in a custom query/action doesn't seem to do anything for me?
> > 
> > 
> > Ah, I understand - the search box doesn't scroll when the new file is created, so it's not immediately apparent that something happened. I think I can aim to make the option scroll to the one just created?
> 
> Maybe if we scroll to the new file + highlight it in blue or something it will be more obvious?

Yes, that makes sense

> Also, if I brought up too many things feel free to just address the ones that feel easier or more important and we can improve the rest later!

No I think they're all relevant and very specific - so extremely useful to refine this and move towards merging!

Janpot:
Some things I noticed after running this PR

* I'm seeing a small line next to the function selection button. Any idea what causes this?

    Screenshot 2023-11-07 at 11 04 19


* as we touched already on the grooming call. It feels a bit strange to have a tab bar. Its main use is to indicate which query has unsaved changes. But it has some disadvantages as well:
     * It's unclear which tabs are queries and which tabs are actions. In the editor itself it's also not really indicated, except on the options page. I feel like it's an important enough concept to indicate it more clearly.
     * It takes up vertical space

  I feel like if we indicate in the left side navigation tree view which query is currently active, and and an unsaved changes indicator, the tab bar becomes mostly unnecessary.

* Opening this menu should automatically focus the search box so that I can immediately start typing
    
    Screenshot 2023-11-07 at 11 28 15

* When creating a new query, it immediately saves the query to the project. Instead it should create it in draft and let me save  or discard it when I'm done. There shouldn't be any query in the project because that makes the runtime immediately start executing it. This is how it currently works and I believe we should keep this behavior.

bharatkashyap:
I've updated to accommodate all our feedback from the grooming meetings - feel free to review partially since this is very big pull request. 

I'll update this with documentation changes once we have approvals!

apedroferreira:
Looks great, I haven't found any bugs when I tried it and it seems intuitive! Will look at the code next.
Here are the only possible issues I found, but they don't need to block this PR, they can be improved later if needed:

- The "Run" and "Save" button locations and types seem fine to me, but design-wise it feels like they need more vertical space around them. Not sure what the best way to improve that would be though.

Screenshot 2023-11-23 at 16 37 41

- Creating a new file in a custom function doesn't let us name the file, right? And the button at the bottom does not take the full-width of the popover so that may feel a bit weird. We can maybe improve the UX here later!

Screenshot 2023-11-23 at 16 37 34

- I see that actions still have a "Run to be available on the page" option in their settings. Isn't that what makes something a query as opposed to an action, though?

apedroferreira (approved):

packages/toolpad-app/src/utils/platform.ts
```diff
@@ -0,0 +1,4 @@
+export function isMac(): boolean {
```
apedroferreira, line 1: These kinds of utilities usually are pretty brittle and don't always work in my experience, do we still need it?

Janpot: Yeah, better to avoid exposing the platform, but abstract it instead. e.g.
```tsx
export function getModifierKey(): string {
  return isMac() ? '⌘' : 'Ctrl'
}
```

Nitpick though :)

bharatkashyap:
> Looks great, I haven't found any bugs when I tried it and it seems intuitive! Will look at the code next. Here are the only possible issues I found, but they don't need to block this PR, they can be improved later if needed:
> 
> * The "Run" and "Save" button locations and types seem fine to me, but design-wise it feels like they need more vertical space around them. Not sure what the best way to improve that would be though.
> 
> Screenshot 2023-11-23 at 16 37 41

> * Creating a new file in a custom function doesn't let us name the file, right? And the button at the bottom does not take the full-width of the popover so that may feel a bit weird. We can maybe improve the UX here later!

- Agree, we can probably batch the design and minor UX improvements in an issue following this PR.

> Screenshot 2023-11-23 at 16 37 34
> * I see that actions still have a "Run to be available on the page" option in their settings. Isn't that what makes something a query as opposed to an action, though?

Yes, that's how one would change an action to be a query

packages/toolpad-app/src/utils/platform.ts
```diff
@@ -1,4 +1,8 @@
-export function isMac(): boolean {
+function isMac(): boolean {
   const userAgent = navigator.userAgent;
   return /Mac|iPod|iPhone|iPad/.test(userAgent);
 }
+
+export function getModifierKey(): string {
+  return isMac() ? '⌘S' : '<kbd>Ctrl </kbd>S';
```
Janpot, line 7:
```suggestion
  return isMac() ? '⌘' : 'Ctrl';
```

Janpot: Actually, this could even be just a constant:

```tsx
export MODIEFIER_KEY = isMac() ? '⌘' : 'Ctrl';
```

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/QueryEditor/index.tsx
```diff
@@ -11,11 +11,17 @@ import * as appDom from '../../../../appDom';
...
+import { getModifierKey } from '../../../../utils/platform';
 import useUnsavedChangesConfirm from '../../../hooks/useUnsavedChangesConfirm';
 
 function SaveShortcutIndicator() {
-  return <span style={{ opacity: 0.55, marginLeft: 6 }}>{isMac() ? '⌘' : <kbd>Ctrl</kbd>}S</span>;
+  return (
+    <span
+      style={{ opacity: 0.55, marginLeft: 6 }}
+      // eslint-disable-next-line react/no-danger
+      dangerouslySetInnerHTML={{ __html: getModifierKey() }}
```
Janpot, line 22: I will never recommend something that requires you to use `dangerouslySetInnerHTML` without explicitly stating so.

bharatkashyap: Noted 😅

Janpot:
I'm trying this out with 

```bash
yarn toolpad dev test/integration/backend-basic/fixture/toolpad
```
But I don't see my existing queries or actions show up in the navigation of the page http://localhost:3000/_toolpad/app/pages/basic:


Screenshot 2023-12-11 at 17 50 22

This page should have many queries and actions

bharatkashyap:
> I'm trying this out with
> 
> ```shell
> yarn toolpad dev test/integration/backend-basic/fixture/toolpad
> ```
> 
> But I don't see my existing queries or actions show up in the navigation of the page http://localhost:3000/_toolpad/app/pages/basic:
> 
> Screenshot 2023-12-11 at 17 50 22
> 
> This page should have many queries and actions

Interesting, I'm running `yarn toolpad dev test/integration/backend-basic/fixture/toolpad --dev` on the latest commit locally and seeing this: 

Screenshot 2023-12-12 at 12 59 04 AM

Janpot:
Also when clicking the one of the "create query" buttons, nothing happens and I get an error in the console

```
index-wJpWncMs.js:40 Uncaught Invariant Violation: handleCreateNode should only be used for queries, which should always belong to a page
```


Also for some reason the styling of the button text feels a bit off. They should probably not be primary color as well.
Screenshot 2023-12-12 at 10 16 00

Janpot:
Ok managed to get it to work. Cleaned out node modules and pulled the latest changes.

Some things I noticed:


https://github.com/mui/mui-toolpad/assets/2109932/14ac03e6-36f4-42d2-a8d2-577c36afa1dc



* Resizing doesn't work
* When opening page parameters, the panel closes, even with unsaved changes

bharatkashyap:
> Ok managed to get it to work. Cleaned out node modules and pulled the latest changes.
> 
> Some things I noticed:
> 
>  Screen.Recording.2023-12-12.at.10.31.22.mov 
> * Resizing doesn't work
> * When opening page parameters, the panel closes, even with unsaved changes

Fixed both of those; the resizing was a strange bug I didn't completely investigate but got around for now. The page parameters dialog I think should be allowed to coexist with an open panel, so I made that behaviour happen.

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/ComponentCatalog/ComponentCatalog.tsx
```diff
@@ -40,11 +40,11 @@ const FUTURE_COMPONENTS = new Map<string, FutureComponentSpec>([
   ['Radio', { url: 'https://github.com/mui/mui-toolpad/issues/744', displayName: 'Radio' }],
 ]);
 
-const WIDTH_COLLAPSED = 40;
+export const COMPONENT_CATALOG_WIDTH_COLLAPSED = 40;
```
Janpot, line 43:
```suggestion
const COMPONENT_CATALOG_WIDTH_COLLAPSED = 40;
```

packages/toolpad-app/src/components/EditableText.tsx
```diff
@@ -73,15 +75,21 @@ const EditableText = React.forwardRef<HTMLInputElement, EditableTextProps>(
 
     const handleChange = React.useCallback(
       (event: React.ChangeEvent<HTMLInputElement>) => {
+        if (readOnly) {
```
Janpot, line 78: If this gets called while the input is readonly, that would mean there is an error in the code, right?
In that case, would it make more sense to make this an invariant? 


```suggestion
        invariant(!readOnly, 'Readonly input should be disabled')
```

Janpot (approved):
Went through all the files. This looks great! I sprinkled a few comments here and there, nothing blocking.

packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -563,8 +563,12 @@ function parseBindings(
     }
 
     if (appDom.isQuery(elm)) {
+      let kind: 'query' | 'action' = 'query';
```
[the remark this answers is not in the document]
bharatkashyap: Understood, I added "action" as a way to progressively perhaps do away with mutation in the code and eventually create a separate node type called "action", since if that is what we want to call it in the UI, should it also not be what we describe it in the codebase for simplicity?

packages/toolpad-app/src/toolpadDataSources/googleSheets/client.tsx
```diff
@@ -81,7 +81,7 @@ function QueryEditor({
 
   const handleSpreadsheetChange = React.useCallback(
     (event: React.SyntheticEvent<Element, Event>, newValue: GoogleDriveFile | null) => {
-      setInput((existing) => {
+      setInput?.((existing) => {
```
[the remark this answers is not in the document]
bharatkashyap: The `onChange` functionality for the new query editor client started living in the `appStateApi`, so we don't need to pass in an `onChange` handler prop anymore

packages/toolpad-app/src/constants.ts
```diff
@@ -18,6 +18,8 @@ export const TOOLPAD_BRIDGE_GLOBAL = '__TOOLPAD_BRIDGE__';
 
 export const VERSION_CHECK_INTERVAL = 1000 * 60 * 10;
 
+export const PAGE_PANEL_WIDTH = 250;
```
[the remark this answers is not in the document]
bharatkashyap: Do you mean directly defining the width in the consuming file instead of creating a constant for this?

packages/toolpad-app/src/toolpadDataSources/mysql/client.tsx
```diff
@@ -12,6 +12,7 @@ function MySQLConnectionParamsInput({
 
 const dataSource: ClientDataSource<SqlConnectionParams, SqlQuery> = {
   displayName: 'MySQL',
+  isEnabled: false,
```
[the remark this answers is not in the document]
bharatkashyap: Agreed, this can be done without

packages/toolpad-app/src/runtime/ToolpadApp.tsx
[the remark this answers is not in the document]
Janpot: Personally I prefer either one way or the other:
* either we call it "action" everywhere and just do a translation to "mutation", right before we read/write to a file
* or we call it "mutation" everywhere and name it "action" just right before we display in the UI

The reason is that this prevents us from mismatching and constant translation all around the codebase. Ideally if the rename is done in a large PR, you start with the latter, just to avoid introducing many unnecessary chamges to unrelated code. Then in a separate PR update the codebase to the former.

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/QueryEditor/QueryEditorPanel.tsx
```diff
@@ -0,0 +1,233 @@
...
+        <Typography fontSize={12} sx={{ alignSelf: 'center' }}>
+          Set query mode:
+        </Typography>
+        <TextField
+          select
+          label="mode"
+          value={draft?.attributes?.mode ?? 'query'}
+          onChange={handleModeChange}
+          sx={{
+            '& .MuiInputLabel-root': { fontSize: 12 },
```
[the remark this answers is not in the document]
Janpot: I noticed this all around the PR, but I only mentioned it once.
Non-blocking as well

packages/toolpad-app/src/toolpadDataSources/googleSheets/client.tsx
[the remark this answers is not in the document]
Janpot: But functionally, how should it behave when no handler is defined. What's the use-case?

packages/toolpad-app/src/constants.ts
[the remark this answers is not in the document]
Janpot: It's still a constant, but no need for it globally. It's used in only one file. And it's the sort of thing that probably should always be local.

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/QueriesExplorer/index.tsx
```diff
@@ -0,0 +1,502 @@
...
+    },
+    [labelText, validateItemName],
+  );
+
+  const handleClick = React.useCallback(() => {
+    invariant(toolpadNodeId, 'DataTreeItem should only be used for nodes with a toolpadNodeId');
+    onSelectNode?.(toolpadNodeId);
+  }, [onSelectNode, toolpadNodeId]);
+
+  const queryCreationMode = React.useMemo(() => {
```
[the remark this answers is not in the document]
bharatkashyap: > Ideally, the DataTreeItem doesn't know if it's a query or a mutation. In a true abstraction, we should be able to introduce a third query type without needing to alter DataTreeItem.

Agreed, naming it `DataTreeItem` was an attempt to begin making it agnostic to the type of item represented but I imagine that'll have to be a separate effort distinct from this one or any future improvements scoped to the query panel

> Also, I don't fully understand, do all tree nodes in the queries tree have the same node id? ids must be unique, otherwise we can't refer to the individual items in the treeview APIs. And in an ideal world they are opaque as well. (Meaning: we can't gain any knowledge about the object just by inspecting the id on its own)

The node ids are separate for each node

packages/toolpad-app/src/toolpad/propertyControls/boolean.tsx
```diff
@@ -14,7 +14,9 @@ function BooleanPropEditor({ propType, label, value, onChange, disabled }: Edito
   return (
     <PropertyControl propType={propType}>
       <FormControlLabel
-        control={<Checkbox checked={!!value} disabled={disabled} onChange={handleChange} />}
+        control={
+          <Checkbox size="small" checked={!!value} disabled={disabled} onChange={handleChange} />
```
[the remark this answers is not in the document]
bharatkashyap: Yep, you're right - this isn't making a difference

packages/toolpad-app/src/components/EditableText.tsx
```diff
@@ -75,9 +76,7 @@ const EditableText = React.forwardRef<HTMLInputElement, EditableTextProps>(
 
     const handleChange = React.useCallback(
       (event: React.ChangeEvent<HTMLInputElement>) => {
-        if (readOnly) {
-          return;
-        }
+        invariant(!readOnly, 'Readonly input should be disabled');
```
Janpot, line 79: There's another instance in this file where this is happening

packages/toolpad-app/src/toolpad/AppEditor/QueryIcon.tsx
```diff
@@ -29,25 +41,16 @@ export default function QueryIcon({ id: iconId, mode, sx }: QueryIconProps) {
   const ModeIcon = modeIconMap.get(mode ?? '');
 
   return (
-    <div style={{ display: 'flex' }}>
+    <div style={{ display: 'flex', gap: DataSourceIcon ? 0.5 : 0 }}>
```
Janpot, line 44:
```suggestion
    <div style={{ display: 'flex', gap: 0.5 }}>
```

The gap has no effect if there is only one child

PR #2394, opened 2023-07-31
Improve docs

prakhargupta1:
- Moved **Create custom component** to the same level. Before and After:

Screenshot 2023-07-31 at 1 28 37 PM Screenshot 2023-07-31 at 1 29 11 PM

- Divided HTTP request and custom functions into separate sections. Before:
Screenshot 2023-07-31 at 2 41 11 PM
After:
Screenshot 2023-08-02 at 10 36 03 AM


prakhargupta1:
There is one issue, I have a valid path http://localhost:3003/toolpad/concepts/queries/
It is parent to HTTP requests and custom functions but is not being shown maybe because of accordion component limitation. Is there any way to fix it?

apedroferreira:
> http://localhost:3003/toolpad/concepts/queries/

From a quick look at different MUI docs it seems like the accordion headings are never supposed to be clickable sections?
So maybe we could just have the more "general" section be the first sub-item, followed by HTTP requests and custom functions at the same level?

apedroferreira (approved):
Not sure about doing something like `- HTTP Requests` because I don't see it being done anywhere else in any MUI docs... Other than that this seems good to me so I will approve already.

I noticed that in the other docs sometimes the first item is called "Overview":
Screenshot 2023-08-01 at 17 02 18

prakhargupta1:
> I noticed that in the other docs sometimes the first item is called "Overview":

Thanks for sharing this, looks perfect! I'll merge it today.

PR #2400, opened 2023-08-01
Chart improvements from npm stats example update

apedroferreira:
Extract chart improvements that I had added in https://github.com/mui/mui-toolpad/pull/2385 so that they can be in the next release without updating the example yet.
Then the updated example can be properly reviewed by just bumping the Toolpad version after the next release.

Features: 
- fix loading/error in nested bindings
- add loading spinner to chart
- dashed grid behind chart
- reduce default chart height


packages/toolpad-app/src/runtime/evalJsBindings.ts
```diff
@@ -252,6 +255,13 @@ export default function evalJsBindings(
           const nestedBindingResult = results[nestedBindingId];
 
           if (nestedBindingResult) {
+            if (!nestedBindingsError) {
```
Janpot, line 258: Would be interesting to add a test for the error/loading propagation (There are some examples of these in the `backend-basic` test) for when we start overhauling the binding engine.

packages/toolpad-components/src/Chart.tsx
```diff
@@ -172,11 +174,26 @@ function Chart({ data = [], height, sx }: ChartProps) {
           })}
         </ComposedChart>
       </ResponsiveContainer>
+      {loading ? (
```
Janpot, line 177: Would it make sense to use the Skeleton, like we do for images? https://github.com/mui/mui-toolpad/blob/5a1a5717e8aee85a428e55bfb18727309c60da89/packages/toolpad-components/src/Image.tsx#L69

Janpot (approved):
Some comments, nothing blocking

packages/toolpad-components/src/Chart.tsx
apedroferreira: The chart components being used in the examples had a spinner like this, I'll see if that looks better and go with the best option.

packages/toolpad-app/src/runtime/evalJsBindings.ts
apedroferreira: I'll give it a try, if it's too much I'll split it into a separate PR.

packages/toolpad-components/src/Chart.tsx
apedroferreira: I think we can keep the spinner here - the chart already shows an outline of where the axes are going to be before it loads the data, and it would be a bit manual and probably unnecessary to make the skeleton rectangle fit inside those axes.
If anything changes when adding the MUI X charts I'll keep this in mind.

packages/toolpad-app/src/runtime/evalJsBindings.ts
apedroferreira: Working on it separately just to keep this simpler.

Issue #2431, opened 2023-08-05
Canvas just shows infinite spinner when running npm run dev on Windows


dprothero:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

Steps:

1. Use Windows 11, from PowerShell
2. `npx create-toolpad-app budget-ui`
3. `cd budget-ui`
4. `npm run dev`

### Current behavior 😯

The tool loads at a url like `http://localhost:3000/_toolpad/app/pages/0BLHdTc` but the canvas just shows a circular spinner. I am unable to drop any components onto the canvas (the mouse pointer turns into a 🚫)

![image](https://github.com/mui/mui-toolpad/assets/1897006/801c6c1c-7150-49b5-9f6c-6aa486a5b36c)

When I went to take the above screenshot (Windows key + Shift + S), the browser view changed to the following:

![image](https://github.com/mui/mui-toolpad/assets/1897006/08b23ab8-53c7-4ca8-9039-89cca3c709ea)

### Expected behavior 🤔

The toolpad canvas should load as shown in the tutorial and I should be able to drag components onto it.

### Context 🔦

I'm just trying to run the tutorial.

### Your environment 🌎

Edition	Windows 11 Pro
Version	22H2
OS build	22621.1992
Experience	Windows Feature Experience Pack 1000.22644.1000.0

Running from PowerShell command prompt (non-admin)

Node.js 18.13.0

@mui/toolpad 0.1.22

Natalie-Perez:
I was experiencing the same issue. Stopping and re-running the project a few times got it to work. I didn't do anything helpful to fix the issue so I'm unsure why it started working.


My environment 🌎
Edition Windows 10 Pro
Version 22H2
OS build 19045.3208
Experience Windows Feature Experience Pack 1000.19041.1000.0

Running from Git Bash command prompt (non-admin)

Node.js 16.14.2

@mui/toolpad 0.1.22

apedroferreira:
I can reproduce similar behavior when running the latest version, this is the error I get:
![Capture](https://github.com/mui/mui-toolpad/assets/10789765/adf68353-13b2-491b-85e3-c634e3e79e7c)
Looking into it to try to figure out the cause, thanks for submitting this issue!

apedroferreira:
This should be fixed in the latest version of Toolpad.
Will wait to see if we get any more similar reports over the next couple of weeks.

[...]

PR #2442, opened 2023-08-07
Fix app not starting in Windows (10)

apedroferreira:
Seems to fix the issue I had in https://github.com/mui/mui-toolpad/issues/2431.

Followed the top answer from https://stackoverflow.com/questions/75883720/504-outdated-optimize-dep-while-using-react-vite.

Not 100% sure if this is the same issue the users had, but I could reproduce it every time, just the first time I ran a new project.
Also I don't have Windows 11 to be able to test there if there are any additional issues.


packages/toolpad-app/src/server/toolpadAppBuilder.ts
```diff
@@ -285,6 +285,7 @@ export function createViteConfig({
           'superjson',
           'zod',
         ],
+        exclude: ['@mui_toolpad_runtime', '@mui_toolpad_canvas'],
```
Janpot, line 288: Does this only work with
```
@mui_toolpad_runtime
```
or also with
```
@mui/toolpad/runtime
```
?

apedroferreira: I can try `@mui/toolpad/runtime` in case it works.

apedroferreira:
I'm actually not sure if this works, the issue only seems to be reproducible if the project uses a remote version of  `@mui/toolpad` such as 0.1.22.

When trying to use local changes the issue does not seem to happen which I think mislead me to think the above solution was working...
So not sure how to test fixes for this besides creating a new release with changes which is not ideal...

Janpot:
> So not sure how to test fixes for this besides creating a new release with changes which is not ideal

does it happen when using the packages built for a PR by codesandboxci?

apedroferreira:
> does it happen when using the packages built for a PR by codesandboxci?

Good idea, I tried and the issue is still there in this PR, at least looks like I can debug it now.

apedroferreira:
Looks like using `@mui/toolpad/runtime` and `@mui/toolpad/canvas` is the actual way to do it, seems to be working now with the Codesandbox packages!

Issue #2453, opened 2023-08-08
Freshly created custom components don't show up in the catalog


Janpot:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

Steps:

1. Click the plus button in the component catalog
2. Give name and click "create"

### Current behavior 😯

a component gets created on disk, but doesn't show up in the catalog

### Expected behavior 🤔

The component should show up in the catalog

### Context 🔦

Update to the tests that shows the problem:

```diff
diff --git a/test/integration/editor/new.spec.ts b/test/integration/editor/new.spec.ts
index 4322257a5..226d0f82d 100644
--- a/test/integration/editor/new.spec.ts
+++ b/test/integration/editor/new.spec.ts
@@ -44,6 +44,9 @@ test('can create new component', async ({ page }) => {
 
   await editorModel.createPage('somePage');
   await editorModel.createComponent('someComponent');
+  await editorModel.componentCatalog.hover();
+  await expect(editorModel.componentCatalog.getByText('someComponent')).toBeVisible();
+  await editorModel.dragNewComponentTo('someComponent');
 });
 
 test('can create/delete page', async ({ page, localApp }) => {
```

PR #2454, opened 2023-08-08
Fix new components not showing up when component folder doesn't exist

Janpot:
Fixes https://github.com/mui/mui-toolpad/issues/2453

### DIFF
--- a/test/integration/editor/new.spec.ts
+++ b/test/integration/editor/new.spec.ts
@@ -24,7 +24,7 @@ test('can place new components from catalog', async ({ page }) => {
 
   // Drag in a first component
 
-  await editorModel.dragNewComponentTo(TEXT_FIELD_COMPONENT_DISPLAY_NAME);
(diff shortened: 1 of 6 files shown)


test/integration/editor/new.spec.ts
```diff
@@ -44,6 +44,10 @@ test('can create new component', async ({ page }) => {
 
   await editorModel.createPage('somePage');
   await editorModel.createComponent('someComponent');
+  await editorModel.componentCatalog.hover();
+  await expect(editorModel.componentCatalog.getByText('someComponent')).toBeVisible();
+  await editorModel.dragNewComponentTo('someComponent');
```
apedroferreira, line 49: I'll rechange the name of this method to `dragNewComponentToCanvas`, it makes it more readable (I had changed the name some day ago).

apedroferreira (approved):
Looks good!
We can change the test name from `can create new component` to `can create and place new component` (I wasn't able to add the suggestion directly).

PR #2455, opened 2023-08-08
Try to improve undo/redo flakyness

Janpot:
Can't really replicate the flakyness locally but I think we should only start the timeout after we've verified that the page has processed the change. The update only being processed on the page when we're already 45ms into the timeout could explain the flakyness I'm seeing in some test results


apedroferreira (approved):
If it makes the test less flaky this makes sense.
I also don't remember exactly why we're still debouncing on undo/redo, we should probably be able to get rid of it since the DOM changes are "atomic".

PR #2439, opened 2023-08-07
Try out windows test in CI

Janpot:
Initial attempt at running tests in windows on circleci

PR #2461, opened 2023-08-09
Clean up env watcher and build contexts on process exit

Janpot:
On exit
* Properly dispose of the .env file watcher in the env file manager
* Properly dispose of the esbuild context in the function manager that builds backend functions
* Properly dispose of the vite dev server

At least to make sure these aren't responsible for the EBUSY errors we see on windows integration tests cleanup


Janpot:
For that particular timeout I already have https://github.com/mui/mui-toolpad/pull/2439/files#diff-2fd1cd5f4afa580dcae47c1599ee25484366978fb7e1161e0cc212e860b84ce8R173

For the tmp folders, one thing you could do is look into something like [`wtfnode`](https://www.npmjs.com/package/wtfnode) in all processes that Toolpad starts to detect open handles on exit.

Anyway, this PR lays some of the groundwork needed to run teardown code in general. I don't know immediately anything that could be holding file handles open, so we can merge this for now as far as I'm concerned.

apedroferreira:
> For the tmp folders, one thing you could do is look into something like [wtfnode](https://www.npmjs.com/package/wtfnode) in all processes that Toolpad starts to detect open handles on exit.

Ok thanks, I'll try that. We can also keep track of the EBUSY errors in an issue, I'll probably create one.

PR #2471, opened 2023-08-10
Add graphQL example

prakhargupta1:
App:

Screenshot 2023-08-18 at 10 07 57 AM


examples/graphQL/toolpad/resources/function.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export async function queryStargazers(owner: string, repository: string) {
+  if (!process.env.GITHUB_TOKEN) {
+    throw new Error(`Env variable GITHUB_TOKEN not configured`);
+  }
+
+  const endpoint = 'https://api.github.com/graphql';
+  const token = process.env.GITHUB_TOKEN;
+
+  const query = `
```
apedroferreira, line 11: We can use https://www.npmjs.com/package/graphql-tag for these template strings.

apedroferreira:
We should just call the example folder `graphql` I think, as it seems that the pattern is to use all lowercase letters and spacing with hyphens when necessary (as in `npm-stats`)

apedroferreira (changes requested):
Looks good, just really minor comments!
Also, based on the screenshot, we could format the followers data so that we show only the `totalCount` in that column? Unless there's more data we might want to show sometimes.

examples/graphQL/toolpad/resources/function.ts
prakhargupta1: I added this and expected syntax highlighting, but nothing happened.

prakhargupta1:
> Looks good, just really minor comments/suggestions! Also, based on the screenshot, we could format the followers data so that we show only the `totalCount` in that column? Unless there's more data we might want to show sometimes.

Couldn't make this work, do I need to change the response array?

prakhargupta1:
> We should just call the example folder `graphql` I think, as it seems that the pattern is to use all lowercase letters and spacing with hyphens when necessary (as in `npm-stats`)

Did graphQL -> graphql. This change is visible in my local but isn't detected by GitHub desktop :/ https://github.com/desktop/desktop/issues/2672#issuecomment-1681471505

apedroferreira:
> Did graphQL -> graphql. This change is visible in my local but isn't detected by GitHub desktop :/ [desktop/desktop#2672 (comment)](https://github.com/desktop/desktop/issues/2672#issuecomment-1681471505)

I just added a commit with the change.
It's an issue on Mac/Linux with case sensitive file names it seems.
I ran `git config core.ignorecase true` in the terminal before renaming the folder, and then ran `git config --unset core.ignoreCase` to put it back to normal.

prakhargupta1:
> I just added a commit with the change.
> It's an issue on Mac/Linux with case sensitive file names it seems.
> I used an intermediary commit with `graphqltemp` as the folder name to make it work.

Thanks for this!

apedroferreira (approved):

PR #2472, opened 2023-08-10
Fix cycle of SIGINT disposal event

apedroferreira:
This `process.on` was causing a cycle of `SIGINT` signals, stopping the app from terminating.
Using `process.once` instead of `process.on` seems to not cause the cycle.

### DIFF
--- a/packages/toolpad-app/cli/index.ts
+++ b/packages/toolpad-app/cli/index.ts
@@ -21,7 +21,7 @@ async function runCommand(cmd: 'dev' | 'start', { dir, ...args }: Omit<RunOption
     cmd,
   });
 
-  process.on('SIGINT', () => {
+  process.once('SIGINT', () => {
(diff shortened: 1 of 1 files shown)


packages/toolpad-app/cli/index.ts
```diff
@@ -21,7 +21,7 @@ async function runCommand(cmd: 'dev' | 'start', { dir, ...args }: Omit<RunOption
     cmd,
   });
 
-  process.on('SIGINT', () => {
+  process.once('SIGINT', () => {
     app.dispose().then(() => {
```
Janpot, line 25: Looks like I didn't push this after all. Let's still do:
```suggestion
    app.dispose().finally(() => {
```

Janpot (approved):
Let's still make that `.then` => `.finally` change

PR #2477, opened 2023-08-11
[docs] Add Stripe script example

### LINKED ISSUE #2422: Add stripe app to Github examples
### Duplicates

### Latest version

### Summary 💡

As discussed yesterday, let's add the stripe app in the Github examples folder

bharatkashyap:

- Closes #2422 

- Updated to download the `PDFs` to the client, and use a Stripe data provider to show a paginated list of invoices:

https://github.com/mui/mui-toolpad/assets/19550456/bd327963-c061-4a75-906d-22acdd696211


bharatkashyap:
An interesting issue I'm facing is that the `listInvoices` function doesn't return anything until I wrap its actual output inside a `JSON.stringify(...)` - cannot figure out why, can anyone else replicate this?

examples/stripe-script/toolpad/resources/functions.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  list.data?.forEach((invoice) => {
+    try {
+      const dest = path.resolve('.', `invoices-${invoice.id}.pdf`);
+      // check if destination is resolvable
+      if (fs.existsSync(dest)) {
+        console.error('Invoice already exists:', dest);
+        return;
+      }
+      exec(`curl '${invoice.invoice_pdf}' -L -o '${dest}'`);
```
apedroferreira, line 45: An alternative to this, which works but does not give a lot of feedback, would be to create a custom component for the download button that uses an anchor tag with the `download` attribute, so that the PDF is downloaded on the browser.
Not sure how easy/possible it would be but it might be an improvement.

examples/stripe-script/toolpad/resources/functions.ts
```diff
@@ -0,0 +1,52 @@
...
+// });
+
+export async function listInvoices(limit: number = 100, starting_after?: string) {
+  const stripe = initStripe();
+  const list = await stripe.invoices.list({
+    limit,
+    starting_after,
+  });
+
+  return JSON.stringify(list.data);
```
apedroferreira, line 26: It's strange that we need to stringify this, it should just work.
Maybe the data returned by the function as an object is too big? Could we maybe try a lower limit or filter out some properties?
I don't have a Stripe token so just speculating, let me know if you want me to try it.

examples/stripe-script/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "private": true,
+  "scripts": {
+    "dev": "toolpad dev",
+    "build": "toolpad build",
+    "start": "toolpad start"
+  },
+  "dependencies": {
+    "@mui/toolpad": "^0.1.22",
+    "child_process": "^1.0.2",
+    "fs": "^0.0.1-security",
```
apedroferreira, line 13: Are these not included with node? (`fs`, `path`, `child_process`...)
Do we need to include them in package.json?

apedroferreira (commented):
I added just some comments but if none of them needs any action I can approve the PR.

examples/stripe-script/toolpad/resources/functions.ts
bharatkashyap: It worked with just returning the `list` instead of `list.data` - not quite sure why but leaving this here to offer clarity in case of attempts of future resolution

examples/stripe-script/toolpad/resources/functions.ts
bharatkashyap: The current button downloads all the invoices in the grid instead of just one, so it won't end up being an improvement, although I agree showing visual feedback during the download process is an improvement.

Personally I would like for a success toast to appear on the page once the action is complete.

apedroferreira (approved):

Issue #2468, opened 2023-08-10
It should be possible to access cookies in backend functions


Janpot:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

I'd like to be able to access cookies of the underlying HTTP requests in backend functions. e.g. to read an authentication cookie and extract the user role.

### Examples 🌈

```tsx
import { getContext } from '@mui/toolpad/server';
import { authenticateRequest } from '../../src/utils/auth';

export default async function getUserRole() {
  // returns information about the context under which this function was called
  const ctx = await getContext()
  const user = await authenticateRequest(ctx.req)
  if (!user) {
    throw new Error(`Not authenticated`)
  }
  return user.role
}
```

PR #2491, opened 2023-08-14
Support reading cookies in backend functions

Janpot:
Fixes https://github.com/mui/mui-toolpad/issues/2468

This proposes to add a `getContext` method which returns the context under which a backend function was called. The context contains a dictionary that maps cookie name to cookie value. We can expand this interface as more use cases arise.


apedroferreira (approved):
Looks good! I think it seems easy to use and makes sense.

PR #2505, opened 2023-08-17
Upgrade to Node 18

apedroferreira:
Upgrade Node version to 18.17.1, the current LTS version.


apedroferreira:
> @apedroferreira I updated this PR:
> 
> * put the engines field at `>=18` for now (codesandboxci is not at latest node 18)
> * update netlify
> * update condesandbox

Got it, thanks!

Issue #2499, opened 2023-08-16
First build attempt fails on newly cloned monorepo


apedroferreira:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

Steps:

1. Clone the monorepo with `gh repo clone mui/mui-toolpad`, for example
2. Run `yarn install`
3. Run `yarn dev`

### Current behavior 😯

`yarn dev` build will fail.
Build output in terminal:

<img width="1791" alt="Screenshot 2023-08-16 at 18 04 26" src="https://github.com/mui/mui-toolpad/assets/10789765/f7d10ec5-ed02-4c71-97db-34d00f4ee5a5">

Exiting and `yarn dev` a second time does not show any errors anymore.

### Expected behavior 🤔

There should be no errord during build.

### Context 🔦

Simulating a new user trying out the monorepo for the first time, following the instructions in the Contributions guide.

### Your environment 🌎

Mac, Node 20

apedroferreira:
> We're hitting simmilar issues as [egoist/tsup#688](https://github.com/egoist/tsup/issues/688). Personally I don't feel like there's a satisfying solution to this at the moment. There are a lot of issues with type generation in `tsup`.

Got it, I guess that https://github.com/mui/mui-toolpad/pull/2515 isn't an ideal solution? 
It does solve the issue, but no need to merge it if you don't like using postinstall scripts like that or if there might be a better way that's better integrating with how `tsup` should work.

Janpot:
> It does solve the issue

Yes, but it also adds some overhead to every `yarn install`, even when it's not needed. Doing a lot of code infra work, I'm running `yarn install` quite often. If possible I'd prefer to not add this overhead. Unless this issue becomes too annoying.

apedroferreira:
> > It does solve the issue
> 
> Yes, but it also adds some overhead to every `yarn install`, even when it's not needed. Doing a lot of code infra work, I'm running `yarn install` quite often. If possible I'd prefer to not add this overhead. Unless this issue becomes too annoying.

Got it, makes sense, I'll close the other PR then.

PR #2515, opened 2023-08-21
Add build step after installing dependencies

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/2499 (possible solution).

This is the exact same script as `test:build`, we can merge them into the same script if it makes sense and we give it a name that works well, or we can keep things as in this PR.


apedroferreira:
Let's find a better solution that doesn't add overhead to `yarn install`

PR #2535, opened 2023-08-23
Refactor Toolpad App Navigation

Janpot:
Extracted from https://github.com/mui/mui-toolpad/pull/2406 to keep the merge manageable.

- extract the app layout component
- make it agnostic from app dom
- export it so it can be used by generated code
- only pass through a subset of the search parameters
- touch up the path matching logic a bit


apedroferreira (approved):
Looks good!

PR #2545, opened 2023-08-25
Remove lodash from @mui/components

Janpot:
As far as I understand this code, all that's needed is a shallow compare of those 4 properties. There's many ways to achieve it. But pulling in whole lodash for a deeply nested comparison seems a bit overkill.

### DIFF
--- a/packages/toolpad-components/src/Form.tsx
+++ b/packages/toolpad-components/src/Form.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { Container, ContainerProps, Box, Stack, BoxProps } from '@mui/material';
 import { LoadingButton } from '@mui/lab';
 import { useNode } from '@mui/toolpad-core';
+import { equalProperties } from '@mui/toolpad-utils/collections';
 import { useForm, FieldValues, ValidationMode, FieldError, Controller } from 'react-hook-form';
 import { SX_PROP_HELPER_TEXT } from './constants';
 import createBuiltin, { BuiltinArgTypeDefinitions } from './createBuiltin';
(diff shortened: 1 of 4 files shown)


packages/toolpad-components/src/Form.tsx
```diff
@@ -19,7 +16,7 @@ export const FormContext = React.createContext<{
 
 interface FormProps extends ContainerProps {
   value: FieldValues;
-  onChange: (newValue: FieldValues) => void;
+  onChange?: (newValue: FieldValues) => void;
```
apedroferreira, line 19: Why should this be optional?

packages/toolpad-components/src/Form.tsx
```diff
@@ -263,7 +260,12 @@ export function useFormInput<V>({
   React.useEffect(() => {
     if (
       form &&
-      !isEqual(validationProps, previousManualValidationPropsRef.current) &&
+      !(
+        validationProps.isInvalid === previousManualValidationPropsRef.current.isInvalid &&
```
apedroferreira, line 264: Sure, it's a bit more verbose though, but if we don't want to use any external utilities for deep equality it will still work.

apedroferreira, line 264: I'm not sure why I called this `previousManualValidationPropsRef` if I did, I guess it could be just `previousValidationPropsRef`?

Janpot: I'm adding a function for this

packages/toolpad-components/src/Form.tsx
Janpot: No reason, leftover I forgot to remove. Updated

apedroferreira (approved):
Looks great! Thanks for this.

PR #2582, opened 2023-08-30
[refactor] Move Toolpad runtime to the RPC mechanism

Janpot:
Updates the data fetching transport mechanism of the Toolpad runtime.

* Aligns editor calls to execute queries with runtime calls to execute queries for less differences between query preview and actual production calls.
* Prepares for integrating the upcoming data-provider interfaces for server collections


Janpot:
Merging this to unblock work on serverside collections. Feel free to retroactively review

Issue #2581, opened 2023-08-30
Support data grid column type date formats


oliviertassinari:
### Summary 💡

Today, developers can set their column type to date:

<img width="277" alt="Screenshot 2023-08-30 at 14 14 54" src="https://github.com/mui/mui-toolpad/assets/3165635/9e375a18-4935-48cb-a426-88947c57cdca">

However, have no options to decide what formatting it should use exactly. It's using the default MUI X column type formatting https://mui.com/x/react-data-grid/column-definition/#column-types, so effectively `value.toLocaleDateString()`.

It would be great to be able to choose from more options, a bit like numbers

<img width="291" alt="Screenshot 2023-08-30 at 14 17 19" src="https://github.com/mui/mui-toolpad/assets/3165635/61d3a11c-dd0b-455f-81cd-90a21b63168b">

### Examples 🌈

- Metabase

<img width="368" alt="Screenshot 2023-08-12 at 20 33 03" src="https://github.com/mui/mui-private/assets/3165635/a63fb82d-5492-460a-a422-e597d77d9dd0">

### Motivation 🔦

Less code to write. For now, I will stick to using dayjs: https://github.com/mui/mui-private/blob/c5f0a6288cbbd052cee7f141d2f117bc186ece6a/tools-private/toolpad/resources/invoice.ts#L77

```jsx
    invoice_date: dayjs(row.invoice_date).format('MMMM D, YYYY'),
```

It's a continuation of #2371.

Janpot:
Taking an initial stab at it in https://github.com/mui/mui-toolpad/pull/2589

PR #2589, opened 2023-08-30
Support date/dateTime formatting

Janpot:
This PR proposes date formatting for the date and dateTime columns, in the flavour of the number formatter. Similarly it is built on top of `Intl.DateTimeFormat` and can be reused for other components just like we reuse the number format for the Metric component

Fixes https://github.com/mui/mui-toolpad/issues/2581

https://github.com/mui/mui-toolpad/assets/2109932/87a79166-5a93-4d77-8d9f-cca7586eb446

_edit_

Reversed the order since the video:

Screenshot 2023-08-31 at 15 15 50


packages/toolpad-core/src/dateFormat.tsx
```diff
@@ -0,0 +1,165 @@
...
+);
+
+const TIME_FORMATS = new Map<DateStyle, Intl.DateTimeFormat>(
+  (['short', 'medium', 'long', 'full'] as const).map((timeStyle) => [
+    timeStyle,
+    new Intl.DateTimeFormat(undefined, { timeStyle }),
+  ]),
+);
+
+const DEMO_DATE = new Date();
```
apedroferreira, line 42: Could also be the current time? Anyway not important at all :D

apedroferreira:
Just tried it, works great!
Not sure if it would be better for showing time (hours/minutes/seconds) to be an option all under the same column type, but it could depend on what obstacles that would cause or how much each of these types could be expanded.

packages/toolpad-core/src/dateFormat.tsx
Janpot: I want a time that always demonstrates whether it's 12h or 24h clock

Janpot:
> Not sure if it would be better for showing time (hours/minutes/seconds) to be an option all under the same column type, but it could depend on what obstacles that would cause or how much each of these types could be expanded.

You mean only having a `dateTime` type and not both `date` and `dateTime` types? I slightly agree with that, but the X grid has both so it could make sense to follow them for now. Switching to a single column type would be the same as deprecating the `date` type.

packages/toolpad-core/src/dateFormat.tsx
Janpot: I've added a comment

apedroferreira:
> > Not sure if it would be better for showing time (hours/minutes/seconds) to be an option all under the same column type, but it could depend on what obstacles that would cause or how much each of these types could be expanded.
> 
> You mean only having a `dateTime` type and not both `date` and `dateTime` types? I slightly agree with that, but the X grid has both so it could make sense to follow them for now. Switching to a single column type would be the same as deprecating the `date` type.

Ok, if the X grid separates them I guess we should do the same.

apedroferreira (approved):

Issue #2384, opened 2023-07-27
Create a new Code Component


dai-chao:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

![image](https://github.com/mui/mui-toolpad/assets/45806795/f7b30cc0-d8fd-40b5-873d-72a7d834533c)
When creating custom components, the icons are fixed. Is it possible to make both the names and icons configurable?
![image](https://github.com/mui/mui-toolpad/assets/45806795/27897b54-fea2-4d7d-9a4d-b5df5eee8949)
The same icon.

### Examples 🌈

![image](https://github.com/mui/mui-toolpad/assets/45806795/f7b30cc0-d8fd-40b5-873d-72a7d834533c)

### Motivation 🔦

![image](https://github.com/mui/mui-toolpad/assets/45806795/f7b30cc0-d8fd-40b5-873d-72a7d834533c)

JerryWu1234:
I can to try this? @Janpot

Janpot:
I'm not entirely sure how the implementation will work. One idea we had was to add an `icon` property in the `createComponent` function, but I'm not sure which format it should be. A React component, or a React node, or a url, or something else.

JerryWu1234:
> I'm not entirely sure how the implementation will work. One idea we had was to add an `icon` property in the `createComponent` function, but I'm not sure which format it should be. A React component, or a React node, or a url, or something else.

@Janpot, How many icons do you want the user to choose?
I saw more than two thousand in the @mui/icon.
It could be a performance problem if there are two thousand.

PR #2607, opened 2023-09-01
add icon

JerryWu1234:
#2384 

image
image

I reference this [code](https://docs.astro.build/en/guides/server-side-rendering/#configuring-individual-routes) that I was contributed before.
I copied the way they [code](https://github.com/withastro/astro/blob/f21599671a90c3327307eb6d2f4d5c02e9137207/packages/astro/src/vite-plugin-scanner/scan.ts#L68) 
 

Janpot:
@JerryWu1234 I don't feel at the moment there is a good solution that doesn't introduce too much tech debt compared to the benefits. I'm inclined to say this issue is not yet up for implementation

JerryWu1234:
> 

Where we can save the `Iconname` on the back-end side is the problem, I think.

Tell me if you have any good idea, however I also will keep going

JerryWu1234:
> @JerryWu1234 I don't feel at the moment there is a good solution that doesn't introduce too much tech debt compared to the benefits. I'm inclined to say this issue is not yet up for implementation

I think it would be better for me if you could share more details about your requirements

brhndursun (approved):

JerryWu1234:
@Janpot closing or upgrading ?

PR #2625, opened 2023-09-04
Allow DataGrid height to be styled

apedroferreira:
Attempt to fix the problems in https://github.com/mui/mui-toolpad/discussions/2613 about stretching the DataGrid.

This PR should make it so that it's more possible to try to set the DataGrid to full page height or style its height, such as by for example by setting `sx: { height: "100vh" }` (even though this kind of overrides the default behavior).

The label was also getting cut-off at the bottom when an element is set to take up the full page height like this, so this included moving the HUD selection hint inwards in those cases.

Table with `sx: { height: "100vh" }` before (the container height did not change with the grid, so the dotted border is shorter and does not match):
Screenshot 2023-09-04 at 22 06 58

Table with `sx: { height: "100vh" }` after:
Screenshot 2023-09-04 at 22 07 43

New label position (bottom and inwards):
Screenshot 2023-09-04 at 22 03 08


Janpot:
I'm hesitant to try to solve https://github.com/mui/mui-toolpad/discussions/2613#discussioncomment-6894994. I don't see how it could fit in the current layout paradigm.

apedroferreira:
Actually this is all outdated by https://github.com/mui/mui-toolpad/pull/2818, closing.

PR #2630, opened 2023-09-05
Support setCookie API

Janpot:
Allow setting cookies from backend functions


docs/data/toolpad/concepts/custom-functions.md
```diff
@@ -154,3 +156,26 @@ export async function myBackendFunction() {
   return user?.id;
 }
 ```
+
+### Write cookies with `context.setCookie`
+
+You can also use the context to set a cookie, which can subsequently used in other backend functions called from your Toolpad application. Example:
```
bharatkashyap, line 162:
```suggestion
You can also use the context to set a cookie, which can subsequently be used in other backend functions called from your Toolpad application. Example:
```

PR #2635, opened 2023-09-05
Fix container components

apedroferreira:
Fixes containers not working as reported in https://github.com/mui/mui-toolpad/discussions/2613

When changing things so that we don't need PageRow components to wrap single elements, I forgot to adjust container components, which relied on this row being created when the first component is dropped.

Also adds a basic test for dropping components in a container component.
Also fixes the `children: []` we've been getting in the .yaml files for components where it doesn't make sense, I can separate this change if it's too much.


Janpot (approved):
Looks good, thank you!

The test case covers the original issue, right?

apedroferreira:
> Looks good, thank you!
> 
> The test case covers the original issue, right?

Yep, in a simple way but it does - it would fail without these fixes.

apedroferreira:
Just noticed there are still some bugs in containers... will add fixes very soon.

Issue #2585, opened 2023-08-30
When outputting yml files, Toolpad should respect the user's prettier config


Janpot:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

I have a prettier configuration in my project. The files that Toolpad ooutputs and updates should respect the user's prettier configuration.
This should apply at least to the files that we intend to be hand-editable as well. It doesn't apply to the build artifacts.

### Motivation 🔦

To avoid users from having to ignore them https://github.com/mui/mui-public/pull/133/files#diff-a0fdacd2ade87c5238dde44378f574f7e123e669acdf8b740878b14b33b20275R5
(cc @oliviertassinari I noticed this on your PR)

JerryWu1234:
Can I make sure how to implement it?
Generate the `yml` file after invoking the local formatting, or by default.  right？

Janpot:
Easiest way would be to format [`newContent`](https://github.com/mui/mui-toolpad/blob/1eb90b1e5ca6739ec89b0e064291df7ad8811ed8/packages/toolpad-utils/src/fs.ts#L55C9-L55C19) with `prettier`. We can use [`resolveConfig`](https://prettier.io/docs/en/api#prettierresolveconfigfilepath--options) with the `filePath` and then [`format`](https://prettier.io/docs/en/api#prettierformatsource-options) `newContent`

JerryWu1234:
![image](https://github.com/mui/mui-toolpad/assets/27261522/1872fdb9-e6c2-4e18-ab72-20d0d160b319)

oliviertassinari:
Should we close this issue? #2638 was merged

Janpot:
We still need to address formatting [new components](https://github.com/mui/mui-toolpad/blob/cf7171620217a881256732fc2a3481f46f28d396/packages/toolpad-app/src/server/localMode.ts#L1181) and [new functions](https://github.com/mui/mui-toolpad/blob/cf7171620217a881256732fc2a3481f46f28d396/packages/toolpad-app/src/server/FunctionsManager.ts#L328) when they are initialized.

JerryWu1234:
I'm going to do both in one pr 
Next, I'll develop a search of the input.

JerryWu1234:
[...]

PR #2638, opened 2023-09-06
respect the user's prettier config when writing yml files

JerryWu1234:
#2585


JerryWu1234:
image

packages/toolpad-utils/src/fs.ts
```diff
@@ -50,12 +50,16 @@ export async function writeFileRecursive(
   await fs.writeFile(filePath, content, options);
 }
 
-export async function updateYamlFile(filePath: string, content: object) {
+export async function updateYamlFile(
+  filePath: string,
+  content: object,
+  callback?: (content: string) => string,
```
Janpot, line 56: I'd keep the `updateYamlFile` interface as it was and call `prettier` inside of it. We'd have to add `prettier` as a dependency to `@mui/toolpad-utils`. Just make sure it's the same version as the rest of the project.

JerryWu1234: My problem.
Actually, I wrote both ways and felt that could reduce dependency.
I will fix it late

Janpot: No worries. There shouldn't be any extra dependencies installed in the user's node_modules fodler if we keep the same version across the whole project.

packages/toolpad-utils/package.json
```diff
@@ -55,11 +55,13 @@
...
   },
   "devDependencies": {
     "@types/invariant": "2.2.35",
-    "@types/react-is": "18.2.1"
+    "@types/react-is": "18.2.1",
+    "@types/prettier": "2.7.3"
   },
   "gitHead": "342439583072ebcaaa3b3587517804e4fdfb9dcf"
-}
+}
```
Janpot, line 67:
```suggestion
}

```

packages/toolpad-utils/src/fs.ts
```diff
@@ -55,7 +69,7 @@ export async function updateYamlFile(filePath: string, content: object) {
   const newContent = oldContent ? yamlOverwrite(oldContent, content) : yaml.stringify(content);
 
   if (newContent !== oldContent) {
```
Janpot, line 71: We can assume that in most cases `oldContent` will be formatted. That means `newContent` needs to be formatted before comparing. Otherwise it would write more often than necessary.

packages/toolpad-utils/src/fs.ts
```diff
@@ -3,8 +3,22 @@ import * as path from 'path';
...
+import * as prettier from 'prettier';
 import { errorFrom } from './errors';
 
+/**
+ * Formats a yaml source with `prettier`.
+ */
+const PRETTIERNAME = 'prettier.config.js';
+
+export function formatYaml(code: string): string {
+  const readConfig = prettier.resolveConfig.sync(PRETTIERNAME);
```
Janpot, line 15: Upcoming `prettier` update will remove all sync apis. So to make this function forward compatible I'd make `formatYaml` an async function and use the async version of `resolveConfig`. 

In general you can assume that in most of the application code we want to use non-blocking APIs.

packages/toolpad-utils/src/fs.ts
```diff
@@ -3,8 +3,22 @@ import * as path from 'path';
...
+  return prettier.format(code, {
+    parser: 'yaml',
```
Janpot, line 17: You may move `parser: 'yaml',` underneath `...readConfig,` to indicate we intend it to be an override and not a default value.

packages/toolpad-utils/src/fs.ts
Janpot: If I understand correctly, the filePath that is supplied to `resolveConfig` is the one of the file that is to be formatted. The prettier config file is then resolved from that location.

Janpot (changes requested):
Thank you, this is a good start. I found a few small problems that need to be solved before it can be merged

packages/toolpad-utils/src/fs.ts
JerryWu1234: It's my first to invoke this function, I try it late

JerryWu1234:
can I ask a question that is not related to this PR?
Can you let me know if you have any questions about nodejs？
Recently I always focused on front-end and I want to learn more about another aspect of the project.
-------
> Thank you, this is a good start. I found a few small problems that need to be solved before it can be merged

I need to go back to my hometown. So this PR might be put off until next Monday.

packages/toolpad-utils/src/fs.ts
```diff
@@ -3,8 +3,21 @@ import * as path from 'path';
...
 import * as yaml from 'yaml';
 import { yamlOverwrite } from 'yaml-diff-patch';
+import * as prettier from 'prettier';
 import { errorFrom } from './errors';
 
+/**
+ * Formats a yaml source with `prettier`.
+ */
+
+export async function formatYaml(code: string, filePath: string): Promise<string> {
```
Janpot, line 13: No need to export it for now

```suggestion
async function formatYaml(code: string, filePath: string): Promise<string> {
```

packages/toolpad-utils/src/fs.ts
```diff
@@ -52,8 +65,9 @@ export async function writeFileRecursive(
 
 export async function updateYamlFile(filePath: string, content: object) {
   const oldContent = await readMaybeFile(filePath);
-  const newContent = oldContent ? yamlOverwrite(oldContent, content) : yaml.stringify(content);
 
+  let newContent = oldContent ? yamlOverwrite(oldContent, content) : yaml.stringify(content);
+  newContent = await formatYaml(newContent!, filePath);
```
Janpot, line 70: `newContent` should always be a string. We want to avoid using the no-null-assertion operator.

```suggestion
  newContent = await formatYaml(newContent, filePath);
```

Janpot (changes requested):
Two small fixes and then this can be merged as far as I'm concerned. (If the CI passes)

JerryWu1234:
> Looking good, two small fixes and then this can be merged as far as I'm concerned. (If the CI passes)

Hello I done 
when would you refactor part of the code file that is about file (create/ delete/ modify)?

Janpot (approved):
thank you, this looks great

PR #2644, opened 2023-09-07
Introduce Data providers

Janpot:
https://github.com/mui/mui-toolpad/assets/2109932/a196172f-df46-4b1c-8d72-25c68431de26

## To do:

- [x] Add a switch in the property controls to switch between row mode and dataprovider mode
- [x] documentation
- [x] test
- [x] reset cursor when pageSize changes
- [ ] ?

Documentation:
https://deploy-preview-2644--mui-toolpad-docs.netlify.app/toolpad/concepts/data-providers/
https://deploy-preview-2644--mui-toolpad-docs.netlify.app/toolpad/reference/api/create-data-provider/


docs/data/toolpad/concepts/data-providers.md
```diff
@@ -0,0 +1,111 @@
...
+
+   export default createDataProvider({
+     async getRecords({ paginationModel: { start = 0, pageSize } }) {
+       const records = DATA.slice(start, start + pageSize);
+       return { records, totalCount: DATA.length };
+     },
+   });
+   ```
+
+## pagination
```
mnajdova, line 41:
```suggestion
## Pagination
```

In my opinion, is better to capitalize the features, as they reference to the functionality, not a specific prop/code equivalent.

docs/data/toolpad/reference/api/create-data-provider.md
```diff
@@ -0,0 +1,91 @@
...
+    return { records, totalCount: DATA.length };
+  },
+});
+```
+
+Data providers expose collections to the Toolpad frontend. They are server-side data structures that abstract the loading and manipulation of a backend collection of records of similar shape. They can be directly connected to data grids to display the underlying data.
+
+## Parameters
+
+- `config` An object describing the data provider capabilities
```
mnajdova, line 29:
```suggestion
- `config: DataProviderConfig` An object describing the data provider capabilities
```

Maybe even with a link to the `DataProviderConfig` would give a much better experience.

docs/data/toolpad/reference/api/create-data-provider.md
```diff
@@ -0,0 +1,91 @@
+# createDataProvider API
```
mnajdova, line 1: We should have a link to this page from the Data providers page.

docs/data/toolpad/concepts/data-providers.md
```diff
@@ -0,0 +1,117 @@
+# Data Providers
+
+<p class="description">Bring tabular data to the frontend with server-side pagination and filtering.</p>
+
+Toolpad functions are great to bring some backend state to the page, but they fall short when it comes to offering pagination and filtering capabilities from the server. Toolpad offers a special construct to enable this use case: Data providers. Data providers abstract server-side collections. They could be database tables, or REST APIs, or any data that represents a set of records that share a common interface. Data providers are defined as server-side objects and can be directly connected to a data grid to make it fully interactive.
```
apedroferreira, line 5: Maybe this can be split into 2 paragraphs, seems like a large chunk of text to read right at the start. :)

docs/data/toolpad/concepts/data-providers.md
```diff
@@ -0,0 +1,117 @@
...
+
+Follow these steps to create a new data provider
```
apedroferreira, line 7:
```suggestion
Follow these steps to create a new data provider:
```

docs/data/toolpad/concepts/data-providers.md
apedroferreira, line 5:
```suggestion
Toolpad functions are great to bring some backend state to the page, but they fall short when it comes to offering pagination and filtering capabilities from the server. Toolpad offers a special construct to enable this use case: Data providers. Data providers abstract server-side collections. They could be database tables, REST APIs, or any data that represents a set of records that share a common interface. Data providers are defined as server-side objects and can be directly connected to a data grid to make it fully interactive.
```

docs/data/toolpad/concepts/data-providers.md
```diff
@@ -0,0 +1,117 @@
...
+      records: page,
+      totalCount,
+    };
+  },
+});
+```
+
+### Cursor based
+
+This is the strategy your data is paginated by when it returns data based on a cursor and a page size. The `getRecords` method will receive `cursor` and `pageSize` values in it `paginationModel` parameter and returns a set of records representing the page. You indicate the cursor of the next page with a `cursor` property in the result. Pass `null` to signal the end of the collection. The You can enable Cursor based pagination by setting `paginationMode` to `'cursor'`.
```
apedroferreira, line 64:
```suggestion
This is the strategy your data is paginated by when it returns data based on a cursor and a page size. The `getRecords` method will receive `cursor` and `pageSize` values in it `paginationModel` parameter and returns a set of records representing the page. You indicate the cursor of the next page with a `cursor` property in the result. Pass `null` to signal the end of the collection. You can enable Cursor based pagination by setting `paginationMode` to `'cursor'`.
```

apedroferreira, line 64:
```suggestion
This is the strategy your data is paginated by when it returns data based on a cursor and a page size. The `getRecords` method will receive `cursor` and `pageSize` values in its `paginationModel` parameter and returns a set of records representing the page. You indicate the cursor of the next page with a `cursor` property in the result. Pass `null` to signal the end of the collection. The You can enable Cursor based pagination by setting `paginationMode` to `'cursor'`.
```

docs/data/toolpad/concepts/data-providers.md
```diff
@@ -0,0 +1,117 @@
...
+
+```tsx
+export default createDataProvider({
```
apedroferreira, line 67: Maybe i'm not super familiar with cursor pagination, but it isn't completely obvious to me right away how to use this cursor pagination.
is the cursor a number? Or can I use the id of a record?
The prisma documentation on this kind of this is very clear, maybe we could use it as inspiration https://www.prisma.io/docs/concepts/components/prisma-client/pagination.

apedroferreira: Oh, also the way Prisma does it kind of just works and we don't have to explicitly choose a pagination model. Not sure if we could do something similar?

apedroferreira: Nevermind I understand this is quite different from Prisma now as it integrates very directly with the Data Grid, these were just some quick ideas while going over the docs and trying to use the feature in a basic way. Feel free to ignore if they don't make sense or don't provide any insights :D

packages/toolpad-app/src/server/functionsDevWorker.ts
```diff
@@ -158,6 +209,21 @@ export function createWorker(env: Record<string, any>) {
 
       return result;
     },
+
+    async intropectDataProvider(
```
apedroferreira, line 213:
```suggestion
    async introspectDataProvider(
```

packages/toolpad-app/src/toolpad/propertyControls/DataProviderSelector.tsx
```diff
@@ -0,0 +1,331 @@
...
+  }, []);
+
+  return (
+    <DataProviderSelectorRoot>
+      <CreateNewDataProviderDialog
+        open={dialogOpen}
+        onClose={handleClose}
+        onCommit={(newName) => onChange(`${newName}.ts:default`)}
+        initialName={dialogValue}
+        existingNames={existingNames}
```
apedroferreira, line 244: Not sure if it would be better to pass a validation function instead, anyway not a big deal.

apedroferreira (commented):
Looks good and it's easy to use, really great feature! Here's some feedback I collected while trying it out, just trying to give as much feedback as possible:

- I set up some static data for records but forgot to include `id` strings, so an error message showed in the Data Grid telling me to do that. Once I included the `id`s and saved the file, the Data Grid did not update without the error until I refreshed the page.
- Maybe we could leave the cursor choice for later if possible instead of having it in the modal right away, and start with index pagination automatically? But this could depend on many things, it's just a thought to reduce initial friction/confusion a bit if someone's trying the feature for the first time.
- If I'm using the data grid for the first time, I notice these options and I try to create a data provider, how can I quickly get to the documentation so I can understand all the options I have? As this can be a complex feature, maybe we can make this easier. Maybe we can have a comment with a link to the docs in the data provider template (and eventually more helpful comments in general).

docs/data/toolpad/concepts/data-providers.md
Janpot: Most writing style guides I could find (e.g. https://libguides.hull.ac.uk/writing/paras) suggest between 100-200 words per paragraph or less than 6 sentences.

apedroferreira: Got it, it's fine to keep it too! Just personal opinion, and website users are particularly lazy readers as opposed to people who read studies, books or newspapers :D

docs/data/toolpad/concepts/data-providers.md
Janpot: I've added another paragraph explaining the value of `cursor` in this pagination model.

> Oh, also the way Prisma does it kind of just works and we don't have to explicitly choose a pagination model. Not sure if we could do something similar?

In prisma you don't need to explicitly specify a pagination model because it doesn't need to signal to consumers how it is to be called. The Toolpad data provider needs to signal to Toolpad how it's to be fetched. Supplying this option configures the interfaces in `getRecords` for the right pagination model.

packages/toolpad-app/src/toolpad/propertyControls/DataProviderSelector.tsx
```diff
@@ -0,0 +1,331 @@
...
+      onClose();
+    },
+  });
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    createProviderMutation.mutate();
+  };
+
+  const nameExists = existingNames.has(newName);
```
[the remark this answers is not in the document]
Janpot: `nameExists` is a boolean and `existingNames` is a `Set`. Therefore `nameExists` is automatically stable, `existingNames` depends on the consumer to provide a stable value. `.has` is a fast operation. Not sure which is better, but  don't believe it matters much. I can put it inside as well.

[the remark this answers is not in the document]
apedroferreira: Got it, it's not important so we can keep it, I think it's fine either way.

Janpot:
> - I set up some static data for records but forgot to include id strings, so an error message showed in the Data Grid telling me to do that. Once I included the ids and saved the file, the Data Grid did not update without the error until I refreshed the page.

I added an error boundary to the grid. Also adding ids automatically when data provider is used, just the way it is done when rows prop is used. This also uncovered some problems around column inference in the columns editor which had the same root cause

> - Maybe we could leave the cursor choice for later if possible instead of having it in the modal right away, and start with index pagination automatically? But this could depend on many things, it's just a thought to reduce initial friction/confusion a bit if someone's trying the feature for the first time.

The choice of pagination mode is made in the code only. It's not an option that can be easily changed from the UI. This initializes the code with the right pagination mode. Initially this wasn't part of data providers but was suggested in the last grooming call.

> - If I'm using the data grid for the first time, I notice these options and I try to create a data provider, how can I quickly get to the documentation so I can understand all the options I have? As this can be a complex feature, maybe we can make this easier. Maybe we can have a comment with a link to the docs in the data provider template (and eventually more helpful comments in general).

- I added a helper text to the pagination mode input with a link to the documentation
- vscode also displays a tooltip with links to the docs when you hover `createDataProvider`: https://github.com/mui/mui-toolpad/pull/2644/files#diff-1df7e8b1e0c8176606fffa17934efa350f0483e6a994b40f5fba2a3c2b61ce23R135
- also added a link to the docs in the generated code as a comment

apedroferreira (approved):
The changes look great! Thanks for addressing the feedback.

PR #2652, opened 2023-09-08
[docs] Sync prism with docs-infra

oliviertassinari:
I saw the issue in #2353

Screenshot 2023-09-08 at 15 29 21

https://deploy-preview-2353--mui-toolpad-docs.netlify.app/toolpad/live-config/#demo

My workaround is to copy and paste the same file from docs-infra. The proper solution would be for docs-infra to import the CSS file with a bundler import, rather than using the static file.

### DIFF
--- a/docs/public/static/styles/prism-okaidia.css
+++ b/docs/public/static/styles/prism-okaidia.css
@@ -1,34 +1,26 @@
 /**
- * https://unpkg.com/prismjs/themes/prism-okaidia.css
- *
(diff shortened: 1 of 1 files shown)


Janpot:
Uff, that code on https://deploy-preview-2353--mui-toolpad-docs.netlify.app/toolpad/live-config/#demo is not intended for the user, it doesn't at all reflect how you'd use that POC

Janpot (approved):

PR #2653, opened 2023-09-08
Fix gap and direction in layout slots

apedroferreira:
Dragging-and-dropping in containers was still not working 100% after the placing the second element.
This PR fixes the layout direction and adds a gap of 1 to the layout slot, just like the top page, making it so that layout slots act as a mini-page.

https://github.com/mui/mui-toolpad/assets/10789765/8667b1a4-2717-4df5-b085-d545c21d50ed

### DIFF
--- a/packages/toolpad-app/src/canvas/getPageViewState.ts
+++ b/packages/toolpad-app/src/canvas/getPageViewState.ts
@@ -114,7 +114,9 @@ export function getNodesViewInfo(rootElm: HTMLElement): {
             const display = window.getComputedStyle(childContainerElm).display;
 
             let flowDirection = 'row';
-            if (display === 'grid') {
+            if (slotType === 'layout') {
+              flowDirection = 'column';
+            } else if (display === 'grid') {
               const gridAutoFlow = window.getComputedStyle(childContainerElm).gridAutoFlow;
               flowDirection = gridAutoFlow === 'row' ? 'column' : 'row';
             } else if (display === 'flex') {
--- a/packages/toolpad-core/src/runtime.tsx
+++ b/packages/toolpad-core/src/runtime.tsx
@@ -4,6 +4,7 @@ import { Emitter } from '@mui/toolpad-utils/events';
 import * as ReactIs from 'react-is';
 import { hasOwnProperty } from '@mui/toolpad-utils/collections';
 import { createProvidedContext } from '@mui/toolpad-utils/react';
+import { Stack } from '@mui/material';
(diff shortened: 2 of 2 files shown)


Janpot (approved):
Perhaps a test screenshot during the hover?

apedroferreira:
> Perhaps a test screenshot during the hover?

The thing is that the hover has 4 directions so might be too many screenshots? 
But I could add it, I guess it's probably worth it, will create a few.
Also I'm changing the test I have added about dropping into containers to account for the additional issues I saw here.

Janpot:
> But I could add it, I guess it's probably worth it, will create a few.

If we can add it in one single test so it doesn't create too much overhead. Adding screenshots here and there is quite cheap if we don't create too many new tests for it, so we shouldn't be too shy adding them. Just as long as we make sure they are stable.

test/visual/components/index.spec.ts
```diff
@@ -43,25 +44,116 @@ test('showing grid while resizing elements', async ({ page, argosScreenshot }) =
...
+  ];
+  const getDropPreviewTopCoordinates = (boundingBox: BoundingBox): [number, number] => [
+    boundingBox!.x + boundingBox!.width / 2,
+    boundingBox!.y + (1 / 3) * boundingBox!.height,
+  ];
+  const getDropPreviewRightCoordinates = (boundingBox: BoundingBox): [number, number] => [
+    boundingBox!.x + (2 / 3) * boundingBox!.width,
+    boundingBox!.y + (2 / 3) * boundingBox!.height,
+  ];
+  const getDropPreviewBottomCoordinates = (boundingBox: BoundingBox): [number, number] => [
```
Janpot, line 94:
```suggestion
  const getDropPreviewBottomCoordinates = (boundingBox: DOMRect): [number, number] => [
```
?

test/visual/components/index.spec.ts
```diff
@@ -43,25 +44,116 @@ test('showing grid while resizing elements', async ({ page, argosScreenshot }) =
...
+  await editorModel.waitForOverlay();
+
+  const appCanvasBoundingBox = await editorModel.appCanvas.locator('body').boundingBox();
+
+  const screenshotConfig = {
+    clip: appCanvasBoundingBox || undefined,
+  };
+
+  const getDropPreviewLeftCoordinates = (boundingBox: BoundingBox): [number, number] => [
+    boundingBox!.x + (1 / 3) * boundingBox!.width,
```
Janpot, line 83: I don't believe any of these no-null assertion operators are necessary
```suggestion
    boundingBox.x + (1 / 3) * boundingBox.width,
```

Janpot (approved):

PR #2671, opened 2023-09-14
Remove unused hook

Janpot:
Removing some dead code. `useRisingEdge` was a hook to run an effect when a value changes from `false` to `true`.


michelengelen (approved):
😂💪🏻

packages/toolpad-app/src/utils/useRisingEdge.ts
```diff
@@ -1,17 +0,0 @@
-import * as React from 'react';
-import useEvent from './useEvent';
```
oliviertassinari, line 2: This looks almost identical to `@mui/utils/useEventCallback`.

Janpot: it's a direct implemention of https://github.com/reactjs/rfcs/blob/useevent/text/0000-useevent.md

oliviertassinari: Is there an opportunity to merge these two helpers? But I'm not sure https://github.com/mui/mui-toolpad/blob/0c05f1a5d655f711b704f04b3f9998a9422f776c/packages/toolpad-app/src/utils/useEvent.ts looks a bit more bundle size and none React 17 compatible. I didn't know about the `useInsertionEffect` hook, interesting.

Janpot: > Is there an opportunity to merge these two helpers?

We could use it potentially. I'm just curious, what is an _enhanced_ effect? And is there an opportunity to better name/document this hook?

> I didn't know about the `useInsertionEffect` hook, interesting.

The RFC [polyfill](https://github.com/reactjs/rfcs/blob/useevent/text/0000-useevent.md#internal-implementation) mentions an effect with comment:
```tsx
// In a real implementation, this would run before layout effects
useLayoutEffect(() => {
```
Which is what `useInsertionEffect` does. In practice we can probably rely on `useLayoutEffect`.

oliviertassinari: > I'm just curious, what is an enhanced effect? And is there an opportunity to better name/document this hook?

A use layout effect but without the server side hydration warning. Yeah, the name is subjective 😁. The name used also impacts the eslint rule, e.g. https://github.com/facebook/react/pull/25285. We would need to update https://github.com/mui/material-ui/blob/89635d1dd83b9533e5853a287df014f10f07f90f/.eslintrc.js#L123 too.

Janpot: Quick google shows mostly `useIsomorphicLayoutEffect` being used as a name. I'm just mostly interested in adding a comment there. Just a few extra points.

* I'll switch to `useEventCallback` in Toolpad
* With this `useEnhancedEffect` hook, I'd be a bit concerned. It's being applied correctly in the case of `useEventCallback`. But in many cases [the warning could be justified](https://gist.github.com/gaearon/e7d97cdf38a2907924ea12e4ebdf3c85) and using this hook would be sweeping the problem under the carpet.
* It's being redefined in some of the docs examples, e.g. https://github.com/mui/material-ui/blob/135d00a4091b3d9ff867ec8488a5a64aa5d33d21/docs/data/joy/customization/theme-colors/RemoveActiveTokens.js#L26. Perhaps this should also import the `@mui/utils` one? (The usage here seems a bit funky as well)

PR #2674, opened 2023-09-14
format IDE code

JerryWu1234:
image
#2585 


packages/toolpad-app/src/toolpad/AppEditor/BindingEditor.tsx
```diff
@@ -501,18 +504,22 @@ export function BindingEditorDialog<V>({
...
   const committedInput = React.useRef<BindableAttrValue<V> | null>(input);
 
-  const handleSave = React.useCallback(() => {
+  const handleSave = React.useCallback(async () => {
     let newValue = input;
 
     if ((input as JsExpressionAttrValue)?.$$jsExpression) {
+      const jsExpression = await tryFormatExpression(
+        (input as JsExpressionAttrValue).$$jsExpression,
+        data!,
```
Janpot, line 513: If the backend call to `getPrettierConfig` fails, or is still loading, `data` will be `undefined`. So using `!` is not the right solution as that is only reserved for when `data` is guaranteed to be defined by the programmer.

Instead, show a warning to the user that explains that the prettier config couldn't be loaded and therefore the code won't be formatted when saving.

JerryWu1234: > If the backend call to `getPrettierConfig` fails, or is still loading, `data` will be `undefined`. So using `!` is not the right solution as that is only reserved for when `data` is guaranteed to be defined by the programmer.
> 
> Instead, show a warning to the user that explains that the prettier config couldn't be loaded and therefore the code won't be formatted when saving.

what's type of warning for user?
such as https://mui.com/material-ui/react-alert/?

JerryWu1234: is there component that i can directly invoke ?

Janpot: An Alert could work. I'd have to see what it looks like. For me I think it's important it's inline, close enough to the save button (i.e. no popup)

JerryWu1234: image
I checked again.
user want to reformat the code but that code doesn't be formated because the network signal not well.
and then user must reload the website.
It will bring a new problem. such as: If the user wishes to format code without reloading the page.

JerryWu1234: image
how about this ?

JerryWu1234: >will do as to give more clues about the reason?

I'm not totally understand. what's clue? 

this error only happens without a network I think

JerryWu1234: image

Janpot: > this error only happens without a network I think

There could be a problem while prettier tries to access the file system, or a bug in our code, or in prettier, or our bundler could have transformed the code wrong, or it could be running on an incompatible node.js runtime that we're unaware about, a solar storm flipping a bit,... anything really 🙂

JerryWu1234: image
```js
const { data } = client.useQuery('getPrettierConfig', []);
```
I see.

do these errors need to be passed to front-end, if back-end occurs error ?

and also I didn't see any `try catch` in reading prettier config

JerryWu1234: @Janpot
please make a final check.

Janpot: @JerryWu1234 [`useQuery`](https://github.com/mui/mui-toolpad/pull/2674/files#diff-67eb614d84fd465e47f1730731e8e1cef4e2e15dcf3294bcfc35546866a82b2cR508) is a wrapper around `react-query`. Next to the `data` property it will return an `error` property.

packages/toolpad-app/src/toolpad/AppEditor/BindingEditor.tsx
```diff
@@ -241,8 +241,9 @@ export function ValueBindingEditor({ value, onChange, error }: ValueBindingEdito
             }}
           >
             <Typography sx={{ mb: 2, color: 'red' }}>
-              Error while reading the prettier configuration: The prettier config could not be
-              loaded and therefore the code would not be formatted
+              Error while reading the prettier configuration:
+              {(error as string) ??
```
Janpot, line 245: The `error` prop is `unknown`, casting it to a `string` potentially results in undesired behavior. `@mui/toolpad-utils/errors` has this handy `errorFrom` function which parses any input into an error object. We use it sometimes to [extract error messages](https://github.com/mui/mui-toolpad/blob/9596a8fe03c5145597ecd337ac336dcfb0ddd0f9/packages/toolpad-components/src/components/ErrorOverlay.tsx#L24) from `unknown` input.

Janpot (approved):
Great work, thanks!

PR #2677, opened 2023-09-14
Unify useDom, useDomLoader and useAppState

Janpot:
Unify the three reducers into a single one `AppState`. Having separate `useDom`, `useDomLoader` and `useAppState`, each with a different subset of the same state feels a bit complicated. We can leave the selection to consumers of the state instead. I'm also trying to add more than just the dom in this app state and the nomenclature gets all messed up.


apedroferreira (approved):
Sounds good!

When the app state became more complex I ended up splitting these by different hooks, as I wasn't sure if we should have undo/redo logic eventually be separated from the DOM loading logic, for example. But it was all open to feedback / iteration.
This is less confusing and they were still not completely decoupled so if you feel like it's fine we can have it all together!

PR #2684, opened 2023-09-15
[docs] Remove unused dependency

oliviertassinari:


oliviertassinari:
Arf, we are screwed, we import it in material-ui/docs/src/modules/components/DemoSandbox.js for the demos of the docs. I think this should be an injection point.

PR #2690, opened 2023-09-15
Global functions explorer and header

apedroferreira:
UI for global functions with explorer using the new `EditableTreeItem` component, and empty views.
We can merge already under the feature flag to keep things more separated.

Right now only works for function resources but we should discuss and see how to make it work for all types of resources, as we have data providers now.

https://github.com/mui/mui-toolpad/assets/10789765/066875c2-a260-4cb9-a0a3-6693389ceaf8

Screenshot 2023-10-12 at 18 56 26

**Possible improvements:**

- [ ] UI button to create a new handler inside a file (append at the end of file)
- [ ] Function templates


packages/toolpad-app/src/toolpad/ToolpadShell/ToolpadNavigation.tsx
```diff
@@ -8,16 +8,17 @@ import {
...
   List,
   Stack,
-  Typography,
   IconButton,
+  Tooltip,
 } from '@mui/material';
 import { Link, matchPath, useLocation } from 'react-router-dom';
 
-const DRAWER_WIDTH = 68; // px
+const DRAWER_WIDTH = 44; // px
```
apedroferreira, line 16: The sidebar was too bulky and a bit ugly, I think we can keep it smaller and simpler.

apedroferreira:
Finishing this feature/PR would take some time and it's probably not the biggest priority for now, it already took me a bit to make the other pending PRs mergeable...
We can either leave this one on hold or I don't mind closing it either and we can still get it back later.

apedroferreira:
This is good to merge, it's behind a feature flag, it was part of the progress towards having a global functions explorer as shown in the description.

packages/toolpad-app/src/toolpad/ToolpadShell/ToolpadNavigation.tsx
Janpot: Could always make it resizable :)

Janpot (approved):
Let's move forward with this

PR #2709, opened 2023-09-22
Move to vitest

Janpot:
* Migrated test configs to vitest format
* Migrated commands to use `vitest run`
* Make sure imports are used everywhere instead of relying on the globals
* Make sure testing library matchers work
* Make sure console logging detection works
* Changed one test which was testing whether an error was logged. Turns out it was faulty detecting an error related to the testing library.


apedroferreira (approved):
I had never heard about `vitest`, seems like a pretty straightforward replacement.
If it solves all the issues with ESM it sounds great!

PR #2728, opened 2023-09-27
New page creation UX

apedroferreira:
- File explorer-like creation UX for new pages, similar to what we have / will have for functions.
- Ability to rename pages and page hierarchy elements by double-clicking, using the same logic/component (will also have this later in functions)
- Sticky headers for these left-side panels, as opposed to the previous collapsible headers, and adjusted spacing/sizing

https://github.com/mui/mui-toolpad/assets/10789765/1e301b55-e1f0-4dfd-a288-0d504ba0139a


apedroferreira:
Please don't review yet, need to change the page creation tests.

Janpot:
cc @flaviendelangle might be interested in this use case, for an x-tree-view demo or for pro features.
* create/edit tree items
* scroll into view after creation

packages/toolpad-app/src/toolpad/AppEditor/PagesExplorer/index.tsx
```diff
@@ -166,22 +193,73 @@ export default function PagesExplorer({ className }: PagesExplorerProps) {
...
+    const appNode = appDom.getApp(dom);
+
+    appStateApi.update((draft) => appDom.addNode(draft, newNode, appNode, 'pages'), {
+      kind: 'page',
+      nodeId: newNode.id,
+    });
+
+    handleCloseCreateNewPage();
+
+    setTimeout(() => {
```
Janpot, line 259: `setTimeout` doesn't feel 100% right to me. I mean, I don't think there is a strong guarantee that the item has rendered. If it needs to run after a render, shouldn't it need to run in an effect?

packages/toolpad-app/src/toolpad/AppEditor/PagesExplorer/index.tsx
```diff
@@ -166,22 +193,73 @@ export default function PagesExplorer({ className }: PagesExplorerProps) {
...
+  }, [handleOpenCreateNewPage, nextProposedName]);
+
+  const handleCloseCreateNewPage = React.useCallback(() => {
+    setNewPageInput('');
+    handleCloseCreateNewPageDialog();
+  }, [handleCloseCreateNewPageDialog]);
+
+  const [anchorEl, setAnchorEl] = React.useState<HTMLButtonElement | null>(null);
+  const createNewInputRef = React.useRef(null);
+  const open = !!anchorEl;
```
Janpot, line 229: `anchorEl` is being reset in an affect when the `newPageInputErrorMsg` turns `null`. This means there is 1 render where `newPageInputErrorMsg` is `null` and `open` is `true`. Could that be the cause of the glitch we see in the video where there is just the icon in the popover?

packages/toolpad-app/src/toolpad/AppEditor/PagesExplorer/index.tsx
```diff
@@ -250,8 +329,65 @@ export default function PagesExplorer({ className }: PagesExplorerProps) {
           aria-level={1}
           labelText="Pages"
           createLabelText="Create page"
-          onCreate={handleCreatePageDialogOpen}
+          onCreate={handleCreateNewPage}
         >
+          {isCreateNewPageOpen ? (
```
Janpot, line 334: Between this and https://github.com/mui/mui-toolpad/pull/2690 is there any opportunity to share logic? I feel like in several places we need an elaborate tree component with create/edit support

bharatkashyap: Agreed on the need to have a common tree component for our "explorer" use cases with the editor. It'll take some planning and dedicated effort to bridge the gap, since currently even the components for the Hierarchy explorer and the (proposed) Queries explorer are separate. Does it make sense to create an issue to track merging all tree components into one?

packages/toolpad-app/src/toolpad/AppEditor/PagesExplorer/index.tsx
apedroferreira: This usually always works in my experience as I just want it to wait for the next render. I did it with effects but in both cases I needed to use state just for this.

packages/toolpad-app/src/toolpad/AppEditor/PagesExplorer/index.tsx
apedroferreira: Yeah this code was copied from the functions explorer so I saw that bug but hadn't really fixed it, should be good now.

packages/toolpad-app/src/toolpad/AppEditor/PagesExplorer/index.tsx
apedroferreira: I've created a `CreateTreeItem` component for now - should be able to use it in the global functions explorer too.

apedroferreira: > Does it make sense to create an issue to track merging all tree components into one?

If we want to create a whole tree view component with our own setting and features to reuse everywhere we can create a separate issue for it and do it, feels a bit early for this PR but should be a good idea long term.

bharatkashyap: Adding it as a discussion item for next to next grooming

packages/toolpad-app/src/components/CreateTreeItem.tsx
```diff
@@ -0,0 +1,151 @@
...
+    InputBaseProps,
+    'value' | 'onChange' | 'autoFocus' | 'onFocus' | 'onBlur' | 'onKeyDown' | 'fullWidth'
+  >;
+}
+
+export default function CreateTreeItem({
+  treeRef,
+  open,
+  suggestedNewItemName = '',
+  onCreate,
```
Janpot, line 22: Not sure how far to take this here. But this component is very focused on the "create" use case, while I'd expect the create logic to share 90% with the edit logic. Would an `EditableTreeItem` which is agnostic of whether it's in a create or edit scenario not make more sense?

apedroferreira: Should I try to make page names editable in the pages explorer maybe, to see how much of the logic I can share and how? And remove the ability to do it from the right-side panel? I guess that most of this logic might be sharable, but there is probably some more specific logic that I'd have to see how it fits.

Janpot:
> What's the reason for using a TreeView in the first place given that pages can't be nested, and collapsing the top-level doesn't seem useful?

We're looking for a UX similar to the vscode file explorer. The data roughly maps on the file system and has similar functionality around create/rename. It's true that pages can't be nested (yet), we do plan to use this pattern in some places in our UI which will have nested items. So far the other places have 1 level of nesting (function files with exported functions). It could be valuable to explore whether there are other UI patterns than a tree view to represent these.

We should probably remove the top-level "Pages" item. This is a leftover from a time when there were more top-level items.

packages/toolpad-app/src/toolpad/AppEditor/PagesExplorer/index.tsx
```diff
@@ -224,41 +337,49 @@ export default function PagesExplorer({ className }: PagesExplorerProps) {
...
+            nodeId="::create::"
+            isEditing
+            suggestedNewItemName={nextProposedName}
+            onEdit={handleCreateNewCommit}
+            onCancel={handleCloseCreateNewPage}
+            validateItemName={validatePageName}
+            sx={pagesTreeItemSx}
+          />
+        ) : null}
+        {sortBy(pages, 'name').map((page) => (
```
Janpot, line 369: This is mutating the `pages` array. Best to Avoid mutating variables in the render function.
It's also running a sort i every render. Feels a bit wasteful.

packages/toolpad-utils/src/collections.ts
```diff
@@ -108,3 +108,18 @@ export function equalProperties<P extends object>(obj1: P, obj2: P, subset?: (ke
   }
   return true;
 }
+
+/**
+ * Sorts an array of objects by the value of one property.
+ */
+export function sortBy<P extends object>(objs: P[], prop: keyof P, isAsc = true): P[] {
+  return objs.sort((a, b) => {
```
Janpot, line 116: None of the other exports of this file mutate the source array. Feels a bit counterintuitive, and I feel like it could lead to bugs. An alternative could be to add a 

```tsx
export function createPropertyComparator<T>(propName: string): (a: T, b: T) => number;
```

and let the consumer decide whether they want to mutate their array or not.

Janpot (commented):
Looks good. I proposed a few tweaks regarding project organization. Other remarks:

* 🤔 For visual consistency, I feel like it would be better if these are aligned

   Screenshot 2023-10-05 at 11 11 46

* Will it be confusing to remove the existing page name field? It's a bit obsolete now, but I feel like we could still offer a way to edit the name in that page options panel. Perhaps less obtrusive as the one we have now. In any case, we can keep it for a follow up PR if we leave the name existing editor in place.

apedroferreira:
> * 🤔 For visual consistency, I feel like it would be better if these are aligned

The issue is that in the hierarchy explorer we need space for the little expand/collapse arrow for nestable elements. I'll see if it works to add some more padding to the pages though, might be fine.

> * Will it be confusing to remove the existing page name field? It's a bit obsolete now, but I feel like we could still offer a way to edit the name in that page options panel. Perhaps less obtrusive as the one we have now. In any case, we can keep it for a follow up PR if we leave the name existing editor in place.

I've actually already removed it... There's pros and cons to removing it or not, on one hand it gives the users maybe a more obvious place to change the name, but on the other hand it might be confusing how it's the same thing as renaming on the left. And it might be easier for users to associate renaming in the left panel with renaming the project files. I guess I can keep the input for now and we'll see how things go from there.

packages/toolpad-app/src/utils/scroll.ts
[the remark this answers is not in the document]
apedroferreira: That's fine, I wasn't really sure where to put this.

packages/toolpad-app/src/toolpad/AppEditor/PagesExplorer/index.tsx
apedroferreira: Yeah, I noticed it was mutable, in this particular case it doesn't have too much impact but it's better to make it immutable and memoize it, I'll do it.

packages/toolpad-utils/src/collections.ts
apedroferreira: I think it's ok to just make this method immutable, simpler and more intuitive. At first I guess I wanted it to be more like `.sort`, but the way that we use it it's probably better for it just to be immutable.

packages/toolpad-app/src/components/ExplorerHeader.tsx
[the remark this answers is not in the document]
apedroferreira: No, I'm using in both the hierarchy explorer and the pages explorer. It's supposed to be the header for any "explorer" panel like this.
Eventually I can make these components part of a larger one that handles any explorer panel including the header and creatable/renamable items. I guess I wanted too keep it more modular first. But maybe I should try it already.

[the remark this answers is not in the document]
apedroferreira: Oh, I can try to move the component above just enough in the project structure probably if it makes sense, might be better than in the `components` folder then.

apedroferreira:
I've added all the changes discussed. Updated UI screenshot:

Screenshot 2023-10-05 at 20 44 46

packages/toolpad-app/src/toolpad/AppEditor/PagesExplorer/index.tsx
```diff
@@ -321,11 +321,7 @@ export default function PagesExplorer({ className }: PagesExplorerProps) {
     [appStateApi, dom],
   );
 
-  const pagesTreeItemSx = {
-    '.MuiTreeItem-label': {
-      ml: -1.5,
-    },
-  };
+  const alphabeticallySortedPages = React.useMemo(() => sortBy(pages, 'name'), [pages]);
```
Janpot, line 324: That's not alphabetically sorted. e.g. `'a' < 'B' === false`.

Not a fan of `sortBy`, I'd just do:

```tsx
[..pages].sort((page1, page2) => page1.name.localeCompare(page2.name)
```

apedroferreira: Uh, that's right about the alphabetical ordering... And nice, I didn't know about `localeCompare` or `toSorted`.
The first approach seems fine, I tried implementing the second abstraction you proposed but I'm not really sure how to? And we might not need that abstraction yet anyway...

Janpot (approved):
Great, thanks for updating!

Janpot:
@apedroferreira The position of the scrollbar seems to be off on the pages section. I don't believe position: sticky/fixed should be used for headers. The UX feels off when the scrollbar extends over the header.

Screenshot 2023-10-07 at 01 36 54

PR #2729, opened 2023-09-27
Refactor cli interface

Janpot:
First in a set of refactors which aims at exposing toolpad as a library to be embedded in your node.js application
This is mostly renaming parameters and replacing `cmd` with `dev`
Working towards a single comprehensive interface that we can expose to users to create their own Toolpad http handler

```tsx
dev: boolean // Dev/prod mode
dir: string // Toolpad project directory
port: number // port under which to run
externalUrl: string // external url
```

Still contemplating how to expose `basePath` and `externalUrl`. Perhaps a `basePath` + `externalOrigin`?

### DIFF
--- a/packages/toolpad-app/src/server/localMode.ts
+++ b/packages/toolpad-app/src/server/localMode.ts
@@ -992,7 +992,6 @@ class ToolpadProject {
   constructor(root: string, options: Partial<ToolpadProjectOptions>) {
     this.root = root;
     this.options = {
-      // cmd: 'start',
       dev: false,
       externalUrl: 'http://localhost:3000',
       ...options,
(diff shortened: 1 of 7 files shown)


packages/toolpad-app/src/server/localMode.ts
```diff
@@ -992,7 +992,7 @@ class ToolpadProject {
   constructor(root: string, options: Partial<ToolpadProjectOptions>) {
     this.root = root;
     this.options = {
-      cmd: 'start',
+      // cmd: 'start',
```
bharatkashyap, line 995: Is there a reason to leave this in commented, or a miss?

bharatkashyap (approved):

packages/toolpad-app/src/server/localMode.ts
Janpot: a miss

PR #2730, opened 2023-09-27
Add global sidebar

apedroferreira:
Add global sidebar in left-side of screen, that allows for navigating between top level concepts such as "Pages" or "Functions" ("Resources" in the future).

I don't really know how to make Material UI look that good aesthetically sometimes, so I'm open to feedback about how it looks visually too.
Uses a temporary global variable as a feature flag.

Issue in Notion: https://www.notion.so/mui-org/Build-functions-independent-of-pages-479aeb651ea14ede9210988777e0ce8b

Extracted from https://github.com/mui/mui-toolpad/pull/2690

https://github.com/mui/mui-toolpad/assets/10789765/c2ec036e-1507-4738-9369-21ad447fa38f


packages/toolpad-app/src/toolpad/Toolpad.tsx
```diff
@@ -54,9 +131,16 @@ export default function Toolpad({ basename }: ToolpadProps) {
           <ErrorBoundary fallbackRender={ErrorFallback}>
             <React.Suspense fallback={<FullPageLoader />}>
               <BrowserRouter basename={basename}>
-                <Routes>
-                  <Route path="/*" element={<AppEditor />} />
-                </Routes>
+                <AppProvider>
+                  <RouteShell>
```
apedroferreira, line 135: The shell was moved outside of all routes so it can be shared between them.

Janpot:
Looks good. I'd try to find better names for `RouteShell` and `NavigationSidebar`. Perhaps we need to find a name for this navigation surface? e.g. "App navigation", or "Toolpad navigation"?

Would it make sense to have this bar in a contrast color? cc @danilo-leal for a comment maybe?

apedroferreira:
> Looks good. I'd try to find better names for `RouteShell` and `NavigationSidebar`. Perhaps we need to declare a name for this navigation surface in our glossary? e.g. "App navigation", or "Toolpad navigation"?
> 
> Would it make sense to have this bar in a contrast color? cc @danilo-leal for a comment maybe?

I've renamed `NavigationSidebar` to `ToolpadNavigation`, I think we can call the editor navigation like that and call the in-app navigation for Toolpad apps "App Navigation"?

In terms of a glossary I only see what we have in this page https://mui.com/toolpad/concepts/building-ui/, where it seems that we could just call this new sidebar "Navigation" if we wanted to... I'm not sure how much users need to name these sidebars in particular, it's something we could discuss and then update the docs accordingly.

As for `RouteShell`, I'm not really sure what else to call it, it's just the `ToolpadShell` with different props for different routes... Any particular ideas?

bharatkashyap:
> https://mui.com/toolpad/concepts/building-ui/

Does it make sense to call them `GlobalNavigation` and `PageNavigation` instead?

As for `RouteShell`, `AppShell` makes the most sense to me since it is connected directly with the `appState`

apedroferreira:
> > https://mui.com/toolpad/concepts/building-ui/
> 
> Does it make sense to call them `GlobalNavigation` and `PageNavigation` instead?
> 
> As for `RouteShell`, `AppShell` makes the most sense to me since it is connected directly with the `appState`

Not sure about `GlobalNavigation`, I guess it's fine since it's explicit. `PageNavigation` sounds good for Toolpad apps, and `AppShell` also sounds good!
I'll either decide from those or we can all decide quickly as a team in the next grooming meeting.

Janpot:
> I'll either decide from those or we can all decide quickly as a team in the next grooming meeting.

Feel free to pick whatever you think is best of those options. Just some more background on why I thought the original names are less optimal:

* NavigationSidebar: we have more than one sidebar that does navigation. This name doesn't really disambiguate.
* RouteShell: This is naming by function, "the shell of the route", but any "shell" on the page is a shell of the route. To properly qualify it, it's more logical to name it after its content, "the shell of the toolpad editor".

apedroferreira:
Ok, I've considered all options and went with `ToopadNavigation` for the global sidebar, `AppPagesNavigation` for the in-app navigation sidebar, and `EditorShell` for the shell around the routes.
Let me know if you disagree with anything and it's fine to change them.

Janpot (approved):

PR #2736, opened 2023-09-28
Correct types of component labels

Janpot:
Avoid using `as string`, it's reducing type safety, we can just narrow down the type.

### DIFF
--- a/packages/toolpad-components/src/Checkbox.tsx
+++ b/packages/toolpad-components/src/Checkbox.tsx
@@ -19,6 +19,7 @@ import {
 
 export type FormControlLabelOptions = {
   onChange: (newValue: boolean) => void;
+  label: string;
   defaultValue: string;
   fullWidth: boolean;
 } & Omit<FormControlLabelProps, 'control' | 'onChange'> &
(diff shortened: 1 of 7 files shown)


packages/toolpad-components/src/Checkbox.tsx
```diff
@@ -17,19 +17,19 @@ import {
...
 } from './Form';
 
-export type FormControlLabelOptions = {
-  onChange: (newValue: boolean) => void;
-  defaultValue: string;
-  fullWidth: boolean;
-} & Omit<FormControlLabelProps, 'control' | 'onChange'> &
-  Omit<CheckboxProps, 'onChange'> &
-  Pick<FormInputComponentProps, 'name' | 'isRequired' | 'isInvalid'>;
+export type CheckboxProps = Omit<FormControlLabelProps, 'control' | 'onChange'> &
```
Janpot, line 20: aligning coding style with the rest of the components

apedroferreira (approved):
Right, this is definitely better, that `as` there wasn't really correct.

apedroferreira:
> Right, this is definitely better, that `as` there wasn't really correct.

Oh, just an optional thing for this, as we don't have to omit `label` the omission there for `value` and `onChange` can also be removed if we want to since those prop types are overriden.

PR #2738, opened 2023-09-28
Remove Chart as future component

Janpot:
This has been implemented. Replaced with a Password component


apedroferreira (approved):
I also had this in the x-charts PR, since it's taking a while I'll see what I can extract from there.

PR #2740, opened 2023-09-29
Support --base option to set a custom base path

Janpot:
Documentation: https://deploy-preview-2740--mui-toolpad-docs.netlify.app/toolpad/concepts/deployment/#custom-base-path

* Introduce a `--base` option to set the base path of the Toolpad application. For now we leave it by default at `/prod`, but if you set it to e.g. `/hello`, the toolpad application will live under `/hello`. The editor will still be hosted under `/_toolpad`
* Removes the `/preview` prefix in favour for the base prefix.

This further sets the stage for:
* hosting Toolpad applications as part of a custom server
* we should support custom servers in dev mode and allow connecting a toolpad editor to an application in a custom server


docs/data/toolpad/concepts/deployment.md
```diff
@@ -72,6 +72,21 @@ If you want to listen on a specific port you can change the start script to:
...
+Toolpad applications can run under a custom base path. Use the `--base` CLI parameter to set a base under which the Toolpad application is hosted.
+
+```bash
+toolpad dev --base /foo
+```
+
+Now the Toolpad application is accessible under `http://localhost:3000/foo`. The `--base` parameter must be supplied to the `build` command. A build always has one specific base path:
+
+```bash
+toolpad build --base /foo
```
apedroferreira, line 86: Would it be useful for the slash in the beginning not to be required?

apedroferreira (approved):
Looks good, just a small idea for slightly better UX.

docs/data/toolpad/concepts/deployment.md
Janpot: > Would it be useful for the slash in the beginning not to be required?

That would take away the option for us to do something useful with paths that don't have a slash. e.g. absolute paths. vite and next.js both expect a leading slash in their base path settings.

apedroferreira: Ok got it, I wasn't familiar with how other tools do it, this is good then!

PR #2746, opened 2023-10-02
Move server files to the src folder

Janpot:
So that we have a single logical place to import them in the CLI and also to export them as library. This also colocates them more with related files and harmonizes the location of worker entrypoints


Janpot:
Merging this as it's just moving files around and I'd like to start building on top of it.

PR #2747, opened 2023-10-02
Allow integrating toolpad applications in a custom server

Janpot:
Export a `unstable_createHandler` which can be used to integrated in a custom server. See example.

Can be imported as ESM, should also work under commonjs. In any case it can always dynamically be imported as ESM under commonjs.

Changes include:

- make serverside code ESM compatible
- refactor the intiialization of options so that it can be shared between cli and custom server
- export `unstable_createHandler`
- remove `projectRoot`. (dead code)
- disable the preview header in custom server (for now at least). It's not ideal because:
  - It contains the theme switcher for the editor. Which theme should it choose? Does it even make sense in the first place to have the editor theme in the preview window?
  - contains a button to open the editor, while there is nothing to open (yet)

Follow up work includes:

-  Allow editing apps running under custom server


Janpot:
Putting up for rereview. I turned this PR into exporting the toolpad handler as it turned out there were more things that needed to be done than was initially proposed

apedroferreira:
I'm getting an error for a missing `exports` property in `package.json` when trying to run the custom server example.

Screenshot 2023-10-05 at 16 30 06

examples/custom-server/index.mjs
```diff
@@ -0,0 +1,13 @@
...
+import express from 'express';
+
+const app = express();
+
+const appBase = '/my-app';
+const handler = await createHandler({
+  dev: process.env.NODE_ENV === 'development',
+  base: appBase,
+});
+app.use('/my-app', handler.handler);
```
apedroferreira, line 11: So in the example `/my-app` is the path that the whole Toolpad runs under including the editor in dev mode, and `/my-app` is the base path where the production app runs?
If so, should they be different to be more obvious?

apedroferreira (changes requested):
Just noticed a couple of things maybe worth mentioning, everything else looks good!

examples/custom-server/index.mjs
Janpot: > including the editor in dev mode

Not the editor, only the application. For editing I'm envisioning the following:

- you start it locally in dev mode.
- you run `yarn toolpad --connect http://localhost:3000/my-app` to open the editor

Janpot:
> I'm getting an error for a missing exports property in package.json when trying to run the custom server example.

How are you running it? The following should work:
- add `examples/custom-server` to the package.json  `workspaces` field
- run `yarn && yarn dev --ignore custom-server`
- run `yarn workspace custom-server dev`

examples/custom-server/index.mjs
apedroferreira: I see, that sounds cool!
Are there cases where the first argument for `app.use` and `appBase` would not be the same, though? Or could they share the same variable?
Anyway that's not important at all so I'm approving already.

apedroferreira (approved):

examples/custom-server/index.mjs
Janpot: it's an example. for demonstration purposes this is clearer.

Issue #492, opened 2022-05-29
Make responsive container of pages configurable


Janpot:
Allow for pages to stretch edge to edge, or configuration the maximum width of the fluid responsive container.

prakhargupta1:
Would this allow the user to make apps responsive?
@apedroferreira @Janpot

Janpot:
Currently the whole page is in a responsive container. If there is demand we can make that configurable. COnverting this issue to a wait for upvotes.

apedroferreira:
> 

The page is "responsive" right now, but we're not adapting anything for smaller screens, which means probably nothing will look very good in smaller screens yet.
I think this is a different thing but not that difficult to do, we could just add the ability to customize horizontal margins in the page, or setting a maximum width, that could work too.

Janpot:
Duplicate of https://github.com/mui/mui-toolpad/issues/1183

PR #2757, opened 2023-10-05
Page layout container fluid

brhndursun:
Closes https://github.com/mui/mui-toolpad/issues/492

We are unable to create fullWidth layouts due to the Container component in PageRoot.
This is why I wanted to introduce a Page Layout parameter without disrupting the existing structure.
The default value for Page Layout is set to Container.
If desired, you can choose Fluid to access 100% of the page width.

I have also updated the documentation and included screenshots.


apedroferreira:
Hi @brhndursun, thanks a lot for your PR, it seems really well done, the code in general seems good and thanks for even adding to the docs!
But before taking a deeper look into the code itself, I'm just wondering: would the aim of this feature be to solve the issue of the fixed page width only? Or are there other scenarios where you think this feature could be useful?
Because otherwise, maybe we could just have a customizable "page width" for any page, which could be set to any pixel value or 100%?

brhndursun:
> Hi @brhndursun, thanks a lot for your PR, it seems really well done, the code in general seems good and thanks for even adding to the docs! But before taking a deeper look, I'm just wondering: would the aim of this feature be to solve the issue of the fixed page width only? Or are there other scenarios where you think this feature could be useful? Because otherwise, maybe we could just have a customizable "page width" for any page, which could be set to any pixel value or 100%?

Hello @apedroferreira .

Hello @appripate. In my opinion, we shouldn't allow users to input any value for the page width as they wish. Because in a simple admin panel, it might not pose a problem, but when dealing with more pages, maintaining consistency in page widths across all of them would become a challenge if each page had to specify its width individually. Perhaps we can solve this with a global parameter, but considering frequency of use, container or full-width usage is more common. Additionally, the Container component works quite well in a responsive manner.

brhndursun:
Saw your edit after I wrote my comment. Thanks in advance.

docs/schemas/v1/definitions.json
```diff
@@ -344,6 +344,21 @@
                 }
               ],
               "description": "Display mode of the page. This can also be set at runtime with the toolpad-display query parameter"
+            },
+            "layout": {
```
apedroferreira, line 348: The team had some discussion on this - instead of `layout`, could we have.`container` page property with 2 possible values for now: `fluid` and `none`? 
`none` would be the full-width unrestricted page you proposed in this PR, and `fluid` would be the previous behavior. We feel that this would align closer to the terminology of the MUI `Container` component https://mui.com/material-ui/react-container/

packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -1199,6 +1204,18 @@ function PageRoot({ children }: PageRootProps) {
         {children}
       </Stack>
     </Container>
+  ) : (
+    <Stack
+      data-testid="page-root"
+      direction="column"
+      sx={{
```
apedroferreira, line 1211: I see no vertical margin here, we should probably keep the `my: 2` from the Stack when there is a container?
No need to write things twice either, let's write the `<Stack` component only once  and make some of the styles change depending on the `container` setting.

packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -1165,7 +1165,7 @@ function RenderedNodeContent({ node, childNodeGroups, Component }: RenderedNodeC
       NodeError={NodeError}
     >
       {isLayoutNode ? (
-        <Component {...wrappedProps} />
+        <Component {...wrappedProps} attributes={node.attributes} />
```
apedroferreira, line 1168: Not sure about passing the attributes to every node here, they're supposed to have the component props only... 
Maybe let's just pass a different version of the `PageRootComponent` to `<RenderedNodeContent` depending on whether there is a container or not? One with the container layout and one with the layout without a container? That way we can just read from `page.attributes` inside the `RenderedPage` component.

packages/toolpad-app/src/toolpad/AppEditor/PageLayoutEditor.tsx
```diff
@@ -0,0 +1,53 @@
...
+interface PageLayoutEditorProps {
+  node: PageNode;
+}
+
+const PAGE_LAYOUT_OPTIONS: { value: PageLayoutMode; label: string }[] = [
+  { value: 'container', label: 'Container' },
+  { value: 'fluid', label: 'Fluid' },
+];
+
+export default function PageLayoutEditor({ node }: PageLayoutEditorProps) {
```
apedroferreira, line 15: This suggestion is completely optional, so don't worry about it: but this control might work better as a toggle as there's only 2 options? Maybe https://mui.com/material-ui/react-toggle-button/
Anyway no need to do this in this PR unless you really want to - we plan to make the same adjustment for the page `display` attribute too.

packages/toolpad-app/src/toolpad/AppEditor/PageLayoutEditor.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const domApi = useDomApi();
+  const [pageLayoutInput, setpageLayoutInput] = React.useState(node.attributes.layout);
```
apedroferreira, line 17:
```suggestion
  const [pageLayoutInput, setPageLayoutInput] = React.useState(node.attributes.layout);
```

apedroferreira (changes requested):
Looks great and the changes are easy to follow, also thanks again for adding this feature as it had already been requested before!
Adding a few requested changes, hopefully shouldn't be too much!

apedroferreira:
I can continue on your PR soon and finish it up, thanks for your initial contribution @brhndursun !

Issue #2360, opened 2023-07-25
Page hierarchy explorer doesn't always select


Janpot:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

Steps:

1. build Toolpad `yarn && yarn release:build`
2. start Toolpad dev on the basic backend fixture: `yarn toolpad dev test/integration/backend-basic/fixture`
3. open the "basic" page
4. try selecting items in the page hierarchy explorer

### Current behavior 😯

1. It doesn't always select the item (see video)
2. hovering over page rows doesn't highlight them
3. the title "components" is very ambiguous. Components for me are like templates, they are blueprints. When they're instantiated on the page I prefer to talk about nodes, or elements.

https://github.com/mui/mui-toolpad/assets/2109932/89208dd8-5a26-43ec-9927-36651b9b380a

### Expected behavior 🤔

1. It always selects the node when I click it
2. hovering over page rows highlights the node in the canvas
3. I propose we change the title to "page hierarchy"

bharatkashyap:
@Janpot Is there any additional information you could share? Context/Browser/error messages in the console? I can't replicate 1. without those.


https://github.com/mui/mui-toolpad/assets/19550456/2b9e58fb-2d79-430f-b40b-7239b293c1b3

Agree, I'll make changes for 2. and 3.

Janpot:
Yes, it seems to happen when I click the text itself, or the vertical center in the whitespace left of the text, basically anything in this area:

![Screenshot 2023-07-25 at 10 40 50](https://github.com/mui/mui-toolpad/assets/2109932/5245ce29-4ede-432e-b1d3-89398e34e501)

Janpot:
I also updated the issue with 4. and 5. that I ran into.

bharatkashyap:
On 3., "Components" makes sense to me given that :

- We call the catalog which contains them the "Component Library" 

- Benchmarking:

| Appsmith | Dynaboard | Retool |
| -------| -----------|  --------- |
[...]

PR #2762, opened 2023-10-06
[core] Use `readonly` for Editable Text

bharatkashyap:
- Partial fix of https://github.com/mui/mui-toolpad/issues/2360 (P2)

https://github.com/mui/mui-toolpad/assets/19550456/cc688589-cd0c-4a59-928f-4bdb5495e682

### DIFF
--- a/packages/toolpad-app/src/components/EditableText.tsx
+++ b/packages/toolpad-app/src/components/EditableText.tsx
@@ -5,7 +5,6 @@ import {
   TypographyVariant,
   SxProps,
   inputBaseClasses,
-  formHelperTextClasses,
   inputClasses,
 } from '@mui/material';
 
@@ -58,6 +57,8 @@ const EditableText = React.forwardRef<HTMLInputElement, EditableTextProps>(
       }
     }, [ref, editable]);
 
+    const readOnly = React.useMemo(() => disabled || !editable, [disabled, editable]);
+
     const handleBlur = React.useCallback(
       (event: React.FocusEvent<HTMLInputElement>) => {
         // This is only triggered on a click away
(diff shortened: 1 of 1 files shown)


Janpot:
What's the difference between a `<EditableText readonly />` and a `<Typography />`?

apedroferreira:
> What's the difference between a `<EditableText readonly />` and a `<Typography />`?
> 
> I feel like there is some code to be shared with #2728 around creating/editing tree items. I'm not sure whether EditableText is the right basis for editable tree items 🤔

I completely replaced EditableText with the new tree item component in my PR, it felt like the best approach :/ But let me know if there are any advantages here that I'm missing and we can switch back.

Janpot:
> I completely replaced EditableText with the new tree item component in the hierarchy explorer in my PR, it felt like the best approach :/ But let me know if there are any advantages here that I'm missing and we can switch back.

No, I agree, it's tempting to try and reuse editable text for editable tree items, but I don't believe it's a good fit.

bharatkashyap:
> I completely replaced EditableText with the new tree item component in the hierarchy explorer in my PR, it felt like the best approach :/ But let me know if there are any advantages here that I'm missing and we can switch back.
> 
> Edit: also maybe it's useful to create a hook to share some similar logic between those 2 components, I didn't really look into that.

Ah, I hadn't noticed that. To me it looks fine to continue with the approach in #2728 for Tree Items.  I feel we can preserve this fix for `EditableText` regardless.

> What's the difference between a `<EditableText readonly />` and a `<Typography />`?
I was previously switching between a `<TextField>` and a `<Typography>` for the `<EditableText>` but found it to be better form to only switch attributes on the same component instead of replacing entire markup blocks - the latter causes subtle layout shifts  and styling inconsistencies which require duplication to be fixed, among other things.

bharatkashyap:
Closing since we aren't using this anymore

Issue #2616, opened 2023-09-02
Allow to compose env variables in HTTP request


oliviertassinari:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

I would like to use my process.env.GITHUB_TOKEN token, so I wired the value:

<img width="405" alt="Screenshot 2023-09-02 at 15 59 53" src="https://github.com/mui/mui-toolpad/assets/3165635/237a6572-497f-470e-b7fc-e1bc0efadd43">

However, I need to have this logic:

```jsx
Authorization: `Bearer ${process.env.GITHUB_TOKEN}`,
```

for the token to be correctly recognized.

Janpot:
One solution could be to allow users to bind parameters, not only to UI state, but also to environment variables. That way we could have the best of both worlds:
- Explicit declaration of which parameters are in use by our queries
- Full freedom in how to compose them inside of the different query properties

apedroferreira:
> One solution could be to allow users to bind parameters, not only to UI state, but also to environment variables.

So I guess that we could allow the user to bind to an expression in these cases, and in these cases that expression would have access to the values of the environment variables. We could try something like that?

Janpot:
> So I guess that we could allow the user to bind to an expression in these cases, and in these cases that expression would have access to the values of the environment variables. We could try something like that?

Just to be sure I expressed myself clearly. What I meant was to enable binding to an environment variable in the parameter bindings:

![Screenshot 2023-09-05 at 09 56 53](https://github.com/mui/mui-toolpad/assets/2109932/62bc7e7b-32b6-4410-9b6b-a642f9e797bc)

So that in the query bindings one could use it like

![Screenshot 2023-09-05 at 09 58 19](https://github.com/mui/mui-toolpad/assets/2109932/f57b8f8e-3785-42fb-bf5c-9887f841bc90)


It's just a proposal, maybe it's still too convoluted. 

What I like about it:
- it is explicit and lintable. We can warn on app startup when a variable is not supplied.
- it centralizes all inputs for a query. 
- it'll be easy to extend with e.g. cookie support.
- more options to compose env variables in bindings. now you can only bind the whole variable, with this new setup you can derive other values from them. (e.g. `btoa(`${parameters.user}:${parameters.password}`);`)

What I don't like
- It adds extra steps before being able to bind to a secret.

<hr>

@JerryWu1234 You can find it when opening the binding editor for a header:

[...]

PR #2767, opened 2023-10-07
a new feature that let user add env variable.

JerryWu1234:
#2616 


JerryWu1234:
image

JerryWu1234:
please help checking before I add a test

apedroferreira:
Good start @JerryWu1234 !
I'm still not 100% sure what the best solution is, but here's what I think we could try now:

What if we change ```params: params: Record<string, string>``` in `execBase` to ```params: [string, BindableAttrValue<string>][]```, so that we always pass the binding entries there instead of the binding values?

That way we don't need the changes you made in [packages/toolpad-core/src/jsRuntime.tsx](https://github.com/mui/mui-toolpad/pull/2767/files#diff-8743c8b985952b64bf101678990a804d780065cb492f3bef76e32f475b425be5) and in [packages/toolpad-app/src/toolpadDataSources/rest/server.ts](https://github.com/mui/mui-toolpad/pull/2767/files#diff-4a7aa1ca4a54a2a0238476a8b5c22e015ba156791d90175b2dcfc49036b0c32e) we just need ```
parameters: resolveBindableEntries(jsRuntime, params, queryScope)```

apedroferreira:
> Good start @JerryWu1234 ! I'm still not 100% sure what the best solution or if this will work without any new issues, but here's what I think we could try:
> 
> What if we change the argument `params: params: Record<string, string>` in `execBase` to `params: [string, BindableAttrValue<string>][]`, so that we always pass the binding entries there instead of the binding values?
> 
> That way we don't need the changes you made in [packages/toolpad-core/src/jsRuntime.tsx](https://github.com/mui/mui-toolpad/pull/2767/files#diff-8743c8b985952b64bf101678990a804d780065cb492f3bef76e32f475b425be5), and in [packages/toolpad-app/src/toolpadDataSources/rest/server.ts](https://github.com/mui/mui-toolpad/pull/2767/files#diff-4a7aa1ca4a54a2a0238476a8b5c22e015ba156791d90175b2dcfc49036b0c32e) we just need `parameters: resolveBindableEntries(jsRuntime, params, queryScope)`

If this suggestion does not work for client-side parameters, we can keep client-side parameters the way they are but handle the server-side parameters separately, the way I said above.

apedroferreira:
Ok, if we can get everything to work in https://github.com/mui/mui-toolpad/pull/2777 we don't need to follow this approach here.

JerryWu1234:
I wil close this pr

PR #2774, opened 2023-10-09
Demonstrate how data providers can be used with prisma

Janpot:
Two simple prisma based data providers. One demostrating index based pagination, the other demonstrating cursor based pagination


examples/with-prisma-data-provider/prisma/migrations/20230704094414_init/migration.sql
```diff
@@ -0,0 +1,19 @@
+-- CreateTable
```
apedroferreira, line 1: Do the example start with an empty database if you run them, and the user has to create the records themselves?
If that's the case could maybe we could seed some dummy data https://www.prisma.io/docs/guides/migrate/seed-database?

examples/with-prisma-data-provider/toolpad/resources/usersByCursor.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+import { createDataProvider } from '@mui/toolpad/server';
+import { PrismaClient } from '@prisma/client';
+
+// Reuse existing PrismaClient instance during development
+(globalThis as any).__prisma ??= new PrismaClient();
+const prisma: PrismaClient = (globalThis as any).__prisma;
+
+export default createDataProvider({
```
apedroferreira, line 14: Btw this is great, I think these examples really help!

apedroferreira (approved):
Looks great!

PR #2777, opened 2023-10-10
Allow binding query parameters to environment variables

JerryWu1234:
#2616 


JerryWu1234:
image

JerryWu1234:
image

JerryWu1234:
image

Janpot:
Ok, I think we can bring env var values to the editor.

apedroferreira:
Agree, it should be fine to show the `env` values in the preview panel, it might even be better in some ways and it show more useful information to the user.
But the queries that are not called in preview mode are still missing values for the parameters that are associated with environment variables, so we still need to make that work...

JerryWu1234:
> Agree, it should be fine to show the `env` values in the preview panel, it might even be better in some ways as it shows more useful information to the user. But the queries that are not called in preview mode are still missing values for the parameters that are associated with environment variables, so we still need to make that work...

image
image

part of query

JerryWu1234:
I have a suggestion.
I saw that the backend will handle [the env variable](https://github.com/JerryWu1234/mui-toolpad/blob/88419053bb4cf8da4e34f6db496d414ff9943584/packages/toolpad-app/src/toolpadDataSources/rest/server.ts#L135).
but now could we handle all the transformation in the editor front-end.
image

@apedroferreira @Janpot

packages/toolpad-core/src/jsServerRuntime.tsx
```diff
@@ -27,10 +27,9 @@ export function createServerJsRuntime(env?: Record<string, string | undefined>):
...
 }
 
-export function useServerJsRuntime(): JsRuntime {
+export function useServerJsRuntime(processEnv: Record<string, string | undefined> = {}): JsRuntime {
   return React.useMemo(() => {
     // process.env is not available in the browser
-    const processEnv = {};
     return createServerJsRuntime(processEnv);
-  }, []);
+  }, [processEnv]);
```
apedroferreira, line 34: I don't think we can do this...
The runtime cannot have access to the environment variables, they should only be accessible on the server.

packages/toolpad-app/src/toolpadDataSources/rest/types.ts
```diff
@@ -125,5 +125,5 @@ export interface FetchResult extends ExecFetchResult<any> {
 }
 
 export type IntrospectionResult = {
-  envVarNames: string[];
+  envVarNames: Record<string, string | undefined>;
```
apedroferreira, line 128: Let's call this just `env` now that it includes the whole environment.

packages/toolpad-app/src/toolpadDataSources/rest/client.tsx
```diff
@@ -530,7 +532,8 @@ function QueryEditor({
                 globalScope={globalScope}
                 globalScopeMeta={globalScopeMeta}
                 liveValue={paramsEditorLiveValue}
-                jsRuntime={jsBrowserRuntime}
+                jsRuntime={jsServerRuntime}
+                envVarNames={envVarNames}
```
apedroferreira, line 536: We can also pass the whole `env` to these props now instead of just the `envVarNames`, and rename all `envVarNames` to `env`.

JerryWu1234:
> We want to be able to assign and use environment variables as Parameters - not really in Url query, 
> But the queries that are not called in preview mode are still missing values for the parameters that are associated with environment variables, so we still need to make that work...

I got it ,  and now it's going on.

image
image

-----

>I think we can try passing the environment variable parameters to execBase as a new argument, maybe we can call it envParams?: : [string, BindableAttrValue<string>][] = []. But this will require refactoring in other places to make sure we always pass the environment bindings until they reach execBase.
OR (different idea) we can try making a new envParams that works the same as searchParams, but we still have to show those in the Parameters section.


there i'm not  totally understand.
Because All of values we got it in the editor front-end.
we are parsed all value and pass to the editor back-end.

such as 

we have a environment variable
 
```bash
SECRET_BAR="Some bar secret"
```

image

I create `parameter` and `headers`, finally I click the preview button, and send a require.

image


```json
{
  "name": "dataSourceFetchPrivate",
  "params": [
    "rest",
    null,
    {
      "kind": "debugExec",
      "query": {
        "method": "GET",
        "headers": [
          [
            "var",
            {
              "$$env": "SECRET_BAR"
            }
          ]
        ],
        "browser": false,
        "searchParams": [
          [
            "aa",
            {
              "$$jsExpression": "`?code=${parameters.sd}`\n"
            }
          ]
        ]
      },
      "params": {
        "sd": "Some bar secret"
      }
    }
  ]
}
```` 

because we already have all of the environment variable.
We can directly parse this variable.
such as 
```json
"headers": [
            [
              "var",
              {
                "$$env": "SECRET_BAR"
              }
          ]

to  

"headers":{
  "var": "Some bar secret"
}

and 

"searchParams": [
          [
            "aa",
            {
              "$$jsExpression": "`?code=${parameters.sd}`\n"
            }
          ]
        ]
to 

"searchParams": {
  "aa": "?code=`Some bar secret"
}
```

final version 

```json

{
  "name": "dataSourceFetchPrivate",
  "params": [
    "rest",
    null,
    {
      "kind": "debugExec",
      "query": {
        "method": "GET",
        "headers": {
          "var": "Some bar secret"
        },
        "browser": false,
        "searchParams": {
          "aa": "?code=Some bar secre"
        }
      },
      "params": {
        "sd": "Some bar secret"
      }
    }
  ]
}
```

is there any problem in this way?
welcome point out problem

Janpot:
We have to make a distinction between the Toolpad editor and the Toolpad application. The latter is what gets served when running `toolpad start`. It's acceptable to have the environment variables available in the editor. Under no circumstance is it acceptable to have environment variables in the end user application. Not hidden, not ever. For the end user applications environment variables must always be resolved serverside.

Within the "private query" we can add a method to retrieve the environment variables of the process and variable names that exist in .env file. Those can be different. The former is used to resolve `$$env` bindings, the latter is used for autocomplete. It's only for preview purposes. To run the actual query, the env variables need to be calculated serverside.

JerryWu1234:
image

JerryWu1234:
@apedroferreira I done

packages/toolpad-app/src/toolpad/AppEditor/BindingEditor.tsx
```diff
@@ -154,7 +154,7 @@ export function EnvBindingEditor({ value, onChange }: EnvBindingEditorProps) {
       <Typography>Assign to an environment variable</Typography>
       <Autocomplete
         freeSolo
-        options={envVarNames}
+        options={env}
```
apedroferreira, line 157: ... then here we can pass the environment variable names.

packages/toolpad-app/src/toolpad/AppEditor/BindingEditor.tsx
```diff
@@ -138,8 +138,8 @@ function JsExpressionPreview({ jsRuntime, input, globalScope }: JsExpressionPrev
 export interface EnvBindingEditorProps extends WithControlledProp<EnvAttrValue | null> {}
 
 export function EnvBindingEditor({ value, onChange }: EnvBindingEditorProps) {
-  const { envVarNames = [] } = useBindingEditorContext();
-
+  const context = useBindingEditorContext();
+  const env = Object.values(context.env ?? {});
```
apedroferreira, line 142: Let's still show the keys here, `Object.keys(env)`, not `Object.values(env)`. 
We can call the variable `envVarNames` like the previous name => `const envVarNames = Object.keys(env)`

apedroferreira (changes requested):
The latest changes are great, thank you! 
We just still need to use the environment variables **names** in packages/toolpad-app/src/toolpad/AppEditor/BindingEditor.tsx, instead of the **values**.

We still also need to solve the main remaining problem: all these changes work for the "Preview" mode in the editor, but if you try it, you will see that they do not work in a live application. It only works when you click the "Preview" button. In all other cases you will see that the values are still `undefined`.
So we also need a way for the running application to be able to read and resolve environment bindings from the parameters - and we can't pass the `env` values there directly like we do in preview, because they can include secrets. We can only pass the bindings (of type `[string, BindableAttrValue<string>][]`), which then need to be resolved securely in the back-end.
I have already written some suggestions to try to solve this part of the problem before, in previous comments. But if you feel like that part is too much I can try to add a solution on top of your PR!

JerryWu1234:
> The latest changes are great, thank you! We just still need to use the environment variables **names** in `packages/toolpad-app/src/toolpad/AppEditor/BindingEditor.tsx`, instead of the **values**.
> 
> We still also need to solve the main remaining problem: all these changes work for the "Preview" mode in the editor, but if you try it, you will see that they do not work in a live application. It only works when you click the "Preview" button. In all other cases you will see that the values are still `undefined`. So we also need a way for the running application to be able to read and resolve environment bindings from the parameters - and we can't pass the `env` values there directly like we do in preview, because they can include secrets. We can only pass the bindings (of type `[string, BindableAttrValue<string>][]`), which then need to be resolved securely in the back-end. I have already written some suggestions to try to solve this part of the problem before, in previous comments. But if you feel like that part is too much I can try to add a solution on top of your PR!

image

I guess you doesn't say this preview. right?

image
this preview 

just make sure

JerryWu1234:
I think definitely we weren't on the same page about `preview`.
two Preview buttons in the editor. LoL.

image

if I were right, you would want to say this preview.

JerryWu1234:
> But if you feel like that part is too much I can try to add a solution on top of your PR!

i will do it

apedroferreira:
> I think definitely we weren't on the same page about `preview`. two Preview buttons in the editor. LoL.

Oh yeah, there's 2 things called preview, so it can be confusing... Thanks for bringing that up, maybe we can think of different names for them :D

I actually meant the "preview" that runs a **query**, the one in the triangle "play button" - right now the environment variables only work when you call them by clicking that button. They don't work in the application itself.

JerryWu1234:
image
image

this way isn't proper.
There was an error when the user refreshed the website and reopened. 
So I have to tweak the code.

JerryWu1234:
image
image

JerryWu1234:
https://github.com/mui/mui-toolpad/assets/27261522/906e206d-a768-40f0-9516-7440f512ac01


https://github.com/mui/mui-toolpad/assets/27261522/02c969ae-ec33-4b08-b7f9-d1e9965ac36c

@apedroferreira I done that both were worked

apedroferreira:
> 20231019-074612.mp4 
>  20231019-074651.mp4 
> @apedroferreira I done that both were worked

Hey @JerryWu1234 ,

I see that you are resolving the environment bindings to their value directly in the parameters. This works but we can't do it because it's not secure, it's actually very dangerous...

Many people using Toolpad may include secret keys in their environment variables, that can provide access to their private services. It's fine if someone who is editing their local Toolpad app in their local environment can see the environment variables values in Toolpad, **but not an external user of their app**.
This means that **we should never give the browser access to the values of the environment variables**. We can show them in the query editor like you did, using the `introspection` query, but not anywhere else, such as in the page content. The `introspection` query is safe to use because it can only be called by a local user in dev mode, but it only works in those conditions.

So we still need the `page.yml` files to show environment variables in parameters as `$$env: ENV_VAR_NAME`, because we should never show the value there.
And in order to retrieve the value when the page calls a query, we need to pass in that $$env: ENV_VAR_NAME` binding to the server when the query is called. The server has access to `process.env`, and in the server-side it is secure to get `process.env[ENV_VAR_NAME]`. 
The bindings for `URL query`, for example, already work server-side, as they are resolved in`rest/server.ts`.
That's why I proposed that kind of solution in previous comments in this PR.

JerryWu1234:
image

JerryWu1234:
image

I think this pr follow your idea.

JerryWu1234:
@apedroferreira I follow your idea. please tell me if there is any missing compared with your idea

packages/toolpad-app/src/toolpadDataSources/rest/server.ts
```diff
@@ -137,20 +137,22 @@ async function execBase(
   connection: Maybe<RestConnectionParams>,
   fetchQuery: FetchQuery,
   params: Record<string, string>,
+  envParams?: Record<string, string>,
```
apedroferreira, line 140: Maybe we don't need this `envParams` argument after all? 
I think `resolveBindableEntries` already resolves env variable parameters by itself. Then we can just pass an empty object `{}` as the scope for the `resolveBindableEntries` below.

packages/toolpad-app/src/toolpadDataSources/rest/server.ts
```diff
@@ -137,20 +137,22 @@ async function execBase(
...
 ): Promise<FetchResult> {
   const har = createHarLog();
   const instrumentedFetch = withHarInstrumentation(fetch, { har });
   const jsRuntime = createServerJsRuntime(process.env);
 
+  const parameter = resolveBindableEntries(jsRuntime, Object.entries(params), envParams!);
```
apedroferreira, line 146: Let's call this variable `resolvedParams`

packages/toolpad-app/src/toolpadDataSources/rest/server.ts
```diff
@@ -137,20 +137,22 @@ async function execBase(
   connection: Maybe<RestConnectionParams>,
   fetchQuery: FetchQuery,
   params: Record<string, string>,
```
apedroferreira, line 139: I guess that we need to update the type of the `params` argument now: besides having binding values directly, it can now also have bindings (type `BindableAttrValue`)

packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -376,6 +376,13 @@ function parseBinding(
       result: { value: bindable },
     };
   }
+
+  if (bindingType === 'env') {
+    return {
+      scopePath,
+      result: { value: bindable },
```
apedroferreira, line 383: Could we make this return a new 
```
return {
      scopePath,
      envVarName: <% (name of environment variable) %>,
    };
```

and then make it so that the final arguments to `execBase` are the same as right now?

apedroferreira (changes requested):
Awesome, this approach seems to work!! Thank you so much!
Just added a few more suggestions, but we're almost there!

Also 2 more things:
1. Could we make the environment bindings not appear in the binding panel and be `undefined`?
Screenshot 2023-10-24 at 23 29 46
2. We can add tests later separately if you prefer, but in the end it would also be nice to have an integration test where we have an environment variable assigned to a parameter, and then we use it in the query headers by binding to the expression: `Bearer ${params.envVarName}`

JerryWu1234:
>We can add tests later separately if you prefer, but in the end it would also be nice to have an integration test where we have an environment variable assigned to a parameter, and then we use it in the query headers by binding to the expression: Bearer ${params.envVarName}

I will add some tests if review is done

JerryWu1234:
> Awesome, this approach seems to work!! Thank you so much! Just added a few more suggestions, but we're almost there!
> 
> Also 2 more things:
> 
> 1. Could we make the environment bindings not appear in the binding panel and be `undefined` in the client-side?
> 
> Screenshot 2023-10-24 at 23 29 46
> 2. We can add tests later separately if you prefer, but in the end it would also be nice to have an integration test where we have an environment variable assigned to a parameter, and then we use it in the query headers by binding to the expression: `Bearer ${params.envVarName}`

image

JerryWu1234:
@apedroferreira I done

packages/toolpad-app/src/toolpadDataSources/rest/server.ts
```diff
@@ -136,21 +136,21 @@ async function execBase(
   project: IToolpadProject,
   connection: Maybe<RestConnectionParams>,
   fetchQuery: FetchQuery,
-  params: Record<string, string>,
+  params: BindableAttrValue<any>,
```
apedroferreira, line 139: I think the correct type now here would be something like `Record<string, string | BindableAttrValue<any>>`?

packages/toolpad-app/src/runtime/ToolpadApp.tsx
apedroferreira: Did you try this? Any outcome?

packages/toolpad-app/src/toolpad/AppEditor/GlobalScopeExplorer.tsx
```diff
@@ -54,10 +54,28 @@ function groupScopeMeta(value: Record<string, unknown>, meta: ScopeMeta): Explor
     structure[group].items.push({
```
apedroferreira, line 54: Could we just `.filter` out the env bindings from `structure[group].items` instead so that they don't show at all?

apedroferreira (changes requested):
Thank you! Some more comments:

JerryWu1234:
@apedroferreira please help check

JerryWu1234:
> Hi @JerryWu1234 , it's almost there but we still need to solve these cases, after that we can probably merge:
> 
> 1. When using bindings on the page, not only do I want the environment variable parameters to be `undefined`, but they should not even appear. They should be hidden.
> 
>  Screen.Recording.2023-11-02.at.14.16.53.mov 
> 2. But in the query editor, I want to be able to see the real value, just like you had it before. Right now it splits the string into an array so the logic there is not working correctly:
> 
>  Screen.Recording.2023-11-02.at.14.17.17.mov

image
image

JerryWu1234:
image

apedroferreira (changes requested):
Thanks a lot for your work so far @JerryWu1234 , I know it has been a long process. :)
You can just add this suggested change and then if there's nothing else you would like to add I think we should be good!

apedroferreira (approved):
Thank you @JerryWu1234 , this was a really important feature!
We might add some refinements later but this solution will work great for now!

PR #2778, opened 2023-10-10
[core] Smoothen edges on the new Editable Tree component

bharatkashyap:
### Before
https://github.com/mui/mui-toolpad/assets/19550456/e4108a43-b1af-4678-a616-5480a2a92eec

### After
https://github.com/mui/mui-toolpad/assets/19550456/e337f9d0-43de-457b-8043-89c83de73c2b

### DIFF
--- a/packages/toolpad-app/src/components/EditableTreeItem.tsx
+++ b/packages/toolpad-app/src/components/EditableTreeItem.tsx
@@ -81,11 +81,9 @@ export default function EditableTreeItem({
       handleCancel();
       return;
     }
-
     if (onEdit) {
       onEdit(itemNameInput);
     }
-
     setItemNameInput('');
     setIsInternalEditing(false);
   }, [handleCancel, itemNameInput, newItemValidationResult.isValid, onEdit]);
@@ -96,6 +94,7 @@ export default function EditableTreeItem({
 
   const handleFocus = React.useCallback((event: React.FocusEvent<HTMLInputElement>) => {
     event.target.select();
+    event.stopPropagation();
   }, []);
 
   const handleBlur = React.useCallback(
@@ -113,6 +112,8 @@ export default function EditableTreeItem({
     (event: React.KeyboardEvent<HTMLInputElement>) => {
       if (event.key === 'Enter') {
         handleConfirm();
+      } else if (event.key === 'Backspace') {
+        event.stopPropagation();
       } else if (event.key === 'Escape') {
         handleCancel();
         event.stopPropagation();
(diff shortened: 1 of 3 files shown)


packages/toolpad-app/src/toolpad/AppEditor/NodeMenu.tsx
```diff
@@ -61,10 +60,10 @@ export default function NodeMenu({
 
   const handleRenameClick = React.useCallback(
     (event: React.MouseEvent) => {
+      onRenameNode?.();
```
apedroferreira, line 63: Maybe we can still pass the node id just in case?

packages/toolpad-app/src/toolpad/AppEditor/NodeMenu.tsx
```diff
@@ -88,12 +87,14 @@ export default function NodeMenu({
           menuProps.onClick?.(event);
         }}
       >
-        <MenuItem onClick={handleRenameClick}>
-          <ListItemIcon>
-            <ModeEditIcon />
-          </ListItemIcon>
-          <ListItemText>{renameLabelText}</ListItemText>
-        </MenuItem>
+        {onRenameNode ? (
```
apedroferreira, line 90: Should duplicate and delete also have this condition?

apedroferreira (approved):
Looks good, thanks a lot for finding the issue and fixing it, I guess I had missed it!

packages/toolpad-app/src/components/EditableTreeItem.tsx
```diff
@@ -113,6 +112,8 @@ export default function EditableTreeItem({
     (event: React.KeyboardEvent<HTMLInputElement>) => {
       if (event.key === 'Enter') {
         handleConfirm();
+      } else if (event.key === 'Backspace') {
+        event.stopPropagation();
```
apedroferreira, line 116: Oh, also maybe we should just stop propagation at the start no matter what they key is, to avoid future issues?

packages/toolpad-app/src/toolpad/AppEditor/NodeMenu.tsx
bharatkashyap: Yep, makes sense

packages/toolpad-app/src/components/EditableTreeItem.tsx
```diff
@@ -96,6 +94,7 @@ export default function EditableTreeItem({
 
   const handleFocus = React.useCallback((event: React.FocusEvent<HTMLInputElement>) => {
     event.target.select();
+    event.stopPropagation();
```
Janpot, line 97: Just out of curiosity, what's the reason to stop propagation here?

bharatkashyap: The issue was that the `autoFocus` triggered the `handleFocus` handler, which somehow also caused the `handleBlur` handler to immediately trigger and cause the input to become unfocused, nullifying the effect of clicking the Rename button. Stopping propagation in the focus handler fixed that.

PR #2783, opened 2023-10-11
Turn integration tests into a yarn workspace

Janpot:
Extracting from https://github.com/mui/mui-toolpad/pull/2546 to reduce the footprint of that PR. We can't rely on dependency hoisting for the tests to find the `@mui/toolpad`package. This turns it into a workspace. It also makes use ESM to fix issues with the `get-port` package


Janpot:
Going to merge this as I could use it for building a custom server integration test

Issue #2789, opened 2023-10-11
Page creation doesn't work with default name


Janpot:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

Steps:

1. click "create page" button
2. click outside of the page explorer without editing the name

### Current behavior 😯

The page disappears

https://github.com/mui/mui-toolpad/assets/2109932/c9d521fa-b76e-459f-84b5-7b96f778db21

### Expected behavior 🤔

The page gets created just like when you give it a custom name

### Your environment 🌎

<details>
  <summary><code>npx @mui/envinfo</code></summary>

```
  Don't forget to mention which browser you used.
  Output from `npx @mui/envinfo` goes here.
```
</details>

apedroferreira:
I think I did this intentionally because I found that I was creating pages when I didn't want to.
Maybe we can remove it though, I wasn't too sure about it.

Janpot:
I guess the current behavior makes sense if the input is not prefilled, like in vscode. The proposed behavior is consistent with e.g. creating a folder in mac finder.

I think either is wide enough accepted as a UX. I'm fine with what you did in the PR

PR #2790, opened 2023-10-11
Fix creating new page with blur and default name + scroll only below explorer headers

apedroferreira:
Fixes https://github.com/mui/mui-toolpad/pull/2728#issuecomment-1751504508 (scroll only below explorer headers)
Fixes https://github.com/mui/mui-toolpad/issues/2789 (make it possible to create new pages with default name on blur)

https://github.com/mui/mui-toolpad/assets/10789765/bf5e7108-bc4c-4375-9dc3-9e9fe0ca41ec


Janpot (approved):
I guess the current behavior makes sense if the input is not prefilled, like in vscode.

PR #2798, opened 2023-10-13
Add tests for custom server

Janpot:
Refactored the test infra a bit so we can choose to have fixtures for custom server or cli.

I designed the primitives to use explicit resource management. This allows us to use the same code for playwright fixtures, and for more classic before/after setup.

- [x] add prod test and dev test for custom server
- [x] Fixed an issue with dev mode not running the function types worker
- [x] Fixed an issue with prod mode not picking up the environment variables

Next up is creating and testing the `toolpad editor http://localhost:3000/my-custom-server` command


Janpot:
Merging this before my holidays. Feel free to review retroactively

test/integration/custom-server/prod.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+  },
+  customServerConfig: {
+    dev: false,
+    env: {
+      SECRET_BAZ: 'Some baz secret',
+    },
+  },
+});
+
+test('custom server production mode', async ({ context, customServer, page }) => {
```
apedroferreira, line 21: Should we extract the test content somewhere so it doesn't have to be written twice?

test/integration/editor/new.spec.ts
```diff
@@ -51,6 +52,11 @@ test('can create and place new component', async ({ page }) => {
 });
 
 test('can create/delete page', async ({ page, localApp }) => {
+  invariant(
```
apedroferreira, line 55: Maybe we should have a reusable utility for this type of invariant with this type of error message too, since it seems to be pretty common.

apedroferreira (commented):
Looks good! Just added a few notes on duplicated code that we can maybe make reusable but nothing major or blocking.

PR #2800, opened 2023-10-13
Add Toolpad editor command for connecting to custom servers

Janpot:
Add `toolpad editor` command which takes a url to a custom server and starts up a Toolpad editor operating on that custom server. Added a smoke test that makes sure this setup works.

The communication between the editor and the canvas currently can be done through direct access to the content window. The communication is synchronous. This is not possible on an iframe on a different origin. The custom server hacks around this limitation by proxying the custom server through the editor server. This is likely not going to be a viable long term solution and we'll have to figure out this communication through message passing.

Follow up to drive this home:
- [ ] https://github.com/mui/mui-toolpad/pull/2878
- [ ] https://github.com/mui/mui-toolpad/pull/2879


test/models/ToolpadEditor.ts
```diff
@@ -137,23 +137,22 @@ export class ToolpadEditor {
 
     await this.componentCatalog.hover();
 
-    let pageRootBoundingBox;
-    await expect(async () => {
-      pageRootBoundingBox = await this.pageRoot.boundingBox();
-      expect(pageRootBoundingBox).toBeTruthy();
-    }).toPass();
+    await expect.poll(() => this.pageRoot.boundingBox()).toBeTruthy();
```
apedroferreira, line 140: I added a utility for this in [test/utils/waitForBoundingBox.ts](https://github.com/mui/mui-toolpad/blob/5a40939a4a8ee5fc5dbc43545d594ad1f1c172e5/test/utils/waitForBoundingBox.ts), maybe should separate it from the PR.
I can probably change it to use `.poll` too.

apedroferreira (approved):

test/models/ToolpadEditor.ts
Janpot: I'm just allergic to the null-assertion operator 😄 
reverted this for now

PR #2812, opened 2023-10-17
Fix flaky tests

apedroferreira:
- Wait for bounding boxes to be available/stable
- Wait for data to load if it's possible in some tests

Includes replacing `Container` with `Box` in some components, unrelated but it seems like `Container` is more appropriate for page-wide wrappers.


bharatkashyap (dismissed):

bharatkashyap (commented):
Seems like its failing CI again

apedroferreira:
> Seems like its failing CI again

Yep, I ran the CI one more time and it failed, will keep trying to figure this out, I still don't know what's causing the issues.

test/utils/clickCenter.ts
```diff
@@ -1,8 +1,6 @@
 import { Page, Locator, expect } from '@playwright/test';
 
 export default async function clickCenter(page: Page, targetLocator: Locator) {
-  await expect(targetLocator).toBeAttached();
-
   const targetBoundingBox = await targetLocator.boundingBox();
```
Janpot, line 4: maybe try something like
```tsx
await expect(async () => {
  const targetBoundingBox = await targetLocator.boundingBox();
  expect(targetBoundingBox).toBeTruthy();
  expect(targetBoundingBox.width).toBeGreaterThan(0)
  expect(targetBoundingBox.height).toBeGreaterThan(0)
}).toPass();
```
(writing on my phone)

apedroferreira: My latest commit seems to fix the flaky tests (partially if not fully, the only one I'm not sure about is `can create new custom components`, we can see after merging). 
I had added that line before as an attempt, but I removed it in that last commit because it wasn't doing anything, the problems were unrelated to that logic there it seems.

apedroferreira: I didn't know about that awaitable `.toPass` though, actually I'll add it too!

apedroferreira: Maybe not, they seemed to make the `data providers` test more stable before the other fixes but it could have been just a coincidence.
But these changes will also be needed for the upcoming resizing improvements, that's also why I tried getting them in anyway. And I've seen some issues with the grid height not matching the dotted highlight borders before.
If there's a reason why we're using absolute position I can revert that.

Janpot: i just wouldn't expect to find a container component there. feels more like a top level thing, given that its behavior depends on viewport width

apedroferreira: Ah, yeah `Container` in components isn't the most appropriate, I've changed all components with a `Container` to a `Box` with 100% width and reverted the DataGrid until if we actually see that we need any changes there.

bharatkashyap (approved):
👌Thanks for taking this up!

PR #2818, opened 2023-10-20
Fix and improve resizing in editor (incl. Add Spacer component)

apedroferreira:
I've tried to bring back this old PR so we can merge it instead of just closing, it's a bit large though... The Spacer component could be separated.

Improves resizing and allows for more control over spacing in the editor overall:
- Some improvements to the editor such as getting rid of single columns in a page (pages are already columns), and spreading columns in pages if a column moves to a page after a deleting/moving an element.
- **Component API changes:** Instead of a `resizableHeightProp`, make it possible to set a `defaultLayoutHeight` and `minimumLayoutHeight` for each component. This means that all components can be resized using the same system, but for now to avoid bugs I've kept the old resizing behavior: only charts and data grid are resizable, and only from the bottom border. This way no new bugs are introduced until we find solutions for some issues in the UX, but this new system is already there.
- Adds a new Spacer component that can be used just to create white space. Maybe it could be useful but we can remove it from this PR too.

**Spacer component demo:**

https://github.com/mui/mui-toolpad/assets/10789765/38141fb6-1ac2-4969-b52f-eda25f9d87d3

**Column spread demo:**

https://github.com/mui/mui-toolpad/assets/10789765/eb2638f3-0582-407d-b8e3-d87ce0a50031

**Improved resizing demo** (all components would be resizable + could be resized from top border, but this feature shown here is commented out for now as it still has some poor UX to improve in some scenarios):

https://github.com/mui/mui-toolpad/assets/10789765/629eab57-80ce-4b11-9b8e-73886cc6a05a


packages/toolpad-app/src/canvas/index.tsx
```diff
@@ -81,6 +81,15 @@ export default function AppCanvas({
     }
   });
 
+  const handleWindowResize = React.useCallback(() => {
+    handleScreenUpdate();
+  }, []);
+
+  React.useEffect(() => {
```
apedroferreira, line 88: Reisizing the screen was causing the selection boxes to stay outdated.

packages/toolpad-app/src/appDom/index.ts
```diff
@@ -816,6 +817,21 @@ export function moveNode<Parent extends AppDomNode, Child extends AppDomNode>(
   return setNodeParent(dom, node, parent.id, parentProp, parentIndex);
 }
 
+export function spreadNode<Child extends AppDomNode>(dom: AppDom, node: Child) {
```
apedroferreira, line 820: If an action in the editor causes a column of elements to be inside the page root, for example, we can spread the contents of the column on the page root directly and get rid of the column element.

packages/toolpad-app/src/server/localMode.ts
```diff
@@ -732,7 +733,14 @@ function mergePageIntoDom(dom: appDom.AppDom, pageName: string, pageFile: Page):
   return dom;
 }
 
-function optimizePageElement(element: ElementType): ElementType {
+function optimizePageElement(
+  element: ElementType,
+  isPageChild = false,
+): ElementType | ElementType[] {
+  if (isPageChild && element.component === PAGE_COLUMN_COMPONENT_ID) {
```
apedroferreira, line 740: Gets rid of page columns inside the page, I saw it in a few apps already and it's unnecessary / shouldn't ever happen.

packages/toolpad-components/src/Chart.tsx
```diff
@@ -205,7 +204,8 @@ export default createBuiltin(Chart, {
   loadingProp: 'loading',
   loadingPropSource: ['data'],
   errorProp: 'error',
-  resizableHeightProp: 'height',
+  defaultLayoutHeight: 360,
```
apedroferreira, line 207: New interface that should be simpler.
Any component is vertically resizable, but we can set a default and minimum height for specific components.

packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/NodeHud.tsx
```diff
@@ -177,7 +196,18 @@ export default function NodeHud({
   isHoverable = true,
   isHovered = false,
 }: NodeHudProps) {
-  const hintPosition = rect.y > HUD_HEIGHT ? HINT_POSITION_TOP : HINT_POSITION_BOTTOM;
+  const { nodeId: pageNodeId, viewState } = usePageEditorState();
+
+  const { nodes: nodesInfo } = viewState;
```
apedroferreira, line 201: This just moves the HUD controls to the inside of the element if they would overflow at the bottom of the window.

packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -1195,6 +1193,8 @@ function RenderedNodeContent({ node, childNodeGroups, Component }: RenderedNodeC
             display: 'flex',
             alignItems: boundLayoutProps.verticalAlign,
             justifyContent: boundLayoutProps.horizontalAlign,
+            height: node.layout?.height ?? componentConfig.defaultLayoutHeight,
```
apedroferreira, line 1196: Element heights are controlled in the runtime instead of the components themselves.

Janpot:
@apedroferreira Can you also add a demo video that demonstrates the new features?

apedroferreira:
> @apedroferreira Can you also add a demo video that demonstrates the new features?

I've added 3 demos to the description, hopefully it helps!

Janpot:
Looking good:
- I see one suspicious regression test: https://app.argos-ci.com/mui/mui-toolpad/builds/2595/78472283. Is this expected?
- I'm not a huge fan of exposing the Spacer component. Ideally we'd treat it as an internal component that gets created automatically when manipulating the UI. I guess we can start like this and remove it down the line?
- Does Spacer collaps? e.g. if you have  "component - spacer - component - spacer component" and you delete the middle component, do you end up with "component - spacer - spacer component" or with "component - spacer component"?

apedroferreira:
> * I see one suspicious regression test: https://app.argos-ci.com/mui/mui-toolpad/builds/2595/78472283. Is this expected?

Yeah, it's because the default height of the data grid changed, I probably changed it to be a multiple of the resizing intervals.

> * I'm not a huge fan of exposing the Spacer component. Ideally we'd treat it as an internal component that gets created automatically when manipulating the UI. But I guess we can start like this and remove it down the line?

Yeah I just didn't want to delete it right away from this PR but we had discussed not having it and it just being created automatically... As we're not sure about it I don't mind removing it and seeing what we come up with later if we come back to this, I guess it's better than introducing something we might not plan to keep. I'll remove it from the PR if you don't disagree.

> * Do Spacer components collapse? e.g. if you have  "component - spacer - component - spacer - component" and you delete the middle component, do you end up with "component - spacer - spacer - component" or with "component - spacer - component"?

Nope, not in this version of them. Again I guess that I can remove it for now anyway.

Janpot (approved):
impressive

PR #2845, opened 2023-10-31
Add preview header to custom server

Janpot:
As discussed in grooming session. The current header is problematic for several reasons:

- it's confusing for users, some think it belongs to the Toolpad application. Then are surprised it's not there in production.
- it's pulling in the full theme of the editor into the application runtime.
- it has a theme switcher that controls the mode of the editor, not the app.
- it's not really usable in the custom server mode as there is no editor running. It's very important to have an undismissable warning in this mode because exposing dev mode publicly is a security issue
- We likelywant to add a header to Toolpad applications at some point in the near future, to accomodate authentication, this will clash visually with our current preview header

To keep things simple, this PR replaces the header with a MUI [Alert](https://mui.com/material-ui/react-alert/). The benefits are:

- it's much clearer that this is not part of the application
- it follows the application theme instead of the editor theme
- we can show it in custom server mode (for now I just remove the edit button, in the future I will add instructions on how to run the editor for a custom server)

Screenshot:

Screenshot 2023-10-31 at 11 59 55

It's the vanilla Alert unmodified, we can refine the look further in follow-up PRs.

cc @mui/core Not sure whether this often comes up with our users, but the Drawer and AppBar are built using fixed positioning. This makes a lot of sense for document-centered webpages, where it feels very natural to have these navigation surfaces floating over the content. In my experience, for more application-centered webpages we tend use more an approach where we start from the browser viewport and progressively split the screen horizontally and vertically to create application surfaces that can scroll independently. It would be very useful to have a mode on surfaces like Drawer and AppBar which allow them to render relative to a predefined container, rather than the browser viewport.


packages/toolpad-app/src/runtime/PreviewHeader.tsx
```diff
@@ -1,35 +1,48 @@
...
               Edit
             </Button>
-          </Stack>
+          )
         }
-      />
-    </ThemeProvider>
+      >
+        <Typography variant="body2">
+          This is a preview version of the application. It&apos;s not suitable for production.
```
bharatkashyap, line 43:
```suggestion
          This is a preview version of the application, not suitable for production. Click "Edit" to return to the editor.
```
```suggestion
          This is a preview version of the application, not suitable for production. Click "Edit" to go to the editor.
```

packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -91,13 +91,13 @@ import api, { queryClient } from './api';
 
 const browserJsRuntime = getBrowserRuntime();
 
-const isPreview = process.env.NODE_ENV !== 'production';
-const isCustomServer = process.env.TOOLPAD_CUSTOM_SERVER === 'true';
-const isRenderedInCanvas =
+const IS_RENDERED_IN_CANVAS =
```
bharatkashyap, line 94: We've typically been using all caps for values exposed from `constants`
```suggestion
const isRenderedInCanvas =
```

packages/toolpad-app/src/runtime/ToolpadApp.tsx
```diff
@@ -91,13 +91,13 @@ import api, { queryClient } from './api';
...
   typeof window === 'undefined'
     ? false
     : !!(window.frameElement as HTMLIFrameElement)?.dataset?.toolpadCanvas;
 
+const SHOW_PREVIEW_HEADER = IS_PREVIEW && !IS_RENDERED_IN_CANVAS;
```
bharatkashyap, line 99: as above
```suggestion
const showPreviewHeader = IS_PREVIEW && !isRenderedInCanvas;
```

packages/toolpad-app/src/runtime/evalJsBindings.ts
```diff
@@ -1,7 +1,15 @@
 import { BindingEvaluationResult, JsRuntime } from '@mui/toolpad-core';
-import { set as setObjectPath } from 'lodash-es';
 import { mapValues } from '@mui/toolpad-utils/collections';
-import { updatePath } from '../utils/immutability';
+// TODO: remove these lodash-es imports
```
bharatkashyap, line 3: Why move this function out of `immutability.ts`?

bharatkashyap (approved):

packages/toolpad-app/src/runtime/evalJsBindings.ts
Janpot: It introduces type unsafety, we shouldn't promote using this style of programming

packages/toolpad-app/src/runtime/ToolpadApp.tsx
Janpot: We typically use all caps for static values that are know at compile time. It tells the call site that this variable abstracts a constant value, and could be inlined without changing semantics if we wanted to.

packages/toolpad-app/src/runtime/PreviewHeader.tsx
Janpot: I'll change the button text to be more actionable instead:

Screenshot 2023-11-02 at 09 18 10

PR #2851, opened 2023-11-01
Add supabase example showcasing list component

prakhargupta1:
App screenshot: 
Screenshot 2023-11-01 at 2 14 01 PM


docs/src/modules/components/ExamplesGrid.js
```diff
@@ -84,6 +84,13 @@ function layouts() {
       href: 'https://github.com/mui/mui-toolpad/tree/master/examples/with-prisma-data-provider',
       source: 'https://github.com/mui/mui-toolpad/tree/master/examples/with-prisma-data-provider',
     },
+    {
+      title: 'With Supabase',
+      description: 'A Toolpad app that fetches data from Supabase and shows in a list component.',
```
apedroferreira, line 89:
```suggestion
      description: 'A Toolpad app that fetches data from Supabase and shows it in a list component.',
```

apedroferreira (approved):
Looks good!
Maybe it will look better if the font for the description is larger than for the properties below, not sure.

PR #2632, opened 2023-09-05
[docs] Add 'Why Toolpad?'

prakhargupta1:
Updated the structure around developer problems and removed extra info.

Preview link: https://deploy-preview-2632--mui-toolpad-docs.netlify.app/toolpad/getting-started/why-toolpad/

PR #2854, opened 2023-11-01
[docs] Why Toolpad? iteration

oliviertassinari:
My take on #2632, what I see as pain.

In https://mui.com/toolpad/getting-started/why-toolpad/#trade-offs-with-pro-code-tools, it feels like the 3 pains with pro-code are about the same thing: low vs. high-level abstraction.

If it was only about the abstraction level, we could argue that a solution without drag and drop would be better, and low-code users would likely only migrate later, only when the tool gets more mature.

Now, with the pain of not having as direct manipulation of the data as it could be possible, having a drag-and-drop UI might give us a unique chance to do AI. And maybe also a way to monetize (taking a cut on top of Open AI)

I have updated the pains with the ones I think better match our solution.

Preview: https://deploy-preview-2854--mui-toolpad-docs.netlify.app/toolpad/getting-started/why-toolpad/

---

As a side note, I used to only write SQL in Metabase, I moved to use their low-code query editor out of laziness 😅, only using SQL when the query is too complex.

### DIFF
--- a/docs/data/toolpad/getting-started/why-toolpad.md
+++ b/docs/data/toolpad/getting-started/why-toolpad.md
@@ -7,41 +7,65 @@
 - **Higher-level API**. Toolpad operates at a higher abstraction level than Material UI. It sets healthy UI constraints that prevent common pitfalls and anti-patterns.
 - **Drag-and-drop builder**. Toolpad comes with a drag-and-drop builder to simplify UI building. The builder saves the UI as a YAML file, so you can even modify and version the visual output on your file system.
 - **Less boilerplate** so you can focus on the essential parts of the app.
-- **Run alongside existing code**. It integrates well with your existing code. You can use your database models, client libraries, secrets, and bespoke components directly.
+- **Run alongside existing code**. Toolpad integrates well with your existing code. You can use your database models, client libraries, secrets, and bespoke components directly.
 - **Single source of truth**. Because Toolpad runs alongside your existing code, it lets you maintain a single source of truth for all business logic.
 - **Integrates well with your development lifecycle**. Your project lives on your file system. You can use your own IDE, version control system, deployment target, CLI tools.
 
-After reviewing multiple products and approaches, a few recurring areas of concern were noticed. Let's take a closer look at them:
-
 ## Trade-offs with pro-code tools
 
-### Compromised delivery speed
+The key problems most pro-code tools have relative to Toolpad:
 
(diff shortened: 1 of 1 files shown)


docs/data/toolpad/getting-started/why-toolpad.md
```diff
@@ -7,41 +7,65 @@
...
-### A lot of boilerplate
+Every organization wants to achieve fast delivery but in order to accommodate various use cases, you are pushed to choose flexible pro-code solutions.
 
+This flexibility comes at the cost of more **boilerplate** code.
 To connect backend data to the browser and set up data-display components, a substantial amount of boilerplate HTML, CSS, and JavaScript glue code is often required.
+As the use cases repeat themselves, you create abstractions but their maintenance burden quickly grows.
+
+Wouldn't it be great if you could use a tool that abstracts the repetitive parts of creating internal tools?
+
+### No direct manipulation of objects
```
Janpot, line 28: Perhaps "interactivity" as terminology would be more familiar?

docs/data/toolpad/getting-started/why-toolpad.md
```diff
@@ -7,41 +7,65 @@
...
+
+By the definition of a pro-code tool, you write code in an IDE and the UI renders in the browser, a different window.
+
+You can't **directly manipulate** the parameters, queries, etc. in the browser. It makes it harder to make changes on the application.
+
+Wouldn't it be great if it was possible to directly manipulate the app created in the browser, manipulating the data the closest possible to the source of truth?
 
-### Lower-level abstraction
+### Require specialized knowledge
```
Janpot, line 36: Essentially shortening the loop between writing code and consulting documentation.

docs/data/toolpad/getting-started/why-toolpad.md
```diff
@@ -7,41 +7,65 @@
...
+The key problems most low-code tools have relative to Toolpad:
 
 ### Extensibility
 
-Low-code tools can be hard to extend. When the tool can't handle a particular task by default, integrating your own code seamlessly is often a challenge. This lack of flexibility often results in having to write custom code inside these apps and storing your business logic within their configuration files.
+Low-code tools can be hard to extend. When the tool can't handle a particular task by default, integrating your own code seamlessly is often a challenge.
+
+This lack of flexibility often results in being stuck or having to write custom code inside these apps and storing the business logic within their configuration files.
+
+Would it be great if the tool would use code in your filesystem as the source of truth?
```
Janpot, line 54: To me the "local first" argument is about ownership. Existing hosted low-code tools store their configurations opaquely on their end. They end up reinventing version control tools and deployment lifecycle tools, which don't integrate well with your existing stack.

Janpot (approved):
Ok, let's move this forward and see if there is a response.
We can tweak the wording further later on.

PR #2878, opened 2023-11-07
Add instructions to start the editor for custom servers

Janpot:
Complements https://github.com/mui/mui-toolpad/pull/2800 with instructions in a popover on the application header, to start the standalone editor.  

Screenshot 2023-11-07 at 17 22 56

### DIFF
--- a/packages/toolpad-app/src/runtime/PreviewHeader.tsx
+++ b/packages/toolpad-app/src/runtime/PreviewHeader.tsx
@@ -149,11 +149,14 @@ function CustomServerInstructions({ basename }: CustomServerInstructionsProps) {
           vertical: 'top',
           horizontal: 'right',
         }}
+        sx={{
+          maxWidth: 1000,
+        }}
       >
         <Box sx={{ p: 2 }}>
           <Typography>
             This application is running under a custom server. Run the standalone Toolpad editor to
-            edit this application.
+            make changes to this application.
           </Typography>
           <CodeView>{`npx @mui/toolpad editor ${appUrl}`}</CodeView>
         </Box>
--- a/packages/toolpad-app/src/runtime/PreviewHeader.tsx
+++ b/packages/toolpad-app/src/runtime/PreviewHeader.tsx
@@ -1,10 +1,172 @@
 import * as React from 'react';
-import { Button, Typography, Box, useTheme, Alert } from '@mui/material';
+import {
+  Button,
+  Typography,
+  Box,
+  useTheme,
+  Alert,
+  ButtonProps,
+  Popover,
+  styled,
+  IconButton,
+  Tooltip,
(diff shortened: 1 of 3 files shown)


packages/toolpad-app/src/runtime/PreviewHeader.tsx
```diff
@@ -1,10 +1,178 @@
...
 import { useMatch } from 'react-router-dom';
+import ContentCopyIcon from '@mui/icons-material/ContentCopy';
 import { IS_CUSTOM_SERVER, PREVIEW_HEADER_HEIGHT } from './constants';
 
-export default function PreviewHeader() {
+interface CopyToClipboardButtonProps extends IconButtonProps {
+  content: string;
+}
+
+function CopyToClipboardButton({ content, onClick, ...props }: CopyToClipboardButtonProps) {
```
apedroferreira, line 26: Looks like this could be a useful shared component besides being just in this file?

apedroferreira (approved):
Looks great!

I just noticed that in the screenshot the copy icon seems a bit off-center vertically.

packages/toolpad-app/src/runtime/PreviewHeader.tsx
Janpot: I don't want to incentivize maintainers importing from outside the runtime folder. That's why I opted for this setup. Maybe in the future.

Janpot:
> I just noticed that in the screenshot the copy icon seems a bit off-center vertically.

🤔  It's not meant to be centered, it's meant to be in the corner

packages/toolpad-app/src/runtime/PreviewHeader.tsx
apedroferreira: Ah got it because it's in the runtime, ok we can keep it.

apedroferreira:
> > I just noticed that in the screenshot the copy icon seems a bit off-center vertically.
> 
> 🤔 It's not meant to be centered, it's meant to be in the corner

Ok, I personally think it would look better centered but if it's intentional feel free to keep it!

Janpot:
> Ok, I personally think it would look better centered but if it's intentional feel free to keep it!

My main hesitation against that would be, what would it look like then for a code viewer that has multiple lines?

apedroferreira:
> > Ok, I personally think it would look better centered but if it's intentional feel free to keep it!
> 
> My main hesitation against that would be, what would it look like then for a code viewer that has multiple lines?

Got it, yeah even in the MUI docs it's always in the corner https://mui.com/material-ui/react-autocomplete/, I didn't think of that scenario. For a single line command maybe centering still looks better aesthetically, but anyway not a big deal, I think we can have it in the corner for consistency.

PR #2879, opened 2023-11-07
Add custom server documentation

Janpot:
Add some docs for the new custom server feature

Complements https://github.com/mui/mui-toolpad/pull/2800

Concept page: https://deploy-preview-2879--mui-toolpad-docs.netlify.app/toolpad/concepts/custom-server/
API reference page: https://deploy-preview-2879--mui-toolpad-docs.netlify.app/toolpad/reference/api/create-handler/


bharatkashyap:
I feel like I'd greatly benefit from a video or an example of any other form demonstrating a custom server while reading this documentation page: is it possible to show one working and embed a video of it?

Janpot:
Thanks, there already is an example, I'm adding a link on the docs page to make it more discoverable:

Screenshot 2023-11-09 at 10 55 58

Janpot:
merging for the release later today. Feel free to further review

docs/data/toolpad/concepts/custom-server.md
```diff
@@ -0,0 +1,56 @@
...
+  }
+}
+```
+
+Now you can use the corresponding commands to interact with Toolpad in the custom server
+
+| command      | description                                                                                                                                                                                           |
+| :----------- | :---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `yarn dev`   | Run the custom server in dev mode. Similar to opening the application preview when running `toolpad dev`                                                                                              |
+| `yarn start` | Running the Toolpad application in production mode. The application must still be built first with `toolpad build`. Make sure to supply the correct base path.                                        |
```
apedroferreira, line 50: These are not inline in my browser, `yarn start`, `yarn build` and `yarn eject` are taking up 2 lines so they look a bit off.

apedroferreira (commented):
Everything looks good, it's simple and easy to understand!
One thing that not sure can be improved is that the base path (`/my-app` in the examples) needs to be written separately/independently in 4 or 5 different places, maybe there's a way to make it simpler/automatic.

apedroferreira:
Just tried the example, some more feedback:
- `yarn dev` didn't tell me that the app is running on `localhost:3001`, I kind of expected that as confirmation that it was running.
- It wasn't immediately obvious to me that I needed to run `yarn dev` before `yarn edit`, and not sure if `yarn edit` should/could provide a more informative error in those cases.

Janpot:
> One thing that not sure can be improved is that the base path (/my-app in the examples) needs to be written separately/independently in 4 or 5 different places, maybe there's a way to make it simpler/automatic.

In this specific example it's written in 2 places:
- once in `createHandler`.
- once when attaching the handler.
They are not guaranteed to always be the same. You can always assign them to a variable. I choose not to because I want the options to `createHandler` to illustrate clearly what is expected as a value.

apedroferreira:
> > One thing that not sure can be improved is that the base path (/my-app in the examples) needs to be written separately/independently in 4 or 5 different places, maybe there's a way to make it simpler/automatic.
> 
> In this specific example it's written in 2 places:
> 
> * once in `createHandler`.
> * once when attaching the handler.
>   They are not guaranteed to always be the same. You can always assign them to a variable. I choose not to because I want the options to `createHandler` to illustrate clearly what is expected as a value.

Yeah makes sense, just mentioning it, I was also referring to the `package.json` file, it's there a few times too. But anyway not very important, just in case it was easy to improve somehow.

PR #2891, opened 2023-11-10
[experiment] Support custom code files for pages

Janpot:
Circumvent the drag&drop editor and write pages directly as code. The content of the page comes from a React component as the default export of the referenced page in `page.yml`.

No interaction with Toolpad backend has been implemented yet, this is just a starting point for page ejection.

https://github.com/mui/mui-toolpad/assets/2109932/5d765ce6-0b6b-4420-b6ed-98c5f63b73ea

The feature is marked as `unstable_codeFile` to warn users it's not intended for production usage


apedroferreira (approved):
Nice idea for flexibility! It could be a starting point as I guess that ejecting would change a page from being all in YML to using this new feature.

PR #2894, opened 2023-11-10
Update instructions for smoke testing the release

Janpot:
Makes more sense to do it before publishing


apedroferreira (approved):
Good idea!

PR #2920, opened 2023-11-20
Use ESM for CLI

Janpot:
- Fixing deprecation warnings in vite 5

    ```
    The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.
    ```


Janpot:
Putting back into draft, this seems to increase the integration test run time consistently by a minute or so

PR #2925, opened 2023-11-21
Add authentication

apedroferreira:
Make it possible to log in using Github or Google, and set a Toolpad application as protected for logged-in users only.
Includes slight UI changes to the navigation sidebar with added logo and title.
Adding this under a feature flag for now - we can handle additional features in separate PRs.

### Demo video

https://github.com/mui/mui-toolpad/assets/10789765/c46ecfb1-958e-4d5b-a3e5-d585934b3f04

### Todo in further PRs

- [x] Handle roles in the user + roles mapping for authorization (I've disabled the checks for now)
- [ ] Protect runtime RPC
- [x] Hide pages in sidebar menu if user does not have the roles to access them
- [ ] Set the `NEXT_AUTH_URL` environment variable on behalf of the user in production apps, according to https://next-auth.js.org/configuration/options (custom base path). This helps some cookies be set as secure in HTTPS URLs, for example. Alternatively, ignore this and set the `useSecureCookies` option, or in last case scenario tell the user to set up the environment variable.
- [ ] Add documentation + update links to these doc pages in authorization UI
- [ ] Add tests
- [ ] Add an example
- [x] Fix `application.yml` not being generated in non-contributor project if issue is confirmed and persists
- [ ] Possible UX improvements for setting environment variables (probably optional for first version, not as high-priority)

### How to try this before we add documentation

The following environment variables can be set:

- `TOOLPAD_AUTH_SECRET`: Required random string used to hash tokens, sign/encrypt cookies and generate cryptographic keys (https://next-auth.js.org/configuration/options#secret). 
A good value can be generated in the CLI with: `$ openssl rand -base64 32`

**Github authentication:**
- `TOOLPAD_GITHUB_ID`
- `TOOLPAD_GITHUB_SECRET`

OAuth client setup instructions: https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/creating-an-oauth-app
The callback URL should be something like `http://www.example.com/prod/api/auth/callback/github`

**Google authentication:**
- `TOOLPAD_GOOGLE_CLIENT_ID`
- `TOOLPAD_GOOGLE_CLIENT_SECRET`
- `TOOLPAD_GOOGLE_AUTH_DOMAIN` (optional, sets an email domain so that only users in that domain can log in) (e.g.: `example.com`)

OAuth client setup instructions: https://developers.google.com/workspace/guides/create-credentials#oauth-client-id
The needed authorized redirect URL should be something like `http://www.example.com/prod/api/auth/callback/google`


apedroferreira:
Most of the work is done, but documenting the remaining work as I'll be out next week:
- [ ] Display mode options to hide/show header bar independently of navigation sidebar
- [ ] Make it possible to still login/logout even if header bar is set to be hidden
- [ ] Effectively protect pages that are set as protected to be accessible by logged-in users only
- [ ] Add documentation, including display mode changes, Google auth setup, Github auth setup and options to restrict to a specific email domain
- [ ] Test auth in a deployed application with custom domain (need to confirm that some options that were added to run the app locally are secure, and if everything works correctly)

Janpot:
We can start by making all pages protected as a default when authentication is configured. We can think about access control as part of the authorization. Toolpad is not intended for public applications anyway.

apedroferreira:
It looks like in production application the `application.yaml` file is not generated, haven't looked deeper into it yet...
Almost ready for review, just need to make the tests pass.

packages/toolpad-app/src/runtime/AppLayout.tsx
```diff
@@ -25,20 +38,29 @@ export interface NavigationEntry {
...
+  activePageSlug,
   pages,
   clipped = false,
   search,
 }: AppPagesNavigationProps) {
   const navListSubheaderId = React.useId();
 
+  const theme = useTheme();
+
+  const productIcon = new URL(
```
apedroferreira, line 57: Does anyone know why I can't import images in the runtime like the editor does in its header to show the logo?
This solution works in dev but in a production app it doesn't. Importing the files directly doesn't work in dev for the runtime.

Janpot: Yes,

* When runtime gets build, the svg gets copied to the dist folder of the runtime, the import gets rewritten to include the content hash.
* When vite imports the runtime, it doesn't seem to do its [asset handling](https://vitejs.dev/guide/assets) on this file.

Let's not block this PR on the problem. We can move this PR forward without the Toolpad icon

Janpot: Looks like the same problem as https://github.com/vitejs/vite/issues/11943

Janpot: Actually, looked a bit deeper into it. It looks like esbuild translates the asset into a path and copies the file alongside the compiled output. Instead I externalized imports that look like assets so that esbuild preserves the import. Now it seems that vite picks up this import and translates it correctly to an asset.

apedroferreira: Awesome, thanks a lot, that was a though one!
Ok, if all checks are green which they should be now I will mark this PR for review.

docs/schemas/v1/definitions.json
```diff
@@ -20,6 +20,17 @@
...
+                "providers": {
+                  "type": "array",
+                  "items": {
+                    "type": "string",
+                    "enum": [
+                      "github",
+                      "google"
+                    ]
+                  },
+                  "description": "The name of the authentication provider to use."
```
apedroferreira, line 32:
```suggestion
                  "description": "Authentication providers to use."
```

packages/toolpad-app/src/server/schema.ts
```diff
@@ -261,6 +261,10 @@ export const applicationSchema = toolpadObjectSchema(
   z.object({
     authorization: z
       .object({
+        providers: z
+          .array(z.enum(['github', 'google']))
+          .optional()
+          .describe('The name of the authentication provider to use.'),
```
apedroferreira, line 267:
```suggestion
          .describe('Authentication providers to use.'),
```

packages/toolpad-app/src/server/schema.ts
```diff
@@ -261,6 +261,10 @@ export const applicationSchema = toolpadObjectSchema(
   z.object({
     authorization: z
```
Janpot, line 262: would it make sense to split `authorization` and `authentication` at this level?

docs/schemas/v1/definitions.json
```diff
@@ -20,6 +20,17 @@
             "authorization": {
               "type": "object",
               "properties": {
+                "providers": {
+                  "type": "array",
+                  "items": {
+                    "type": "string",
```
Janpot, line 26: This should probably be an object, so that we can add configuration to it. I'd also make `authentication` separate from `authorization`

packages/toolpad-app/src/runtime/auth.tsx
```diff
@@ -1,51 +1,65 @@
...
     () => new Set(asArray(allowedRole ?? [])),
     [allowedRole],
   );
 
   let reason = null;
   if (!user) {
-    reason = 'User is not authenticated.';
-  } else if (!user.roles || user.roles.length <= 0) {
+    if (!isSigningIn) {
+      window.location.replace(`${basename}${AUTH_SIGNIN_PATH}`);
```
Janpot, line 28: No do side-effects in the render method

packages/toolpad-app/src/runtime/auth.tsx
```diff
@@ -1,51 +1,65 @@
...
+
+  if (!user.roles || user.roles.length <= 0) {
     reason = 'User has no roles defined.';
   } else if (!user.roles.some((role) => allowedRolesSet.has(role))) {
     const rolesList = user?.roles?.map((role) => JSON.stringify(role)).join(', ');
     reason = `User with role(s) ${rolesList} is not allowed access to this resource.`;
   }
 
-  return reason ? (
+  // @TODO: Once we have roles we can add back this check.
```
Janpot, line 53: What if I only want authentication, no authorization?

packages/toolpad-app/src/server/EnvManager.ts
```diff
@@ -101,7 +101,7 @@ export default class EnvManager {
   }
 
   // eslint-disable-next-line class-methods-use-this
-  getEnv() {
+  async getEnv() {
```
Janpot, line 104:
```suggestion
  getEnv() {
```

packages/toolpad-app/src/server/FunctionsManager.ts
```diff
@@ -254,7 +254,7 @@ export default class FunctionsManager {
 
   private async createRuntimeWorker() {
     const oldWorker = this.devWorker;
-    this.devWorker = createDevWorker(this.project.envManager.getEnv());
+    this.devWorker = createDevWorker(await this.project.envManager.getEnv());
```
Janpot, line 257:
```suggestion
    this.devWorker = createDevWorker(this.project.envManager.getEnv());
```

packages/toolpad-app/src/runtime/auth.tsx
apedroferreira: I'll make sure it will still work, for that the idea is that you just need to set up an auth provider.

packages/toolpad-app/src/server/schema.ts
```diff
@@ -259,12 +259,23 @@ elementSchema = baseElementSchema
...
   'application',
   z.object({
-    authorization: z
+    authentication: z
       .object({
         providers: z
-          .array(z.enum(['github', 'google']))
+          .array(
+            z.object({
+              id: z
```
Janpot, line 267: I'd call this property "provider".

Janpot (approved):
I'd rename "id" to "provider". nothing else blocking

apedroferreira:
> I'd rename "id" to "provider". nothing else blocking

Ok, I've added that change + blocking unauthenthicated users from pages at the HTTP level, it's all at https://github.com/mui/mui-toolpad/pull/2925/commits/414ba2b1ad27bad8aff9255548494373ec49778b.

packages/toolpad-app/src/runtime/SignInPage.tsx
```diff
@@ -16,14 +16,14 @@ export default function SignInPage() {
   const { signIn, isSigningIn } = React.useContext(AuthContext);
 
   const [errorSnackbarMessage, setErrorSnackbarMessage] = React.useState<string>('');
-  const [latestSelectedProvider, setLatestSelectedProvider] = React.useState<AuthProvider | null>(
-    null,
-  );
+  const [latestSelectedProvider, setLatestSelectedProvider] = React.useState<
+    AuthProvider['provider'] | null
```
Janpot, line 20:
```
type AuthPrivider = 'Google' | 'GitHub' // just use the same names as in auth.js

interface AuthProviderConfig {
  provider: AuthProvider
}
```

apedroferreira: Yeah that could be better, should be good now: https://github.com/mui/mui-toolpad/pull/2925/commits/9bbfeec826c153c0976d1af949dea345651de940

Janpot (approved):

PR #2945, opened 2023-11-29
mui#2923 Autocomplete will occur an error when set same value

JerryWu1234:

### DIFF
--- a/packages/toolpad-app/src/toolpad/propertyControls/SelectOptions.tsx
+++ b/packages/toolpad-app/src/toolpad/propertyControls/SelectOptions.tsx
@@ -203,6 +203,7 @@ function SelectOptionsPropEditor({
               ) : null}
               <TextField
                 fullWidth
+                error
                 sx={{ my: 1 }}
                 variant="outlined"
                 inputRef={optionInputRef}


JerryWu1234:
image
image

JerryWu1234:
@apedroferreira help me review this ?

packages/toolpad-app/src/toolpad/propertyControls/SelectOptions.tsx
```diff
@@ -47,17 +48,38 @@ function SelectOptionsPropEditor({
     return null;
   }, [editingIndex, value]);
 
+  const watchInputValue = React.useCallback(
+    (event: React.KeyboardEvent<HTMLInputElement>) => {
+      const inputText = (event.target as HTMLInputElement).value;
+      if (
+        value.some((item) =>
+          (item as SelectOption)?.value
```
apedroferreira, line 56: Can you check if `typeof item === 'string'` instead of using `as`?

packages/toolpad-app/src/toolpad/propertyControls/SelectOptions.tsx
```diff
@@ -29,6 +29,7 @@ function SelectOptionsPropEditor({
   value = [],
   onChange,
 }: EditorProps<(string | SelectOption)[]>) {
+  const [addOptionState, setAddOptionState] = React.useState(false);
```
apedroferreira, line 32: What if we use this state for any possible error message, so we could call this `[optionErrorMessage, setOptionErrorMessage]`, and it would be a `string` with an error message instead of a boolean? 
That way we can use it for more things later on, if we need to.

packages/toolpad-app/src/toolpad/propertyControls/SelectOptions.tsx
```diff
@@ -47,17 +48,38 @@ function SelectOptionsPropEditor({
     return null;
   }, [editingIndex, value]);
 
+  const watchInputValue = React.useCallback(
```
apedroferreira, line 51: I would call this function `validateOptionValue` or something like that, and if it found any duplicates it would set the error message to "Must not have duplicate values", otherwise an empty string ('').

packages/toolpad-app/src/toolpad/propertyControls/SelectOptions.tsx
```diff
@@ -81,11 +103,24 @@ function SelectOptionsPropEditor({
 
   const handleOptionChange = React.useCallback(
     (newOption: string | SelectOption) => {
+      const errorState = value.some((item) =>
```
apedroferreira, line 106: We should be able to reuse the same function here as in `watchInputValue` above, can you please make a reusable function that is used in both places?

Also, repeated logic inside this function such as `(newOption as SelectOption)?.value ?? newOption` could be assigned to a variable ahead of time so you don't have to write it twice.

apedroferreira (changes requested):
Good job validating on the 2 different inputs! I left a few suggestions but this is a good start.

There are also a few more cases where it seems that you need to clear the error message or it does not reset, can you please cover all of these? Some examples, not sure if there's any more cases than these:
- User closes the modal, by either pressing "Close" or clicking outside
- User clicks "Delete all" to delete all options
- User presses the back button from an option

packages/toolpad-app/src/toolpad/propertyControls/SelectOptions.tsx
JerryWu1234: What message should be set for this?

`Do not input <kbd>same</kbd> value`,
`Do not input <kbd>same</kbd> value`
There are two errors message, with some element marks in the message

JerryWu1234: If you just want to clarify what it means, I think the variable name is also a key point,

apedroferreira: `optionErrorMessage` would have to be a `string` instead of a `boolean` so that we can set any error message.
"Do not enter duplicate value" could work as an error message!

apedroferreira: If we need to include tags such as `<kbd>` you can use `React.ReactNode` as the type for the error message, instead of a `string`.

JerryWu1234:
@apedroferreira help to review the rest

packages/toolpad-app/src/toolpad/propertyControls/SelectOptions.tsx
```diff
@@ -202,15 +245,23 @@ function SelectOptionsPropEditor({
...
                 inputRef={optionInputRef}
                 onKeyUp={handleOptionTextInput}
                 label={'Add option'}
                 helperText={
-                  <span>
-                    Press <kbd>Enter</kbd> or <kbd>Return</kbd> to add
-                  </span>
+                  optionErrorMessage ? (
+                    <span>
+                      Do not input <kbd>same</kbd> value
```
bharatkashyap, line 258:
```suggestion
                      This option already exists
```

packages/toolpad-app/src/toolpad/propertyControls/SelectOptions.tsx
```diff
@@ -129,6 +164,14 @@ function SelectOptionsPropEditor({
               <Stack gap={1} py={1}>
                 <TextField
                   label="Value"
+                  error={!!optionErrorMessage}
+                  helperText={
+                    optionErrorMessage ? (
+                      <span>
+                        Do not input <kbd>same</kbd> value
```
bharatkashyap, line 171:
```suggestion
                       This option already exists
```

packages/toolpad-app/src/toolpad/propertyControls/SelectOptions.tsx
```diff
@@ -47,17 +48,43 @@ function SelectOptionsPropEditor({
     return null;
   }, [editingIndex, value]);
 
+  const switchErrorState = React.useCallback(
```
bharatkashyap, line 51: This is a little too complicated: why create a callback with another callback as its parameter?

bharatkashyap (changes requested):
Some notes around the structure of the changes

packages/toolpad-app/src/toolpad/propertyControls/SelectOptions.tsx
JerryWu1234: > This is a little too complicated: why create a callback with another callback as its parameter?

frist all
`typeof item !== 'string' ? item.value === newOptionValue : item === newOptionValue,`
` typeof item !== 'string' ? item.value === inputText : item === inputText,`
this condition is only different in both. and the callback as parameter is easy understand

JerryWu1234:
@apedroferreira @bharatkashyap help me to review this PR

bharatkashyap (changes requested):
```tsx
 switchErrorState((item) =>
        typeof item !== 'string' ? item.value === newOptionValue : item === value,
      );
```

There is some replication of the above code. 

The idea with this feature is that you accept an input value, compare it with the array of existing options, and flip the state of the error in case a duplicate is found. You can refactor this into a single, memoized function that does all of this, with `inputText` being the only required parameter. 

In my opinion,  separate `validateOptionValue` and `switchErrorState` functions are not needed.

JerryWu1234:
> ```tsx
>  switchErrorState((item) =>
>         typeof item !== 'string' ? item.value === newOptionValue : item === value,
>       );
> ```
> 
> There is some replication of the above code.
> 
> The idea with this feature is that you accept an input value, compare it with the array of existing options, and flip the state of the error in case a duplicate is found. You can refactor this into a single, memoized function that does all of this, with `inputText` being the only required parameter.
> 
> In my opinion, separate `validateOptionValue` and `switchErrorState` functions are not needed.

all done

packages/toolpad-studio/src/toolpad/propertyControls/SelectOptions.tsx
```diff
@@ -47,17 +48,35 @@ function SelectOptionsPropEditor({
...
+      } else {
+        setOptionErrorMessage('');
+      }
+    },
+    [value],
+  );
+
   const handleOptionTextInput = React.useCallback(
     (event: React.KeyboardEvent<HTMLInputElement>) => {
+      validateOptionValue((event.target as HTMLInputElement).value);
```
bharatkashyap, line 67: Does it make sense to only validate the option when the user attempts to add, versus running the validation on every keystroke?

packages/toolpad-studio/src/toolpad/propertyControls/SelectOptions.tsx
```diff
@@ -47,17 +48,35 @@ function SelectOptionsPropEditor({
     return null;
   }, [editingIndex, value]);
 
+  const validateOptionValue = React.useCallback(
+    (inputValue: string) => {
+      const errorState = value.some((item) =>
+        typeof item !== 'string' ? item.value === inputValue : item === inputValue,
+      );
+      if (errorState) {
```
bharatkashyap, line 56:
```suggestion
setOptionErrorMessage(errorState ? "This option already exists: : "")
```

Issue #2887, opened 2023-11-10
Support server-side row editing


prakhargupta1:
Add CRUD operations to the Toolpad [data providers](https://mui.com/toolpad/concepts/data-providers/).

The data provider interface can be augmented with `updateRecord(id, fields)` methods. When available the frontend component should automatically provide edit buttons in the data grid which turns the row editable and calls said method in the backend.

PR #2978, opened 2023-12-06
Support delete in data providers

Janpot:
Closes https://github.com/mui/mui-toolpad/issues/2889

Add optional `deleteRecord` method to the dataprovider interface. If provided, the grid will have a delete button on every row.

https://github.com/mui/mui-toolpad/assets/2109932/205de6ce-e7b6-496c-b3b9-172c2202d9b1

Error and loading state:

https://github.com/mui/mui-toolpad/assets/2109932/b39ec304-e053-4560-87b3-dc57d7f77fc5

Updated example: https://github.com/mui/mui-toolpad/blob/04fdbf9fdcbc267d5d1d372d008538f92e17cd8b/examples/with-prisma-data-provider/toolpad/resources/crud.ts#L24

Documentation: https://deploy-preview-2978--mui-toolpad-docs.netlify.app/toolpad/concepts/data-providers/#deleting-rows

To Do:

- [x] documentation
- [x] visual loading state
- [x] error state
- [x] test

PR #3001, opened 2023-12-15
Support updating rows in the data provider

Janpot:
Closes https://github.com/mui/mui-toolpad/issues/2887

Adds support for an `updateRecords` property in the data provider. When this property exists, theconnected data grid will have a row edit button. After clicking save, the backend `updateRecords` method will be called with the updated values.

Loading state is reflected in the save icon. error state is reflected in the same way as we do for row deletion: https://github.com/mui/mui-toolpad/pull/2978

https://github.com/mui/mui-toolpad/assets/2109932/e84ab669-d6e8-46e9-88af-581e51951fc5

docs: https://deploy-preview-3001--mui-toolpad-docs.netlify.app/toolpad/concepts/data-providers/#row-editing

### To Do:

- [x] Add test


Janpot:
> Is the case when the ID field isn't id sufficient enough to handle before this lands? To me, it's okay for this to go and us to error gracefully

I'm correcting a bug in this PR where if the first row has `id: 0` it would mistakenly flag it as having no explicit id. This bug must be fixed for the test to run.

PR #3002, opened 2023-12-15
Add users with roles and finish role-based page authorization

apedroferreira:
Still under a feature flag, as runtime RPC protection is needed next and can be done in a separate PR.

**List of features:**
- Assign roles to users individually by email (I considered several options and this seemed like the best one, eventually we could allow for importing CSVs and such. It does not seem possible in `@auth/core` to map roles from a service and I'm not sure if that would even be desirable or straightforward for users).
- Disable users from accessing pages if they don't have the correct roles (client-side only, as our routing for these pages is client-side? Should we also try to block by HTTP somehow or it's not worth it?).
- Hide inaccessible pages in sidebar + show message that page is inaccessible if user somehow navigates to them.

**Demo:**

https://github.com/mui/mui-toolpad/assets/10789765/97c37db6-b3cd-428a-ad21-b936003c59c2


docs/schemas/v1/definitions.json
```diff
@@ -47,6 +75,31 @@
                     ]
                   },
                   "description": "Available roles for this application. These can be assigned to users."
+                },
+                "users": {
```
Janpot, line 79: I think we can start with authentication only for this PR. 

I don't think it makes sense at the moment for the toolpad configuration to act as a user pool. This would require developers to have to hardcode each user in the configuration. And likely be a privacy problem for public repos.

apedroferreira: Ok, I can keep most of this PR on hold if that's the best course and see what can be merged for now.
We can also have a more in-depth discussion about this and see if there's any more solutions we can agree on.

packages/toolpad-app/src/server/auth.ts
```diff
@@ -27,6 +44,18 @@ export function createAuthHandler(base: string): Router {
...
             clientSecret: process.env.TOOLPAD_GITHUB_SECRET,
+            async profile(profile) {
+              const roles = await getProfileRoles(profile.email ?? '', project);
+
+              return {
+                ...profile,
+                id: profile.email ?? String(profile.id),
+                name: profile.name,
+                email: profile.email,
+                image: profile.avatar_url,
```
apedroferreira, line 55: id: profile.id.toString(),
name: profile.name ?? profile.login,
email: profile.email,
image: profile.avatar_url,

packages/toolpad-app/src/server/auth.ts
```diff
@@ -38,6 +67,17 @@ export function createAuthHandler(base: string): Router {
                 response_type: 'code',
               },
             },
+            async profile(profile) {
+              const roles = await getProfileRoles(profile.email, project);
+
+              return {
+                id: profile.email,
```
apedroferreira, line 74:
```suggestion
                id: profile.sub,
```?

apedroferreira:
Replaced by https://github.com/mui/mui-toolpad/pull/3077, can revisit later if we decide to add our own role assignments.

PR #3017, opened 2023-12-21
Remove components from appDom

Janpot:
Manage components separately from the dom.

* Solve ambiguities in the code on the flow between file system and what's in the runtime. Half of the flow was in the dom, half was out. This also caused too many updates of the editor when things in the project folder changes.
* Move further towards isolating the dom as model for the pages
* Fixed an issue along the way where the theme didn't update when it was changed manually on disk


Janpot:
@bharatkashyap the visual regression tests are flaky around the new query editor

PR #3027, opened 2023-12-24
[docs] Lint next.config.mjs

oliviertassinari:
This PR and https://github.com/mui/material-ui/pull/40301 should ensure #3026 can't happen again the next time we introduce a new required param in `withDocsInfra`.


Janpot (approved):
nice, tagging @mui/docs-infra to port X docs to use esm and type checking for their next config.

PR #3034, opened 2023-12-27
Protect runtime RPC with authorization

apedroferreira:
On query requests to the runtime RPC, require that a user has the required roles according to the page the query is called from.
Includes a few refactors too.


Janpot:
* I don't think we need to prioritize this until we've taken the time to change the dataprovider to not use "resource-file+function"  to call its method, but rather "page-file+datagrid". Because with the current method it's possible for every user to call arbitrary functions anyway, even if you change execQuery.
* I'm not sure we should change the rpc implementation for this. I think we could put authentication information in the server context so that users can get this info with `getContext()`. We should be able to piggy-back on this system to implement strict access enforcement on the server functions. But again, as long as datasources provide an interface that allows calling arbitrary functions, this last thing is pretty much wasted effort.

So I'd say:
* Forget about RBAC in the rpc layer for now, or refactor the dataprovider first instead.
* Expand `getContext` with auth information. this could be a nice user facing feature.

apedroferreira:
> * I don't think we need to prioritize this until we've taken the time to change the dataprovider to not use "resource-file+function"  to call its method, but rather "page-file+datagrid". Because with the current method it's possible for every user to call arbitrary functions anyway, even if you change execQuery.
> * I'm not sure we should change the rpc implementation for this. I think we could put authentication information in the server context so that users can get this info with `getContext()`. We should be able to piggy-back on this system to implement strict access enforcement on the server functions. But again, as long as datasources provide an interface that allows calling arbitrary functions, this last thing is pretty much wasted effort.
> 
> So I'd say:
> 
> * Forget about RBAC in the rpc layer for now, or refactor the dataprovider first instead.
> * Expand `getContext` with auth information. this could be a nice user facing feature.

Sounds good, thanks! I'll probably close this one for now then, and can add to the context in a further PR after the others are merged and authentication is released.

apedroferreira:
Closing, I'll create issues later for the features that were suggested instead of this implementation.

Issue #2865, opened 2023-11-03
Canvas overlay doesn't update after resize


Janpot:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

Steps:

1. select an element in the canvas
2. resize the canvas

### Current behavior 😯

https://github.com/mui/mui-toolpad/assets/2109932/b87930e4-dd27-48ab-82c7-d824b0236142

The canvas overlay doesn't update

### Expected behavior 🤔

It should update the size of the node hud.

### Your environment 🌎

<details>
  <summary><code>npx @mui/envinfo</code></summary>

```
  Don't forget to mention which browser you used.
  Output from `npx @mui/envinfo` goes here.
```
</details>

Janpot:
I bissected the commits on master and the issue seems to have been introduced in https://github.com/mui/mui-toolpad/pull/2760

apedroferreira:
I had a fix here https://github.com/mui/mui-toolpad/pull/2818, but I guess your suggestion is probably better, I hadn't realized this was a recent regression.

PR #3044, opened 2024-01-03
Remove react-devtools

Janpot:
All components have a wrapper now, we don't need react devtools anymore to reify toolpad nodes to the browser dom element, we can just add a react ref to the wrapper and register. This simplifies the runtime, and it frees us from digging around in react-devtools internals.

also seems to close https://github.com/mui/mui-toolpad/issues/2865


apedroferreira (approved):
LGTM

PR #3047, opened 2024-01-03
Add universal required email configuration for authentication

apedroferreira:
Make it possible to add multiple email address patterns to be matched by authenticated users, either for the Github or Google authentication providers.
Also takes in consideration all verified emails in Github.
Also makes the authentication/authorization dialog be tabbed, brought this from https://github.com/mui/mui-toolpad/pull/3002.

https://github.com/mui/mui-toolpad/assets/10789765/fbcf546a-60e5-4e08-8e63-ca8f9006eb1a


packages/toolpad-app/src/server/auth.ts
```diff
@@ -44,12 +84,31 @@ export function createAuthHandler(base: string): Router {
...
+            }
             if (account?.provider === 'google') {
               return Boolean(
                 profile?.email_verified &&
                   profile?.email &&
-                  (!process.env.TOOLPAD_GOOGLE_AUTH_DOMAIN ||
-                    profile.email.endsWith(`@${process.env.TOOLPAD_GOOGLE_AUTH_DOMAIN}`)),
+                  (requiredEmails.length === 0 ||
+                    requiredEmails.some(
+                      (requiredEmail) => new RegExp(requiredEmail).test(profile.email!) ?? false,
```
Janpot, line 110: I'm not a huge fan of using regexes for this. Seems a bit like a footgun to me. e.g. users forgetting to add `$` on the end, or forgetting to escape `.`. Wouldn't >95% of the usecases be covered by just restricting the email domain? Can you benchmark other products on how configurable this is?

apedroferreira: Ok I'll give a look into how other tools do it, but last case scenario we can just check if the user email ends with the specified domain preceded by @, instead of regex.

packages/toolpad-app/src/server/auth.ts
```diff
@@ -27,6 +31,42 @@ export function createAuthHandler(base: string): Router {
           GithubProvider({
```
Janpot, line 31: Does this provider have to be recreated on every request? Would it make sense to put it in its own module?

apedroferreira: I was just following some examples and didn't consider that, it should be better to move it outside, will do!

apedroferreira: Things didn't work when I moved the provider to a separate variable, I think it should be fine to do it like this as that's how the examples I have seen work too.

Janpot: > Things didn't work when I moved the provider to a separate variable

Any idea why? This seems like a strange thing. Can you investigate?

apedroferreira: Sure I'm taking a look at how it works internally, this library has some weird things sometimes though, it seems to be a bit of a work in progress and most things aren't documented at all.

apedroferreira:
Renaming the property to `restrictedDomains` too, it sounds better.

packages/toolpad-app/src/server/auth.ts
apedroferreira: Nevermind it wasn't about the library this time - I was trying to set the providers as global variables but I don't think the environment variables had been initialized there. If I set the variables inside the `createHandler` function it works this fine, this should be enough? https://github.com/mui/mui-toolpad/pull/3047/commits/39c0f006c403dd400f6d9611a81cd8766b29a422

Janpot: That's what I assumed. What's preventing us from initializing variables earlier? It feels very counterintuitive and error-prone to have half the program work with uninitialized environment

apedroferreira: Isn't the environment set with `this.envManager.start()`? This runs after the `auth.ts` file is imported, that's why it's not there for a global variable in that file.

apedroferreira: Maybe we can load the Toolpad app's environment before any imports when the server starts if you think it's worth it. Sounds like an unrelated issue though, I don't see what's wrong with initializing the providers in `createAuthHandler`.

apedroferreira: I can also load the providers in a global function if that's better.

Janpot: > Maybe we can load the Toolpad app's environment before any imports when the server starts if you think it's worth it

Just

```tsx
import 'dotenv/config'
```
before anything else

apedroferreira: Ok, it works if I add that line to `toolpadAppServer.ts`, as that file seems to run before `index.ts` - it might be because the worker imports that file first, I can look deeper into it and try in another file if this one doesn't seem ideal for any reason.
https://github.com/mui/mui-toolpad/pull/3047/commits/07015a045461e47395b092807171c3aec9b42277

packages/toolpad-app/typings/@auth.d.ts
```diff
@@ -0,0 +1,5 @@
+export declare module '@auth/core/types' {
+  interface Profile {
+    verifiedEmails?: string[];
```
Janpot, line 3: If this isn't public API, should we really rely on it?

packages/toolpad-app/src/server/auth.ts
```diff
@@ -1,64 +1,127 @@
...
+      if (!profile.email) {
+        // If the user does not have a public email, get another via the GitHub API
+        // See https://docs.github.com/en/rest/users/emails#list-public-email-addresses-for-the-authenticated-user
+        const githubRes = await fetch('https://api.github.com/user/emails', {
+          headers,
+        });
+
+        if (githubRes.ok) {
+          const emails: GitHubEmail[] = await githubRes.json();
+          profile.email = (emails.find((e) => e.primary) ?? emails[0]).email;
```
Janpot, line 43: Can't we just look for the first verified email that satisfies our restrictedDomains setting and use that as the email? Then set the `email_verified` boolean to `true`?

apedroferreira: That was the default logic but it would be an improvement, i'll change it and fallback to primary email.

packages/toolpad-app/typings/@auth.d.ts
apedroferreira: With your suggestion `verifiedEmails` shouldn't be needed anymore so I'll remove it, it was only for verifying the restricted domains with the Github provider.

packages/toolpad-app/src/server/auth.ts
apedroferreira: I've added the changes here https://github.com/mui/mui-toolpad/pull/3047/commits/fd18d7c074bc53f1e826e61235a34089af2237ea, let me know if it's good now!

Janpot (approved):

PR #3048, opened 2024-01-03
Improve authorization middleware

apedroferreira:
Improve authorization middleware to cover everything under `/prod` except static files and Toolpad dev server.
This way, the same middleware can be used to redirect to sign in and protect the RPC, among other things.


packages/toolpad-app/src/server/auth.ts
```diff
@@ -154,14 +154,14 @@ export async function createAuthPagesMiddleware(project: ToolpadProject) {
 
     const hasAuthentication = authProviders.length > 0;
 
+    const isPageRequest = req.get('sec-fetch-dest') === 'document';
     const signInPath = `${base}/signin`;
 
-    let isRedirect = false;
+    let isAuthorized = true;
     if (
+      (!project.options.dev || isPageRequest) &&
```
apedroferreira, line 162: In dev mode it only redirects pages so that the user in preview sees the same behavior there will be in production. Can't hide RPC requests behind authentication in dev mode for example, as the editor needs to make them.

Janpot (approved):

Janpot:
Looks good. Annoying that we have a parallel dev and prod implementation. A more ideal implementation would be one single function which switches on/off certain middleware based on a `dev` flag. That would guarantee consistency in ordering of the middlewares between the two environments. The challenge would be to avoid loading `vite` in prod. That's what causes high memory spikes on startup, even if a vite devserver isn't initialized.

PR #3055, opened 2024-01-04
Update monorepo

apedroferreira:


apedroferreira:
Actually I'll probably just move the old deduplicate script to our repo.

.circleci/config.yml
```diff
@@ -92,9 +92,6 @@ jobs:
       - run:
           name: Should not have any git not staged
           command: git add -A && git diff --exit-code --staged
-      - run:
```
Janpot, line 95: Let's merge https://github.com/mui/mui-toolpad/pull/2546 first instead of removing this check

apedroferreira: Sounds good, I was going to copy the old script to our repo instead of removing, but we can merge that first and all good!

PR #3056, opened 2024-01-04
Authentication/authorization tests

apedroferreira:
Tests for authentication and authorization, using the new `credentials` authentication provider that takes username/password combinations (enabled just in test mode in this PR), and some mock users.

Includes some refactors too, I saw many things to improve/fix while implementing tests.


packages/toolpad-app/src/server/auth.ts
```diff
@@ -98,18 +163,45 @@ export function createAuthHandler(project: ToolpadProject): Router {
     tenantId: process.env.TOOLPAD_AZURE_AD_TENANT_ID,
   });
 
+  const credentialsProvider = CredentialsProvider({
```
apedroferreira, line 166: I can move the mock results somewhere else maybe, if it feels weird to have them here like this.

packages/toolpad-app/package.json
```diff
@@ -41,7 +41,7 @@
     }
   },
   "dependencies": {
-    "@auth/core": "0.24.0",
+    "@auth/core": "0.20.0",
```
apedroferreira, line 44: There's an issue in newer versions of `@auth/core` and I don't see any good solution proposed yet in https://github.com/nextauthjs/next-auth/issues?q=cannot+parse+action, will wait until they provide a fix or solution.

Janpot: Is there an exact issue?

test/playwright/tmp-0xS4E0/fixture/toolpad/.gitignore
```diff
@@ -0,0 +1 @@
+.generated
```
Janpot, line 1: Accidentally committed these?

packages/toolpad-app/package.json
apedroferreira: Nevermind it was an easy fix after all, I hadn't tried this for some reason: https://github.com/mui/mui-toolpad/pull/3056/commits/4445f9d00020f36a282cb1e29ca41bd7b53620c4

Janpot (approved):

Issue #2888, opened 2023-11-10
Support server-side row creation


prakhargupta1:
Add CRUD operations to the Toolpad [data providers](https://mui.com/toolpad/concepts/data-providers/).

The data provider interface can be augmented with a `createRecord(fields)` method. When available the frontend component should automatically provide a create button in the data grid which adds an empty editable row and a commit button that calls the `createRecord` in the backend with the new fields

JerryWu1234:
do you have more details?
I want to take this PR

Janpot:
We just merged https://github.com/mui/mui-toolpad/pull/3001. Creation will be similar.

JerryWu1234:
> We just merged #3001. Creation will be similar.

I got it.
let me try it.

JerryWu1234:
@Janpot this can assign me ? let me easy to find out when i want to find this issue

Janpot:
@JerryWu1234 I'll be picking this up myself

PR #3058, opened 2024-01-05
Add serverside row creation

Janpot:
Closes https://github.com/mui/mui-toolpad/issues/2888

This PR introduces the following changes:

* when the data provider contains a `createRecord` method, the grid will have an "add record" button in the toolbar
* clicking this button will add a row at the top of the grid in draft mode
* when clicking "update" after editing the row, the grid will call `createRecord` on the provider with the new values
* all crud operations are disabled for all rows when one is in editing mode
* a row in editing has to be explicitly updated or canceled, no focus events
* createRecord and updateRecord must now return the created or updated record so the grid can update UI accordingly. This is important in scenarios where the server has to generate defaults for something
* add a notification system for success/failure of the crud operations. When creating a new record, a link is created in the notification which scrolls the grid to the newly created item.

changes to public API:
* dataprovider `updateRecord` can now optionally return a record to replace the one from the grid.
* dataprovider now supports `createRecord`, which has the same signature as `updateRecord` minus the `id` parameter.

https://github.com/mui/mui-toolpad/assets/2109932/1d687c64-77ef-4ce2-8e02-2004bb8525a4

docs: https://deploy-preview-3058--mui-toolpad-docs.netlify.app/toolpad/concepts/data-providers/#row-creation

- [x] docs
- [x] tests


bharatkashyap:
A video might be helpful to those tracking progress/interested in this feature?

Janpot:
🙂 The PR is not reviewable yet

bharatkashyap (approved):

PR #3063, opened 2024-01-07
Update some documentation for pnpm

Janpot:
- [ ] I've read and followed the [contributing guide](https://github.com/mui/mui-toolpad/blob/master/CONTRIBUTING.md#sending-a-pull-request) on how to create great pull requests.
- [ ] I've updated the relevant documentation for any new or updated feature.
- [ ] I've linked relevant GitHub issue with "Closes #<issue id>".
- [ ] I've added a visual demonstration in the form of a screenshot or video.

PR #3065, opened 2024-01-08
[core] Use pnpm

Janpot:
Third try


Janpot:
This is a combination of https://github.com/mui/mui-toolpad/pull/2546 and https://github.com/mui/mui-toolpad/pull/3063 with no other changes. The problem that caused us to revert was fixed by https://github.com/mui/mui-toolpad/pull/3074

PR #3076, opened 2024-01-12
Handle errors during data provider introspection

Janpot:
Make sure we give feedback when the data provider introspection fails. Currently it just shows as empty, this PR makes it show an error in the rows area. We can also render the grid as loading while introspecting.


bharatkashyap (approved):
👍🏻

PR #3067, opened 2024-01-09
Add authorization docs

apedroferreira:
Authorization docs. Can be iterated on.

Removes the feature flag for authentication/authorization too - I think we can fully release the feature by merging https://github.com/mui/mui-toolpad/pull/3056 first and then this PR.

Previews:
- https://deploy-preview-3067--mui-toolpad-docs.netlify.app/toolpad/concepts/authentication/
- https://deploy-preview-3067--mui-toolpad-docs.netlify.app/toolpad/concepts/rbac/

PR #3077, opened 2024-01-12
Add RBAC with Azure AD authentication provider

apedroferreira:
- Add RBAC based on provider roles.
- Add Azure Active Directory (AD) as authentication provider, the first one with roles enabled.
- Add role mapping from provider role names to Toolpad role names, set via UI in `application.yml`.
- Disallow users from accessing pages if they don't have the correct roles (client-side only, as our routing for these pages is client-side? Should we also try to block by HTTP somehow or is it not worth it?). In a separate PR I am disallowing queries from being called if user does not have the necessary roles, if possible, just need to finish it up if needed, it's outdated right now (https://github.com/mui/mui-toolpad/pull/3034)
- Hide inaccessible pages in sidebar + show message that page is inaccessible if user somehow still navigates to them.

Will add to documentation in the authentication/authorization documentation PR: https://github.com/mui/mui-toolpad/pull/3067
Also to cover in authentication/authorization tests in: https://github.com/mui/mui-toolpad/pull/3056

### Video

https://github.com/mui/mui-toolpad/assets/10789765/616fdc07-64b9-4b44-b235-f00d127b6db9

It's possible to assign multiple provider roles to a Toolpad role by separating the roles with commas such as `admin, reader,writer`.


package.json
```diff
@@ -98,7 +98,7 @@
   "engines": {
     "npm": "please-use-yarn",
     "node": ">=18",
-    "pnpm": "8.7.0"
+    "pnpm": ">=8.7.0"
```
apedroferreira, line 101: Is this fine or should I use this specific version of `pnpm`? I had a more recent one installed.

apedroferreira:
I saw something else wrong in the video - the admin-only page is showing when it shouldn't, feel free to wait before reviewing until I add some fix.

packages/toolpad-app/src/server/auth.ts
```diff
@@ -2,13 +2,20 @@ import express, { Router } from 'express';
...
 import { OAuthConfig } from '@auth/core/providers';
 import { asyncHandler } from '../utils/express';
 import { adaptRequestFromExpressToFetch } from './httpApiAdapters';
 import { ToolpadProject } from './localMode';
 import * as appDom from '../appDom';
+import { AuthProvider } from '../types';
+
+const SKIP_VERIFICATION_PROVIDERS: AuthProvider[] = [
+  // Azure AD should be fine to skip as the user has to belong to the organization to sign in
+  'azure-ad',
```
apedroferreira, line 17: Also they don't return any data that allows us to determine if emails are verified, from what I searched.

docs/schemas/v1/definitions.json
```diff
@@ -82,6 +83,26 @@
                     ]
                   },
                   "description": "Available roles for this application. These can be assigned to users."
+                },
+                "roleMappings": {
```
Janpot, line 87: Intuitively this feels like a provider specific property. e.g. how would this work under Google/GitHub?
I'd sort of have expected something like:

```yml
providers:
  - provider: 'azure-ad'
    roles:
      - source: 
          - my-azure-role
          - my-other-azure-role
        target: my-toolpad-role
```

Also, UI doesn't really have proirity right now. I'm fine to release this with config only

package.json
Janpot: I was hoping we could pin the pnpm version, just like we used to do with yarn

apedroferreira: So I should use 8.7.0? Why are we not using a more recent version? 
Anyway I can switch, no problem, I guess it's safer against weird issues if we all use the same version.

docs/schemas/v1/definitions.json
apedroferreira: You're right, I'll make that change, seems better. The UI is already done so should be easy to keep now.

apedroferreira: I've changed the schema here: https://github.com/mui/mui-toolpad/pull/3077/commits/20755776f3c4e76aaed649ceda54469e735dc31b

The schema definition is much better like this too.

Janpot (approved):
👍  Looking good

PR #3085, opened 2024-01-15
Load dotenv asap on the serverworker

Janpot:


packages/toolpad-app/src/server/appServerWorker.ts
```diff
@@ -1,3 +1,4 @@
+import 'dotenv/config';
```
apedroferreira, line 1: I forgot that we also had the option to load the env in the CLI commands, just mentioning it in case it's better than importing like this in multiple files.

apedroferreira (approved):
Looks good!

packages/toolpad-app/src/server/appServerWorker.ts
Janpot: It was the only way I could make it work. But you are right, we should never build things that we don't understand. I've done some further root cause analysis and after inspecting the bundles I noticed some imports before `dotenv/config`. I immediately remembered [an issue](https://github.com/evanw/esbuild/issues/399) I'd seen before. I disabled code splitting and now the ordering is predictable. Adding a big fat comment to explain why the setting is there.

Issue #2917, opened 2023-11-20
Add a public folder to host static content within the Toolpad application


Janpot:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Would like to be able to host static content within the Toolpad application, just like http://localhost:3000/static/movies.json file. Application frameworks like vite/next.js/remox offer the use of a "public" folder which statically hosts its content on the application

### Examples 🌈

allow creating a "public" folder which hosts its content on the Toolpad application.

Open question:
- how do we avoid clashes with Toolpad internal routes (present and future ones)

### Motivation 🔦

https://discord.com/channels/1131323012554174485/1173687498308866058/1173687498308866058

JerryWu1234:
we can add a mark that like "_".
such as: `/__public`

Janpot:
To not deviate too much from other frameworks, I think we can just host everything from the public folder under the root. e.g. `./toolpad/public/myData.json` will be hosted as `/myData.json` on the Toolpad application.

JerryWu1234:
> * how do we avoid clashes with Toolpad internal routes (present and future ones)

if follow you said, we don't avoid clashes with internal routes.
because Internet route didn't have an extension

Janpot:
> if follow you said, we don't avoid clashes with internal routes.

Yes, in hindsight, I think we can live with that

JerryWu1234:
May I pick this up?

Janpot:
yes

JerryWu1234:
@Janpot 
![image](https://github.com/mui/mui-toolpad/assets/27261522/3ee8b565-d035-4359-a62f-1282b0101b6b)
I have a question. how to handle if there is collision with the toolpad-app directory ?

Janpot:
public folder takes precedence. We can add a warning in dev mode if we detect a collision

JerryWu1234:
I have done this issue,please for review

maya9997:
anyone can help me how to run this repo not created from cli ? 
@JerryWu1234 
@Janpot

PR #3098, opened 2024-01-18
specify public for static accessed

JerryWu1234:
#2917 

### DIFF
--- a/packages/toolpad-app/src/server/index.ts
+++ b/packages/toolpad-app/src/server/index.ts
@@ -315,6 +315,10 @@ async function createToolpadHandler({
   const publicPath = path.resolve(currentDirectory, '../../public');
   router.use(express.static(publicPath, { index: false }));
 
+  const childrenPublic = path.resolve(project.getRoot(), './public');
+  console.log(childrenPublic, '>>>>>>');
+  router.use(express.static(childrenPublic, { index: false }));
+
   const appHandler = await createToolpadAppHandler(project);
 
   router.use(project.options.base, appHandler.handler);


JerryWu1234:
image

JerryWu1234:
@Janpot help review

packages/toolpad-app/src/server/index.ts
```diff
@@ -315,6 +316,16 @@ async function createToolpadHandler({
   const publicPath = path.resolve(currentDirectory, '../../public');
   router.use(express.static(publicPath, { index: false }));
 
+  const childrenPublic = path.resolve(project.getRoot(), './public');
+  const sameNameStatic = await compareSameLayerFolder(publicPath, childrenPublic);
```
Janpot, line 320: This is not necessary.
Over time we will add a warning when there is a _Toolpad page_ overlap with the public folder. a first version does not need this warning as I'm concerned

The `publicPath` thing at the moment is mostly there because it hosts the sample movies.json file we use as a default in the REST queries. I want to remove it over time. No need to add warnings there.

JerryWu1234: >public folder takes precedence. We can add a warning in dev mode if we detect a collision

doesn’t the first version care the warning ?
could we leave the code of `compareSameLayerFolder`? I have added  a lot of the tests 😂

JerryWu1234: I just want to confirm, can I simply delete this comparison code?
`const sameNameStatic = await compareSameLayerFolder(publicPath, childrenPublic);`

Janpot: > doesn’t the first version care the warning ?

It wouldn't catch a collision with a Toolpad page. It would only catch collisions with the contents of https://github.com/mui/mui-toolpad/tree/master/packages/toolpad-app/public Which we should probably remove because it contains files relevant to the Toolpad editor, not the Toolpad app

JerryWu1234: > > doesn’t the first version care the warning ?
> 
> It wouldn't catch a collision with a Toolpad page. It would only catch collisions with the contents of https://github.com/mui/mui-toolpad/tree/master/packages/toolpad-app/public Which we should try remove in the future because it contains files relevant to the Toolpad editor, not the Toolpad app

Which kind of page(path) does it belong to a `Toolpad page` ？ 
Sorry I have a little bit confused.

JerryWu1234: "There is only this link for user access. Would you like to inspect whether there is a collision with 'http://localhost:3000/_toolpad/app/pages/page'?"

Janpot: Yes, but as I said before, it can be a separate effort. Better focus on creating docs for this feature.

JerryWu1234: > Yes, but as I said before, I prefer it to be a separate effort. I's better to focus on creating docs for this feature.

image

>"There is only this link for user access. Would you like to inspect whether there is a collision with 'http://localhost:3000/_toolpad/app/pages/page'?"

I have done this

JerryWu1234:
@Janpot  please help review

PR #3100, opened 2024-01-18
Open up the slots API for custom components (WIP)

Janpot:
Allow users to define their own slots in custom components:
```tsx
import { Box } from '@mui/material';
import { createComponent, UnstableSlots } from '@mui/toolpad/browser';

export interface MyWrapperProps {
  content?: React.ReactNode;
  msg?: string;
}

function MyWrapper({ msg }: MyWrapperProps) {
  return (
    <Box sx={{ flex: 1, display: 'flex', flexDirection: 'row' }}>
      {msg}
      <UnstableSlots prop="content" />
    </Box>
  );
}

export default createComponent(MyWrapper, {
  argTypes: {
    content: {
      type: 'element',
    },
    msg: {
      type: 'string',
    },
  },
});
```

* marked as unstable for now

### TODO

- [x]  this depends on https://github.com/mui/mui-toolpad/pull/3099
- [ ] make it work for any property, regardless of whether it's named "children" or marked with type: 'template'
- [ ] page hierarchy explorer doesn't work with props that aren't name "children"
- [x] migrate built-in components to use this
- [ ] docs
- [x] tests


Janpot:
@apedroferreira For a List version of this I was thinking of the following API:

```tsx
import { Template, TemplateInstance } from '@mui/toolpad-core';
import { List as MuiList, ListItem } from '@mui/material';

function List() {
  return (
    <MuiList>
      {Array.from(Array(10), (_, index) => (
        <ListItem key={index}>
          {index === 0 ? (
            <Template prop="item" />
          ) : (
            <TemplateInstance prop="item" id={index} scope={{ i: index }} />
          )}
        </ListItem>
      ))}
    </MuiList>
  );
}
```

With these APIs we can remove all the weird property wrapping that we do in the runtime and give users a clean API to implement these features themselves. wdyt?

apedroferreira:
> With these APIs we can remove all the weird property wrapping that we do in the runtime and give users a clean API to implement these features themselves. wdyt?

Seems more readable than the current component instead of handling the logic in different places/files, so it's probably a good change!
Not sure what other use cases there would be for `TemplateInstance` though, but it makes sense and seems easy to understand. Maybe the `id` prop of `TemplateInstance` could just be called `index` and take a number, which would automatically include `i` in the scope too? Unless there's more cases where we want it to be more flexible.

Janpot:
> Not sure what other use cases there would be for TemplateInstance though, but it makes sense and seems easy to understand. Maybe the id prop of TemplateInstance could just be called index and take a number, which would automatically include i in the scope too? Unless there's more cases where we want it to be more flexible.

it essentially is a render prop, which have many use cases. In any case, it would feel a bit strange to me to restrict these components to create index based lists only. I can certainly imagine use cases where more information is passed than the index. or where there is no obvious index, e.g. while traversing an object.

apedroferreira:
> > Not sure what other use cases there would be for TemplateInstance though, but it makes sense and seems easy to understand. Maybe the id prop of TemplateInstance could just be called index and take a number, which would automatically include i in the scope too? Unless there's more cases where we want it to be more flexible.
> 
> it essentially is a render prop, which have many use cases. In any case, it would feel a bit strange to me to restrict these components to create index based lists only. I can certainly imagine use cases where more information is passed than the index. or where there is no obvious index, e.g. while traversing an object.

Yeah, those were just some first thoughts but it should be good to have it be more flexible/agnostic.

PR #3114, opened 2024-01-22
Update CONTRIBUTING.md

apedroferreira:
Fix contribution guide section about running Toolpad for local development with linked packages.
It should also work with `pnpm` and `link:`, not sure if that's preferable?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -110,13 +110,13 @@ pnpm install
 
    1. Replace `<your-local-toolpad-monorepo>` with the path to the toolpad monorepo on your file system. Make sure to keep `portal:`.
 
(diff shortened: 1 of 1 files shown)


Janpot:
> Should we use yarn instructions here? Or update the instructions to pnpm?

the latter if possible, otherwise the former

apedroferreira:
> > Should we use yarn instructions here? Or update the instructions to pnpm?
> 
> the latter if possible, otherwise the former. `portal:` dependencies don't exist in pnpm, there is [`injected`](https://pnpm.io/package_json#dependenciesmetainjected) but I don't believe it works as seamless as `portal:`. Might have to reinstall on every change (which should be faster than with yarn though)

Just using `"@mui/toolpad-core": "link:../mui-toolpad-pedro/packages/toolpad-core"` has been working well for me I think.

Issue #315, opened 2022-04-25
Toolpad should support exporting components as a URL


newguy-123:
Export ESM to allow embedding like Framer Handshake:
https://www.framer.com/developers/handshake/

Janpot:
Some things to think about:
* What exactly will be exported? a page? an app? 
* How does data fetching work? 
* We're doing internal apps, what does this mean for authentication? Can you import a component of a private app?
* ...

prakhargupta1:
From internal apps perspective, the developer might want to embed Toolpad app in their existing internal tool. I such a scenario, can they pass on auth params to grant access to Toolpad app? This is clearly more work for the developer, but they don't want other teams to log into a different platform.
So a single page should be exported. 


https://www.jotform.com/help/how-to-embed-apps-to-a-website/
https://docs.appsmith.com/advanced-concepts/embed-appsmith-into-existing-application

From external apps perspective. I think making a publicly accessible through a URL >> making it embeddable.

buremba:
Great emphasis on the the hybrid approach and the integration with internal tools. I'm also doing proof of concept with Toolpad and it looks great so far. 

For the hybrid approach, I wonder if we can find a way to let the user switch between UI and code easily. I see that you call the YML & JSON as "output" and React & ESM as "export". Rather than treating the components created in UI as packages, we might also consider exporting to MDX from the internal JSON format and parsing MDX to the same JSON format for better readability and compatibility. MDX already supports compiling to JSX modules that can be imported in ESM. Since MDX can import modules and define `export` or Provider components, we can also compile $jsExpression easily. WDYT?

Janpot:
> I see that you call the YML & JSON as "output" and React & ESM as "export".

I don't see the yaml/json files just as "output", they also serve as "input". They are the underlying data model that describes a toolpad application. It's hand editable but that's not the intention for day-to-day use. You can hand edit it as a power user to fulfil tasks such as bulk editing large applications, generate an application from higher level requirements,... The main requirement for it is to be declarative. We've experimented a bit and it doesn't seem likely that it's very feasable to use a non-declarative language as the data model for a Toolpad application (Also see [utopia.app](utopia.app) or [blocksUI](https://blocks-ui.com/))

We see "export" or "eject" as a one-way action only, producing runnable code where the focus would be on one of the following use cases:
[...]

PR #3125, opened 2024-01-25
Next.js integration: run Toolpad apps as React server components

Janpot:
WIP

```tsx
// ./app/my-toolpad-app/[[...slug]]/page.tsx
import { ToolpadApp } from '@mui/toolpad/next';

export default function ToolpadPage() {
  return <ToolpadApp base="/my-toolpad-app" dir="./myToolpadAppDir" />;
}
```


packages/toolpad-app/src/exports/next.tsx
```diff
@@ -0,0 +1,18 @@
...
+import ToolpadAppClient from './next-client';
+import { loadDom } from '../server/localMode';
+import createRuntimeState from '../runtime/createRuntimeState';
+
+export interface ToolpadAppServerProps {
+  base: string;
+  dir: string;
+}
+
+export async function ToolpadApp({ base, dir = './toolpad' }: ToolpadAppServerProps) {
```
buremba, line 13: How does the app decide which page to run? If it's the URL path concerning `base`, is it possible to abstract the URL from the app and modify the props as follows?

```
export interface ToolpadAppServerProps {
  dir: string;
  page: string;
}
```

It could help us dynamically render any page without worrying about the paths, giving user more control IMO. WDYT?

Janpot: Yes, that makes sense. This is somewhat on hold because it depends on a bunch of [low-level work](https://github.com/mui/mui-toolpad/pull/3171) that needs to happen first. We'll also have to abstract Toolpad routing to be able to plug in the Next.js router.

buremba: > Another option I'm thinking of is to expose another component that renders a single page instead of the whole app.
Yes, that would be even more useful for [my use case](https://github.com/mui/mui-toolpad/issues/315#issuecomment-1805702316) where I have a backend to pull the MDX file and I can transform it on the fly render the page without even shell. Further, the users can use Toolpad to generate the yml where I can provide a serializer for them.

PR #3135, opened 2024-01-26
Add authentication examples

apedroferreira:
Add examples of authentication for Github and Google, but without the needed environment variables which would need to be added by the person who tries the example.
The examples have `.env.example` files that show what environment variables can be set, plus the READMEs link to the relevant documentation.


examples/auth/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+<p class="description">An app that shows how to set up authentication in Toolpad.</p>
+
+Example showcasing how to set up authentication providers in a Toolpad app. [The docs](https://mui.com/toolpad/concepts/authentication/).
+
+<a target="_blank">
+  <img src="https://mui.com/static/toolpad/marketing/authentication.png" alt="Authentication" style="aspect-ratio: 131/88;" width="524">
+</a>
+
+## How to run
```
Janpot, line 11: Would have to explain here as well how to set up the tokens.
Is it possible to try this example with e.g. only Google? It may make sense to make 2 examples, one for each rpovider.

examples/auth/toolpad/application.yml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: application
+spec:
+  {
+    authentication:
+      { providers: [{ provider: github }, { provider: google }], restrictedDomains: [] },
```
Janpot, line 6: 🤔 Why does this file use `{` and `}` everywhere?

apedroferreira: I probably changed this manually and not really used to writing in YAML :D, I should be able to change it in the UI now.

examples/auth/README.md
apedroferreira: I guess so, but since the providers can work together and not to be redundant I thought I would have them together... Both options could be valid, so not sure, I guess I can split them since it's only 2 providers and they should be pretty popular.

Janpot: I feel like the example loses its purpose if I have to set up both providers to try it out. Too much friction.

CONTRIBUTING.md
```diff
@@ -62,7 +62,7 @@ If your application has dependencies other than `@mui/toolpad`, you have to temp
 1. Make sure to start the build in watch mode again and the run the app with
 
    ```bash
-   pnpm toolpad dev examples/qr-generator --dev
+   pnpm toolpad dev examples/qr-generator/toolpad --dev
```
apedroferreira, line 65: If I don't include `toolpad` it creates the files outside the `toolpad` folder, that's not expected, right? Or am I doing something wrong?

examples/auth/README.md
apedroferreira: Got it, I've split the examples and linked from the readmes to the OAuth client / provider specific instructions too.

docs/src/modules/components/ExamplesGrid.js
```diff
@@ -106,6 +106,13 @@ function layouts() {
       href: 'https://github.com/mui/mui-toolpad/tree/master/examples/charts',
       source: 'https://github.com/mui/mui-toolpad/tree/master/examples/charts',
     },
+    {
+      title: 'Google Authentication',
```
apedroferreira, line 110: I guess we can show Google only here as it should be the most popular provider.

Janpot (approved):
👌

CONTRIBUTING.md
apedroferreira: It works well after this change and it matches the command that was already in step 3 above it, so I'm pretty sure this should be the way to write it.

PR #3136, opened 2024-01-26
Replace basic authentication with credentials authentication provider

apedroferreira:
We don't have to merge this one as not sure if it's a good idea and apparently the current basic authentication works well on iframes but this one might not work as well? Anyway the idea is here, if we close it it's ok.
We also don't necessarily have to remove the previous basic authentication, these features could coexist...

As has been briefly discussed before, here's how we could replace our current basic authentication feature with using the new authentication features, with a "Credentials" authentication provider that I already had to set up for authentication tests anyway.

https://github.com/mui/mui-toolpad/assets/10789765/8e47e59a-b9ea-40a6-88c8-b46f0a5d5891


Janpot:
As per last grooming meeting:
- to test whether this would break iframe embedded apps like we use in notion

apedroferreira:
> As per last grooming meeting:
> 
> * to test whether this would break iframe embedded apps like we use in notion

Looks like that even if I whitelist requests from iframes, which are all being blocked by the auth middleware right now, the CSRF protection doesn't allow the sign-in to work.
So I guess we either ditch this PR or we don't use it as a replacement for the basic auth we already have.

PR #3157, opened 2024-02-01
Add authenticated session to app server context

apedroferreira:
Add authenticated `session` to server context so that it can be retrieved in custom functions.
Had to move some things around between `@mui/toolpad-core` and `@mui/toolpad-utils`, let me know if it could be set up differently.
Also updated docs with an example, and covered in tests.


bharatkashyap (approved):

apedroferreira:
Looks like with the recent changes we can redirect to the sign-in page with React Router: https://github.com/mui/mui-toolpad/pull/3157/commits/172d2c385c363a1c959d3ffb219dc42d8027257c

About the flaky Firefox test mentioned in https://github.com/mui/mui-toolpad/pull/3179#issuecomment-1933796998 I just added those errors to the ignore list in this test, i think it should be fine here? 
From what I found it looks like in the last check that sees if the users gets redirected to the sign-in page, that redirect cancels some network requests and in Firefox that leads to showing some console errors. This discussion seems to be related to that https://bugzilla.mozilla.org/show_bug.cgi?id=1280189#c7, and seems like they're just browser differences...

docs/data/toolpad/reference/api/get-context.md
```diff
@@ -33,14 +33,15 @@ a `ServerContext` containing information on the context the backend function was
...
-| :---------- | :-------------------------------------- | :------------------------------------------------ |
-| `cookies`   | `Record<string, string>`                | A dictionary mapping cookie name to cookie value. |
-| `setCookie` | `(name: string, value: string) => void` | Use to set a cookie `name` with `value`.          |
+| Name        | Type                                                               | Description                                                                         |
+| :---------- | :----------------------------------------------------------------- | :---------------------------------------------------------------------------------- |
+| `cookies`   | `Record<string, string>`                                           | A dictionary mapping cookie name to cookie value.                                   |
+| `setCookie` | `(name: string, value: string) => void`                            | Use to set a cookie `name` with `value`.                                            |
+| `user`      | `{ name: string; email: string; avatar: string; roles: string[] }` | Get current [authenticated](/toolpad/concepts/authentication/) logged-in user data. |
```
Janpot, line 44: * Let's add a separate table for this?
* Would it make sense to call this `session` instead of `user`?

apedroferreira: `session` sounds good, it's what it's called in `@auth/core` too, `user` might be a bit too generic.

apedroferreira: I've made these changes, ended up having `session` in context which has a `user` property with this data.
Also covered this in one of the authentication tests (since the user needs to be logged-in to see the context session data).

Janpot:
@apedroferreira Would it make sense to also add user info in the page state for components to bind to? e.g. like https://docs.retool.com/reference/apps/global/objects/current_user

apedroferreira:
> @apedroferreira Would it make sense to also add user info in the page state for components to bind to? e.g. like https://docs.retool.com/reference/apps/global/objects/current_user

It would be nice, I've created an issue https://github.com/mui/mui-toolpad/issues/3209.
Marked it as a good first issue but not 100% sure if I'm underestimating how difficult it would be.

PR #3170, opened 2024-02-06
Add isCanvas into app host context

Janpot:
Working towards allowing the canvas to render without relying on vite devserver, this decouples the canvas implementation from having to run in an iframe


apedroferreira (commented):
Got it, looks good!

PR #3171, opened 2024-02-06
[core] Add canvas mode that doesn't rely on vite

Janpot:
First batch of changes that renders the canvas in the same react tree as the editor instead of loading a page in an iframe. Currently behind a flag `EXPERIMENTAL_INLINE_CANVAS` because more work is required to make it the main rendering mode. But aiming to already merge this as to avoid accumulating many big architectural changes in one PR.

Now the editor is hosted under the same path as the app:
- /{basepath}/app/pages/... => application
- /{basepath}/editor/pages/... => editor (new)

things that are now fixed

- Both editor and app are now built in the same vite config, meaning we won't get clashing deps optimization anymore
- The canvas is rendered directly in the react tree of the editor, so now pages load immediately in the editor

add to the .env file to play with this:
```
.env
EXPERIMENTAL_INLINE_CANVAS=1
```

To do (in follow up):
- [x] fix monaco
- [x] fix expression editor not seeing scope
- [ ] change links to new editor url
- [ ] make tests green
  - [ ] backand basis
    - [ ] function editor reload
  - [ ] code components
    - [ ] can create new component
  - [ ] duplication
  - [ ] editor
    - [ ] deletelast
    - [ ] index
      - [ ] code editor autocomplete
  - [ ] rest basic
    - [ ] rest editor basic
  - [ ] undo redo
    - [ ] basic
      - [ ] test basic undo redo
- [ ] remove legacy and simplify

PR #3176, opened 2024-02-07
[core] Define routes statically

Janpot:
Define the routes statically. In preparation for moving to `react-router` new API with [`createBrowserRouter`](https://reactrouter.com/en/main/routers/create-browser-router). Which allows for lazy-loading routes. This will be interesting in the long term future to do code splitting between pages, but necessary in the short term to host the editor alongside the application for https://github.com/mui/mui-toolpad/pull/3171

Statically means that the whole routing is now defined in:

```tsx
<Routes>
  <Route path="/signin" element={<SignInPage />} />
  <Route path="/" element={<ToolpadAppLayout />}>
    <Route path="/pages/:pageName" element={<RenderedPage />} />
    <Route path="/pages" element={<DefaultPageNavigation />} />
    <Route path="/" element={<DefaultPageNavigation />} />
    <Route path="*" element={<PageNotFound />} />
  </Route>
</Routes>
```

In this PR:
- defines the routes statically
- fixes error when no pages are available in the app
- uses outlet to render child routes in app layout

### DIFF
--- a/packages/toolpad-app/src/runtime/ToolpadApp.tsx
+++ b/packages/toolpad-app/src/runtime/ToolpadApp.tsx
@@ -1529,11 +1529,11 @@ function shouldShowPreviewHeader(appHost: AppHost): boolean {
   return !!appHost.isPreview && !appHost.isCanvas;
 }
 
-interface ToolpadPagesLayoutProps {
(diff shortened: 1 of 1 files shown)


bharatkashyap (commented):
👍🏻

PR #3177, opened 2024-02-07
Fix drag and drop

apedroferreira:
I've noticed recently that the drag-and-drop is buggy in some cases, it can probably be quite annoying and turn off some users... Added a fix, will just add a test to cover it so it doesn't happen again, not sure what changed to cause this problem as this code hasn't been touched in a long time...

Before: 

https://github.com/mui/mui-toolpad/assets/10789765/4797726b-8fae-468e-a3e1-8ba83934150e

After: 

https://github.com/mui/mui-toolpad/assets/10789765/9ede58a2-aa99-4fe7-b02c-a163f355771c

I've seen some bad bugs with renaming nodes/pages too, will take a look at those next and create an issue if it's not quick.

### DIFF
--- a/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/RenderOverlay.tsx
+++ b/packages/toolpad-app/src/toolpad/AppEditor/PageEditor/RenderPanel/RenderOverlay.tsx
@@ -590,7 +590,8 @@ export default function RenderOverlay({ bridge }: RenderOverlayProps) {
             parentAwareBaseRect = {
               x: isPageChild ? 0 : baseRect.x,
               y: hasPositionGap ? baseRect.y : baseRect.y - parentGap,
-              width: isPageChild && parentRect ? parentRect.width : baseRect.width,
+              width:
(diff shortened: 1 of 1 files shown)


Janpot:
maybe a test?

test/visual/components/index.spec.ts
```diff
@@ -112,20 +114,58 @@ test('showing drag-and-drop previews', async ({ page, argosScreenshot }) => {
   await argosScreenshot('drop-preview-left', screenshotConfig);
 
   await page.mouse.move(...getDropPreviewTopCoordinates(inputBoundingBox!), {
-    steps: 10,
+    steps: 1,
```
apedroferreira, line 117: 10 steps was misleading, it wasn't really screenshotting the final state after the drag.

apedroferreira:
> maybe a test?

Yep I was gonna add it, added it now.

Janpot (approved):
I suppose these screenshot changes are expected?

apedroferreira:
> I suppose these screenshot changes are expected?

Nope they were wrong, these step numbers seem to influence certain screenshots... I fixed it now.

PR #3187, opened 2024-02-09
Fix editable tree item bugs

apedroferreira:
Fix some bugs seen in editable tree items:

- In pages explorer, since the items were changed to show page display names instead of node names, the renaming feature was confusing and could even cause pages to be deleted unintentionally. Fix: Changed the renaming to update display name instead of node name.
- In page hierarchy explorer, somehow clicking near the outside of an item triggered a blur event that cleared the editable input. Fix: I removed a state update to the value of this input that didn't seem like it was doing anything.
- When deleting a page that is not the active page from the explorer, Toolpad changed to an empty app screen.

**Before:**

https://github.com/mui/mui-toolpad/assets/10789765/c9071693-1158-460b-8827-18282908496b

**After:**

https://github.com/mui/mui-toolpad/assets/10789765/ef648dc0-29aa-458e-a6cd-6393fa6d5ab9

### DIFF
--- a/packages/toolpad-app/src/components/EditableTreeItem.tsx
+++ b/packages/toolpad-app/src/components/EditableTreeItem.tsx
@@ -12,7 +12,7 @@ import {
 
 export interface EditableTreeItemProps extends Omit<TreeItemProps, 'label'> {
   labelText?: string;
-  renderLabel?: (children: React.ReactNode, isEditing: boolean) => React.ReactNode;
+  renderLabel?: (children: React.ReactNode) => React.ReactNode;
   suggestedNewItemName?: string;
   isEditing?: boolean;
   onEdit?: (newItemName: string) => void | Promise<void>;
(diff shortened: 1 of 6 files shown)


Janpot:
Interesting. It feels a bit counterintuitive to me. :Like, you have a page "users". And you rename it to "employees". Now you have "pages/users/page.yml" in the file system and "employees" in the editor file tree. Maybe we need to change the editor file tree back to use page name? And then show the display name in a tooltip?

apedroferreira:
> Interesting. It feels a bit counterintuitive to me. :Like, you have a page "users". And you rename it to "employees". Now you have "pages/users/page.yml" in the file system and "employees" in the editor file tree. Maybe we need to change the editor file tree back to use page name? And then show the display name in a tooltip?

Yeah originally when we added display names I thought the pages explorer would stay the same.
Let's change it back to showing the file names then but still show the display names, a tooltip might work!

Janpot:
> Yeah originally when we added display names I thought the pages explorer would stay the same.

👍 I changed that, didn't give it enough thought

> a tooltip might work

👍 another alternative is `pageName (Display Name)`. But maybe that would create too long strings.

apedroferreira:
> > Yeah originally when we added display names I thought the pages explorer would stay the same.
> 
> 👍 I changed that, didn't give it enough thought
> 
> > a tooltip might work
> 
> 👍 another alternative is `pageName (Display Name)`. But maybe that would create too long strings.

Maybe this works? Just not a lot of space, but in most cases probably still enough. https://github.com/mui/mui-toolpad/pull/3187/commits/fbb838eb89e59b39b37dc51d39e722567899224e

https://github.com/mui/mui-toolpad/assets/10789765/60bda73f-5224-4e64-8baa-26c61a782d89

apedroferreira:
Ok, looks like a Tooltip does the job, just not as immediately visible but I think that's fine?

https://github.com/mui/mui-toolpad/assets/10789765/d8d34f2a-b89c-42f3-9e62-2e4e74990df6

bharatkashyap:
> Ok, looks like a Tooltip does the job, just not as immediately visible but I think that's fine?
> 
> 
> 
> https://github.com/mui/mui-toolpad/assets/10789765/d8d34f2a-b89c-42f3-9e62-2e4e74990df6
> 
> 
> 
> 

This feels like the best option to me 👍

Janpot (approved):

PR #3188, opened 2024-02-09
Fix project loading several times

Janpot:
The editor project is loading several times due to suspense. This leads to event handlers created multiple times. This change creates the project in an effect instead.

Fix needed for EXPERIMENTAL_INLINE_CANVAS


apedroferreira (approved):
Cool! I had noticed this.

PR #3193, opened 2024-02-10
Fix test for code components under new editor runtime

Janpot:
The new editor that runs side by side with the app runtime now reloads when a component is added. This is caused by https://github.com/vitejs/vite/issues/12912. Before only the app runtime reloaded, but now the editor and app are one single react tree. This PR teaches the tests to anticipate this reload.

I suppose we can live with this for now, the whole editor became more responsive at the cost of a single reload whenc reating a new component. The ways to fix it are either
* fix https://github.com/vitejs/vite/issues/12912
* change Toolpad custom components to require a barrel file that collects all of them so that Toolpad doesn't have to create a virtual module to import them. We will likely require you to do this in Next.js.

### DIFF
--- a/test/integration/codeComponents/index.spec.ts
+++ b/test/integration/codeComponents/index.spec.ts
@@ -35,7 +35,7 @@ test('can create new custom components', async ({ page, localApp }) => {
 
   await editorModel.goto();
 
-  await editorModel.pageRoot.waitFor();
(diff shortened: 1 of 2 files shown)


apedroferreira (approved):
> the whole editor became more responsive at the cost of a single reload when creating a new component

Yeah seems like a good trade-off.

PR #3194, opened 2024-02-10
Create predictably named test directories

Janpot:
Avoid vite configs changing and triggering dependency optimization.

This replaces the random test directory name with one that's based on the parallel index of the current worker. If amount of workers is set at 1 (which we already do), the tests will always operate on the same folder and the dependency optimization won't invalidate.

### DIFF
--- a/test/playwright/localTest.ts
+++ b/test/playwright/localTest.ts
@@ -10,7 +10,7 @@ import archiver from 'archiver';
 import * as url from 'url';
 import getPort from 'get-port';
 import * as execa from 'execa';
-import { PageScreenshotOptions, test as baseTest } from './test';
+import { PageScreenshotOptions, WorkerInfo, test as baseTest } from './test';
 import { waitForMatch } from '../utils/streams';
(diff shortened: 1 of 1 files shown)


test/playwright/localTest.ts
```diff
@@ -58,9 +58,23 @@ interface LocalServerConfig {
   env?: Record<string, string>;
   base?: string;
 }
+export async function getTemporaryDir() {
```
apedroferreira, line 61: Do we still need this function? I see it's only being used in a custom server test now, not sure if it could be replaced there too?

apedroferreira (approved):

apedroferreira:
Oh, unrelated but could we gitignore these temp test folders?

test/playwright/localTest.ts
Janpot: yes, it's using a random directory that's different from the test directory

Janpot:
> Oh, unrelated but could we gitignore these temporary test folders?

I've thought about it as well, but ideally you remove them instead of ignoring

PR #3203, opened 2024-02-11
[code infra] Remove issue cleanup workflow

Janpot:
This creates friction for me 1 out of 3 issues I create. I can't imagine our community not running into this regularly.


oliviertassinari:
Can we compress the header of move the fields to the bottom? The problem for me without this action is that I can't read the meat of the issue without scrolling down.

PR #3207, opened 2024-02-13
Reroute "open in editor" link in app preview

Janpot:
Make sure the new editor opens when clicking "open in editor" in the app preview.

Also deprecate the editor command for custom servers.

Also in this PR:
- add some margin to the togglebutton prop control
- fix HMR and react refresh when running with the `--dev` command
  - point to typescript sources instead of build output
  - remove non-component exports from Toolpad.tsx


Janpot:
> Does it make sense here to add a screenshot/videos of the changed behaviour to the PR description?

The only thing that changes is the url of the editor. And the speed at which it opens

PR #3208, opened 2024-02-13
Make sure the inline canvas uses the correct font

Janpot:


bharatkashyap (approved):
Looks good to me, not sure if there is any way of optimising this for performance

PR #3221, opened 2024-02-16
Fix unintentional page rows being created

apedroferreira:
This must have been caused a while ago in https://github.com/mui/mui-toolpad/pull/2635 but I hadn't noticed until now, or maybe it only made itself visible now due to the recent fixes:

When dragging elements just outside a page row, new page rows with page rows inside were being created and it could get quite messy. This PR adjusts the logic and seems to be a good fix from what I've tested.

Added a visual test to cover this bug.

**Before:**

https://github.com/mui/mui-toolpad/assets/10789765/9a3abace-20db-45f9-8c59-bc354d1a13be

**After:**

https://github.com/mui/mui-toolpad/assets/10789765/f2071d56-985d-4b7e-8dff-a340edd6bd69


Janpot (approved):
This is great

PR #3235, opened 2024-02-23
[code-infra] Use the @mui/internal-markdown package

michaldudak:
Used the @mui/internal-markdown package from npm instead of the monorepo.
Renamed all the references from @mui/markdown to @mui/internal-markdown.

Uses changes from https://github.com/mui/material-ui/pull/41240


michaldudak:
It looks like there's an issue with pnpm. I submitted a bug report: https://github.com/pnpm/pnpm/issues/7698

Issue #3158, opened 2024-02-02
Add engines field to package.json


Janpot:
### Steps to reproduce

Link to live example: (required)

Steps:
1. use node 14
2. install Toolpad
3.

### Current behavior

No problem

### Expected behavior

Toolpad doesn't run on Node.js 14. Installation should fail. `@mui/toolpad` package.json should contain an `engines` field with `>=18`

### Your environment

<details>
  <summary><code>npx @mui/envinfo</code></summary>

```
  Don't forget to mention which browser you used.
  Output from `npx @mui/envinfo` goes here.
```
</details>

**Search keywords**: engines

Janpot:
That is the package.json of the monorepo root. This is the package.json of `@mui/toolpad`: https://github.com/mui/mui-toolpad/blob/master/packages/toolpad-app/package.json

PR #3254, opened 2024-02-27
[core] add engines field to @mui/toolpad package.json

JerryWu1234:
closes https://github.com/mui/mui-toolpad/issues/3158


packages/toolpad-app/package.json
```diff
@@ -165,5 +165,10 @@
   "typings": "./index.d.ts",
   "publishConfig": {
     "access": "public"
+  },
+  "engines": {
+    "npm": "please-use-yarn",
+    "node": ">=18",
```
Janpot, line 171: Only the `node` field is necessary, we don't care which package manager they use.

Janpot (approved):
Thank you

PR #3262, opened 2024-02-29
Improve errors in Chart component

apedroferreira:
While looking a bit into propagating query errors to component binding results, ran into some semi-related issues in charts.

**Edit**: I moved this back into draft as I thought the fix wouldn't break any tests but it does. This logic is a bit complicated, I wasn't able to find a good solution so far / get to the root of the issue.

**Before:**
- Charts example was showing errors in some charts before loading

https://github.com/mui/mui-toolpad/assets/10789765/4f0d70a5-c86c-4a2a-b04a-7d4009f894f0

**After:**
- Charts load without errors

https://github.com/mui/mui-toolpad/assets/10789765/669c2f7c-8703-4233-b996-e7452520e8f2

### DIFF
--- a/packages/toolpad-app/src/runtime/evalJsBindings.ts
+++ b/packages/toolpad-app/src/runtime/evalJsBindings.ts
@@ -171,12 +171,12 @@ export default function evalJsBindings(
     if (expression) {
       const computed = computationStatuses.get(expression);
       if (computed) {
-        if (computed.result) {
+        if (!computed.result) {
+          throw new Error(`Cycle detected "${scopePath}"`);
+        } else if (!computed.result.error) {
           // From cache
           return computed.result;
         }
-
-        throw new Error(`Cycle detected "${scopePath}"`);
       }
 
       // use null to mark as "computing"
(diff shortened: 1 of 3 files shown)


packages/toolpad-app/src/runtime/evalJsBindings.ts
```diff
@@ -171,12 +171,12 @@ export default function evalJsBindings(
     if (expression) {
       const computed = computationStatuses.get(expression);
       if (computed) {
-        if (computed.result) {
+        if (!computed.result) {
```
apedroferreira, line 174: The reason why some charts were showing errors before loading was because they were using a cached result, but using this result didn't let them recognize that the query was loading, unlike in the charts with uncached results here...
Here i am not caching if there is an error. Maybe there is a better solution?

packages/toolpad-components/src/Chart.tsx
```diff
@@ -171,8 +171,8 @@ function Chart({ data = [], loading, error, sx }: ChartProps) {
 
   return (
     <Box sx={{ ...sx, position: 'relative', height: '100%', width: '100%' }} aria-busy={loading}>
-      {displayError ? <ErrorOverlay error={displayError} /> : null}
-      {loading && !error ? (
+      {displayError && !loading ? <ErrorOverlay error={displayError} /> : null}
```
apedroferreira, line 174: This is just an adjustment, doesn't address any of the reported issues in this PR.

PR #3264, opened 2024-03-01
[DataGrid] fix errors not showing up in the rows area

Janpot:
This used to be an overlay, but we've hijacked the NoRowsOverlay for this, so we need to make the grid visible

adding a vrt in https://app.argos-ci.com/mui/mui-toolpad/builds/2684/79662068


apedroferreira (approved):
Nice!

Issue #3251, opened 2024-02-26
[authentication] Support Firebase auth


Janpot:
### Search keywords

firebase auth

### Latest version

- [X] I have tested the latest version

### Summary

Support firebase auth: https://authjs.dev/reference/adapter/firebase

### Motivation

https://discord.com/channels/1131323012554174485/1131329982669525015/1211027645584646164

apedroferreira:
This was a database adapter that uses a Firestore database for sessions instead of JWTs, not an auth provider, so we decided to not do it for now.
Not sure if that means we should close this issue or leave it on hold.

PR #3266, opened 2024-03-01
Add Firebase adapter

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/3251


apedroferreira:
TODO:

- [ ] Find a good solution in terms of setting environment variables to use Firebase
- [ ] Documentation updates

apedroferreira:
On hold for now as the fact that the Firebase integration of `@auth/core` was a database adapter was a misunderstanding as we don't plan to support that use case for now.

PR #3278, opened 2024-03-06
[website] Add star count fallback

bharatkashyap:
- Seeing this on the landing page Screenshot 2024-03-06 at 2 48 40 PM

- With a rate limiting API error being the reason for it:

```json
{
    "message": "API rate limit exceeded for 49.228.103.47. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)",
    "documentation_url": "https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"    
}
```

- Tweak the Button UI to make the star count optional (empty if an error occurs):
Screenshot 2024-03-07 at 10 26 38 PM

- Prevent a NaN on the star count

- Log to console and do not cache in case of error

Live: https://deploy-preview-3278--mui-toolpad-docs.netlify.app/toolpad/

### DIFF
--- a/docs/src/components/landing/GithubStars.js
+++ b/docs/src/components/landing/GithubStars.js
@@ -15,13 +15,12 @@ export default function GithubStars() {
   const fetchStars = React.useCallback(async () => {
     setFetching(true);
     const response = await fetch('https://api.github.com/repos/mui/mui-toolpad');
-    const data = await response.json();    
+    const data = await response.json();
     setFetching(false);
     if (typeof window !== 'undefined') {
-      localStorage.setItem('mui-toolpad-stars', `${Date.now()}_${data.stargazers_count}`);          
+      localStorage.setItem('mui-toolpad-stars', `${Date.now()}_${data.stargazers_count}`);
     }
     setStars(data.stargazers_count ?? FALLBACK_STAR_COUNT);
-    
   }, []);
 
   React.useEffect(() => {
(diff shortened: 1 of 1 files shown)


Janpot:
Let's start by checking whether the request actually returned a 200.

```tsx
// initialize stars to -1 and display an error state when (!fetching && stars < 0)

setFetching(true);
try {
  const url = 'https://api.github.com/repos/mui/mui-toolpad'
  const response = await fetch(url);
  if (!response.ok) {
    // check for errors
    throw new Error(`HTTP ${response.status} while fetching ${url}`)
  }
  const data = await response.json();
  setStars(data.stargazers_count);
} finally {
  // make sure this always! runs by placing it in a finally block
  setFetching(false);
}
```

bharatkashyap:
> fake

I wouldn't say this is fake necessarily, nor arbitrary. Just outdated. What would you be in favour of showing in place of the star count if the request fails:

- an error message such as this? 
Screenshot 2024-03-07 at 10 09 20 AM

- Or, let it be empty instead with perhaps an error border around it and a label showing the error message beneath the button?
Screenshot 2024-03-07 at 10 18 45 AM

Janpot:
Without a star count, this is just a link to our repository. We say "Star" and an empty star icon. Similar to:

Screenshot 2024-03-07 at 09 53 49

We just print an error to the console, the user doesn't care.

Janpot (approved):

PR #3287, opened 2024-03-11
Fix flaky auth test with console error "Failed to fetch"

apedroferreira:
Caused by a CSRF token request for auth that fails with "Failed to fetch", only happens in headless mode, probably because the requests there happen fast and previous pending auth requests end up making this one fail.
Waiting for network requests to complete seems to fix the issue.
In a real application this will probably never happen and even if it does it's just a console error.


package.json
```diff
@@ -69,7 +69,7 @@
     "format-util": "1.0.5",
     "globby": "14.0.1",
     "jsdom": "24.0.0",
-    "lerna": "8.1.2",
+    "lerna": "7.4.2",
```
apedroferreira, line 72: Using lerna 7 until this gets fixed: https://github.com/lerna/lerna/issues/3964

Janpot (approved):
ah that makes sense, great catch. suggestion: consider waiting for something in the page content if that works as well. It'd be a more user facing signal.

apedroferreira:
> ah that makes sense, great catch. suggestion: consider waiting for something in the page content if that works as well. It'd be a more user facing signal.

I tried as Playwright also doesn't recommend using this method, but the other options didn't work.

PR #3290, opened 2024-03-13
Fix flaky authentication test: Wait for network idle in tests again

apedroferreira:
Last flaky test fix attempt with waiting for elements to render did not work in 100% of cases, seems like we really need to wait for network idle. Reverting back to that fix.


Janpot:
Does this mean we can remove the console ignores again?

apedroferreira:
> Does this mean we can remove the console ignores again?

I saw the Firefox errors come up again so that's why I re-added the console ignores, but maybe this previous approach was fixing them too.
I'll try to remove them again later just to see, I'm not sure if they were actually related to this flaky test (probably not as those errors were already happening before).

PR #3291, opened 2024-03-13
Toolpad core project setup

apedroferreira:
Setting up Toolpad core library.
Very simple for now with just a very basic `AppProvider`, but should be good for a first release and to start building more  features.


Janpot (approved):
Looking good

PR #3301, opened 2024-03-18
[code-infra] Use @mui/docs from npm

michaldudak:
To be updated once https://github.com/mui/material-ui/pull/41206 is merged and published.


Janpot:
Still needs publishing

oliviertassinari:
This change broke all the links, they are missing their leading , it doesn't work

Before https://deploy-preview-3301--mui-toolpad-docs.netlify.app/toolpad/
After: https://deploy-preview-3301--mui-toolpad-docs.netlify.app/toolpad/

Strange

Janpot:
>  `process.env.__NEXT_TRAILING_SLASH` is not true on the server-side, strange. I can only see two options possible:

oh wow, ok. Let's transpile it

PR #3329, opened 2024-03-26
Add custom component example

Janpot:
Add example that demonstrates creating a custom component


examples/custom-component/components/Clock.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react';
+import { createComponent } from '@toolpad/studio/browser';
+
+export interface ClockProps {
+  msg: string;
+}
+
+function Clock({ msg }: ClockProps) {
+  const [time, setTime] = React.useState(() => new Date().toLocaleTimeString());
```
apedroferreira, line 9: Maybe some whitespace between the state, effect and render makes it more readable?

examples/custom-component/components/Clock.tsx
```diff
@@ -0,0 +1,30 @@
...
+  return (
+    <div>
+      {msg}: {time}
+    </div>
+  );
+}
+
+export default createComponent(Clock, {
+  argTypes: {
+    msg: {
```
apedroferreira, line 25: Would `label` be a more descriptive name for this prop?

apedroferreira (approved):
Looks good, just left some minor comments!
Also not sure if we should exemplify more arg types, for example, but I think this is good enough as a basic example.

PR #3336, opened 2024-03-28
Add rename, duplicate and delete for Page Hierarchy

### LINKED ISSUE #3168: Ability to rename nodes
### Summary

We used  to be able to rename nodes, it seems that we can't anymore

### Examples

_No response_

### Motivation

I noticed this in https://github.com/oliviertassinari/mui-olivier-tools/pull/6. I wasn't should about using `select.value` for my binding, I was going to rename the node from `select` to `org`.

*BTW, it was super fast to build, I guess if I had used Pro-code, it would have been as fast, but for a person who doesn't know a lot about frontend but still has an engineer mind, damn 👌.*

**Search keywords**: rename nodes

asif-choudhari:
- Added Menu to the tree items. Menu is hidden until the item is hovered.
- Actions added in the Menu:
   - Rename
   - Duplicate
   - Delete
-  Closes #3168 

https://github.com/mui/mui-toolpad/assets/108984774/2120d8c4-0016-4142-ba7e-18553e8238f5

- [ ] I've read and followed the [contributing guide](https://github.com/mui/mui-toolpad/blob/master/CONTRIBUTING.md#sending-a-pull-request).


packages/toolpad-studio/src/toolpad/AppEditor/HierarchyExplorer/index.tsx
```diff
@@ -1,22 +1,53 @@
...
+  [`
+  & .${treeItemClasses.content}:hover .${classes.treeItemMenuButton},
+      & .${classes.treeItemMenuOpen}
+    `]: {
+    visibility: 'visible',
+  },
+});
+
+interface StyledTreeItemProps extends TreeItemProps, EditableTreeItemProps {
+  labelTextSx?: SxProps;
```
Janpot, line 41: To simplify this a bit, I think you can just supply values for these directly to the `NodeMenu` component. I see we're exporting `CustomTreeItemProps`, I don't think this is necessary, we can keep the component local to this module for now.

Janpot:
Looks great. Had just one suggestion about simplifying the component a bit. I don't think we intend for this `CustomTreeItem` to be reusable. It's exported for no reason.

Janpot (approved):
Thank you!

PR #3343, opened 2024-03-28
Add DashboardLayout component to @toolpad/core (old, PR has been split)

apedroferreira:
Docs/demo: https://deploy-preview-3343--mui-toolpad-docs.netlify.app/toolpad/core/components/dashboard-layout/

(once we have a navigation adapter we can make the demo sidebar links work with local state to provide better examples)
(API docs generation to be added next in separate PR)

---

Ready with component tests, supports `branding` with logo and name for header, and `navigation` with nested items in sidebar. Over time we can expand on this and add more features / improvements, of course.

Please check the types in [packages/toolpad-core/src/AppProvider/AppProvider.tsx](https://github.com/mui/mui-toolpad/pull/3343/files#diff-d72d0f71caf7ef5c3f7a2250be04d58d3dc741a4e944a9c4a9f5bf1f518e6b5f) + component props ([packages/toolpad-core/src/AppProvider/AppProvider.tsx](https://github.com/mui/mui-toolpad/pull/3343/files#diff-d72d0f71caf7ef5c3f7a2250be04d58d3dc741a4e944a9c4a9f5bf1f518e6b5f) and [packages/toolpad-core/src/layout/DashboardLayout/DashboardLayout.tsx](https://github.com/mui/mui-toolpad/pull/3343/files#diff-16914c8076bb1bbac0a0309370f80d3a528d427b0e7dd9c62c4472575f4760d7)) for the chosen API.

For now, `AppProvider` can be imported as a named import from `@toolpad/core/AppProvider` and `DashboardLayout` the same way from `@toolpad/core/DashboardLayout`. They can also be imported directly from `@toolpad/core` along with some maybe useful types such as `Branding` and `Navigation`.

Made it so that `pnpm dev` can be run to build Toolpad core + watch changes, and then a new Next.js project in `playground/toolpad-core-nextjs` can be run to develop and see changes live. We can add other projects/frameworks to the playground folder later.

Added documentation page for the DashboardLayout component with demos. The demos can also be used for development but in some cases a real application is better, of course.

Also added simple contribution instructions for Toolpad Core.


packages/toolpad-core/src/layout/DashboardLayout.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+const DRAWER_WIDTH = 240;
+
+interface DashboardLayoutProps {
+  children: React.ReactNode;
+}
+
+export default function DashboardLayout({ children }: DashboardLayoutProps) {
+  return (
+    <NoSsr>
```
apedroferreira, line 36: Theming isn't working without `NoSsr`, trying to understand these types of issues better there seem to be a lot of them...

packages/toolpad-core/src/client/AppProvider/AppProvider.tsx
```diff
@@ -0,0 +1,38 @@
...
+  navigation,
+}: AppProviderProps) {
+  return (
+    <AppRouterCacheProvider options={{ enableCssLayer: true }}>
+      <ThemeProvider theme={theme}>
+        {/* CssBaseline kickstarts an elegant, consistent, and simple baseline to build upon. */}
+        <CssBaseline />
+        <BrandingContext.Provider value={branding}>
+          <NavigationContext.Provider value={navigation}>
+            {/* @TODO: Can probably remove NoSsr once we have non CSS-in-JS solution. */}
```
Janpot, line 31: MUI should be able to ssr https://mui.com/material-ui/guides/server-rendering/

packages/toolpad-core/src/client/layout/DashboardLayout/DashboardLayout.tsx
```diff
@@ -0,0 +1,198 @@
...
+  List,
+  ListItem,
+  ListItemButton,
+  ListItemIcon,
+  ListItemText,
+  ListSubheader,
+  Stack,
+  Toolbar,
+  Typography,
+} from '@mui/material';
```
oliviertassinari, line 18: flat imports

packages/toolpad-core/src/client/layout/DashboardLayout/DashboardLayout.tsx
```diff
@@ -0,0 +1,198 @@
...
+// Must use named imports or app crashes when the component is imported in a Next.js app
+// eslint-disable-next-line no-restricted-imports
+import { ExpandLess as ExpandLessIcon, ExpandMore as ExpandMoreIcon } from '@mui/icons-material';
```
oliviertassinari, line 21: Flast imports

packages/toolpad-core/src/client/layout/DashboardLayout/DashboardLayout.spec.tsx
```diff
@@ -0,0 +1,135 @@
+/**
+ * @vitest-environment jsdom
+ */
+
+import * as React from 'react';
+import { describe, test, expect, afterEach } from 'vitest';
```
oliviertassinari, line 6: Test files should be `.test.tsx`. `spec.tsx` is for TypeScript tests.

packages/toolpad-core/src/client/layout/DashboardLayout/DashboardLayout.tsx
```diff
@@ -0,0 +1,198 @@
...
+import { styled } from '@mui/system';
+import Link from 'next/link';
```
oliviertassinari, line 23: No coupling with any router

```suggestion
```

Janpot: @apedroferreira perhaps to get a baseline version merged we can start with plain `a` tags? Then progressively improve with a pluggable router adapter and/or Link component?

Janpot: What I've been prototyping with is an [adapter](https://github.com/Janpot/dashboard-framework-prototype/blob/412060d835deffc3bb348e2f77bbf7ae1291d945/src/lib/dash/filter.tsx#L134-L188) that is modeled as a subset of the new [navigation API](https://developer.mozilla.org/en-US/docs/Web/API/Navigation_API). That way hopefully we can some day just delegate directly to `window.navigation` as the default router adapter. It's a challenge though to bridge to this from the different Next.js routers.

apedroferreira: I'm also using `@mui/material-nextjs/v14-appRouter` but the pages router should use `'@mui/material-nextjs/v14-pagesRouter'` and Vite shouldn't need anything... So I guess we need to allow the user to choose a framework somehow and have a component that uses one of these according to that setting.

Janpot: that is what the top level provider is for

apedroferreira: I made this version framework agnostic for now but if I use `<a>` tags I need to use something like `window.location.pathname` to compare routes, and this shows SSR errors in Next.js.
I can work on some kind of framework-aware navigation adapter in this PR already to solve this issue, or we can merge this and I can work on it next.

apedroferreira:
I've added review fixes and some discussed changes such as a different build process using babel and a playground for development.
I've also updated the PR description accordingly.

Not really sure what is happening with the errors in CI, let me know if you do!

packages/create-toolpad-app/src/generateProject.ts
```diff
@@ -139,24 +100,66 @@ export default function generateProject(
...
+
+  const inter = Inter({
     weight: ["300", "400", "500", "700"],
     subsets: ["latin"],
     display: "swap",
   });
-  
-  const theme = createTheme({
+
+  const baseTheme = createTheme({
```
apedroferreira, line 113: It would be ideal to have a design to match this theme to, I kind of made it look like the other template we have but not always sure about the right colors for example.

packages/toolpad-core/babel.config.js
```diff
@@ -0,0 +1,150 @@
+const path = require('path');
```
apedroferreira, line 1: Got these mostly from Base UI, let me know if there's anything important we should be doing differently as I don't fully understand these build scripts.

scripts/build.mjs
```diff
@@ -0,0 +1,128 @@
...
+          description: `Valid bundles: "${validBundles.join('" | "')}"`,
+          type: 'string',
+        })
+        .option('largeFiles', {
+          type: 'boolean',
+          default: false,
+          describe: 'Set to `true` if you know you are transpiling large files.',
+        })
+        .option('out-dir', { default: './build', type: 'string' })
+        .option('watch', { type: 'boolean' })
```
apedroferreira, line 120: I've added this `watch` argument so we can build on changes during development.

Janpot:
I think to prove that the build is working correctly, this PR should show that it can be displayed in a demo component in the docs as well as in  the Toolpad studio application.

packages/toolpad-core/src/types/Navigation.types.ts
```diff
@@ -0,0 +1,13 @@
...
+  icon: React.ReactNode;
+  routes?: (Omit<NavigationRoute, 'routes'> & { path: string })[];
+}
+
+export interface NavigationSection {
+  title: string;
+  routes: NavigationRoute[];
+}
+
+export type Navigation = NavigationSection[];
```
Janpot, line 13: When I designed the interface to represent navigation I tried to optimize for a variety of use case:

flat list of pages

```
- page 1
- page 2
- page 3
```
=>
```tsx
[
  { name: 'page-1', title: 'Page 1' },
  { name: 'page-2', title: 'Page 2' },
  { name: 'page-3', title: 'Page 3' }
]
```
flat list with sub headers

```
header 1
------
- page 1
- page 2
header 2
------
- page 3
```
=>
```tsx
[
  { kind: 'header', title: 'Header 1' },
  { name: 'page-1', title: 'Page 1' },
  { name: 'page-2', title: 'Page 2' },
  { kind: 'header', title: 'Header 2' },
  { name: 'page-3', title: 'Page 3' }
]
```
deeply nested

```
- page 1
- folder 1
  - page 2
  - folder 2
    - folder 3
      - page 3
- folder 3
  - page 4
```
=>
```tsx
[
  { name: 'page-1', title: 'Page 1' },
  { 
    name: 'folder-1', 
    title: 'Folder 1',
    children: [
      { name: 'page-2', title: 'Page 2' },
      {
        name: 'folder-2',
        title: 'Folder 2',
        children: [
          { 
            name: 'folder-3', 
            title: 'Folder 3',
            children: [
              { name: 'page-3', title: 'Page 3' },
            ]
          }
        ]
      }
    ]
  },
  { 
    name: 'folder-3', 
    title: 'Folder 3',
    children: [
      { name: 'page-4', title: 'Page 4' }
    ]
  }
]
```

and any combination of pages, folders, and subheaders.

I believe the interface I came up with is the simplest structure that supports all these use cases. Can you give examples for each of those with the interfaces you've designed?

packages/toolpad-core/src/types/Navigation.types.ts
```diff
@@ -0,0 +1,13 @@
+export interface NavigationRoute {
```
Janpot, line 1: I think I prefer to keep the word "route" out of the nomenclature, this is not a router

packages/toolpad-core/src/types/index.ts
```diff
@@ -0,0 +1,2 @@
+export type * from './Branding.types';
```
Janpot, line 1: Why the .types.ts extension? And is it really necessary to create a separate top level folder for types? I feel like in core/X types are grouped logically with their components.

I feel like it would make most sense to start like MUI core, where all components are exported top level. And we can start organizing them once we're reaching a sizable amount. wdyt?

packages/toolpad-core/src/context/BrandingContext.ts
```diff
@@ -0,0 +1,6 @@
+import * as React from 'react';
```
Janpot, line 1: I think we should try to avoid grouping things by function (e.g. "contexts", "modules", "types", "controllers", "views"). That doesn't scale.
If it's internal only, I'd nest it under the `AppProvider`. If we will export this as an API, I think we can do like base does it, under a folder named as the hook that will eventually consume this context: https://github.com/mui/base-ui/tree/master/packages/mui-base/src/useTabs

packages/toolpad-core/src/types/Navigation.types.ts
apedroferreira: The interface I came up with supports multiple lists with different titles too, and up to one level of nesting. 
It would be easy to make it infinitely nestable, but a second level of nesting already starts looking a bit cramped in the sidebar.

Example with 2 lists and an item with nested items (I think this covers all the cases?):
```
const NAVIGATION: Navigation = [
  {
    title: 'Primary items',
    routes: [
      {
        label: 'Dashboard1',
        path: '/dashboard1',
        icon: <DashboardIcon />,
      },
      {
        label: 'Orders1',
        path: '/dashboard/orders1',
        icon: <ShoppingCartIcon />,
        items: [
          {
            label: 'Nested Orders 1',
            path: '/dashboard/orders1/nested1',
            icon: <ShoppingCartIcon />,
          },
          {
            label: 'Nested Orders 2',
            path: '/dashboard/orders1/nested2',
            icon: <ShoppingCartIcon />,
          },
        ]
      },
    ],
  },
  {
    title: 'Secondary items',
    items: [
      {
        label: 'Dashboard 2',
        path: '/dashboard2',
        icon: <DashboardIcon />,
      },
      {
        label: 'Orders2',
        path: '/dashboard/orders2',
        icon: <ShoppingCartIcon />,
      },
    ],
  },
];
```

I can make the titles optional so that not every list needs a title.
I guess it covers the same cases as your suggestion? I find my format easier to read but i don't mind changing it.
One con of your suggestion would maybe be that the relation between a header and its items is not so clear (it's determined by the order in the array).

Janpot: > I think this covers all the cases?

How does this format support a flat list of pages without sections or subfolders?

> One con of your suggestion would maybe be that the relation between a header and its items is not so clear (it's determined by the order in the array).

The relation is similar to the relation between titles in html and paragraphs. It's not a hierarchical relationship, the headers are to be seen as markup. The flat list keeps it extensible for adding more markup in the future

apedroferreira: > How does this format support a flat list of pages without sections or subfolders?

If we don't want a title I could titles optional and it would be:

```
const NAVIGATION: Navigation = [
  {
    items: [
      {
        label: 'Dashboard1',
        path: '/dashboard1',
        icon: <DashboardIcon />,
      },
      {
        label: 'Orders1',
        path: '/dashboard/orders1',
        icon: <ShoppingCartIcon />,
      },
    ],
  },
];
```

So yeah there's the `items` property that feels a bit extra in this case, could be seen as a con.

> The relation is similar to the relation between titles in html and paragraphs. It's not a hierarchical relationship, the headers are to be seen as markup. The flat list keeps it extensible for adding more markup in the future. (e.g. look at the MUI docs, one could imagine the sidebar ads are another type of markup)

True, it's more extensible, didn't consider that. I don't mind going with it, shouldn't even be a big change!

Janpot: > If we don't want a title I could make title optional so it would be:

I'm not a fan of having the "title" have both the function of overwriting the name, as well as turn on/off certain features.
How do you differentiate between a folder and a subheader then? One being collapsible and the other separating groups of pages on the same level?


🤔 A horizontal separator is probably a better example for another markup element than the ad

apedroferreira: The title does not turn on/off features, it's just the heading for each top-level list of navigation items and shows up above them.
Collapsible items are navigation items with the `items` property themselves.
The dividers show up automatically between navigation lists.
I added a screenshot to the PR where this shows.

In the header navigation I assume we could use the same system but no titles.

Anyway let's go with your approach.

Janpot: > Collapsible items are navigation items with the items property themselves.

I'm still not fully getting it. How do you differentiate between a section and a folder? Aren't both of them navigation items with the `items` property?

e.g. subheader nested in a folder, or vice-versa:

```
- page1
subheader1
- page2
-- divider --
- page3
- folder1
  - page4
  - subheader2
  - folder2
    - page5
    -- divider --
    - page6
  - page7
```

> The dividers show up automatically between navigation lists.

Ok, but that doesn't allow for dividers between any two arbitrary items.

apedroferreira: > I'm still not fully getting it. How do you differentiate between a section and a folder? Aren't both of them navigation items with the items property?

In my first approach it's only possible to add headings at the top-level (they're just the `title`), so not inside collapsible items, so you can't have sections inside folders...
So I guess that your approach in more versatile in that way. I'm trying it out!

Janpot: Oh I see, your top-level is different from nested levels

apedroferreira: I've also just realized that using your suggested approach for subheaders can be tricky with the MUI List component https://mui.com/material-ui/react-list/ since it takes a `subheader` as a prop for the list.
We can probably use subheaders without the prop, we'll just lose the automatic sticky headers.

Janpot: > since it takes a subheader as a prop for the list.

The content of this prop has to be wrapped in a `<ListSubHeader>` component, and these can also be used inside of the list, just like `<Divider>`. See https://mui.com/material-ui/react-list/#sticky-subheader.

Not 100% sure what would be the best approach. We could declare a navigation item with a title and children but no declared path segment as a section. OTOH a flat list could be more versatile towards the future.

docs/data/toolpad/core/components/dashboard-layout/dashboard-layout.md
```diff
@@ -22,9 +22,9 @@ The `navigation` prop in the `AppProvider` allows for setting any type of naviga
 
 {{"demo": "DashboardLayoutNavigation.js", "height": 640, "iframe": true}}
 
-## API
+<!-- ## API
```
apedroferreira, line 25: Let's leave API docs generation to another PR as it doesn't seem super simple and we can discuss how to do it.

apedroferreira:
Not sure about the remaining CI errors - looks like we may need bigger instances for Circle CI if it's running out of memory?
And not sure why the Netlify build fails, `pnpm docs:build` works locally, not sure if it's some cache that can be cleared or something.

.circleci/config.yml
```diff
@@ -62,7 +62,7 @@ commands:
             pnpm --version
       - run:
           name: Install js dependencies
-          command: pnpm install
+          command: pnpm install --ignore-scripts
```
apedroferreira, line 65: This seems to fix the out of memory issues in CI but might be making some integrations tests fail, looking more into it.

Janpot: I tried a few updates to this branch, cleaned up the dependencies a bit and removed the playground folder. That branch seems to pass CI https://github.com/apedroferreira/mui-toolpad/pull/1

apedroferreira: I see, yeah I guess some of the babel packages I copied over might not be needed...
As for the playground, if we get rid of it do you have any alternative in mind? When I tried to make Toolpad Core work in development in projects outside the monorepo I always had some weird issues.

Janpot: We can try figuring out which part of the playground package.json is the culprit by bissecting the dependencies and  see what happens in CI. Timeconsuming, but probably the only way

apedroferreira: It was probably Next.js having a different version than the docs, should be good now.

apedroferreira:
Submitting for review again, CI is finally fixed.

.npmrc
```diff
@@ -1,5 +1,5 @@
 # Enforce pnpm
 
-engine-strict = true
-auto-install-peers = true
+engine-strict=true
+auto-install-peers=true
 strict-peer-dependencies=false
```
oliviertassinari, line 5: It seems to me that a space is supposed to be present from https://docs.npmjs.com/cli/v9/configuring-npm/npmrc.

.eslintignore
```diff
@@ -4,6 +4,7 @@ pnpm-lock.yaml
 /docs/export
 /docs/schemas
 /packages/**/dist
+/packages/**/build
```
oliviertassinari, line 7: Could be a global rule?

```suggestion
build
```

docs/data/toolpad/core/components/dashboard-layout/DashboardLayoutBranding.js
```diff
@@ -0,0 +1,43 @@
...
+    icon: <ShoppingCartIcon />,
+  },
+];
+
+const BRANDING = {
+  logo: <img src="https://mui.com/static/logo.png" alt="MUI logo" />,
+  title: 'MUI',
+};
+
+export default function DashboardLayoutBranding() {
```
oliviertassinari, line 26: Missing TypeScript version?

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -0,0 +1,233 @@
...
+import { BrandingContext, Navigation, NavigationContext, NavigationPageItem } from '../AppProvider';
+
+const IS_CLIENT = typeof window !== 'undefined';
+
+const DEFAULT_DRAWER_WIDTH = 320;
+
+const LogoContainer = styled('div')({
+  position: 'relative',
+  height: 40,
+  img: {
```
oliviertassinari, line 30: Does this mean?

```suggestion
  '& img': {
```

docs/data/toolpad/core/components/dashboard-layout/DashboardLayoutBasic.js
```diff
@@ -0,0 +1,70 @@
...
+import Box from '@mui/material/Box';
+import Typography from '@mui/material/Typography';
+import DashboardIcon from '@mui/icons-material/Dashboard';
+import ShoppingCartIcon from '@mui/icons-material/ShoppingCart';
+import BarChartIcon from '@mui/icons-material/BarChart';
+import DescriptionIcon from '@mui/icons-material/Description';
+import LayersIcon from '@mui/icons-material/Layers';
+import { AppProvider } from '@toolpad/core/AppProvider';
+import { DashboardLayout } from '@toolpad/core/DashboardLayout';
+import theme from '../theme';
```
oliviertassinari, line 11: It don't think we should have relative imports for demos, it will break once exported to codesandbox.

oliviertassinari:
It feels like some of the changes in this PR could be extracted to another one that isn't about the DashboardLayout component so that other people can work in parallel on other building blocks.

packages/create-toolpad-app/src/generateProject.ts
```diff
@@ -9,84 +9,44 @@ export default function generateProject(
...
+  import type { Navigation } from "@toolpad/core";
   import theme from '../theme';
+
+  const NAVIGATION: Navigation = [
+    {
+      kind: 'header',
+      title: 'Main items',
+    },
+    {
+      path: '/page',
```
Janpot, line 23: I believe a good guiding principle in designing this interface would be to make sure that the individual items are easy to move around, or copy. Just the same way how folders on a file system are easy to move around. The disign principle makes sense since very often these navigation structures will mimic a folder structure in the project.

By making these `path` parameters absolute, moving them around will be more difficult. i.e. moving a nested item under one top-level to an other will require you to rewrite all paths of nested items underneath.

Instead I'd considder letting navigation items only define the segment for which they are responsible, and build the full path by chaining them recursively.

There may be use-cases where users may want to target a completely different path than the one in the hierarchy, but this should be the special case. I wouldn't tackle this until a user makes a compelling case on why this is strictly necessary.

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -0,0 +1,233 @@
...
+}: DashboardSidebarSubNavigationProps) {
+  // Rerender once to update with client-side navigation
+  const [hasMounted, setHasMounted] = React.useState(false);
+
+  const [expandedSidebarItemIds, setExpandedSidebarItemIds] = React.useState<string[]>([]);
+
+  React.useEffect(() => {
+    setHasMounted(true);
+
+    if (IS_CLIENT) {
```
Janpot, line 52: Why is this needed? Effects don't run serverside.

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -0,0 +1,233 @@
...
+        return (
+          <React.Fragment key={navigationItemId}>
+            {navigationItem.path ? (
+              <a href={navigationItem.path} style={{ color: 'inherit', textDecoration: 'none' }}>
+                {listItem}
+              </a>
+            ) : (
+              listItem
+            )}
+            {IS_CLIENT && hasMounted && navigationItem.children ? (
```
Janpot, line 142: `IS_CLIENT` can never be `false` if `hasMounted` is `true`.


```suggestion
            {hasMounted && navigationItem.children ? (
```

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -0,0 +1,233 @@
...
+  subNavigation: Navigation;
+  depth?: number;
+}
+
+function DashboardSidebarSubNavigation({
+  subNavigation,
+  depth = 0,
+}: DashboardSidebarSubNavigationProps) {
+  // Rerender once to update with client-side navigation
+  const [hasMounted, setHasMounted] = React.useState(false);
```
Janpot, line 45: I'm not a big fan of this method. This will prevent rendering certain elements on the first render, even if it's never being hydrated. Have you tried the `useIsSsr` hook that's available in `@mui/utils`?

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -0,0 +1,233 @@
...
+        const nestedNavigationCollapseIcon = isNestedNavigationExpanded ? (
+          <ExpandLessIcon />
+        ) : (
+          <ExpandMoreIcon />
+        );
+
+        const listItem = (
+          <ListItem>
+            <ListItemButton
+              selected={IS_CLIENT && hasMounted && window.location.pathname === navigationItem.path}
```
Janpot, line 123: Same as before.

```suggestion
              selected={hasMounted && window.location.pathname === navigationItem.path}
```

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -0,0 +1,233 @@
...
+          </React.Fragment>
+        );
+      })}
+    </List>
+  );
+}
+
+interface DashboardLayoutProps {
+  children: React.ReactNode;
+  sidebarWidth?: number;
```
Janpot, line 159: Does this really have to be configurable? And if it were configurable, wouldn't it make sense to be responsive? What if we just come up with good defaults based on the breakpoint and wait for making it configurable on user feedback?

Does the drawer collapse on narrower screens?

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -0,0 +1,233 @@
...
+  hideTitle?: boolean;
+}
+
+export function DashboardLayout({
+  children,
+  sidebarWidth = DEFAULT_DRAWER_WIDTH,
+  hideTitle = false,
```
Janpot, line 166: I feel like at this point it's hard to say whether this behavior should be enabled through a specific prop, or we should let it be solved by slots API.
Perhaps at this stage we should concentrate more on coming up with the best possible default experience, for all breakpoints and think of customization once we've collect a few use cases.

packages/toolpad-core/src/AppProvider/AppProvider.tsx
```diff
@@ -1,23 +1,52 @@
...
+
+export type Navigation = NavigationItem[];
+
+export const BrandingContext = React.createContext<Branding | null>(null);
+
+export const NavigationContext = React.createContext<Navigation>([]);
+
 interface AppProviderProps {
   children: React.ReactNode;
   theme: Theme;
```
Janpot, line 38: Does it have to be mandatory? We could provide a default.
```suggestion
  theme?: Theme;
```

packages/toolpad-core/src/AppProvider/AppProvider.tsx
```diff
@@ -1,23 +1,52 @@
...
+  branding?: Branding | null;
+  navigation: Navigation;
```
Janpot, line 40: Default navigation can be `[]`? i.e. I should be allowed to drop in a dashboard layout and concentrate on the refining the branding before adding pages. Or vice versa.

```suggestion
  navigation?: Navigation;
```

packages/toolpad-core/.npmrc
```diff
@@ -0,0 +1 @@
+enable-pre-post-scripts=true
```
Janpot, line 1: Do we need this? I see no pre or post scripts in the package.json. And we probably shouldn't rely on them anyway if the package manager doesn't support it out of the box. We'll do task orchestration with `nx` if we need to. X/core will be using it vor that purpose as well in the future.

packages/toolpad-core/babel.config.js
```diff
@@ -0,0 +1,144 @@
...
+
+const productionPlugins = [
+  ['babel-plugin-react-remove-properties', { properties: ['data-mui-test'] }],
+];
+
+module.exports = function getBabelConfig(api) {
+  const useESModules = api.env(['regressions', 'stable', 'rollup']);
+
+  const defaultAlias = {
+    '@toolpad/core': resolveAliasPath('./packages/toolpad-core/src'),
```
Janpot, line 19: No aliasing in babel. it's not the right tool for that. I see we're using `vitest` for testing, if you need aliasing, do it in `vitest`.

You can also remove the error codes macro. We'll add this optimization later if we need it. We're going to figure out a more portable method to support it.

In fact `babel.config.js` is only for top-level configuration. If possible, we should remove this file altogether.

packages/toolpad-core/package.json
```diff
@@ -1,75 +1,90 @@
...
   },
+  "sideEffects": false,
   "publishConfig": {
     "access": "public"
   },
   "engines": {
     "node": ">=18"
-  }
+  },
+  "files": [
```
Janpot, line 84: Aren't we building into a `build` folder? Like core?

package.json
```diff
@@ -14,7 +14,7 @@
     "markdownlint": "markdownlint-cli2 \"**/*.md\"",
     "prettier": "pretty-quick --ignore-path .eslintignore",
     "prettier:all": "prettier --write . --ignore-path .eslintignore",
-    "dev": "dotenv cross-env FORCE_COLOR=1 lerna -- run dev --stream --parallel --ignore docs",
+    "dev": "dotenv cross-env FORCE_COLOR=1 lerna -- run dev --stream --parallel --ignore docs --ignore toolpad-core-nextjs",
```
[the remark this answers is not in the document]
Janpot: We will move towards `nx` for orchestration over time

Janpot:
> It feels like some of the changes in this PR could be extracted to another one that isn't about the DashboardLayout component so that other people can work in parallel on other building blocks. 

Agreed. The goal in the PR should be to set up the bare minimum of the basic layout component and provider so that we can test/prove the project setup works. I'd strip the configuration options from these components and concentrate on nailing the project setup so we can get this trough.
We can then concentrate on API subsequently.

apedroferreira:
> > It feels like some of the changes in this PR could be extracted to another one that isn't about the DashboardLayout component so that other people can work in parallel on other building blocks.
> 
> Agreed. We could make the goal of the PR to set up the bare minimum of the basic layout component and provider so that we can test/prove the project setup works. Perhaps we should strip the configuration options from these components and concentrate on nailing the project setup so we can get this trough? We can then concentrate on API in follow up PR.

The current state just evolved naturally from review feedback + improving the build to fix some bugs with the component itself. I can probably split project setup from the component at this point though, at least, will see what else I can split.

packages/toolpad-core/package.json
apedroferreira: The base repo uses a `build` folder but this is how the files are listed there.

docs/data/toolpad/core/components/dashboard-layout/DashboardLayoutBasic.js
apedroferreira: Hm ok just wanted to reuse the same theme in all demos, will see how else it can be done.

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
apedroferreira: It was just a nice easy to add option, didn't look into responsiveness though.

> Does the drawer collapse on narrower screens?

No.

packages/toolpad-core/src/AppProvider/AppProvider.tsx
apedroferreira: If you use `create-toolpad-app` it comes with a default theme, but it's not built into the library.
We could add it as a default theme though and not have it set by `create-toolpad-app`, i guess that sounds good.

Janpot: We can have it created by `create-toolpad-app`, but that doesn't mean this property has to be mandatory?

apedroferreira: First options I can think of:

1. Users can provide separate themes for light and dark mode;
2. Theme can take a function with `mode: PaletteMode` as an argument.

I was trying to keep it simple for now thinking we can still make breaking changes afterwards.
I guess option 2 should be more flexible as in option 1 it would be harder to change parts of the theme conditionally based on the mode. But probably we should also allow just passing in a theme in a straightforward way.

packages/toolpad-core/.npmrc
apedroferreira: Changing buils script to ""build": "pnpm prebuild && pnpm build:node && pnpm build:stable && pnpm build:types && pnpm build:copy-files" if that's better.

packages/create-toolpad-app/src/generateProject.ts
apedroferreira: I've done this but before it was easy to define if a collapsible navigation item links to a page or not itself by having a `path` property or not.
If we use the `path` of a collapsible item as the base path for the nested items inside it, we need to have the `path` there whether the collapsible item is a link or not, to serve as the base path for the nested items.

So I'm trying to find the best solution for this... Maybe even collapsible items should never be links themselves, not sure, or we could add a new property just for this but it may be overkill?

apedroferreira:
Completed/split in other PRs.

PR #3386, opened 2024-04-12
[release] v0.1.54

bharatkashyap:


CHANGELOG.md
```diff
@@ -1,5 +1,88 @@
 # Changelog
 
+## v0.1.54
+
+<!-- generated comparing v0.1.53..master -->
+
+_Apr 12, 2024_
```
apedroferreira, line 7: Sorry for the late review, but looks like this is doubled in the changelog :D

apedroferreira (commented):
Saw an issue in the changelog! We can fix it now, there should be no problem.

CHANGELOG.md
bharatkashyap: Ah damn! Thanks, I'll fix it now

PR #3399, opened 2024-04-14
[core] Remove engine please-use-pnpm

oliviertassinari:
This seems to duplicate with:
https://github.com/mui/mui-toolpad/blob/d21b49eeae89228cccf316aa578c13f4fe24208a/package.json#L12 but only catch npm, not yarn, seems better to remove this, it will match with the other repositories too.


Janpot (approved):
yes, this is obsolete now

PR #3403, opened 2024-04-15
Revert monorepo update

Janpot:
This regressed some of the icons:
Screenshot 2024-04-15 at 12 01 53


Janpot:
Tagging code infra as this is something to look into whether it applies to X/base

Issue #3410, opened 2024-04-16
Environment variables are not propagated to HTTP query parameters


Janpot:
### Search keywords

query parameter variable

### Latest version

- [X] I have tested the latest version

### Steps to reproduce

This happens when an environment variable is set, not through the env file. It doesn't get populated in the parameters of HTTP query when bound.

Steps:
1. start Toolpad with `FOO=bar toolpad-studio dev`
2. in a page, create a query
3. add a parameter "foo" to the query
4. bind "foo" to the variable FOO
5. use `parameters.foo` somewhere in the query
6. try running it

You can also open the reproduction and try directly:
1. check out https://github.com/Janpot/repro-env/
2. `pnpm install && pnpm dev`
3. open "query"
4. press "run"
5. check the devtools url

### Current behavior

shows the url with `?hello=undefined`

https://github.com/mui/mui-toolpad/assets/2109932/ce5c0143-af85-4d4e-9870-8aaadc72aa4d

### Expected behavior

shows the url with `?hello=bar` from the environment

### Context

https://discord.com/channels/1131323012554174485/1227982224662728724

### Your environment

<details>
  <summary><code>npx @mui/envinfo</code></summary>

```
  Don't forget to mention which browser you used.
  Output from `npx @mui/envinfo` goes here.
```
</details>

PR #3413, opened 2024-04-17
Fix environment binding preview

Janpot:
Fixes https://github.com/mui/mui-toolpad/issues/3410

The env var file is only a subset of the available vars.

* Use the env var file keys for the warning message
* Use the actual env for evaluation of preview
* Show a preview of the actual value
* Fix issues with the tab switching to expression when you empty the env value

Screenshot 2024-04-17 at 18 40 39

### DIFF
--- a/packages/toolpad-studio/src/server/EnvManager.ts
+++ b/packages/toolpad-studio/src/server/EnvManager.ts
@@ -96,8 +96,8 @@ export default class EnvManager {
     this.watcher.on('change', handleChange);
   }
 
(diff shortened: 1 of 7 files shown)


bharatkashyap:
- Does it make sense to have a more standard warning?

Screenshot 2024-04-18 at 3 00 50 PM

bharatkashyap (approved):

Janpot:
> Does it make sense to have a more standard warning?

Maybe, but probably not within the scope of this PR. The message was already there, I just fixed some bugs around the binding editor for env vars.

PR #3416, opened 2024-04-18
[docs] Improved first-app documentation.

prakhargupta1:
Docs link: 

I have kept the same info about building an interactive dog app, as I think it is easy to follow. Highlighted actions in the images and did some minor edits.


bharatkashyap:
Not sure why Vale isn't flagging this, but let's remove all the "We"s and any other first-person plurals in the guide: https://github.com/mui/material-ui/blob/next/.vale.ini#L18

bharatkashyap (changes requested):
- I think the point of improving this is to have a first app which demonstrates more of the features and power of Toolpad, and also takes the users through the happy path.

- I would: 
    - choose a more complex, real-world use case (Perhaps replace one of the other examples with this?)
    - make sure the data used is parseable without transforming
    - make sure the column headers are readable

- Also, let's remove all the "We"s and any other first-person plurals in the guide: https://github.com/mui/material-ui/blob/next/.vale.ini#L18

prakhargupta1:
The point of improving it is to ensure users reach the end goal faster by helping them quickly see the desired actions. IMO, a simple app that touches on the basic and most used features of Toolpad should be good.

bharatkashyap:
> The point of improving it is to ensure users reach the end goal faster by helping them quickly see the desired actions. IMO, a simple app that touches on the basic and most used features of Toolpad should be good.

I think the app should probably try and at-least attempt to solve a real world use case. Would someone look at the dog app and be like, okay I can use this for my problems? We need to be able to demonstrate that investing the effort of downloading and running Toolpad locally is worth it, and for that the first app is our pitch, given that this is the page linked on from the landing page. Why not replace this with a PDF generator, or a basic admin app, or a simple dashboard?

PR #3424, opened 2024-04-19
[core] Update monorepo

oliviertassinari:
Propagate https://github.com/mui/material-ui/pull/41901


oliviertassinari:
It fails because `@mui/docs` wasn't released with a version that is compatible with @mui/monorepo yet.

package.json
```diff
@@ -39,7 +39,7 @@
   },
   "devDependencies": {
     "@argos-ci/core": "1.5.5",
-    "@mui/monorepo": "github:mui/material-ui#f3cb496c999acbc8f19533e38df20be12e56d059",
+    "@mui/monorepo": "github:mui/material-ui#bbfb2cd742ed3aca3d0c8f6dc3e890ed4fc1d7f2",
```
oliviertassinari, line 42: Should be 

```suggestion
    "@mui/monorepo": "github:mui/material-ui#next",
```

here, otherwise, renovate will try to use the default branch, master

Janpot:
> I still struggle to connect the dots to why npm dependency > git dependency for docs infra.

The problems you encounter are caused by moving docs to the published variant before fully removing  the monorepo dependency. Now you end up with two code dependencies that have some shared assumptions about which version is used, but which are imported through two different release cycles.

My [initial preference](https://github.com/mui/material-ui/pull/41206#issuecomment-2000043424) was to keep aliasing it from the monorepo dependency until we could fully replace, but the team is eager to move forward with publishing.

Similar problems will appear when introducing a different release process for internal and external packages. Which I also seem to be unable to convince the team is a bad idea.

In general it's simple: a hybrid of monorepo+aliasing and published packages will cause more problems than either of those on their own. I am going to be clear about the following, we're going to have a worse system until:
1. we remove the monorepo dependency completely
2. we adopt the same release cycle for internal packages and external ones

I also [agree](https://github.com/mui/material-ui/pull/40889#issuecomment-1952931951) that we should have tried a git dependency first, at least as a stopgap solution.

Janpot (approved):

oliviertassinari:
Arf, something is off with the latest release https://github.com/mui/mui-x/pull/12861#issuecomment-2067785682

package.json
oliviertassinari: Actually, my point is not true, it looks like Renovate uses the GitHub repository default branch. Let's do the opposite, it looks clearer and avoid the need to update the branch as it changes: https://github.com/mui/mui-x/commit/906ad7d286e7fa0663047cc90543d287e6da41ad.

Janpot: Yeah, this has actually worked super well for us so far.

Janpot (changes requested):
Revoking my review, I didn't look at the CI

Janpot (approved):

Janpot (approved):
monorepo is up to date now, so I just kept the change to the config.
Added a few more comments around @mui/docs as well

oliviertassinari:
It looks great, thanks

PR #3425, opened 2024-04-19
[core] Use the root dependency

oliviertassinari:
It seems that we don't need this, it duplicates the root, and makes it possible to have a separate version of the mono repo (not helping to move the codebase all at once).


Janpot (approved):
This was a hack that solved some issues we had with dependency hoisting under `yarn`. If it works now, I'm fine with it.

oliviertassinari:
Right, it seems that this duplication has been here since the beginning: #542. The CI is green, looks like we are good.

PR #3441, opened 2024-04-21
[code-infra] Closer sync with eslint config of codebase

oliviertassinari:
A step to simplify the eslint config in the codebase.


.eslintrc.js
```diff
@@ -94,15 +93,21 @@ module.exports = {
...
+        'material-ui/no-hardcoded-labels': 'off',
+      },
+    },
     {
       files: ['examples/**/*'],
       rules: {
         // We use it for demonstration purposes
         'no-console': 'off',
-        // Personal preference
-        'no-underscore-dangle': 'off',
```
oliviertassinari, line 105: Seems marginally used

.eslintrc.js
```diff
@@ -25,37 +50,11 @@ module.exports = {
...
-        patterns: [
-          {
-            group: ['lodash-es/*'],
-            message: 'Use `import { debounce } from "lodash-es";` instead.',
-          },
-        ],
-      },
-    ],
+    '@typescript-eslint/return-await': 'off',
+    'no-restricted-imports': ['error', noRestrictedImports],
```
oliviertassinari, line 57: moved to a variable, `noRestrictedImports`, to be able to extend it in specific paths.

.eslintrc.js
```diff
@@ -94,15 +93,21 @@ module.exports = {
         ],
       },
     ],
+    'material-ui/no-hardcoded-labels': 'off', // We are not really translating the docs/website anymore
   },
   overrides: [
+    ...baseline.overrides,
```
oliviertassinari, line 99: The main difference

examples/with-prisma-data-provider/toolpad/prisma.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-underscore-dangle */
```
Janpot, line 1: In general we don't add eslint disable comments in examples, but turn off the rule in an eslint override.

Reason is that eslint is not used in the example, so it shouldn't contain eslint comments after I download it.

docs/data/toolpad/studio/components/list/List.js
```diff
@@ -1,6 +1,8 @@
+/* eslint-disable filenames/match-exported */
 import * as React from 'react';
 import { List } from '@toolpad/studio-components';
 
+// TODO fix name
```
Janpot, line 5: What's wrong with the name?

.eslintrc.js
Janpot: There is a reason. examples don't contain eslint. So the code shouldn't contain eslint comments.

Janpot:
Did you see that this eslint is in the process of [deprecating](https://eslint.org/blog/2022/08/new-config-system-part-2/) its config format?

.eslintrc.js
oliviertassinari: This makes sense. Should we move this rule to the share eslint config then. It feels like we want this for all /examples folders through MUI codebase.

docs/data/toolpad/studio/components/list/List.js
oliviertassinari: filenames/match-exported rule wants the component to be called List but List is already in the scope.

I left the TODO to keep this a small time investment, my main objective was for new code to be less likely to bend the shared eslint rules. For sure, we can fix it in this PR too.

oliviertassinari:
> The new format seems considerably easier to understand in terms of extending lint configs.

I briefly looked at it, but saw we are on ESLint v8 not v9 so ignored. I hated these files so much: https://github.com/mui/material-ui/blob/next/packages/api-docs-builder/.eslintrc.js. I'm happy it won't be possible anymore.

.eslintrc.js
Janpot: 👍 I'm good with that as well

docs/data/toolpad/studio/components/list/List.js
Janpot: I updated the naming everywhere.

Janpot (approved):
Fixed the naming and brought back the rule for examples. With clearer comment. Want to merge this as I see PRs coming in that would fail the new naming rules.

PR #3458, opened 2024-04-26
Fix performance regression

Janpot:
After we removed the second build process, the canvas content and the editor started running in the same React tree. The whole editor rendering now affects the responsiveness of the input components. 

This PR keeps the input state local and updates the editor state in a transition to make the rendering of these state updates through the editor interruptible.

Also removing some dead code

### DIFF
--- a/packages/toolpad-studio/src/toolpad/AppEditor/PageEditor/NodeAttributeEditor.tsx
+++ b/packages/toolpad-studio/src/toolpad/AppEditor/PageEditor/NodeAttributeEditor.tsx
@@ -43,16 +43,26 @@ export default function NodeAttributeEditor<P extends object>({
 }: NodeAttributeEditorProps<P>) {
   const domApi = useDomApi();
 
+  const initialPropValue = (node as any)[namespace]?.[name];
+  const [propValue, setPropValue] = React.useState<BindableAttrValue<unknown> | null>(
+    initialPropValue,
+  );
+
+  React.useEffect(() => {
+    setPropValue(initialPropValue);
+  }, [initialPropValue]);
+
   const handlePropChange = React.useCallback(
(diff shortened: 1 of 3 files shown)


apedroferreira:
> but I believe there will still be places in the UI where an input depends on it synchronously updating the state. We should eliminate this assumption.

Yeah, we have some controlled inputs in the right-side sidebar that are reading the state to define their value and so they don't perform as well as they could, it's definitely something we should improve.

packages/toolpad-studio/src/toolpad/AppEditor/PageEditor/NodeAttributeEditor.tsx
```diff
@@ -43,17 +43,27 @@ export default function NodeAttributeEditor<P extends object>({
...
+    setPropValue(initialPropValue);
+  }, [initialPropValue]);
+
   const handlePropChange = React.useCallback(
     (newValue: BindableAttrValue<unknown> | null) => {
-      domApi.update((draft) =>
-        appDom.setNodeNamespacedProp(draft, node, namespace as any, name, newValue),
-      );
+      setPropValue(newValue);
+      React.startTransition(() => {
```
apedroferreira, line 58: Nice use of this transition API, we could probably use it more in this type of updates.

apedroferreira (approved):

packages/toolpad-studio/src/toolpad/AppEditor/PageEditor/NodeAttributeEditor.tsx
Janpot: Ideally we use it just once in the `update` method and eliminate all synchronous coupling between app state and its update method

PR #3526, opened 2024-05-06
Fix/default editable grid

### LINKED ISSUE #3524: [studio] Cell shouldn't be editable by default
### Search keywords

-

### Latest version

### Steps to reproduce

1. Open https://tools-public.mui.com/prod/pages/baseUiNpmKpis
2. 🤷‍♂️

https://github.com/mui/mui-toolpad/assets/3165635/fcccf85c-6035-40d3-a2ad-6401ca0a900e

### Current behavior

_No response_

### Expected behavior

_No response_

### Context

- Retool does this right:

SCR-20240505-qafn

### Your environment

<details>
  <summary><code>npx @mui/envinfo</code></summary>

```
  Don't forget to mention which browser you used.
  Output from `npx @mui/envinfo` goes here.
```
</details>

b4s36t4:
Closes #3524


packages/toolpad-studio-components/src/DataGrid.tsx
```diff
@@ -479,11 +479,11 @@ function getNarrowedColType(type?: string): GridColType | undefined {
   return (type && type in DEFAULT_COLUMN_TYPES ? type : undefined) as GridColType | undefined;
 }
 
-export function parseColumns(columns: SerializableGridColumns): GridColDef[] {
+export function parseColumns(columns: SerializableGridColumns, isEditable?: boolean): GridColDef[] {
```
Janpot, line 482: `parseColumns` acts as a conversion between "serializable columns" and "X grid columns". It's being reused across the codebase for this purpose.
Let's instead solve this downstream. Where we iterate a second time to create the rendered coluns:
https://github.com/mui/mui-toolpad/blob/02cc9a97b76098d51651eb48cf10c12ccdc7c8fe/packages/toolpad-studio-components/src/DataGrid.tsx#L1264-L1279

b4s36t4: @Janpot , say we're having `id` column it shouldn't be editable at all right, should I move the `isIdColumn` logic to this function? 

If we update the `editable` here `id` column may probably get override right?

Janpot: How about only setting `isEditable` if it isn't already set? And remove `isEditable: true` from the `baseColumn` in `parseColumns`.

b4s36t4: Yea, that's a good approach. Will update accordingly.

b4s36t4:
@Janpot I have updated the changes you suggested, can you please verify.

packages/toolpad-studio-components/src/DataGrid.tsx
```diff
@@ -1262,7 +1270,18 @@ const DataGridComponent = React.forwardRef(function DataGridComponent(
...
 
   const renderedColumns = React.useMemo<GridColDef[]>(() => {
-    const result = [...columns];
+    const mapEditableToColumns = columns.map((column) => {
+      // Don't touch the existing columns with `editable` value, example: `id` column
+      if (column.editable || column.editable === false) {
+        return column;
+      }
+
+      column.editable = isRowUpdateModelAvailable;
```
Janpot, line 1279: I think we should avoid mutating items in `.map`. Perhaps something like the following would work?
```tsx
return column.editable === undefined ? { ...column, editable: isRowUpdateModelAvailable } : column;
```

Janpot:
> @Janpot I have updated the changes you suggested, can you please verify.

Can you look into why the tests fail?

PR #3536, opened 2024-05-09
Generate API docs for Toolpad Core

apedroferreira:
Generate API docs for Toolpad Core - first the proptypes must be generated with `pnpm proptypes` and then API docs built with `pnpm docs:build:api`.

Restructured the documentation a bit kind of just to be more like the other MUI libraries I was taking the scripts from and what these scripts expect, but it's all for discussion of course!

API links from component pages are broken - if this PR looks good I think we can fix it here in the MUI monorepo: https://github.com/mui/material-ui/blob/next/packages/markdown/prepareMarkdown.js

**API Page:** https://deploy-preview-3536--mui-toolpad-docs.netlify.app/toolpad/core/api/app-provider/


.npmrc
```diff
@@ -1,5 +1,5 @@
 # Enforce pnpm
 
-engine-strict = true
-auto-install-peers = true
+engine-strict=true
+auto-install-peers=true
 strict-peer-dependencies=false
```
oliviertassinari, line 5: It seems to me that a space is supposed to be present from https://docs.npmjs.com/cli/v9/configuring-npm/npmrc.

apedroferreira: Oh ok I was changing these for consistency as I had seen different spacing in different places, will make all of them include the spacing then.

packages/toolpad-core/src/AppProvider/AppProvider.tsx
```diff
@@ -135,7 +134,543 @@ AppProvider.propTypes /* remove-proptypes */ = {
    * [Theme](https://mui.com/material-ui/customization/theming/) used by the app.
    * @default baseTheme
    */
-  theme: PropTypes.any,
+  theme: PropTypes.shape({
```
apedroferreira, line 137: wow

apedroferreira:
TODO:
- [ ] Fix proptypes generation
- [ ] Do not list demos in callout for every component with API links

docs/data/toolpad/core/pages.ts
```diff
@@ -60,86 +60,44 @@ const pages: MuiPage[] = [
...
-          {
-            pathname: '/toolpad/core/components/line-chart',
-            title: 'Line Chart',
+            pathname: '/toolpad/core/react-app-provider',
+            title: 'App Provider',
           },
         ],
       },
       {
-        pathname: '/toolpad/core/components/filter-group',
```
apedroferreira, line 89: Removing these but we can restore them back in the same way once we add more components, they might have been causing issues with the scripts.

docs/data/toolpad/core/pages.ts
```diff
@@ -60,86 +60,44 @@ const pages: MuiPage[] = [
...
-        subheader: 'Filters',
+        pathname: '/toolpad/core/layout-group',
+        subheader: 'Layout',
         children: [
           {
-            pathname: '/toolpad/core/components/select-filter',
-            title: 'Select',
+            pathname: '/toolpad/core/react-dashboard-layout',
```
apedroferreira, line 79: Not sure if we should follow this `react-...` naming pattern for pages, I saw all the other MUI libraries are doing it? But we can do it differently too.

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -49,6 +48,60 @@ const LogoContainer = styled('div')({
   },
 });
 
+function ToolpadLogo({ size = 40 }: { size?: number }) {
```
apedroferreira, line 51: Moved this component inside here as that's kind of how the scripts expect files to be organized, but it does take up a lot of lines...

Janpot (approved):

Janpot:
Thanks! I removed an alias as it broke the demo formatter.

PR #3554, opened 2024-05-14
Add DashboardLayout component to @toolpad/core

apedroferreira:
Isolated DashboardLayout component and its docs from https://github.com/mui/mui-toolpad/pull/3343 and implemented review feedback.

**Docs/demo:** https://deploy-preview-3554--mui-toolpad-docs.netlify.app/toolpad/core/components/dashboard-layout/

_(once we have a navigation adapter we can make the demo sidebar links work with local state to provide better examples)
(API docs generation to be added next in separate PR)_

### Summary

Please check the types in [packages/toolpad-core/src/AppProvider/AppProvider.tsx](https://github.com/mui/mui-toolpad/pull/3554/files#diff-d72d0f71caf7ef5c3f7a2250be04d58d3dc741a4e944a9c4a9f5bf1f518e6b5f) for the chosen API.

Possible to import types such as `Branding` and `Navigation` from `@toolpad/core`.
Also added `@toolpad/core/themes` where the `baseTheme` we use can be imported from, and that we use as a default theme. Eventually we can provide more themes?

### Addressed PR comments

- https://github.com/mui/mui-toolpad/pull/3343#discussion_r1597305323
- https://github.com/mui/mui-toolpad/pull/3343#discussion_r1597305679
- https://github.com/mui/mui-toolpad/pull/3343#discussion_r1597306617
- https://github.com/mui/mui-toolpad/pull/3343#discussion_r1598049998
- https://github.com/mui/mui-toolpad/pull/3343#discussion_r1598054706
- https://github.com/mui/mui-toolpad/pull/3343#discussion_r1598057358
- https://github.com/mui/mui-toolpad/pull/3343#discussion_r1598078678
- https://github.com/mui/mui-toolpad/pull/3343#discussion_r1598080006
- https://github.com/mui/mui-toolpad/pull/3343#discussion_r1598091993
- https://github.com/mui/mui-toolpad/pull/3343#discussion_r1598097992
- https://github.com/mui/mui-toolpad/pull/3343#discussion_r1598099632


docs/data/toolpad/core/components/dashboard-layout/DashboardLayoutBasic.js
```diff
@@ -0,0 +1,69 @@
+import * as React from 'react';
```
Janpot, line 1: Was this hand-written? I think we should port the `docs:typescript:formatted` script from core

docs/data/toolpad/core/api/components/dashboard-layout.md
```diff
@@ -0,0 +1,3 @@
+# Dashboard Layout API
```
Janpot, line 1: For the API, let's port the docs generator from core?

docs/data/toolpad/core/components/dashboard-layout/DashboardLayoutBasic.js
apedroferreira: I hadn't noticed that, I guess I can include it in this PR if it's simple.

docs/data/toolpad/core/api/components/dashboard-layout.md
apedroferreira: We can do that in this separate PR, right? https://github.com/mui/mui-toolpad/pull/3536

docs/data/toolpad/core/components/dashboard-layout/DashboardLayoutBasic.js
apedroferreira: I've added it here, seems to work fine https://github.com/mui/mui-toolpad/pull/3554/commits/9465e7462b7b45f4938a89689f1bfd14c610e7fb

docs/data/toolpad/core/api/components/dashboard-layout.md
apedroferreira: Ah, the X one, I was checking the Base one, will take a look.

apedroferreira:
This should be ready as a first version of the `DashboardLayout` - of course we can add and improve a lot in the future.
Will add the generated API docs separately, still finishing work on that but it would be easier to manage if this one is merged.

docs/data/toolpad/core/pages.js
Janpot, line 1: I don't believe this file is expected here, see my other comment below

packages/toolpad-core/src/AppProvider/AppProvider.tsx
```diff
@@ -1,20 +1,58 @@
...
+
+export interface Branding {
+  title?: string;
+  logo?: React.ReactNode;
+}
+
+export interface NavigationPageItem {
+  kind?: 'page';
+  title: string;
+  path?: string;
```
Janpot, line 15: With the nomenclature `path` I'd expect it to be the full path. Maybe we need to come up with another name e.g. like "slug"? That way we can later still add `path` if we want to add an override in the future.

```suggestion
  slug: string;
```

scripts/coreTypeScriptProjects.js
Janpot, line 1: 🤔  The file is already in the project, with different casing

Janpot (approved):

PR #3558, opened 2024-05-16
[core] Initial DataGrid component

Janpot:
Bring in DataProvider and DataGrid from the prototype.

Supports all CRUD operations, and serverside/clientside pagination

[docs preview](https://deploy-preview-3558--mui-toolpad-docs.netlify.app/toolpad/core/react-data-grid/)


docs/data/toolpad/core/components/all-components.md
```diff
@@ -0,0 +1,5 @@
+# Toolpad components
```
apedroferreira, line 1: I already created something like this at `toolpad/core/all-components`. But we can easily move it and rewords thing, I copied the structure from Base UI but don't have a particular preference.

docs/data/toolpad/core/components/data-grid/BasicDataProvider.js
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react';
+import { createDataProvider } from '@toolpad/core/DataProvider';
+import { DataGrid } from '@toolpad/core/DataGrid';
+import Box from '@mui/material/Box';
+
+const myData = createDataProvider({
```
apedroferreira, line 6: I went over this page and the examples seem using arrays directly seem good, but I wonder if they could be even more clear if they used some kind of mock API with its own methods.
It could be closer to what users will do with real APIs, or maybe it would be more redundant, so not sure, just a thought. I think this works great for now!

The one example that took a bit more for be to understand was the pagination one.

docs/data/toolpad/core/components/data-grid/FieldInferrence.js
```diff
@@ -0,0 +1,26 @@
...
+        { id: '1', name: 'John' },
+        { id: '2', name: 'Jane' },
+      ],
+    };
+  },
+  // paste fields here:
+});
+// preview-end
+
+export default function FieldInferrence() {
```
apedroferreira, line 20: The correct word is just `Inference` I think.

docs/data/toolpad/core/components/data-grid/data-grid.md
```diff
@@ -0,0 +1,100 @@
...
+productId: toolpad-core
+title: App Provider
+components: DataGrid, DataProvider
+---
+
+# Data Grid
+
+<p class="description">Data Grid component for CRUD applications.</p>
+
+Toolpad core extends the X grid with CRUD functionality. It abstracts the manipulations in a data provider object. The data provider object describes the shape of the data and the available manipulations. When you pass the data provider to a grid it is automatically configured as a CRUD grid. All properties of the X grid are also available and can be used to override the data provider behavior.
```
apedroferreira, line 11:
```suggestion
Toolpad Core extends the X grid with CRUD functionality. It abstracts the manipulations in a data provider object. The data provider object describes the shape of the data and the available manipulations. When you pass the data provider to a grid it is automatically configured as a CRUD grid. All properties of the X grid are also available and can be used to override the data provider behavior.
```

Should we usually capitalize like this?

docs/data/toolpad/core/components/data-grid/data-grid.md
```diff
@@ -0,0 +1,100 @@
...
+
+Where core and X components focus on the user interface. Toolpad core components start from a definition of the data. It centralizes data loading, filtering, pagination, field formatting, mutations, access control and more.
```
apedroferreira, line 13:
```suggestion
Where core and X components focus on the user interface, Toolpad core components start from a definition of the data. It centralizes data loading, filtering, pagination, field formatting, mutations, access control and more.
```

docs/data/toolpad/core/pages.ts
```diff
@@ -81,28 +82,23 @@ const pages: MuiPage[] = [
...
+        subheader: 'Data',
         children: [
           {
-            pathname: '/toolpad/core/api/app-provider',
-            title: 'AppProvider',
-          },
-          {
-            pathname: '/toolpad/core/api/dashboard-layout',
-            title: 'DashboardLayout',
+            pathname: '/toolpad/core/components/data-grid',
```
apedroferreira, line 90: I changed these component pages to be like `toolpad/core/components/data-grid` in the PR for generating API docs, just to follow the convention I saw in other MUI libraries. But I wasn't sure about it, so we can have the like this and remove the `react-` part.

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -274,18 +274,6 @@ function DashboardLayout(props: DashboardLayoutProps) {
             </Stack>
           </a>
           <Box sx={{ flexGrow: 1 }} />
-          {/* <Stack>
```
apedroferreira, line 277: Nice, I wanted to remove this too as it keeps shifting, we can add it back later :D

packages/toolpad-core/src/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,1041 @@
+'use client';
```
apedroferreira, line 1: I have been putting the `use client` directive in the index file only.
But we can change it for the other components if this works better.

packages/toolpad-core/src/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,1041 @@
...
+  // │ To update them, edit the TypeScript types and run `pnpm proptypes`. │
+  // └─────────────────────────────────────────────────────────────────────┘
+  /**
+   * The ref object that allows Data Grid manipulation. Can be instantiated with `useGridApiRef()`.
+   */
+  apiRef: PropTypes.shape({
+    current: PropTypes.object.isRequired,
+  }),
+  /**
+   * If `true`, the Data Grid height is dynamic and follow the number of rows in the Data Grid.
```
apedroferreira, line 735:
```suggestion
   * If `true`, the Data Grid height is dynamic and follows the number of rows in the Data Grid.
```

apedroferreira (changes requested):
Looks great!
Didn't really comment anything major, just some grammar when i went through it and some other small things.

docs/data/toolpad/core/components/all-components.md
Janpot: Yep, this was a merge gone bad

docs/data/toolpad/core/components/data-grid/BasicDataProvider.js
Janpot: I'll think about a solution. In the meantime it should probbaly not be a blocker for this PR

packages/toolpad-core/src/DataGrid/DataGrid.tsx
Janpot: Yes, let's keep them at the smallest scope possible. Probbaly even debatable if libraries should have them at all

mnajdova (commented):
A bit of feedback, can we pin the panel for adding item, I scrolled and suddenly forgot that I had the panel open and wondered why the button is disabled. If we don't pin it, maybe we can make the button always enabled and scroll to the panel if clicked twice (I would go with the pin options from these two).

One more thing, maybe we can add some separator between the fields in the panel and maybe add a disabled state if the user cannot enter some field (e.g. the id is auto-generated).

When using both update and delete, it would be great if the update icons don't cover the remove icon.

Screenshot 2024-05-27 at 10 08 21

The rest looks great 👌

mnajdova (changes requested):
I also noticed that the API links are broken, they point to toolpad-core/api/component instead of toolpad/core/api/component. It works as expected on master.

Janpot:
> A bit of feedback, can we pin the panel for adding item, I scrolled and suddenly forgot that I had the panel open and wondered why the button is disabled.

At the moment, the add row logic is built on top of an empty row in editing mode, would like to pin this row but that's a X pro feature. Potential solutions I can think of:

1. keep using an empty draft row, but disable scrolling when this UI is active
2. build instead as a form in a drawer that overlays the datagrid

For me, personally, 2. makes a lot of sense. For tables with lots of columns, vertical scroll in a form feels more natural than horizontal scrolling in an editable row.

mnajdova:
> For me, personally, 2. makes a lot of sense. For tables with lots of columns, vertical scroll in a form feels more natural than horizontal scrolling in an editable row.

Right makes sense, I was too focused on this particular use-case.

> The idea would be to keep the delete action available even when the row is in update mode? Basically, show 3 icons?

I would vote for this yes. Not a hard push tough, but it seemed a bit awkward that the icon was not there.

Janpot:
Ok, I'll try it out with a form in follow-up. In the meantime I've disabled scrolling when the create flow is activated.
Another advantage would be: The current method also sort of forces us to use row editing mode only. If the creation flow does not build on top of editing, we can leave users the choice whether they want row or cell editing.

Regarding the row actions, I was trying to avoid having to make the column wider than two icons. But we can make that three. Another option is to to have in editing mode a save icon, and a menu that contains a cancel action and a delete action. If we can enable cell editing mode again, we can even reduce it to a delete icon only.

mnajdova:
> Regarding the row actions, I was trying to avoid having to make the column wider than two icons. But we can make that three. Another option is to to have in editing mode a save icon, and a menu that contains a cancel action and a delete action. If we can enable cell editing mode again, we can even reduce it to a delete icon only.

Sounds good, feel free to tackle in a follow up. We can maybe ask someone from the design team for an opinion.

Janpot:
> I also noticed that the API links are broken, they point to toolpad-core/api/component instead of toolpad/core/api/component. It works as expected on master.

This was fixed on master in the meantime, merged these changes in this branch

apedroferreira:
> I also noticed that the API links are broken, they point to toolpad-core/api/component instead of toolpad/core/api/component. It works as expected on master.

I've added the fix in the monorepo https://github.com/mui/material-ui/pull/42362

PR #3561, opened 2024-05-16
Updates to project setup

Janpot:
several fixes brought from https://github.com/mui/mui-toolpad/pull/3558
- memory issues in `@toolpad/studio-components`
- add typescript formatted demos @apedroferreira I needed this, it should correspond to what you have as well
- align tsconfigs around project references
- make sure we can use `@toolpad/utils` in `@toolpad/core`
- make sure we can use `@toolpad/core` in the docs


scripts/coreTypeScriptProjects.js
```diff
@@ -0,0 +1,12 @@
+import path from 'path';
```
apedroferreira, line 1: You included this file twice with different casing :)

apedroferreira (approved):
Looks great! Just saw a duplicated file.

docs/src/modules/utils/babel-plugin-jsx-preview.js
```diff
@@ -1,62 +1,89 @@
...
+    // Trim blank JSXText to normalize
+    // return (
+    //   <div />
+    // )
+    // and
+    // return (
+    //   <Stack>
+    //     <div />
+    // ^^^^ Blank JSXText including newline
+    //   </Stacke>
```
apedroferreira, line 38:
```suggestion
    //   </Stack>
```

PR #3578, opened 2024-05-21
[docs] Fix demo codesandbox dependencies

Janpot:
Make sure the demos have correct dependencies in PRs

### DIFF
--- a/packages/toolpad-core/package.json
+++ b/packages/toolpad-core/package.json
@@ -50,9 +50,6 @@
     "@emotion/styled": "11.11.5",
     "@toolpad/utils": "workspace:*",
     "client-only": "^0.0.1",
-    "invariant": "2.2.4",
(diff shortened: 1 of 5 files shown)


apedroferreira (approved):

packages/toolpad-core/package.json
```diff
@@ -48,15 +48,13 @@
   "dependencies": {
     "@emotion/react": "11.11.4",
```
apedroferreira, line 49: I probably moved some of these back to `dependencies` by mistake, I guess they can be left out...
If the `@emotion` libraries are peer dependencies of `@mui/material` as they are could they be left out too?

Janpot: We shouldn't import from `@emotion/react` at all. I've removed the import in `AppProvider` as well

PR #3587, opened 2024-05-24
Add Next.js app router Toolpad Core playground

apedroferreira:
Add playground app for Next.js app router, navigation is still a bit wonky and empty but should be a starting point.


bharatkashyap (approved):
Tried it out, looks good to me! 👍🏻

PR #3588, opened 2024-05-24
Add Next.js pages router app to Toolpad Core playground

apedroferreira:
- Add playground app for Next.js pages router
- Use Next.js app provider in both playground apps, and the respective cache provider
- Improve documentation for `AppProvider` component including Next.js version
- Improve `DashboardLayout` documentation examples / navigation

Docs: 
- https://deploy-preview-3588--mui-toolpad-docs.netlify.app/toolpad/core/react-app-provider/
- https://deploy-preview-3588--mui-toolpad-docs.netlify.app/toolpad/core/react-dashboard-layout/
- https://deploy-preview-3588--mui-toolpad-docs.netlify.app/toolpad/core/api/app-provider/


packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -18,7 +18,12 @@ import Toolbar from '@mui/material/Toolbar';
 import Typography from '@mui/material/Typography';
 import ExpandLessIcon from '@mui/icons-material/ExpandLess';
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
-import { BrandingContext, Navigation, NavigationContext, NavigationPageItem } from '../AppProvider';
+import {
+  BrandingContext,
+  Navigation,
+  NavigationContext,
+  NavigationPageItem,
+} from '../AppProvider/AppProvider';
```
apedroferreira, line 26: Importing from `AppProvider/index` had an error in SSR as it was missing some exports from the file itself.

playground/toolpad-core-nextjs-pages-router/src/pages/_app.tsx
```diff
@@ -0,0 +1,39 @@
...
+  },
+  {
+    slug: '/dashboard/orders',
+    title: 'Orders',
+    icon: <ShoppingCartIcon />,
+  },
+];
+
+export default function App({ Component, pageProps }: AppPropsWithLayout) {
+  const getLayout = Component.getLayout ?? ((page) => page);
```
Janpot, line 34: To reduce boilerplate, I would create a 

```tsx
function getDefaultLayout(page: React.ReactElement) {
  return <DashboardLayout>{page}</DashboardLayout>;
}
```

and

```suggestion
  const getLayout = Component.getLayout ?? getDefaultLayout;
```

And remove the `getLayout` from the pages. We can then add a `getLayout` to the auth pages to remove the `DashboardLayout`.

playground/toolpad-core-nextjs-pages-router/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "private": true,
+  "scripts": {
+    "dev": "next dev",
+    "lint": "next lint"
+  },
+  "devDependencies": {
+    "@emotion/react": "11.11.4",
+    "@emotion/styled": "11.11.5",
+    "@mui/icons-material": "5.15.18",
+    "@mui/material": "5.15.18",
```
Janpot, line 13: Don't we also need `@mui/material-nextjs`?

playground/nextjs-pages-router/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "playground-nextjs-pages-router",
```
Janpot, line 2: I don't want to have to type this al the time 😄 
```suggestion
  "name": "playground-nextjs-pages",
```

playground/nextjs/src/app/orders/page.tsx
Janpot, line 1: So originally the idea was (notice the parentheses):
`/app/(dashboard)/orders/page.tsx`
that way we can add dashboard layout under 
`/app/(dashboard)/layout.tsx`
and authentication pages that don't adopt this layout
`/app/auth/page.tsx`

packages/toolpad-core/src/nextjs/AppProviderNextApp.tsx
```diff
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { usePathname, useSearchParams, useRouter } from 'next/navigation';
+import { AppRouterCacheProvider } from '@mui/material-nextjs/v14-appRouter';
```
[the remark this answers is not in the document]
apedroferreira: Hm ok, we can remove it for now.

playground/nextjs/src/app/orders/page.tsx
apedroferreira: Ok I'll try that!

Janpot (approved):
Let's merge this, but I think we need to improve the demo page for the app provider. We need to explain that this is meant to go in a layout component. Now it looks like it goes together with the dashboard layout.

I'd also like to see a bit more text in the demos explaining positioning the components in the library and their purpose. We're not only documenting usage, we also need to sell a philosophy.

apedroferreira:
> Let's merge this, but I think we need to improve the demo page for the app provider. We need to explain that this is meant to go in a layout component. Now it looks like it goes together with the dashboard layout.
> 
> I'd also like to see a bit more text in the demos explaining positioning the components in the library and their purpose. We're not only documenting usage, we also need to sell a philosophy.

Got it, yeah that all makes sense, will address all the feedback about the `AppProvider` documentation (and possibly the `DashboardLayout` too in terms of adding more text) in a separate PR!
I guess some docs are better than no docs so still merging :D

PR #3638, opened 2024-06-04
[AppProvider] Create basic router adapters

Janpot:
* Initial router implementation for nextjs app.
* Live [demo](https://deploy-preview-3638--mui-toolpad-docs.netlify.app/toolpad/core/introduction/tutorial/#create-a-new-page) on the tutorial 
* Add to playground
* Initial (untested) implementation for nextjs pages
* Unified approach for serving both app and pages with a single component

@apedroferreira Basic router implementation as a starting point.

Some potential tweaks we could do to the playground:
- Rename `toolpad-core-next` folder to just `nextjs` and the plaground package to `playground-next`
- Change `dashboard` folder to a `(dashboard)` folder to remove that segment from the url
- Can the homepage of the playground be part of the DashboardLayout?


packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -157,6 +147,17 @@ function DashboardSidebarSubNavigation({
     [],
   );
 
+  const handleLinkClick = React.useMemo(() => {
```
apedroferreira, line 150: Why not a `.useCallback` here? Not saying it's wrong, just curious as it's what I would use.

packages/toolpad-core/src/AppProvider/AppProvider.tsx
```diff
@@ -30,11 +48,14 @@ export type NavigationItem = NavigationPageItem | NavigationSubheaderItem | Navi
 
 export type Navigation = NavigationItem[];
 
+// TODO: hide these contexts from public API
```
apedroferreira, line 51: I don't think we need to export the contexts in `AppProvider/index`, if was probably just an oversight from me.

apedroferreira (approved):
This looks really great, looks like there wasn't a need to use hooks after all.

We're just gonna need to fix the API docs for the `AppProvider`, and we can also use this local state router in the examples here https://deploy-preview-3638--mui-toolpad-docs.netlify.app/toolpad/core/react-dashboard-layout/, but none of that has to be done in this PR, I can even take care of those.

I can test them on the the playgrounds too, but it would probably be better to merge this first as you already unified the Next.js app providers.

About the playground tweaks, yeah I was changing them so that the home page is the dashboard already, I guess that makes more sense and we probably don't need a `/dashboard` route?

packages/toolpad-core/src/AppProvider/AppProvider.tsx
apedroferreira: Or we can just move the to their own file.

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
Janpot: Because `useCallback` always returns a function, and here I want it to be able to be `undefined`

packages/toolpad-core/src/AppProvider/AppProvider.tsx
Janpot: We could probably also benefit from creating a context for each of the router properties as they may update at different rates

PR #3646, opened 2024-06-06
[test] Data grid CRUD tests

Janpot:
Add tests for the CRUD operations in the data grid. For each of the operations, go through the full flow:

* create/update/delete item
* verify availability and disabled state of all possible actions
* verify snackbar and subsequent filter + add fixes for mixed id types
* Fix typing issues with forwardRef (just removing it for now)
* Fix @toolpad/utils description and README


packages/toolpad-core/src/DataGrid/DataGrid.test.tsx
```diff
@@ -3,25 +3,251 @@
...
+    ?.find((theRow) => Number(theRow.dataset.rowindex) === rowIndex);
+  if (!row) {
+    throw new Error(`Row ${rowIndex} not found`);
+  }
+  return row;
+}
+
+function getCell(root: HTMLElement, rowIndex: number, colIndex: number): HTMLElement {
+  const row = getRow(root, rowIndex);
+  if (!row) {
```
apedroferreira, line 36: Doesn't `getRow` already throw an error if there's no row?

apedroferreira (approved):

Issue #3586, opened 2024-05-24
[DashboardLayout] Add theme palette mode switcher


Janpot:
### Summary

Currently the DashboardLayout component comes with a default light theme. We should add the possibility for different color schemes and if multiple are defined, we should add a theme switcher.

API proposal:

```tsx
type AtLeastOne<T> = { [K in keyof T]: Pick<T, K> }[keyof T];

interface AppProviderProps {
  // ...
  themes?: AtLeastOne<{
    light: Theme;
    dark: Theme;
  }>;
  // unstable_adoptPaletteMode?: boolean;
}
```

* `themes` undefined => provide default light and dark themes + add theme switcher
* `themes` defined, only one provided => use only the provided theme + remove theme switcher
* `themes` defined, both provided => set both themes + add theme switcher

Additional suggestion:

Add `unstable_adoptPaletteMode`, if it's `true`, we hide the theme switcher and we adopt the mode from the surrounding theme. This could be useful for the docs to remove the iframe from the demos.

**Search keywords**: theme switcher

apedroferreira:
Some benchmarking:
- Refine (https://refine.dev/docs/3.xx.xx/api-reference/mui/customization/mui-custom-theme/#dark-mode-toggle) - single theme, logic must be manually created by the user with React state
- Clerk - could not find a theme switcher utility
- React-admin (https://marmelab.com/react-admin/AppTheme.html) - takes `theme` and `darkTheme` as props of the root component
- Vuetify (https://vuetifyjs.com/en/features/theme/#typescript) - function that takes an object with a `theme` property and nested `themes` where multiple themes can be added as object properties, and a default theme can be set.

These were the first solutions that came to mind for me for this problem: https://github.com/mui/mui-toolpad/pull/3343#discussion_r1598691276

In conclusion, I think the initially proposed solution with 
```
interface AppProviderProps {
  // ...
  themes?: AtLeastOne<{
    light: Theme;
    dark: Theme;
  }>;
  // unstable_adoptPaletteMode?: boolean;
}
```
seems fine and intuitive to use, and it's possibly the one that abstracts away the most / the simplest.
It's very literal and would work with the theme switcher seamlessly.

> themes defined, only one provided => use only the provided theme + remove theme switcher

What about passing just one theme directly? I feel like this would be more natural, but on the other hand it does create 2 quite different ways to define the theming (it's less consistent).

[...]

PR #3674, opened 2024-06-12
[core] Add theme switcher to dashboard layout

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/3586

Add theme switcher to `DashboardLayout` component header.

https://github.com/mui/mui-toolpad/assets/10789765/468ac58f-c589-4214-adf3-9eb8550d69f6

As decided below, the theme can be set as `Theme | CssVarsTheme | { light: Theme, dark: Theme }; `, and the switcher only shows if using both `light` and `dark` themes both for the CSS variables or the standard option.

Also moved all React contexts to `shared/context` while trying to deal with some circular dependency issues + trying not to export them with the package, but this might not have been necessary in the end. Let me know if you think we should organize that differently.

Documented theming + the switcher in the `AppProvider` and `DashboardLayout` docs, and tried to improve the documentation for both these components in general as it had been discussed:

**Component Pages:**
- https://deploy-preview-3674--mui-toolpad-docs.netlify.app/toolpad/core/react-app-provider
- https://deploy-preview-3674--mui-toolpad-docs.netlify.app/toolpad/core/react-dashboard-layout

**API Pages:**
- https://deploy-preview-3674--mui-toolpad-docs.netlify.app/toolpad/core/api/app-provider
- https://deploy-preview-3674--mui-toolpad-docs.netlify.app/toolpad/core/api/dashboard-layout


apedroferreira:
There is an issue during SSR when using the dark mode where the server can't read which mode it's supposed to use, so the app renders in light mode until the client-side sets the correct value for the dark mode.

From what I've tried/researched there's 2 possible solutions:
1. Use cookies - will eventually work but it's complicated as it has to be done in different ways for the app router and the pages router. The app router needs to use a server component and the pages router an alternative solution - these can be difficult and a bit messy to integrate with the library.
2. Leave it as is and probably Pigment.css will provide a more elegant solution for this?

Janpot:
3. We should be ale to leverage [css variables](https://mui.com/material-ui/experimental-api/css-theme-variables/usage/) like we do on the homepage.

apedroferreira:
All good in functionality, just missing documentation now.

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -204,6 +209,16 @@ function DashboardLayout(props: DashboardLayoutProps) {
 
   const branding = React.useContext(BrandingContext);
   const navigation = React.useContext(NavigationContext);
+  const { paletteMode, setPaletteMode, isDualTheme } = React.useContext(PaletteModeContext);
+
+  const [hasMounted, setHasMounted] = React.useState(false);
```
Janpot, line 214: Can you use `useSsr` from `@toolpad/utils`? It would avoid double renders when mounting client side.

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -220,12 +235,25 @@ function DashboardLayout(props: DashboardLayoutProps) {
...
+              <Typography
+                variant="h6"
+                sx={{ color: (theme) => (theme.vars ?? theme).palette.primary.main }}
+              >
                 {branding?.title ?? 'Toolpad'}
               </Typography>
             </Stack>
           </a>
           <Box sx={{ flexGrow: 1 }} />
+          {hasMounted && isDualTheme ? (
```
Janpot, line 247: If possible I'd like to render as much as possible during SSR. We can basicaly show everything except the icon. (Technically I guess under the css vars theme we can render both icons and use css vars to show one or the other)


Instead of turning on/off the whole thing during ssr, can we do something equivalent to
```tsx
{isDualTheme ? (
  <Tooltip title={isSsr ? null : `${paletteMode === 'dark' ? 'Light' : 'Dark'} mode`} enterDelay={500}>
    <IconButton
      aria-label={isSssr ? 'Change theme' : `Switch to ${paletteMode === 'dark' ? 'light' : 'dark'} mode`}
      onClick={toggleMode}
    >
      {isSsr ? <EmptyIcon /> : paletteMode === 'dark' ? <LightModeIcon /> : <DarkModeIcon />}
    </IconButton>
  </Tooltip>
) : null}
```

This will avoid a horizontal layout shift

packages/toolpad-core/src/themes/baseDarkTheme.ts
```diff
@@ -0,0 +1,129 @@
...
+    text: {
+      primary: defaultDarkTheme.palette.grey['100'],
+    },
+  },
+  typography: {
+    h6: {
+      fontWeight: '700',
+    },
+  },
+  components: {
```
Janpot, line 23: It feels wrong to me to add overrides in the theme. Some of these components are likely to appear in user code (Toolbar, List, IconButton). I don't think we should mandate any override to the user. Can't we style these components locally in the DashboardLayout?

playground/nextjs-pages/src/pages/_document.tsx
```diff
@@ -8,7 +8,7 @@ import {
 
 export default function Document(props: DocumentProps & DocumentHeadTagsProps) {
   return (
-    <Html lang="en">
+    <Html lang="en" data-color-scheme="light">
```
Janpot, line 11: What's the difference between `data-color-scheme` and `data-mui-color-scheme`?

test/setupVitest.ts
```diff
@@ -13,3 +13,21 @@ failOnConsole({
 });
 
 afterEach(cleanup);
+
+// Mocks
+
+if (typeof window !== 'undefined') {
```
Janpot, line 19: You also need to check whether `matchMedia` exists. We run these tests in real browsers as well

docs/data/toolpad/core/components/app-provider/AppProviderBasic.tsx
```diff
@@ -38,7 +39,11 @@ export default function AppProviderBasic() {
 
   return (
     // preview-start
```
Janpot, line 41: Just to note, these comments are not absolutely necessary, the script would likely output the same content automatically in this demo.

docs/data/toolpad/core/components/app-provider/AppProviderBasic.tsx
```diff
@@ -38,7 +39,11 @@ export default function AppProviderBasic() {
...
-    <AppProvider navigation={NAVIGATION} router={router}>
+    <AppProvider
+      navigation={NAVIGATION}
```
Janpot, line 43: In the navigation I also see
```
// Add the following new item:
```
I presume we don't need that comment in the context of this demo?

docs/data/toolpad/core/components/app-provider/AppProviderTheme.tsx
```diff
@@ -0,0 +1,177 @@
...
+  {
+    slug: '/page-2',
+    title: 'Page 2',
+    icon: <TimelineIcon />,
+  },
+];
+
+const defaultDarkTheme = createTheme({ palette: { mode: 'dark' } });
+
+const customTheme: Theme = createTheme(defaultDarkTheme, {
```
Janpot, line 30: I really love the look of this demo, but I feel like the MUI theme is the wrong vehicle to configure these customizations. These are customizations that apply to the application as a whole. I feel like we should look more into a `slots`/`slotProps` setup similar to how X [promotes it](https://mui.com/x/react-date-pickers/custom-field/). That way you can target custumizations e.g. the List under the dashboard layout only.

docs/data/toolpad/core/components/app-provider/app-provider.md
```diff
@@ -6,20 +6,90 @@ components: AppProvider
...
+
+  return (
+    <AppProvider>
+      <Component {...pageProps} />
+    </AppProvider>
+  );
+}
+```
+
+## Theming
```
Janpot, line 79: I would also add a separate "Navigation" section that explains the `navigation` prop and an example of a hierarchy that is a bit more extensive with subheaders and nested items. Ideally something realistic for a dashboard that end users could relate to.

This would also be fine to do in follow-up

docs/data/toolpad/core/components/app-provider/app-provider.md
```diff
@@ -6,20 +6,90 @@ components: AppProvider
...
+It is not mandatory that every application is wrapped in an `AppProvider`, but it is highly recommended for most apps that use Toolpad.
 
 ## Basic
 
-Wrap the whole application with the `AppProvider` component to enable many of Toolpad's features.
+Wrap an application page with the `AppProvider` component.
+
+Ideally, the `AppProvider` should wrap every page in the application, therefore in most projects it should be imported and placed in the file that defines a **shared layout** for all pages.
+
+In the following (purely demonstrative) example, an `AppProvider` component wrapping the page provides it with a default theme, and a `DashboardLayout` placed inside it gets its navigation and routing features automatically set based on the props passed to the `AppProvider`.
```
Janpot, line 21: Not sure it brings much to the table 🙂 
```suggestion
In the following example, an `AppProvider` component wrapping the page provides it with a default theme, and a `DashboardLayout` placed inside it gets its navigation and routing features automatically set based on the props passed to the `AppProvider`.
```

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
apedroferreira: I followed the documented examples for a theme switcher in CSS vars and don't remember if there was an extra reason for doing this, will look into it, if it's possible to server-side render the switcher it sounds good.

packages/toolpad-core/src/themes/baseDarkTheme.ts
apedroferreira: I guess the original idea was for these here to apply automatically also when the user uses these components... but let me see what I can do, maybe we can just scope these to the components used by the framework.

playground/nextjs-pages/src/pages/_document.tsx
apedroferreira: Not sure if they should be the same actually, in the App Router I had noticed that `data-mui-color-scheme` was the correct one but in the docs it for Pages Router it's just `data-color-scheme`, will look into it more deeply to make sure it's correct in both cases.

docs/data/toolpad/core/components/app-provider/AppProviderTheme.tsx
apedroferreira: Didn't fully understand this comment, this file here is for exemplifying how to set up a custom theme for the whole application, mostly some palette colors... Do you mean this just about the `components` in this theme example?

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
Janpot: As far as I know the only reason is hydration mismatches. But not all usage of our component will be hydrated, that's why we should use `useSsr`. And even when hydrated, we should only stabilize what causes the hydration error. i.e. the icon.

docs/data/toolpad/core/components/app-provider/app-provider.md
apedroferreira: Yeah this one was because you mentioned before that this example with a `DashboardLayout` directly inside an `AppProvider` could be misleading as it's not how most users would/should use them, so I wasn't even sure if I should include this example. I'm fine with removing that part!

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
apedroferreira: Not sure if it's possible to make the correct icon show during server render as the icons are rendered by Javascript, as opposed to the theme which is just CSS.
I went with a solution kinda like you suggested, it seems great!

bharatkashyap:
How would we detect the "current" palette mode in other components? I would imagine a hook that would allow us to do something like 
```tsx  
 const mode = usePaletteMode() 
 ```
 would be important and very useful to integrate with the theme switcher for the rest of the components.

**Comment by apedroferreira** [comment]
> How would we detect the "current" palette mode in other components? I would imagine a hook that would allow us to do something like
> 
> ```tsx
>  const mode = usePaletteMode() 
> ```
> 
> would be important and very useful to integrate with the theme switcher for the rest of the components.

We have context as can be seen in the `DashboardLayout` component in https://github.com/mui/mui-toolpad/blob/ad9bddb719a81cd113b73e82d304ae5ead1a0e8c/packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx#L81

I guess that should do the job?

**Comment by Janpot** [comment]
for css variables theme: https://mui.com/material-ui/experimental-api/css-theme-variables/usage/#toggle-between-light-and-dark-mode

we can create a shim with the exact same api for when non-css variables themes are used

**Comment by apedroferreira** [comment]
If you mean in components made by users we can create abstractions that just use this context under the hood, yes, shimming that hook also sounds ok if possible!

File: docs/data/toolpad/core/components/dashboard-layout/dashboard-layout.md [hunk 11]
```diff
@@ -8,22 +8,35 @@ components: AppProvider, DashboardLayout
 
 <p class="description">The dashboard layout component provides a customizable out-of-the-box layout for a typical dashboard page.</p>
 
+The `DashboardLayout` component is a quick, easy way to provide a standard full-screen layout with a header and sidebar to any dashboard page, as well as ready-to-use and easy to customize navigation and branding.
+
+Many features of this component are configurable through the [AppProvider](https://mui.com/toolpad/core/react-app-provider/) component that should wrap it.
```
Janpot, line 13: To follow-up outside of this PR: 
Would it make sense to provide all the available functionality that is read from the app provider as props as well? That way the suer has the choice to use with app provider and stand alone. If we let the props fall-back to the app provider contexts we have a way to offer following functionalities:
- use with app provider
- use standalone with the props
- use with app provider but overwrite specific context

docs/data/toolpad/core/components/dashboard-layout/DashboardLayoutBranding.tsx
```diff
@@ -37,19 +49,20 @@ export default function DashboardLayoutBranding() {
...
-    <AppProvider navigation={NAVIGATION} branding={BRANDING} router={router}>
+    // preview-start
+    <AppProvider
+      navigation={NAVIGATION}
+      branding={{
+        logo: <img src="https://mui.com/static/logo.png" alt="MUI logo" />,
+        title: 'MUI',
+      }}
+      router={router}
+      theme={{ light: baseLightTheme, dark: baseDarkTheme }}
```
Janpot, line 60: Do we need this `property` here? If we need it, I'd expect it to set the primary color to one that matches the logo.

I also don't think we should promote anything other than css vars themes. The only place where we mention or use legacy theme should be where we document specifically them.

docs/data/toolpad/core/components/dashboard-layout/dashboard-layout.md
apedroferreira: Yeah I was planning to do that next and then edit the docs accordingly, we can prioritize the props that are passed otherwise default to the app provider.

docs/data/toolpad/core/components/dashboard-layout/DashboardLayoutBranding.tsx
```diff
@@ -37,19 +49,20 @@ export default function DashboardLayoutBranding() {
   }, [pathname]);
 
   return (
-    <AppProvider navigation={NAVIGATION} branding={BRANDING} router={router}>
+    // preview-start
+    <AppProvider
+      navigation={NAVIGATION}
+      branding={{
```
Janpot, line 55: For a follow-up conversation:

Would it make sense to move these into the theme as [custom tokens](https://mui.com/material-ui/experimental-api/css-theme-variables/customization/#adding-new-theme-tokens)?

docs/data/toolpad/core/components/app-provider/app-provider.md
```diff
@@ -6,20 +6,90 @@ components: AppProvider
...
+
+An `AppProvider` can set a visual theme for all elements inside it to adopt via the `theme` prop. This prop can be set in a few distinct ways with different advantages and disadvantages:
+
+1. [CSS variables theme](https://mui.com/material-ui/experimental-api/css-theme-variables/overview/): the default and recommended theming option for Toolpad applications, as it is the only option that prevents issues such as [dark-mode SSR flickering](https://github.com/mui/material-ui/issues/27651) and supports both light and dark mode with a single theme definition. The provided default theme in Toolpad is already in this format.
+2. [Standard Material UI theme](https://mui.com/material-ui/customization/theming/): a single standard Material UI theme can be provided as the only theme to be used.
+3. **Light and dark themes**: two separate Material UI themes can be provided for light and dark mode in an object with the format `{ light: Theme, dark: Theme }`
+
+{{"demo": "AppProviderTheme.js", "height": 500, "iframe": true}}
+
+### Predefined themes
```
Janpot, line 89: I'm not 100% sure about this. Isn't the predefined theme just the default MUI theme? I'm not sure providing different themes should be in the scope of `@toolpad/core`. Or is there a specific reason?

In any case, I think we need to exclusively promote css vars themes and leave the legacy functionality for when people bring their own existing theme.

docs/data/toolpad/core/components/dashboard-layout/DashboardLayoutBranding.tsx
apedroferreira: You mean the `theme` property?
There's actually some conflict when using the CSS vars theme in the docs examples which is why I'm using `theme={{ light: baseLightTheme, dark: baseDarkTheme }}`... Maybe the class names for the iframes can be given a different name or something and that will work, I'll check out what we can do there.

Janpot: > You mean the theme property?

Yes, can't this demo just rely on the default theme?

docs/data/toolpad/core/components/app-provider/app-provider.md
apedroferreira: Doesn't necessarily need to be in the scope, I just wanted to possibly provide some ready-to-use themes that will look good. And that default theme just has a few color changes to look a bit better, but I guess the default theme should look decent enough.

To keep things simple let's keep themes outside the library then, and will rechecked how the default MUI theme looks - if it looks good we can keep it as the default.

docs/data/toolpad/core/components/app-provider/AppProviderTheme.js
```diff
@@ -0,0 +1,78 @@
...
+      <Typography>Dashboard content for {pathname}</Typography>
+    </Box>
+  );
+}
+
+DemoPageContent.propTypes = {
+  pathname: PropTypes.string.isRequired,
+};
+
+const customTheme = createTheme({
```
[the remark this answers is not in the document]
apedroferreira: We could do it, it's actually a small change so might include it in this PR.

apedroferreira:
Current issues to fix (just keeping track...):
- [ ] When using CSS vars theme in docs the docs initial theme is being changes if theme is changed in demos
- [ ] Adjust tests to work with CSS vars theme
- [ ] Probably incompatibility between v6 CSS variable themes being used by the docs and the v5 themes the library uses

Janpot (approved):
Let's merge the current state and follow up on future PRs

docs/data/toolpad/core/components/dashboard-layout/DashboardLayoutBranding.tsx
apedroferreira: It's using the default CSS variables theme now including in docs examples, but had to add a `window` prop to use the docs demo window instead of the whole page's.
Let me know if you think any alternative would be better and we can change it, even if in a separate PR.

apedroferreira:
Merging, thanks!
Some bugs left to fix which are mentioned above.

packages/create-toolpad-app/src/generateProject.ts
```diff
@@ -279,9 +194,9 @@ export default function generateProject(
       'react-dom': '^18',
       next: '^14',
       '@toolpad/core': 'latest',
-      '@mui/material': '^5',
+      '@mui/material': '^6',
```
Janpot, line 197: @apedroferreira I believe this will broken, this version doesn't exist on npm.

apedroferreira: Ok thanks for finding it, I'm working on bug fixes so will make sure the CLI and generated project work along with them.

Janpot: I changed these ones to `'next'` in https://github.com/mui/mui-toolpad/pull/3807

apedroferreira: Oh, didn't know you were releasing or might have checked more carefully. Thanks!

PR #3712, opened 2024-06-26
[core] Remove update-monorepo.yml

oliviertassinari:
We seem to never use this workflow: https://github.com/mui/mui-toolpad/actions/workflows/update-monorepo.yml. It was created in #2243. I guess we don't need this? It's flagged as having a potential security issue https://github.com/mui/mui-toolpad/security/code-scanning/70.


Janpot:
I started working on this but we ended with renovatebot updating this dependency.

oliviertassinari:
Ah makes sense. What did you envision that could be improved over Renovate?

Janpot:
> What did you envision that could be improved over Renovate?

At the moment, nothing. At the time I believe we [had to](https://github.com/renovatebot/renovate/issues/5640) manually update the monorepo dependency and I started building an action to automate it. Then after pnpm migration and pinning the monorepo dependency, I found renovate started handling updates properly. I just forgot about the half built github action.

PR #3751, opened 2024-07-03
[code-infra] Use `@mui/docs` where possible

LukasTy:
Related to https://github.com/mui/material-ui/pull/42833.

Would allow dropping last re-exports of `MarkdownElement`, `Link` and `HighlightedCode`.

~Also, ditch `@mui/docs` aliasing from monorepo as there is already a dependency with that name (related to https://github.com/mui/mui-x/pull/13713).~
We agreed to keep the alias to avoid potential problems with out-of-sync `@mui/docs` and `@mui/monorepo`.


docs/package.json
```diff
@@ -25,7 +25,7 @@
     "@emotion/server": "11.11.0",
     "@emotion/styled": "11.11.5",
     "@mui/base": "5.0.0-beta.40",
-    "@mui/docs": "6.0.0-dev.240424162023-9968b4889d",
+    "@mui/docs": "6.0.0-alpha.13",
```
LukasTy, line 28: `import { SectionTitle } from '@mui/docs/SectionTitle';` probably worked only because of the present alias.
Bumped the dependency version where this component is exported.

Janpot:
The alias for `@mui/docs` was intential as there is still a high coupling between docs code and `@mui/docs`. Unless this unblocks a specific problem I would keep it until all imports from `@mui/monorepo/docs/...` are removed

Janpot (approved):

Janpot:
Thank you 🙏

PR #3756, opened 2024-07-04
[core] Fix a few performance issues in Toolpad production apps

Janpot:
On top of https://github.com/mui/mui-toolpad/pull/3755

Address a few low-hanging fruits from the lighthouse report

* non-render-blocking fonts
* smaller font file
* viewport meta tag
* compression

Before:
Screenshot 2024-07-04 at 16 23 21

After:
Screenshot 2024-07-04 at 16 19 32

### DIFF
--- a/packages/toolpad-studio/src/server/toolpadAppBuilder.ts
+++ b/packages/toolpad-studio/src/server/toolpadAppBuilder.ts
@@ -8,6 +8,7 @@ import * as appDom from '@toolpad/studio-runtime/appDom';
 import type { ComponentEntry, PagesManifest } from './localMode';
 import { INITIAL_STATE_WINDOW_PROPERTY } from '../constants';
 import viteVirtualPlugin, { VirtualFileContent, replaceFiles } from './viteVirtualPlugin';
+import { FONTS_URL } from '../runtime/constants';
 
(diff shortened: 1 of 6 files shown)


packages/toolpad-studio/src/server/toolpadAppBuilder.ts
```diff
@@ -26,11 +27,10 @@ function getHtmlContent(entry: string) {
     <html lang="en">
       <head>
         <meta charset="utf-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1" />
         <title>Toolpad</title>
-        <link
-          rel="stylesheet"
-          href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700&display=swap"
-        />
+        <link rel="preload" href="${FONTS_URL}" as="style" onload="this.onload=null;this.rel='stylesheet'">
```
apedroferreira, line 32: Sorry for not reviewing earlier!
All looks good, not sure if these would also help with font loading performance:
```
<link rel="preconnect" href="https://fonts.googleapis.com" />
<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
```

Janpot: I don't think it would make a difference, the browser will already be loading the css asap.

PR #3757, opened 2024-07-04
[security] Slightly improve CSP header

oliviertassinari:
A starting point for a CSP rule with permissive rules that we can narrow down overtime:  https://stackoverflow.com/questions/35978863/allow-all-content-security-policy

Context: https://groups.google.com/a/mui.com/g/team-store/c/6tkWMOQ2ho0.


Janpot:
The problem seems to be that the iframe lives on about:blank which likely has some elevated policies being a chromium internal page. I didn't investigate deeper, just changed the test to go to https://example.com first and now it seems to pass.

Janpot (approved):

oliviertassinari:
Oh, I didn't think about this. You nailed it 👍, thanks

PR #3789, opened 2024-07-16
[docs] Improve changelog format

Janpot:
Add a template and instructions to the release docs that should result in better changelogs.


RELEASE.md
```diff
@@ -29,6 +29,36 @@
...
+
+### Core
+
+- Commit message (#pr-number) @author
+- ...
+
+<!-- CONTRIBUTORS -->
+```
+
+Strip the `[tag]` from the commit lines and order them in each section. Sections that don't have change can be removed altogether. If necessary, improve, redact or correct commit messages.
```
bharatkashyap, line 60:
```suggestion
Strip the `[tag]` from the commit lines and order them in each section. Sections that don't have changes can be removed altogether. If necessary, improve, redact or correct commit messages.
```

apedroferreira:
Looks great! Just noticed 2 things:
- The item numbering is off now, it starts again from 1 after this item
- In the previous list item there is spacing on the left side so that everything is aligned, I guess this item this item should be the same?

apedroferreira (approved):

PR #3794, opened 2024-07-17
[core] Tweak navigation API

Janpot:
As discussed during grooming

- Rename 'slug' to 'segment'.
- Remove leading slash from route segments.
- Extract navigation context.
- Extract Link component and use polymorphism in the List Button instead of nested anchor.

https://deploy-preview-3794--mui-toolpad-docs.netlify.app/toolpad/core/introduction/


packages/toolpad-core/src/DashboardLayout/DashboardLayout.test.tsx
```diff
@@ -63,30 +63,34 @@ describe('DashboardLayout', () => {
...
+        segment: 'integrations',
         title: 'Integrations',
         icon: <LayersIcon />,
       },
     ];
 
     const user = userEvent.setup();
     render(
-      <NavigationContext.Provider value={NAVIGATION}>
+      <AppProvider navigation={NAVIGATION}>
```
apedroferreira, line 91: Not important but for consistency I guess we could do the same with the `BrandingContext` in this file?

packages/toolpad-core/src/contexts/NavigationContext.ts
```diff
@@ -0,0 +1,33 @@
+import * as React from 'react';
+
+/**
+ * @ignore - internal component.
+ */
+
+export interface NavigationPageItem {
+  kind?: 'page';
+  title?: string;
```
apedroferreira, line 9: Are we intentionally making the `title` optional?
This title is being used for the list item titles (required prop in `ListItemText`, and also as part of the unique ids used for navigation items. But for the ids we could maybe use the item index instead - not sure why it's using the title but it was probably just the first approach.

apedroferreira (approved):
Looks great, thanks!
About the `title` comment I can adjust accordingly afterwards if needed after this is merged.

packages/toolpad-core/src/contexts/NavigationContext.ts
Janpot: we can fall back to the segment.

apedroferreira: Ok, I thought about that. We can do that fallback + I have to try it out but it should be possible to just remove the title from those ids.

PR #3809, opened 2024-07-19
[core] Fix theming bugs

apedroferreira:
Attempting to fix remaining theming bugs:

- [x] Docs and app use same storage key for theme mode
- [x] Console error (mismatching style tag) when app loads from server in dark mode if `defaultMode` is not set to `dark` (even though there seems to be no flash with wrong theme, I don't think this happened before MUI v6 either)


apedroferreira:
I'm pretty sure the console error mentioned above is because the playgrounds aren't using v6 of Material UI, so it should eventually be solved once https://github.com/mui/material-ui/issues/43001 is solved.
We can merge the fix for the other bug and this should be good!

Janpot (approved):

apedroferreira:
@Janpot just adding a few more small changes in a few minutes

PR #3838, opened 2024-07-24
[core] Fix docs segments and make navigation item segments optional

apedroferreira:
Was fixing some segments in docs and maybe we can make them optional since they're already being checked in a few places in `DashboardLayout` and falling back to `''`?


playground/nextjs-pages/src/pages/_app.tsx
```diff
@@ -23,12 +23,11 @@ const NAVIGATION: Navigation = [
     title: 'Main items',
   },
   {
-    segment: '/',
```
apedroferreira, line 26: This one can be omitted in this case, not sure if we want that.

Janpot (commented):
Does it error when there are duplicates?

e.g. 
* two page entries that have don't have `segment` defined
* What should be the meaning of 
    ```tsx
    [
      { kind: 'page', segment: 'hello', children: [ { kind: 'page' } ] }
    ]
    ```

Or more importantly. I was hoping we could build in the restriction of the navigation structure that for each possible path in the url there would exist at most one item in the whole hierarchy that corresponds to it. e.g. so that we can set the selected item correctly. Or so that we can match the breadcrumbs for page container. Is there any way we should guarantee this relationship? Or warn?

Janpot (approved):

apedroferreira:
> Does it error when there are duplicates?
> 
> e.g.
> 
> * two page entries that have don't have `segment` defined
> * What should be the meaning of
>   ```tsx
>   [
>     { kind: 'page', segment: 'hello', children: [ { kind: 'page' } ] }
>   ]
>   ```
> 
> Or more importantly. I was hoping we could build in the restriction of the navigation structure that for each possible path in the url there would exist at most one item in the whole hierarchy that corresponds to it. e.g. so that we can set the selected item correctly. Or so that we can match the breadcrumbs for page container. Is there any way we should guarantee this relationship? Or warn?

It works even with items that go to the same path, as for showing a selected item it just checks if their path matches the current pathname, and the item ids don't use the segment or titles, just indexes. So even if it's probably not ideal that a user sets duplicate paths, it's still possible and it works well.

I see that the `PageContainer` throws an error in these situations though, I guess you might have ran into some case that requires it. I could also throw a similar error in the same situation for these other components, probably, if we should make them consistent.

```
[
  { kind: 'page', segment: 'hello', children: [ { kind: 'page' } ] }
]
```

The item will lead to `hello /` which is probably what I would expect even if in practice it's a bit weird for the user to set up something like that. Let me know if you'd prefer any alternative.

Janpot:
> It works even with items that go to the same path, as for showing a selected item it just checks if their path matches the current pathname, and the item ids don't use the segment or titles, just indexes. So even if it's probably not ideal that a user sets duplicate paths, it's still possible and it works well.

It feels wrong to me to show two active items. I think we should at least show a warning in the console (in dev mode) that this is not supported. And maybe show only 1 active item e.g. the first one it encounters.
Maybe `PageContainer` shouldn't error, but warn as well and use the same behavior as the navigation.

apedroferreira:
> > It works even with items that go to the same path, as for showing a selected item it just checks if their path matches the current pathname, and the item ids don't use the segment or titles, just indexes. So even if it's probably not ideal that a user sets duplicate paths, it's still possible and it works well.
> 
> It feels wrong to me to show two active items. I think we should at least show a warning in the console (in dev mode) that this is not supported. And maybe show only 1 active item e.g. the first one it encounters. Maybe `PageContainer` shouldn't error, but warn as well and use the same behavior as the navigation.

Ok I think we can make it work like that, I'll try it out!

apedroferreira:
I've addressed the latest feedback with commit 1388f2798ceb6ce40e12d453049f55eb6fc95eb6

- Add top-level check for any duplicated navigation paths as an effect in `AppProvider`, console warn for each duplicate found
- Remove errors checks/logic about duplicated paths from `PageContainer` component
- Only show the first item as selected in UI if 2 navigation items link to the same path

packages/toolpad-core/src/AppProvider/AppProvider.tsx
```diff
@@ -110,6 +121,18 @@ function AppProvider(props: AppProviderProps) {
     window: appWindow,
   } = props;
 
+  React.useEffect(() => {
```
Janpot, line 124: To avoid an unnecessary effect, how about the following: During render, for each rendered navigation item, calculate the pathname, then look it up in a `Set`, if it already exists in the set, warn and don't mark it active. Add the pathname to the set.

playground/nextjs-pages/src/pages/_app.tsx
bharatkashyap: Might be good to omit for ease of understandability

apedroferreira:
- Added validation during render but it required a bit of a hack using `useMemo` where `useEffect` would be more normal for ideal behavior... couldn't find a good cleaner solution so far.
- I think it's best to show multiple items as selected after all if the user sets up duplicated paths - users should fix it quickly anyway, and if we pick just the first one for example every method that checks selected items needs to be aware of the whole navigation (at all its levels) to know which the selected item is and be able to do comparisons and it seems like more trouble than it's worth.

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -356,6 +349,13 @@ function DashboardLayout(props: DashboardLayoutProps) {
     }
   }, []);
 
+  // If useEffect was used, the reset would also happen on the client render after SSR which we don't need
+  React.useMemo(() => {
+    uniqueItemPaths = new Set();
```
Janpot, line 354: I don't think this can be safely done with concurrent mode. a render can be interrupted by a higher priority one and reset these.

Perhaps you can store the set in a ref and pass that as a prop to child components? then reset the ref in an effect.

apedroferreira: Refs work fine, I just wasn't sure about adding props just for this, but yeah global variables can fail in some rare cases so this is definitely better.

Changes here: https://github.com/mui/mui-toolpad/pull/3838/commits/c9269c6f81b377912c83f0f4e5bf9af5b6a37e53

Janpot: > (Neither would it work correctly with the theoretical case of having this component twice in the tree).

Actually have to come back on this, it's not theoretical, we are rendering this component multiple times in a page on the docs 😄

apedroferreira: I'll merge in a bit if this current approach with props seems good.

PR #3860, opened 2024-07-28
[cli] Bootstrap authentication from `create-toolpad-app`

### LINKED ISSUE #3861: Add CLI step for authentication

bharatkashyap:
- Closes #3861 

### Adds
- Authentication prompt questions to `create-toolpad-app`
Screenshot 2024-07-28 at 7 32 15 PM

### To Do
- [x] Bootstrap `.env.local` with empty secrets but valid keys
- [x] Handle`nextjs-pages` router branch 

### Questions
- [x] Instead of using strings to bootstrap files, create template folders for different scenarios and copy them based on user selection, instead of writing string content into files? This is [what `create-next-app` does](https://github.com/vercel/next.js/tree/canary/packages/create-next-app/templates), and it seems to be more readable and maintanable.
   - Keep using strings for now; it might scale better versus creating separate folder for each permutation of initial selection, and plays better with the webcontainers demo


packages/create-toolpad-app/src/templates/dashboardLayout.ts
```diff
@@ -0,0 +1,12 @@
+export default `import * as React from 'react';
```
Janpot, line 1: I could see us make each of those into a common shape like:

```tsx
interface File {
  content: string
}

type Template = (option: Options) => File
```

That way a template has input, and can also return metadata beyond content

Janpot (approved):

bharatkashyap:
### Update

- Added information about `.env.local` being created when auth options are selected, linking to Auth.js docs
Screenshot 2024-08-12 at 6 07 48 PM


- Added tests to check for auth options

packages/create-toolpad-app/src/templates/auth/nextjs-app/signInPage.ts
```diff
@@ -0,0 +1,51 @@
...
+          // Source: https://github.com/vercel/next.js/issues/49298#issuecomment-1542055642
+          // Detect a \`NEXT_REDIRECT\` error and re-throw it
+          if (error instanceof Error && error.message === 'NEXT_REDIRECT') {
+            throw error;
+          }
+          // Handle Auth.js errors
+          if (error instanceof AuthError) {
+            return {
+              error:
+                error.type === 'CredentialsSignin'
```
Janpot, line 33: If you don't have the credentials provider, this code path will never be taken

packages/create-toolpad-app/src/generateProject.ts
```diff
@@ -1,363 +1,171 @@
...
+import document from './templates/nextjs-pages/document';
+
+// Auth specific files for all apps
+import credentialsProvider from './templates/auth/credentialsProvider';
+import callbacks from './templates/auth/callbacks';
+import providerImport from './templates/auth/providerImport';
+import oAuthProvider from './templates/auth/oAuthProvider';
+import providerSetup from './templates/auth/providerSetup';
+import providerMap from './templates/auth/providerMap';
+import authImport from './templates/auth/import';
```
Janpot, line 33: IMO this goes way too far. Totally incomprehensible to figure out what's going on in an output file without jumping to three different source files.

packages/create-toolpad-app/src/templates/auth/envLocal.ts
```diff
@@ -0,0 +1,22 @@
...
+# Generate a secret with \`npx auth secret\`
+# and replace the value below with it
+AUTH_SECRET=secret
+
+# Add the CLIENT_ID and CLIENT_SECRET from your OAuth provider
+# to the .env.local file
+${oAuthProviders
+  .map(
+    (provider) => `
+${provider?.toUpperCase()}_CLIENT_ID=
```
Janpot, line 15: Only null-coalesce things when it makes sense. This one can't be null according to the types. And, if it could be `null` for some reason, that would result in `"null_CLIENT_ID"` which would be a bug anyway.


```suggestion
${provider.toUpperCase()}_CLIENT_ID=
```

Besides, you'd probably want to convert to constant case, to avoid `"azure-devops"` from turning into `"AZURE-DEVOPS"`

```tsx
// This function can go in @toolpad/utils/strings
function kebabToConstantCase(input: string): string {
  return input.split('-').map(part => part.toUpperCase()).join('_')
}
```

packages/create-toolpad-app/src/types.ts
```diff
@@ -0,0 +1,13 @@
+import { PackageJson } from './templates/packageType';
+
+export type SupportedAuthProvider = 'Credentials' | 'Google' | 'GitHub' | 'Facebook';
```
Janpot, line 3: Would it make sense to just use the same ids that auth.js uses?

packages/create-toolpad-app/src/templates/auth/auth.ts
```diff
@@ -0,0 +1,74 @@
...
+  return `import NextAuth from 'next-auth';
+  ${providers
+    .map((provider) => `import ${provider} from 'next-auth/providers/${provider?.toLowerCase()}';`)
+    .join('\n')}
+  import type { Provider } from 'next-auth/providers';
+
+  const providers: Provider[] = [
+    ${oAuthProviders
+      .map(
+        (provider) => `${provider}({
```
Janpot, line 16: You could branch here on whether it's the credentials provider to avoid having to traverse the provider array twice in
```tsx
const hasCredentialsProvider = providers.includes('Credentials');
const oAuthProviders = providers.filter((provider) => provider !== 'Credentials');
```

packages/create-toolpad-app/src/templates/nextjs-pages/indexPage.ts
```diff
@@ -0,0 +1,25 @@
+const indexPage = `import Link from 'next/link';
+import * as React from 'react';
+import Typography from '@mui/material/Typography';
+import Box from '@mui/material/Box';
+
+export default function HomePage() {
+  return (
+    <Box
+      sx={{
```
Janpot, line 9: TBH, if we feel the need to add any styles here, it feels like we're doing something wrong with `PageContainer`. How does it look when this is just a `<div>`?

packages/create-toolpad-app/src/templates/rootPage.ts
```diff
@@ -0,0 +1,25 @@
+const rootPage = `
+  "use client";
+  import Link from 'next/link';
+  import { Container, Typography, Box } from '@mui/material';
```
Janpot, line 4:
```suggestion
  import Container from '@mui/material/Container';
  import Typography from '@mui/material/Typography';
  import Box from '@mui/material/Box';
```

packages/create-toolpad-app/src/templates/auth/nextjs-pages/signIn.ts
Janpot, line 1: 'packages/create-toolpad-app/src/templates/auth/nextjs-pages/signIn.ts'

Just a note, but this folder structure makes it a bit harder to share things between pages router and app router. e.g. parts of the auth stuff will be identical.

packages/create-toolpad-app/src/templates/nextjs-pages/indexPage.ts
bharatkashyap: This is to be able to center the message on the screen; we can remove the styles from the `div` and instead move it to the `<Typography>`

packages/create-toolpad-app/src/templates/auth/nextjs-pages/signIn.ts
bharatkashyap: I've placed the parts that are identical in 'packages/create-toolpad-app/src/templates/auth/...' and then moved the router specific files to the respective folders

packages/create-toolpad-app/src/templates/nextjs-pages/indexPage.ts
Janpot: I would just not center the message

examples/core-auth-nextjs-pages/src/pages/index.tsx
```diff
@@ -1,21 +1,12 @@
...
-        my: 4,
-        display: 'flex',
-        flexDirection: 'column',
-        justifyContent: 'center',
-        alignItems: 'center',
-      }}
-    >
-      <Typography variant="h4" component="h1" sx={{ mb: 2 }}>
+    <div>
+      <Typography variant="h4" component="h1" sx={{ m: 2 }}>
```
Janpot, line 7: just text will be fine

```suggestion
```

examples/core-auth-nextjs-pages/src/pages/orders/index.tsx
```diff
@@ -1,21 +1,12 @@
...
-        my: 4,
-        display: 'flex',
-        flexDirection: 'column',
-        justifyContent: 'center',
-        alignItems: 'center',
-      }}
-    >
-      <Typography variant="h4" component="h1" sx={{ mb: 2 }}>
+    <div>
+      <Typography variant="h4" component="h1" sx={{ m: 2 }}>
```
Janpot, line 7:
```suggestion
```

Issue #3836, opened 2024-07-24
[docs] Toolpad Core DashboardLayout demo improvements


apedroferreira:
### Related page

https://mui.com/toolpad/core/react-dashboard-layout

### Kind of issue

Other

### Issue description

As mentioned in https://github.com/mui/mui-toolpad/pull/3776, we can try improving the `DashboardLayout` component docs as follows:
- making each code preview shorter so that it doesn't take up as much scroll space, and ideally the code preview is not tall enough to be scrollable
- split each feature more in a separate demo

### Context

Improving Toolpad Core docs.

**Search keywords**: components demos dashboard layout

oliviertassinari:
A benchmark for the docs of this: https://github.com/azouaoui-med/react-pro-sidebar.

PR #3864, opened 2024-07-30
Improve DashboardLayout navigation docs

apedroferreira:
Improve `DashboardLayout` component docs as requested in https://github.com/mui/mui-toolpad/issues/3836
Closes https://github.com/mui/mui-toolpad/issues/3836

**Docs:** https://deploy-preview-3864--mui-toolpad-docs.netlify.app/toolpad/core/react-dashboard-layout

- Split navigation items into separate examples
- Some segment fixes added while testing, even though I might already be adding them in another PR too (https://github.com/mui/mui-toolpad/pull/3838)
- Make demo code in examples shorter to make scrolling less frustrating (and make them more easy to read)
- Fix initially expanded navigation items (it was not working apparently) + make them work at any nested level, not just the top level

It's not as much as the "Files changed" count makes it seem - just a lot of duplicated changes there.


packages/toolpad-core/src/shared/navigation.tsx
```diff
@@ -8,3 +8,33 @@ export const isPageItem = (item: NavigationItem): item is NavigationPageItem =>
...
   return isPageItem(item) ? item.title ?? item.segment : item.title;
 };
+
+export function hasSelectedNavigationChildren(
+  navigationItem: NavigationItem,
+  basePath: string,
+  pathname: string,
+): boolean {
+  if (isPageItem(navigationItem) && navigationItem.children) {
+    const navigationItemFullPath = `${basePath}${basePath && !navigationItem.segment ? '' : '/'}${navigationItem.segment ?? ''}`;
```
apedroferreira, line 18: An utility for getting this path is coming in https://github.com/mui/mui-toolpad/pull/3838/files

docs/data/toolpad/core/components/dashboard-layout/DashboardLayoutNavigationActions.js
```diff
@@ -78,60 +93,45 @@ function DashboardLayoutNavigationActions(props) {
   // Remove this const when copying and pasting into your project.
   const demoWindow = window !== undefined ? window() : undefined;
 
+  const popoverMenuAction = (
```
bharatkashyap, line 96: I would recommend using tabs on this demo since it becomes very long to read through if expanded

Janpot: > You could perhaps benefit from some code sharing as well.

I'm not sure demos should do much code sharing. Ideally they're self-contained units. And as simple as possible. i.e. as simple as can be to convey the aspect that is being demoed.

bharatkashyap: Agree with ideally them being self-contained, but perhaps we need to optimise for readability if they grow too long?

Janpot: Advantage of single file demoes is that they can be simply copy+pasted from in one action (explains some of the popularity of shadcn). Splitting them across multiple files requires more context switching on behalf of the user to get the example integrated in their own code base. 
I disagree that splitting up in multiple files increases readability, to me it mostly creates extra friction to navigate the code. Instead we should think more in terms of "what can we remove from the code to get to the essence?" and focus on that to increase readability.
I believe the feature of multiple files in demoes should be used very sparingly. i.e. only when the split is mandated by the runtime (_document.js, _app.js) or perhaps to hold a demo dataset.

bharatkashyap: I think if we're optimising for copy and paste-ability, we should ideally: advertise it, and make the demo code blocks much larger. Given the size of our demo code blocks, I think the sense that this is meant to be copy and pasted doesn't immediately come out. (For instance, https://ui.shadcn.com/blocks as a benchmark has "Copy and Paste" as a verb in the header, and a very large code block). 

| Shadcn Blocks | Us |
|--------|--------|
| Screenshot 2024-08-01 at 5 58 57 PM | Screenshot 2024-08-01 at 5 58 26 PM | 

One possible solution is to come up with a flag that can render the MUI Docs Layout with navigation hidden, so that code can be much wider.

Within the context of our docs, I think that longer code can become harder for skimming. This is my opinion, okay to drop this if single files being okay for readability is the more popular opinion 👍🏻

apedroferreira: About this in specific, I tried to make the preview content as short as possible while still showcasing each scenario, but when intentionally expanded I feel like it's ok for things to be longer as long as most of the code is relevant to the example.

apedroferreira: > One possible solution is to come up with a flag that can render the MUI Docs Layout with navigation hidden, so that code can be much wider.

Yeah, seems like in some of these pages we might benefit from taking the full-width of the page as shadcn does, but so far I guess it hasn't been super necessary.

Janpot (approved):
This looks great, if possible I'd try to avoid scrolling in the demo preview area

docs/data/toolpad/core/components/dashboard-layout/DashboardLayoutNavigationActions.js
Janpot: > I think if we're optimising for copy and paste-ability, we should ideally: advertise it, and make the demo code blocks much larger.

We don't optimize for copy and paste-ability in an absolute sense. Definitely not the demo previews. They are too small for most demos to achieve that, they are to help the user decide whether the particular demo from a code point of view is going to be applicable to their problem. You could say the previews are mostly optimized for scanneability, that's why they should be short, to the point and not scrollable. If such a preview is not possible, then there should be likely just no preview. The expanded version should be more optimized for copy-pasteability. 

shadcn is different a different case, they basically act as a snippets library. Our demos are not snippets, we show the integration of our libraries. When I copy+paste from our docs I usually only copy portions, not the whole demo.

apedroferreira:
> This looks great, if possible I'd try to avoid scrolling in the demo preview area

Not sure what that means, nothing is scrollable in my screen I think...
Merging now but let me know and can improve it more later.

PR #3872, opened 2024-08-01
[core] Replace `CssVarsProvider` with `ThemeProvider`

siriwatknp:
To test https://github.com/mui/material-ui/pull/43115 and get feedback from Toolpad team.

## Context

The core team is merging the `CssVarsProvider` and `extendTheme` into `ThemeProvider` and `createTheme` to prevent the confusion of using the APIs. The detail is in this [issue](https://github.com/mui/material-ui/issues/43033).

## Changes

- Remove the CssVarsProvider because now ThemeProvider can handle with/without `colorSchemes` node.
- I see that Toolpad has a custom logic to receive a theme with `{ light: Theme, dark: Theme }` structure so this is untouched.
- The dualTheme can be checked by reading `{ allColorSchemes } = useColorScheme()` if `allColorSchemes.length > 1`.

For more detail of the implementation, see the Material UI PR: https://github.com/mui/material-ui/pull/43115

---


packages/toolpad-core/src/AppProvider/AppProvider.tsx
```diff
@@ -191,7 +202,7 @@ AppProvider.propTypes /* remove-proptypes */ = {
   }),
   /**
    * [Theme or themes](https://mui.com/toolpad/core/react-app-provider/#theming) to be used by the app in light/dark mode. A [CSS variables theme](https://mui.com/material-ui/experimental-api/css-theme-variables/overview/) is recommended.
-   * @default extendTheme()
+   * @default createTheme()
```
siriwatknp, line 205:
```suggestion
   * @default createTheme({ cssVariables: true, colorSchemes: { light: true, dark: true } })
```

The CSS variables feature is not enabled by default.

apedroferreira:
Having `allColorSchemes` seems good to me if we plan to support any number of color schemes with any names.

Since CSS vars themes are becoming standard I wouldn't even support an object with `{ light: Theme, dark: Theme }` in Toolpad anymore. Users would need to use a CSS vars theme if they wanted dual theming / a theme switcher and this way they would never be surprised by running into theme flash issues when server-side rendering, for example. I guess if we did this the implementation would be pretty straightforward?

siriwatknp:
> Since CSS vars themes are becoming standard I wouldn't even support an object with `{ light: Theme, dark: Theme }` in Toolpad anymore

Yes, if that's the case the `AppThemeProvider` can be removed. It will be much simpler.

Janpot:
@siriwatknp I updated the implementation a bit to make sure `useColorScheme` is never called when there is a v5 theme.

Janpot (approved):

packages/toolpad-core/package.json
```diff
@@ -60,8 +60,8 @@
     "prop-types": "15.8.1"
   },
   "devDependencies": {
-    "@mui/icons-material": "6.0.0-beta.2",
-    "@mui/material": "6.0.0-beta.2",
+    "@mui/icons-material": "next",
```
Janpot, line 63: Everything labeled as "next" will eventually get a stable v6 specifier

docs/data/toolpad/core/components/app-provider/AppProviderBasic.js
```diff
@@ -25,7 +25,11 @@ const NAVIGATION = [
   },
 ];
 
-const demoTheme = extendTheme({
+const demoTheme = createTheme({
+  cssVariables: {
+    colorSchemeSelector: 'data-toolpad-color-scheme',
```
apedroferreira, line 30: If this selector needs to be added every time I guess we could override it ourselves, or make it fully configurable. As right now I guess the `AppProvider` theming wouldn't work with any other selector.

apedroferreira (commented):
This looks great, thanks!

PR #3874, opened 2024-08-01
[docs] Improve Core tutorial

bharatkashyap:
- Bring the `core-tutorial` example in sync with the docs page
- Improve the structure of `core-tutorial` to make the tutorial work better, for instance: Add `PageContainer` to `dashboard/layout` instead of `dashboard/page/layout`
- Miscellaneous small improvements to the tutorial docs page
- Also add a `NavigateButton` to the example to improve interactivity on the first button in dev mode:

| **Before** | **After** |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/28a484b8-e979-40a4-a039-0c1e29d0b5fd" /> | <video src="https://github.com/user-attachments/assets/1e229cb0-7b36-4f12-8dc2-edbafbb7f8dc" /> |


Janpot:
> Improve the structure of core-tutorial to make the tutorial work better, for instance: Add PageContainer to dashboard/layout instead of dashboard/page/layout

I did this intentionally, to show how the composability works together with app router layouts

packages/create-toolpad-app/src/generateProject.ts
```diff
@@ -42,63 +42,67 @@ export default function generateProject(
...
   `;
 
-  const dashboardPage = `
+  const navigateButtonContent = `'use client';
+import * as React from 'react';
+import Link from 'next/link';
+import { LoadingButton } from '@mui/lab';
+import { Box } from '@mui/material';
+
+export default function NavigateButton() {
```
apedroferreira, line 86: I think it doesn't make sense to make this a separate file? Seems to specific, better for it to be in the intro page only?

examples/core-tutorial/app/layout.tsx
```diff
@@ -9,19 +10,21 @@ const NAVIGATION: Navigation = [
...
+    segment: 'page',
     title: 'Page',
     icon: <DashboardIcon />,
   },
 ];
 
 export default function RootLayout({ children }: Readonly<{ children: React.ReactNode }>) {
   return (
-    <html lang="en">
+    <html lang="en" data-toolpad-color-scheme="light">
```
apedroferreira, line 21: Matching the playground should be good here!

packages/create-toolpad-app/src/generateProject.ts
bharatkashyap: This is a client component, so – I could be wrong  – it must be in its own file with "use client" at the top

bharatkashyap:
> > Improve the structure of core-tutorial to make the tutorial work better, for instance: Add PageContainer to dashboard/layout instead of dashboard/page/layout
> 
> I did this intentionally, to show how the composability works together with app router layouts

The issue with this was that adding a second page required adding a `layout.tsx` as well to the page folder; the breadcrumbs and title would not automatically appear to the newly created page. (This is what happened for me, perhaps I missed a step while following the original tutorial)

examples/core-tutorial/app/layout.tsx
bharatkashyap: I think the playground app also contains the same attribute in the `app/layout.tsx` file, isn'it? That's what I can see here: https://github.com/mui/mui-toolpad/blob/master/playground/nextjs/src/app/layout.tsx#L27

Janpot:
> The issue with this was that adding a second page required adding a layout.tsx as well to the page folder; the breadcrumbs and title would not automatically appear to the newly created page. (This is what happened for me, perhaps I missed a step while following the original tutorial)

Usually, the features of the CLI should drive the tutorial content, not the other way around. But I'm not heavily against the change, we can go with it for now.

examples/core-tutorial/app/layout.tsx
apedroferreira: Yep that's what I meant, was just confirming that it seems good!

packages/create-toolpad-app/src/generateProject.ts
apedroferreira: I think you can have `'use client'` as the first line inside the component?

But if in terms of file structure you think this makes more sense anyway it's not a big deal!

apedroferreira (approved):

PR #3893, opened 2024-08-05
[docs] Fix issues with DashboardLayout in Toolpad Core homepage

apedroferreira:
Noticed some issues with the dashboard layout in the Toolpad Core homepage:

- [x] When the mobile menu opens the app bar becomes empty
- [x] Unnecessary spacing above `PageContainer`
- [x] Spacing was off in code examples and did not match the actual example code

This PR should fix these issues - it's just not fixing the fact that the grid inside the `PageContainer` is changing width when the menu is expanded - if that's not addressed in https://github.com/mui/mui-toolpad/pull/3848 I can try to fix that too.

https://deploy-preview-3893--mui-toolpad-docs.netlify.app/toolpad/core/

https://github.com/user-attachments/assets/7ef674a8-2bf4-453d-913a-9409225d02cb

PR #3905, opened 2024-08-07
[core] Fix some DashboardLayout bugs and make some docs examples more consis…

apedroferreira:
Fixes the issues mentioned here https://github.com/mui/mui-toolpad/pull/3893#issuecomment-2269324520:

- Fix scrollbar color not always being set when used in iframes with different color schemes (such as home page)
- Fix some possible issues when the layout is used in iframes (it required adding some styles that in most cases would not be needed, though, with `100vw` to set an explicit width)
- Make some docs examples more consistent with other examples in the page they're in

https://deploy-preview-3905--mui-toolpad-docs.netlify.app/toolpad
https://deploy-preview-3905--mui-toolpad-docs.netlify.app/toolpad/core/react-dashboard-layout/


docs/data/toolpad/core/components/account/AccountWithDashboard.js
```diff
@@ -1,15 +1,63 @@
...
+import PropTypes from 'prop-types';
 import Box from '@mui/material/Box';
-import { AppProvider, DashboardLayout } from '@toolpad/core';
+import Typography from '@mui/material/Typography';
+import { extendTheme } from '@mui/material/styles';
+import DashboardIcon from '@mui/icons-material/Dashboard';
+import { AppProvider } from '@toolpad/core/AppProvider';
+import { DashboardLayout } from '@toolpad/core/DashboardLayout';
+
+const NAVIGATION = [
```
apedroferreira, line 10: Yeah it's adding all this boilerplate here too but this way it matches the other examples...

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -466,9 +466,15 @@ function DashboardLayout(props: DashboardLayoutProps) {
...
+      <Box
+        component="main"
+        sx={{
+          flexGrow: 1,
+          minWidth: { xs: isMobileNavigationOpen ? '100vw' : 'auto', md: 'auto' },
+        }}
+      >
         <Toolbar />
-        <div>{children}</div>
+        {children}
```
apedroferreira, line 477: I guess we don't need a `div` here (unrelated though)

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -466,9 +466,15 @@ function DashboardLayout(props: DashboardLayoutProps) {
       >
         {drawerContent}
       </Drawer>
-      <Box component="main" sx={{ flexGrow: 1 }}>
+      <Box
+        component="main"
+        sx={{
+          flexGrow: 1,
+          minWidth: { xs: isMobileNavigationOpen ? '100vw' : 'auto', md: 'auto' },
```
Janpot, line 473: Let's add a comment to the github issue on mui core repository that tracks the issues with the responsive drawer demo.

Janpot (approved):
Looks good. Just add a comment so that we remember what this hack is for and know when we can revert it.

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
apedroferreira: Raised the issue in https://github.com/mui/material-ui/issues/43244 and added todo comments in https://github.com/mui/mui-toolpad/pull/3905/commits/5658dc02bbb7ddbdae6a513e1ad1fa8813d0521e

PR #3970, opened 2024-08-20
[docs] Add signed in state as default on `<Account />` docs

### LINKED ISSUE #3969: Add Account example on the component page in the logged in state
Image

* Show an example of the logged in state first
* Mention that the logged in button is shown when there is no active session available, but show the logged in state as a primary example.

**Search keywords**:

bharatkashyap:
- Fixes #3969 

[preview](https://deploy-preview-3970--mui-toolpad-docs.netlify.app/toolpad/core/react-account/)


apedroferreira (changes requested):
For some reason in https://deploy-preview-3970--mui-toolpad-docs.netlify.app/toolpad/core/react-account/ the demo appears logged out to me at the start. If I click the code editor for example, it changes to a logged-in state.

Also I'm not sure if the preview code shown for the first demo is explanatory enough without being expanded, but not sure about this... I guess it's useful for people to know the format of the `session` object from there, but as for the `signIn` function I guess that's covered in another component so it's fine?

bharatkashyap:
> For some reason in https://deploy-preview-3970--mui-toolpad-docs.netlify.app/toolpad/core/react-account/ the demo appears logged out to me at the start. If I click the code editor for example, it changes to a logged-in state.
> 
> Also I'm not sure if the preview code shown for the first demo is explanatory enough without being expanded, but not sure about this... I guess it's useful for people to know the format of the `session` object from there, but as for the `signIn` function I guess that's covered in another component so it's fine?

You're right, although I can't replicate during local development. Looking into this.

apedroferreira (approved):
Looks good!

Does the script check for unstaged files only? Maybe that would be better if not but not even sure, anyway I think this is a helpful change, helps not always having to run the docs scripts when working on the docs!
As an alternative, `docs:dev` could maybe also just watch for live changes in those files and run the script every time they change, but that might be too heavy so then in that case it should probably just generate files for the documentation that changed. But that's just a possible idea for a future improvement!

bharatkashyap:
Removed the script proposal in favour of using the `--watch` mode on `pnpm docs:typescript:formatted`, which is proposed to be better documented and improved by the code-infra team.

Janpot:
Few questions
* Can we have it demo signed in and signed out? Either in the same demo or in a different one.
* Why is there such a huge vertical margin in the demo?
* For the customization demo, can we show a signed in one next to a signed out one?

bharatkashyap:
> Few questions
> 
> * Can we have a demo signed in and one signed out? Either in the same demo or in a different one.
I'll update the "States" sub section to have two demos, one titled "Signed In" and one titled "Signed out"

> * Why is there such a huge vertical margin in the demo?

I'm using the [`bg: gradient`](https://next.mui.com/experiments/docs/demos/#quot-bg-quot-gradient) property which adds the margins by default, but I can switch to using `bg: outlined`


> * For the customization demo, can we show a signed in one next to a signed out one?

Do you mean showing the avatar next to the Login button, something like this?

Screenshot 2024-08-29 at 8 40 05 PM

Janpot:
> Do you mean showing the avatar next to the Login button, something like this?

yes, it underneath each other. or with a label even

bharatkashyap:
> > Do you mean showing the avatar next to the Login button, something like this?
> 
> yes, it underneath each other. or with a label even

Screenshot 2024-08-29 at 9 04 30 PM ?

Issue #3981, opened 2024-08-22
Auth.js v4 example


bharatkashyap:
More than 2/3 of their users are on latest v4

**Search keywords**:

PR #3982, opened 2024-08-22
[docs] Add `next-auth` v4 example

bharatkashyap:
- Closes https://github.com/mui/mui-toolpad/issues/3981


examples/core-auth-nextjs-pages-nextauth-4/src/pages/auth/signin.tsx
```diff
@@ -0,0 +1,96 @@
...
+}
+
+export default function SignIn({
+  providers,
+}: InferGetServerSidePropsType<typeof getServerSideProps>) {
+  const router = useRouter();
+  return (
+    <SignInPage
+      providers={providers}
+      signIn={async (provider, formData, callbackUrl) => {
```
apedroferreira, line 26: I would separate this function from the rendering not to mix up the logic with the view so much, makes it more readable. Just personal preference though!

apedroferreira (approved):

apedroferreira:
Oh just one more thing does it have to use the Next.js pages router? I thought most examples should just use the app router as it's the most recent one.
Still, the example looks great!

apedroferreira:
Also, how can I run the example in my development workflow? `pnpm install` seems to install in the whole project scope of the monorepo.
If we don't have instructions for doing that with a Toolpad Core example I guess we could add them later separately?
Also not sure if that means that Core and Studio examples should be separated in different folders.

bharatkashyap:
> Oh just one more thing does it have to use the Next.js pages router? I thought most examples should just use the app router as it's the most recent one. Still, the example looks great!

I used the pages router assuming most people using `next-auth@4` would be interested in maintaining older apps they created with the pages router, but I can add a version with the app router as well. Perhaps we can introduce the choice to choose between `next-auth` versions in the CLI, as well. Wdyt @Janpot?

bharatkashyap:
> Also, how can I run the example in my development workflow? `pnpm install` seems to install in the whole project scope of the monorepo. If we don't have instructions for doing that with a Toolpad Core example I guess we could add them later separately? Also not sure if that means that Core and Studio examples should be separated in different folders.

For testing during development, the method would be akin to testing the playground apps, i.e., 
  - modifying the `@toolpad/core` dependency in the example to `workspace/*`
  - adding the example to the workspace in `pnpm-workspace.yml`, like `examples/<example-name>`
  - run the project in a separate process through `pnpm --filter @toolpad/core dev`
  - running `pnpm dedupe && pnpm --filter examples/<example-name> dev` to run the example
  
 Don't know if this merits space in the contributing instructions?

apedroferreira:
> > Oh just one more thing does it have to use the Next.js pages router? I thought most examples should just use the app router as it's the most recent one. Still, the example looks great!
> 
> I used the pages router assuming most people using `next-auth@4` would be interested in maintaining older apps they created with the pages router, but I can add a version with the app router as well. Perhaps we can introduce the choice to choose between `next-auth` versions in the CLI, as well. Wdyt @Janpot?

That's a good point, feel free to stick with that option then, I guess we can't really know for sure what's best for now.

apedroferreira:
> > Also, how can I run the example in my development workflow? `pnpm install` seems to install in the whole project scope of the monorepo. If we don't have instructions for doing that with a Toolpad Core example I guess we could add them later separately? Also not sure if that means that Core and Studio examples should be separated in different folders.
> 
> For testing during development, the method would be akin to testing the playground apps, i.e.,
> 
> * modifying the `@toolpad/core` dependency in the example to `workspace/*`
> * adding the example to the workspace in `pnpm-workspace.yml`, like `examples/<example-name>`
> * run the project in a separate process through `pnpm --filter @toolpad/core dev`
> * running `pnpm dedupe && pnpm --filter examples/<example-name> dev` to run the example
> 
> Don't know if this merits space in the contributing instructions?

Probably not super urgent but the current instructions there are about Toolpad Studio only.
We should probably cover all of this in a single more generic section now.

CONTRIBUTING.md
```diff
@@ -23,6 +23,52 @@ _If you're looking into contributing to the docs, follow the [instructions](#bui
...
+
+   ```bash
+   cd playground/nextjs
+   ```
+
+   ```bash
+   pnpm dev
+   ```
+
+### Test changes to Toolpad Core on a live application
```
apedroferreira, line 54: Isn't this basically the same information as the section above?

CONTRIBUTING.md
```diff
@@ -23,6 +23,52 @@ _If you're looking into contributing to the docs, follow the [instructions](#bui
...
+   pnpm --filter @toolpad/core dev
+   ```
+
+2. Run the playground app
+
+   ```bash
+   pnpm --filter playground-nextjs dev
+   ```
+
+You should now be able to test changes to Toolpad Core on a live application. Follow similar steps to run apps in the `examples` folder after adding them to the workspace in `pnpm-workspace.yaml`.
```
apedroferreira, line 70: This is the part that should be more step by step and covered for both Core and Studio in my opinion.
But it's not really related with this PR, we can just create an issue!

apedroferreira:
> Updated to add instructions to run the playground apps (and examples) for Toolpad Core.
> 
> I've explored integrating `next-auth` v4 with the app router, and the chief problem is that they do not provide a server-side `signIn` equivalent function; the recommended approach is to use your own, wherein there continue to be multiple issues: [nextauthjs/next-auth#45](https://github.com/nextauthjs/next-auth/issues/45)

Stick with the pages router should be cool!

Issue #3983, opened 2024-08-22
DashboardLayout slot for Toolbar/Actions area


apedroferreira:
* Make top-right area that contains the Account button configurable.
* Let's call it Toolbar or actions or actions area,...
* By default it contains the account button if authentication is enabled, or empty if not
* In the future we can add e.g. search bar if (hypothetical) search provider is available. or Bell icon if (hypothetical) alerts provider is available.
* Spiritually let's model it after the [DataGrid Toolbar](https://mui.com/x/react-data-grid/export/#custom-toolbar)

**Search keywords**:

PR #3984, opened 2024-08-22
[core] Add toolbarActions and toolbarAccount slots to DashboardLayout

apedroferreira:
Closes https://github.com/mui/mui-toolpad/issues/3983

Adds 2 new slots to the `DashboardLayout` component:
- `toolbarActions`: for customizing toolbar actions to the left of the theme switcher;
- `toolbarAccount`: to be able to replace/customize the `Account` component to the right of the theme switcher.

Tried to follow how some other MUI libraries do things and the slots we already have in the `SignInPage` component.

Also adds a "Slots" section in the API docs pages for `DashboardLayout` and `SignInPage` by exporting Slots type interfaces in the respective components (they just needed to be named `DashboardLayoutSlots`, for example, to work automatically).

The docs for DashboardLayout now include a simple example of how to add a search bar to the toolbar actions.

https://deploy-preview-3984--mui-toolpad-docs.netlify.app/toolpad/core/react-dashboard-layout/
https://deploy-preview-3984--mui-toolpad-docs.netlify.app/toolpad/core/api/dashboard-layout/
https://deploy-preview-3984--mui-toolpad-docs.netlify.app/toolpad/core/api/sign-in-page/


bharatkashyap (approved):
Looks good! I thought it would improve the feature if we added a section in the `DashboardLayout` docs page with an example on how to use this prop with a custom component - what do you think? 

Not a blocker and can be done in a follow up

apedroferreira:
> Looks good! I thought it would improve the feature if we added a section in the `DashboardLayout` docs page with an example on how to use this prop with a custom component - what do you think?
> 
> Not a blocker and can be done in a follow up

I took a look at the MUI X docs and I think they usually cover this kind of things just with the API reference pages.
Not sure if showing demos for every single feature like this isn't overkill as opposed to just the most common use cases - could be something up for discussion.

Janpot:
> I took a look at the MUI X docs and I think they usually cover this kind of things just with the API reference pages.
Not sure if showing demos for every single feature like this isn't overkill as opposed to just the most common use cases - could be something up for discussion.

I believe the components pages are also about guiding people towards the right solution for their problem, rather than only demoing our features. If there's a specific use-case that is well solved by this slot then it would make sense to show that as a demo.

apedroferreira:
TODO: Add demo for slot with search bar example

apedroferreira:
Ended up adding 2 new separate slots actually: `toolbarActions` and `toolbarAccount`, as these are separated by the theme switcher and anyway logically it seems to make sense.

packages/toolpad-core/src/DashboardLayout/ToolbarActions.tsx
```diff
@@ -0,0 +1,10 @@
+'use client';
+
+/**
+ * @ignore - internal component.
+ */
+function ToolbarActions() {
```
apedroferreira, line 6: Placeholder for later, but could just be null directly too.

apedroferreira:
Gonna merge this one, lmk if you have any further feedback!

PR #3992, opened 2024-08-23
[core] Refactor `<Account />` 

### LINKED ISSUE #3971: [core] Polish `<Account />` component and docs
Image

 
- [ ] unlock the ability to add content between name/email and the popup actions
- [ ] ability to customize name/email section
- [ ] ability to customize the actions section
- [ ] Fix issues with the `<Popover />` anchor on the docs

To consider:
* slot for the avatar
* slot for the log in button

**Search keywords**:

bharatkashyap:
-  Fixes #3971 
- Adds `slots` for `signInButton` and `signOutButton` and `menuItems`
- Fixes an `anchorEl` issue with the previous `<Account />` 
- Adds a custom menu demo to the docs

https://deploy-preview-3992--mui-toolpad-docs.netlify.app/toolpad/core/react-account/


Janpot:
This component could be much more than a simple menu. It feels like a regression to me.

Screenshot 2024-08-27 at 10 32 24
to
Screenshot 2024-08-27 at 10 32 29

bharatkashyap:
> This component could be much more than a simple menu. It feels like a regression to me.
> 
> Screenshot 2024-08-27 at 10 32 24 to Screenshot 2024-08-27 at 10 32 29

I'll explore continuing to use the `Popover`, but I feel like the account/user button component is almost always going to be a list of options and text visually separated by dividers and icons on a temporary surface. Seems apt to be a `Menu`.

prakhargupta1:
I checked two instances of it being a menu (my best guess) component. Since, it's already done I would suggest that we continue with it. In the future, we can revisit it.  

Zendesk:
Screenshot 2024-08-29 at 1 21 13 PM

HiBob:
Screenshot 2024-08-29 at 1 20 54 PM

Janpot:
Nothing should prevent you from just rendering a [`MenuList`](https://mui.com/material-ui/react-menu/#icon-menu) in the slot. Why would we restrict every usage of this component to be this one?

Screenshot 2024-08-28 at 15 18 31

Janpot:
Nothing should prevent you from just rendering a [MenuList](https://mui.com/material-ui/react-menu/#icon-menu) in the slot. Why would we restrict every usage of this component to be this one?

Screenshot 2024-08-28 at 15 18 31

prakhargupta1:
> Nothing should prevent you from just rendering a [MenuList](https://mui.com/material-ui/react-menu/#icon-menu) in the slot. Why would we restrict every usage of this component to be this one?

Ah ok, so I guess it's about whether to give the option of a slot as menu item versus changing it to a popover component that can show any UI (like the Google UI you shared) through a slot. 
I still think, it could be done in the next iteration, based on user demand.

Janpot:
> Ah ok, so I guess it's about whether to give the option of a slot as menu item versus changing it to a popover component that can show any UI (like the Google UI you shared) through a slot.

No, it's already a popover, this PR is changing it to a menu

prakhargupta1:
> No, it's already a popover, this PR is changing it to a menu

Oh ok, in that case, I agree, a popover would be a more versatile choice.

bharatkashyap:
Updated to use `Popover` instead of `Menu` by default

packages/toolpad-core/src/Account/Account.tsx
```diff
@@ -47,109 +65,140 @@ export interface AccountProps {
  * - [Account API](https://mui.com/toolpad/core/api/account)
  */
 function Account(props: AccountProps) {
-  const { slotProps, signInLabel = 'Sign In', signOutLabel = 'Sign Out' } = props;
-  const session = React.useContext(SessionContext);
-
+  const { slots, slotProps, signInLabel = 'Sign In', signOutLabel = 'Sign Out' } = props;
```
Janpot, line 68: How does X handle customizing their labels in e.g. the datagrid?

packages/toolpad-core/src/Account/Account.tsx
```diff
@@ -47,109 +65,140 @@ export interface AccountProps {
...
     <React.Fragment>
-      <IconButton
-        aria-describedby={id}
-        onClick={handleClick}
-        aria-label="Current User"
-        {...slotProps?.iconButton}
-      >
-        <SessionAvatar session={session} sx={{ width: 32, height: 32, ...slotProps?.avatar }} />
-      </IconButton>
+      <Box sx={{ display: 'flex', alignItems: 'center', textAlign: 'center' }}>
```
Janpot, line 109: We could consider starting adopting the habit of avoiding `Box` components?
```
styled('div'){{ display: 'flex', alignItems: 'center', textAlign: 'center' }}
```

docs/data/toolpad/core/components/account/account.md
```diff
@@ -12,18 +12,20 @@ The `Account` component is a quick and easy way to display an account management
...
+{{"demo": "AccountDemo.js", "bg": "outlined" }}
 
 ## Customization
 
-### Components
+### Slot Props
 
 `Account` can take different labels for the sign in and sign out buttons through the `signInLabel` and `signOutLabel` props. Deeper changes can be made by passing in `slotProps` to the underlying components.
 
-{{"demo": "AccountCustom.js", "bg": "gradient" }}
```
Janpot, line 25: Pure visually, I wonder if it makes sense to show those demos inside of an empty `AppBar`.

packages/toolpad-core/src/Account/Account.tsx
bharatkashyap: The closest example I can find in the grid is customising the pagination labels https://mui.com/x/react-data-grid/pagination/#page-size-options

Janpot: https://mui.com/x/react-data-grid/localization/#locale-text

bharatkashyap: Do you mean we should drop the `signInLabel` and `signOutLabel` props in favour of a `LabelsProvider` of some sort?

Janpot: The grid supports a `localeText` property. We could opt for following the same convention? It also reads those from the theme. We can add support for that later.

packages/toolpad-core/src/Account/Account.tsx
```diff
@@ -1,39 +1,67 @@
...
+}));
 
 export interface AccountProps {
   /**
-   * Props to pass to the constituent components in the Account component.
-   * @default {}
-   * @example { signInButton: { color: 'primary' }, signOutButton: { color: 'secondary' } }
+   * The components used for each slot inside.
+   */
+  slots?: {
```
apedroferreira, line 34: You need to export the Slots in their own interface to automatically generate the documentation for them.
Take a look at https://github.com/mui/mui-toolpad/blob/master/packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx, line 253, should be a simple change!

docs/pages/toolpad/core/api/account.json
```diff
@@ -1,13 +1,21 @@
 {
   "props": {
-    "signInLabel": { "type": { "name": "string" }, "default": "'Sign In'" },
-    "signOutLabel": { "type": { "name": "string" }, "default": "'Sign Out'" },
+    "localeText": {
+      "type": { "name": "shape", "description": "{ signInLabel: string, signOutLabel: string }" },
+      "default": "DEFAULT_LABELS"
```
apedroferreira, line 5: Does MUI X also list constants as defaults like this, even if they're not accessible for users?
Just curious, as not sure if there's a better way, but if it's following an existing pattern I guess it's fine!

apedroferreira (changes requested):
Looks cool! Got some feedback mostly on docs/design, nothing major except generating the slots documentation automatically by exporting a type interface!

### Some design feedback on this demo:

Screenshot 2024-09-06 at 18 51 14

- The items aligned to the center don't look very good, it's better if they're aligned to the left.
- There's too much space between the picture and the title and description, I think you can decrease the margin a bit! Not that important though if this is a default.
- A Popover that opens another popover doesn't seem great, I guess a single list should work and be simpler too.

### Some feedback on the new slots:

On a first read I'm not exactly sure what the `menuItems` slot replaces. Is it everything inside the Popover? Just the middle area? Maybe we could make it a bit clearer, even if just with the text in the docs.

Also, we're using `localeText` inside `slotProps` to customize labels, right? The current version of the preview docs seems not to be updated with that.

docs/pages/toolpad/core/api/account.json
bharatkashyap: Yep: https://github.com/mui/mui-x/blob/master/packages/x-data-grid/src/constants/localeTextConstants.ts

bharatkashyap:
> Looks cool! Got some feedback mostly on docs/design, nothing major except generating the slots documentation automatically by exporting a type interface!

Agree with almost everything here and addressed it!

> A Popover that opens another popover doesn't seem great, I guess a single list should work and be simpler too.

Added this to demonstrate how one would be able to create such a menu if they want to. It's not an uncommon UX pattern, and I feel it should be okay to include as an example in the docs? Let me know what you think about that.

docs/data/toolpad/core/components/account/account.md
```diff
@@ -26,12 +26,16 @@ When signed out, the component renders as an inline sign in button within the da
 
 ## Customization
 
-### Components
+### Slot Props
 
-`Account` can take different labels for the sign in and sign out buttons through the `signInLabel` and `signOutLabel` props. Deeper changes can be made by passing in `slotProps` to the underlying components.
+The underlying `signInButton`, `signOutButton` and `iconButton` components can be customized by passing in `slotProps` to the `Account` component.
+
+Labels for the sign in and sign out buttons through the `localeText` prop.
```
apedroferreira, line 33: I think this sentence about the labels is missing something ("can be customized?").

PR #3929, opened 2024-08-12
[docs] Improve textfield docs and re-establish links to Material UI components

prakhargupta1:
Address: https://github.com/mui/mui-toolpad/issues/3675#issuecomment-2178553160

PR #4012, opened 2024-08-28
[studio] Explain props of text-field component with demos

prakhargupta1:
Addressing the text field part from https://github.com/mui/mui-toolpad/pull/3929 in this separate PR.


Janpot (approved):
Thank you

PR #4013, opened 2024-08-28
[docs] Link docs from component demos on Toolpad Core landing page

prakhargupta1:
Preview: https://deploy-preview-4013--mui-toolpad-docs.netlify.app/toolpad/


bharatkashyap:
I think we could go with this kind of an overlay on the code for the links: much more unmissable and attractive:
Screenshot 2024-08-29 at 1 47 59 PM

prakhargupta1:
I don't think it will bring much difference.

docs/src/components/landing/ToolpadFeaturesSwitcher.tsx
```diff
@@ -57,6 +61,12 @@ function ComponentItem({
             {description}
           </Typography>
         )}
+        {link && (
+          <Link href={link.toString()} underline="hover" variant="body2" target="_blank">
+            View docs
```
Janpot, line 66:
```suggestion
            View the docs
```

to mimic X https://mui.com/x/

Janpot (commented):
> I think we could go with this kind of an overlay on the code for the links: much more unmissable and attractive:

Personally I find a link more intuitive than a button as a way to get to a page in our documentation. This follows the [X landing page](https://mui.com/x/), it feels better to me than to core landing page.

docs/src/components/landing/ToolpadFeaturesSwitcher.tsx
```diff
@@ -3,7 +3,9 @@ import dynamic from 'next/dynamic';
 import Box from '@mui/material/Box';
 import Stack from '@mui/material/Stack';
 import Typography from '@mui/material/Typography';
+import Link from '@mui/material/Link';
```
Janpot, line 6: This page lives in the same Next.js app as the docs, so it's an internal link. We could avoid a full navigation if we use the Next.js router. Is it possible to instead use
```suggestion
import { Link } from '@mui/docs/Link';
```

docs/src/components/landing/ToolpadFeaturesSwitcher.tsx
```diff
@@ -57,6 +61,12 @@ function ComponentItem({
             {description}
           </Typography>
         )}
+        {link && (
+          <Link href={link.toString()} underline="hover" variant="body2" target="_blank">
```
Janpot, line 65:
```suggestion
          <Link href={link.toString()} underline="hover" variant="body2">
```

prakhargupta1: Got it, that's how it's in MUI X.

Janpot (approved):

PR #4063, opened 2024-09-09
[core] Add `passkey` provider support and example

bharatkashyap:
- Fixes https://github.com/mui/mui-toolpad/issues/3989
- https://deploy-preview-4063--mui-toolpad-docs.netlify.app/toolpad/core/react-sign-in-page/#passkey


docs/data/toolpad/core/components/sign-in-page/PasskeySignInPage.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+import { AuthProvider, AppProvider, SignInPage } from '@toolpad/core';
+import { useTheme } from '@mui/material/styles';
+// preview-start
+const providers = [{ id: 'passkey', name: 'Passkey' }];
+// preview-end
+
+const signIn: (provider: AuthProvider) => void = async (provider) => {
```
apedroferreira, line 8: Would it be possible for the demo to still show some kind of feedback when we submit the form like the other ones do?

apedroferreira (approved):
Seems all good!
Just 2 non-blocking comments/suggestions but not 100% sure about them as you might have had a reason to do things that way.

Janpot (approved):

PR #4076, opened 2024-09-11
[infra] Adds reusable workflow for closing messages

michelengelen:
related issue: https://github.com/mui/mui-public/issues/156

### DIFF
--- a/.github/workflows/closed-issue-message.yaml
+++ b/.github/workflows/closed-issue-message.yaml
@@ -1,0 +1,16 @@
+name: Add closing message to issue
+
+on:
+  issues:
+    types:
+      - closed
+
+permissions: {}
+
+jobs:
+  add-comment:
+    name: Add closing message
+    uses: mui/mui-public/.github/workflows/issues_add-closing-message.yml@master
+    permissions:
+      contents: read
+      issues: write


Janpot:
@michelengelen What I meant was for example: I closed [this issue](https://github.com/mui/material-ui/issues/43700). Which doesn't generate a notification. But then I still get pinged 1 minute later for the closing message.

michelengelen:
> @michelengelen What I meant was for example: I closed [this issue](https://github.com/mui/material-ui/issues/43700). Which doesn't generate a notification. But then I still get pinged 1 minute later for the closing message. I'm subscribed to the issue, so I get notifications, even if I'm not tagged.

Well, yeah ... that one is unfortunately necessary for this... its more a fault of how github handles notifications and so on though ... there is very little customizability for notifications in place

oliviertassinari:
@Janpot Right, so it does this:

SCR-20240916-qlny

I think a Gmail filter does the trick. Guide updated: https://www.notion.so/mui-org/GitHub-guide-a2d05d63152244cfa568927deb34b25b?pvs=4#103cbfe7b66080758757c4e30bb6f9f1. I have implemented it on my end.

PR #4077, opened 2024-09-11
[code-infra] Replace issue cleanup workflow with centralized one

michelengelen:
related issue: https://github.com/mui/mui-public/issues/193


michelengelen:
> > It simply unifies the cleanup approach we have on all mui repos
> 
> Cleanup of what? From the issue description I can't get any idea on what this PR is going to change to our repo and from the linked issue neither. Would it be possible to be a bit more descriptive, or point to a description?

Sry if you did not know what this change is about.

It will technically change nothing only use the reusable workflow from the mui-public repo instead. 

The original workflow was there for triaging and formatting the issue body when a new issue got opened. In an effort to centralize and normalize that we moved a few workflows to the mui-public repo and made them reusable, so we have 2 benefits:
1. whenever we need to change something in those workflows we only have to do it once
2. we normalize/unify our approach and wording across the public repos

This has been discussed a few times already and also mentioned on the monthly meeting, but I do agree that I could have done a better job at providing a description. Sry about that!
I'll also add the above section to the linked issue.

Janpot (approved):
Thank you, got it now. I initially overlooked that this PR is replacing a workflow, not just adding one. Sry for being a nuisance 🙂.

PR #4113, opened 2024-09-18
Update/correct toolpad repository links in package.json

Janpot:
Use shorthand as per https://docs.npmjs.com/cli/v10/configuring-npm/package-json#repository
Add `directory` property to each


oliviertassinari (commented):
Different from the other package.json, but fair, it works as well

apedroferreira (approved):

PR #4114, opened 2024-09-18
[PageContainer] Make PageContainer customizable for dynamic routes

Janpot:
Closes https://github.com/mui/toolpad/issues/4100

Adds a `title` and `breadCrumbs` override. Also supplies a `useActivePage` to retrieve the current value for `title` and `breadCrumbs` so they can be extended.

https://deploy-preview-4114--mui-toolpad-docs.netlify.app/toolpad/core/react-page-container/#dynamic-routes


bharatkashyap (approved):
Great feature addition! Intuitive to use for the most part.

I think we could benefit greatly from a code snipper or demo showcasing how this `useActivePage` could be used in a Next.js app with the `next/navigation` primitives. Does it make sense to add one on this page itself, or a separate page to document `useActivePage`?

Janpot:
Added a snippet to the docs on how this would integrate with the next.js app router

Janpot:
> What am I doing wrong?

It looks like you're calling `useActivePage` outside of `ReactRouterAppProvider`. The hook needs an app provider to know which pages are available.

tekknow:
Ah, I see.  That explains why it works in the page itself which is in the NAVIGATION array, but any other pages that I react-router navigate() to do not.  I want to have top-level pages work as normal from an item in the left navigation panel, which they do, but sublevel pages where I navigate() away to, do not.  They also lose the <PageContainer> which provides the page title and breadcrumbs.  The following page describes what I'm trying to do, including a sandbox showing what works and what doesn't.  [](https://stackoverflow.com/questions/79541966/how-to-navigate-to-a-new-page-in-material-ui-toolpad-and-change-its-toolbar-in-rd).  I've spent weeks on this and am totally frustrated!

Janpot:
> I've spent weeks on this and am totally frustrated!

`useActivePage` should probably print a warning if it's not used under a page context.

PR #4119, opened 2024-09-19
Integrate Toolpad Core in Toolpad Studio runtime

apedroferreira:
Closes https://github.com/mui/toolpad/issues/4060
Closes https://github.com/mui/toolpad/issues/4155

https://github.com/user-attachments/assets/fb67c7d7-bda0-44c0-8736-14b3d0195399

https://github.com/user-attachments/assets/a68b5ca5-e8c8-48e2-a74b-de9425b096e2

- Includes fix in Studio for only scrolling to selected elements in editor when needed (it was really disrupting UX)
- Includes and documents new specific `AppProvider` with `react-router-dom` routing out of the box
- Adds `sx` prop to `DashboardLayout` that can be used to customize `height`, `width` or any other styles. The default layout dimensions are `100vh` and `100vw` but can be overriden. The layout header and sidebar do not use fixed positions anymore, but absolute ones instead, and vertical scrolling happens in the layout content only.

https://deploy-preview-4119--mui-toolpad-docs.netlify.app/toolpad/core/introduction/base-concepts
https://deploy-preview-4119--mui-toolpad-docs.netlify.app/toolpad/core/react-app-provider
https://deploy-preview-4119--mui-toolpad-docs.netlify.app/toolpad/core/react-dashboard-layout/#full-screen-content

https://github.com/mui/toolpad/issues/4045 and https://github.com/mui/toolpad/issues/4149 will be covered separately for a fuller documentation of the new `AppProvider` for React Router. Left TODO comments in both issues.


packages/toolpad-studio/src/server/toolpadAppBuilder.ts
```diff
@@ -270,6 +270,14 @@ if (import.meta.hot) {
               : // load compiled
                 path.resolve(currentDirectory, '../exports'),
           },
+          {
+            find: '@toolpad/core',
```
apedroferreira, line 274: I had to add an alias here to get `@toolpad/core` imports to work in studio, not sure why it's not needed for `@toolpad/studio-components` for example, as those just worked?

Janpot: pnpm installs the core `build` folder into studio node modules because of its package.json `publishConfig.directory` setting. During the core build command a package.json is generated in that build folder. the build folder is what gets published on npm.

The dev command of builds the sources into the build folder, but it doesn't copy the package.json. That's why module resolution doesn't work. You made it work by alisaing to the core main folder, which does have a package.json.

I added some commits that just call the copy command in the dev command. It's not perfect, it won't watch for changes to package.json, but at least it unblocks us in the PR.

apedroferreira:
> This horizontal scrolling is something to investigate:

Yeah still fixing a few things.

scripts/docs/buildCoreApiDocs/config/getComponentImports.ts
```diff
@@ -23,14 +23,25 @@ export function getComponentImports(name: string, filename: string) {
...
+    const hasReactRouterDOMVersion = fs.existsSync(`${reactRouterDOMRelativePath}/${name}.tsx`);
+
     return [
       `import { ${name} } from '@toolpad/core/${name}';${
         hasNextJsVersion
           ? `\nimport { ${name} } from '@toolpad/core/nextjs/${name}'; // Next.js`
           : ''
+      }${
+        hasReactRouterDOMVersion
+          ? `\nimport { ${name} } from '@toolpad/core/react-router-dom/${name}'; // react-router-dom`
```
Janpot, line 36: We don't support imports deeper than one level. This mist be
```tsx
import { AppProvider } from '@toolpad/core/react-router-dom'
```

cc @bharatkashyap

Janpot (approved):
Looks good, let's just add that test that covers the React warning we see in https://stackblitz.com/run?file=Demo.tsx,index.tsx

Janpot:
The theme toggle doesn't work

https://github.com/user-attachments/assets/e1199d97-68f0-4187-b97c-06add2635639

apedroferreira:
> The theme toggle doesn't work
> 
>  Screen.Recording.2024-10-09.at.16.22.07.mov

Works for me...
And no changes related to that as far as I know...
Can you change it in the other demos?

Janpot:
> Can you change it in the other demos?

Yes, I could

But now it seems to be fixed. I did expect all of them to change when I switch one 🤔

apedroferreira:
> > Can you change it in the other demos?
> 
> Yes, I could
> 
> But now it seems to be fixed.

Weird, I guess we can ignore for now and it might come up again?

> I did expect all of them to change when I switch one 🤔

This used to happen before, I can try to make it work like that again later if we decide it would be best.

PR #4121, opened 2024-09-20
[code-infra] Enable React compiler eslint plugin

Janpot:
* Enable React compiler eslint plugin for everything except studio.
* Fix a few warnings
* TIL: it's ok to set state during render in [specific cases](https://react.dev/reference/react/useState#storing-information-from-previous-renders)

@apedroferreira There is one error left:
```
/Users/janpotoms/Projects/toolpad/packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
  439:5  error  React Compiler has skipped optimizing this component because one or more React ESLint rules were disabled. React Compiler only works when your components follow all the rules of React, disabling them may result in unexpected or incorrect behavior  react-compiler/react-compiler
```
Would you mind taking a look at it?


apedroferreira:
> Would you mind taking a look at it?

Taking a look.

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -435,8 +436,9 @@ function DashboardLayout(props: DashboardLayoutProps) {
 
   // If useEffect was used, the reset would also happen on the client render after SSR which we don't need
   React.useMemo(() => {
-    selectedItemIdRef.current = '';
-    // eslint-disable-next-line react-hooks/exhaustive-deps
+    if (navigation) {
```
apedroferreira, line 439: Having the selected item come from a context will eventually get rid of this so I guess this workaround should work for now.

Janpot: Have you seen that I added a bunch of helpers to optimize translating paths to navigation items and vice versa
https://github.com/mui/toolpad/blob/master/packages/toolpad-core/src/shared/navigation.tsx#L61-L166

it's being used by the new `PageContainer` [breadcrumbs](https://github.com/mui/toolpad/blob/master/packages/toolpad-core/src/useActivePage/useActivePage.ts) calculation. I expect a lot of it to be reusable for DashboardLayout. It adds some global caching to avoid recalculating these path => item maps over and over again.

apedroferreira (approved):

PR #4126, opened 2024-09-20
[code-infra] Update path-to-regexp to 6.3.0

Janpot:
Update `path-to-regexp` to 6.3.0 which contains a fix for https://github.com/advisories/GHSA-9wv6-86v2-598j, but doesn't suffer from https://github.com/vitest-dev/vitest/issues/6540

Few fixes to the setup I'm passing along
* Add react plugin to vite
* Clean up the workaround for mocking of external modules during vitest browser beta  (this was causing flakyness)
* Remove the alias for icon esm. This has been fixed in `@mui/icons-material@6.1.0`

Closes https://github.com/mui/toolpad/issues/4125

### DIFF
--- a/packages/toolpad-core/package.json
+++ b/packages/toolpad-core/package.json
@@ -59,9 +59,10 @@
     "@mui/lab": "6.0.0-beta.10",
     "@mui/utils": "6.1.1",
     "@toolpad/utils": "workspace:*",
+    "@vitejs/plugin-react": "4.3.1",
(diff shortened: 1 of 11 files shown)


apedroferreira (approved):
Looks good, thanks!
If we might not be able to merge https://github.com/mui/toolpad/pull/4074 for a while I guess I can move the test improvements there to a different PR.

PR #4181, opened 2024-10-02
[account] New slots and sub-components

### LINKED ISSUE #4179: [Account] Rename `menuItems` slot to `popoverContent`
### Steps to reproduce

"menuItems" is confusing, it sounds like the only thing you can render in there is a menu.

Maybe we it could use extra demo with content that is not just a menu?

### Current behavior

_No response_

### Expected behavior

_No response_

### Context

https://discord.com/channels/1131323012554174485/1248515176626524200/1290737976934010880

### Your environment

<details>
  <summary><code>npx @mui/envinfo</code></summary>

```
  Don't forget to mention which browser you used.
  Output from `npx @mui/envinfo` goes here.
```
</details>

**Search keywords**: account customize

### LINKED ISSUE #4180: [account] Add demo for account-switcher
- From: https://github.com/mui/toolpad/issues/4150#issuecomment-2371303344

- Add a demo with a popover showing how to build UI that allows switching accounts or tenants from within the `<Account />` component

**Search keywords**:

bharatkashyap:
- Closes #4179 
- Closes #4180 

https://deploy-preview-4181--mui-toolpad-docs.netlify.app/toolpad/core/react-account/#customization
https://deploy-preview-4181--mui-toolpad-docs.netlify.app/toolpad/core/api/account-preview/

### To Do

- [x] we change the `slot` structure of the `<Account />` component to to have `preview` and `popoverContent`
- [x] we add an `AccountPreview` component with `condensed` and `expanded` variants
- [x] we export composable components: 
     - [x] popup header => default contains account details (extended)
     - [x] popup footer => default contains signout button
     - [x] account preview (with variant condensed/extended)
     - [x] sign in button
     - [x] sign out button
- [x] add an internal `localeContext` between all sub-components to have the `localeText` prop work on `Account />` without prop drilling

### Breaking Change
- This removes the `menuItems` slot on the `Account` component and adds `popoverContent` instead


Janpot:
Assuming you're still working on the content of that second demo, I noticed that the justification of the header should probably be to the left:


Screenshot 2024-10-02 at 17 39 29

=>

Screenshot 2024-10-02 at 17 40 13

bharatkashyap:
> Assuming you're still working on the content of that second demo, I noticed that the justification of the header should probably be to the left:

Agreed, and I think we should probably also have an `userDetailsContainer` slot to override this section completely,  and be consistent (since every other section of the component will now have a slot to override it): 

Screenshot 2024-10-02 at 5 56 37 PM

Should also help solve the original pain which inspired the issue, based on https://discord.com/channels/1131323012554174485/1290737976934010880/1291004940369789023

Updated the PR to include this slot and slotProps for it

Janpot:
> I think we should probably have a userDetailsContainer slot (and accompanying slotProps) to customise this section or override it completely.

Can we just make it the default behavior and work on more slots in a separate PR?

docs/data/toolpad/core/components/account/account.md
```diff
@@ -30,16 +30,57 @@ When signed out, the component renders as an inline sign in button within the da
...
+
+##### Account Switcher
+
+{{"demo": "AccountSlotsAccountSwitcher.js", "bg": "gradient"}}
+
+The `content` prop can take any React component, so you can use it to display information instead of adding menu items:
+
+##### Crypto Wallet
+
+{{"demo": "AccountSlotsInfo.js", "bg": "outlined" }}
```
Janpot, line 61: This feels a bit like an unrealistic demo. Can't we just improve the account switcher instead, make it nice looking with avatar and create button, not that menu?

bharatkashyap: Agreed on improving the account switcher to make it nicer looking; at the same time I think we can keep this just for information purposes given it mirrors multiple crypto UX patterns

apedroferreira (approved):
> The content slot will by default take a component which takes a session prop

I guess that if we add a `useSession` hook this will not be needed? Anyway I guess it's not in this PR so that seems fine!

Everything looks fine here to me but this is without having checked https://github.com/mui/toolpad/pull/4227 yet. 
Also it seems like the `content` slot would make some other slots not do anything when used? Not sure if that's something to avoid, but if X for example is using the same kind of logic for some of their slots I guess it's fine.

docs/data/toolpad/core/components/account/AccountSlotsInfo.tsx
```diff
@@ -0,0 +1,119 @@
...
+const demoSession = {
+  user: {
+    name: 'Bharat Kashyap',
+    email: 'bharatkashyap@outlook.com',
+    image: 'https://avatars.githubusercontent.com/u/19550456',
+  },
+};
+
+const mockData = {
+  address: '0x1234...5678',
```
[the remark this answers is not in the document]
bharatkashyap: Agreed, changed this

docs/data/toolpad/core/components/account/account.md
```diff
@@ -40,6 +40,18 @@ Labels for the sign in and sign out buttons can be customized through the `local
...
 
-You can pass in your own items to the `Account` menu through the `menuItems` slot to add additional menu items in the space between the user's account details and the sign out button, to create larger, more complex menus:
+You can pass in your own components to completely override the default components inside the `Account` popover through the `slots` prop.
 
-{{"demo": "AccountSlots.js", "bg": "gradient"}}
+#### Content
+
+Use the `content` slot to add any additional options in the space between the user's account details and the sign out button, to create larger, more complex menus:
+
+##### Account Switcher
```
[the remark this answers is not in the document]
bharatkashyap: I guess with the new slot structure they might seem redundant; to me the idea is to clearly demonstrate that it's possible to use the `Account />` component without adding menu items, which is why the wallet example helps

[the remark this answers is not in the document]
apedroferreira: Yeah feel free to keep it, it was just my thoughts but doesn't hurt to have more examples either.

docs/data/toolpad/core/components/account/AccountCustomLocaleText.js
```diff
@@ -0,0 +1,46 @@
+import * as React from 'react';
+import { Account, AuthenticationContext, SessionContext } from '@toolpad/core';
```
apedroferreira, line 2: Here https://github.com/mui/toolpad/pull/4268 I'll be adding a rule not to import directly from `@toolpad/core, so we can do:

```tsx
import {
  AuthenticationContext,
  SessionContext,
} from '@toolpad/core/AppProvider';
import { Account } from '@toolpad/core/Account';
```

docs/data/toolpad/core/components/account/account.md
```diff
@@ -30,16 +30,45 @@ When signed out, the component renders as an inline sign in button within the da
...
 ### Slot Props
 
-The underlying `signInButton`, `signOutButton` and `iconButton` components can be customized by passing in `slotProps` to the `Account` component.
+The `AccountPreview` component has two variants, `condensed` (the default) and `expanded`. You can change the variant used inside `<Account />` by passing in custom props through `slotProps`:
 
-Labels for the sign in and sign out buttons can be customized through the `localeText` prop.
+{{"demo": "AccountCustomSlotProps.js", "bg": "outlined" }}
 
-{{"demo": "AccountCustom.js", "bg": "outlined" }}
+### Labels
```
apedroferreira, line 50: Maybe we can order sections like "slot props", "slots" and only then the "Labels"? Feels more logical to have the slot sections together, maybe.

docs/data/toolpad/core/components/account/account.md
```diff
@@ -30,16 +30,45 @@ When signed out, the component renders as an inline sign in button within the da
...
+
+You can wrap the default sub-components – such as `SignOutButton` and `AccountPreview` – and wrap them in `AccountPopoverHeader` and `AccountPopoverFooter` components to create custom account popovers, as shown in the following demos:
+
+##### Crypto Wallet
+
+You can wrap the default `SignOutButton` with your own action buttons inside `AccountPopoverFooter` to customize the popover footer section:
+
+{{"demo": "AccountSlotsWallet.js", "bg": "outlined" }}
+
+##### Account Switcher
```
apedroferreira, line 70: Perhaps we could show the account switcher before the crypto wallet? The switcher feels a bit less complex visually.

docs/data/toolpad/core/components/account/AccountSlotsWallet.js
```diff
@@ -0,0 +1,126 @@
...
+import CopyIcon from '@mui/icons-material/ContentCopy';
+
+import {
+  Account,
+  AccountPreview,
+  AuthenticationContext,
+  SessionContext,
+  SignOutButton,
+} from '@toolpad/core';
+import { AccountPopoverFooter } from '../../../../../../packages/toolpad-core/src/Account/AccountPopoverFooter';
```
apedroferreira, line 22: Really long import, shouldn't this component be importable from `@toolpad/core/Account` probably?

docs/data/toolpad/core/components/account/AccountSlotsWallet.js
```diff
@@ -0,0 +1,126 @@
...
+          <Typography variant="h6" fontWeight="bold">
+            {mockData.balance}
+          </Typography>
+          <Typography variant="body2" color="text.secondary">
+            {mockData.usdBalance}
+          </Typography>
+        </Stack>
+      </Stack>
+      <Divider />
+      <AccountPopoverFooter sx={{ gap: 2, px: 2, justifyContent: 'center' }}>
```
apedroferreira, line 83: I thought that we were going to have a separate slot for customizing the header or footer, for example, instead of just one slot for all the popover content, and it felt more intuitive for me to use and more consistent with how MUI generally allows this kind of customization, as opposed to an `AccountPopoverFooter` component that takes children.

Thse components by themselves seem like really simple generic wrappers too.

This might also mean that maybe `signOutButton` shouldn't be a slot, just a component that can optionally be imported to be used/composed in the footer slot.

Also in this example in particular the preview isn't helpful enough, I would need to expand and find `AccountPopoverFooter` to see how it's being used. Maybe the preview could be more helpful somehow?

packages/toolpad-core/src/Account/Account.tsx
```diff
@@ -1,48 +1,37 @@
...
+  /**
+   * The component used for the account preview
+   * @default AccountPreview
+   */
+  preview?: React.ElementType;
+  /**
+   * The component used for the content of account popover
+   * @default Popover
+   */
+  popoverContent?: React.ElementType;
```
apedroferreira, line 24: I think there's a naming mismatch between some properties in `slots` and `slotProps` right now. Shouldn't they always exactly match?

Or I might be wrong, maybe it was just something I assumed in the other components I've worked on... It feels more intuitive that they match...

packages/toolpad-core/src/Account/AccountPreview.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+  test('displays condensed variant by default', () => {
+    render(
+      <AppProvider authentication={auth} session={session}>
+        <AccountPreview />
+      </AppProvider>,
+    );
+
+    const avatar = screen.getByRole('img', { name: 'John Doe' });
+    expect(avatar).toBeInTheDocument();
+    expect(avatar).toHaveAttribute('src', 'https://example.com/avatar.jpg');
```
apedroferreira, line 39: To make sure it's the condensed version you might have to check that the text isn't there too maybe?

packages/toolpad-core/src/internals/demo.tsx
```diff
@@ -1,6 +1,9 @@
 import * as React from 'react';
+import { DocsContext as Docs } from '../shared/context';
```
apedroferreira, line 2: I've made a few small changes to the internals too here https://github.com/mui/toolpad/pull/4268, also just trying to fix that we had both `internal` and `internals` folders. Let me know if you agree with those.

It's just these 2 commits: https://github.com/mui/toolpad/pull/4268/commits/9896d0d6e1f308dad0c5fec15c92295e59377db9 and https://github.com/mui/toolpad/pull/4268/commits/6dfee2f38f8a0526b6ebed24ee1b007deed0e038

apedroferreira (changes requested):
I've added more comments than usual but in general I think it looks good!

The usage for `AccountPreview` and the `preview` slot for the condensed/expanded variants seem good to me, as well as the `SignOutButton` that can be used.

docs/data/toolpad/core/components/account/AccountSlotsWallet.js
apedroferreira: Oh, I just saw that the `PageContainer` has a similar approach to this so anyway it was just my opinion.
Not strongly against doing it this way in the end.

bharatkashyap: Yep, we were going by the `PageContainerToolbar` way of doing things since it is also the new way of composition/customization that X is adopting; lesser slots and more wrappable/composable components

packages/toolpad-core/src/internals/demo.tsx
bharatkashyap: Got it, I'll pull those into this PR

packages/toolpad-core/src/Account/Account.tsx
bharatkashyap: Agreed, it should have in this case as well but I missed one `slot` and one `slotProp` which is causing confusion here; addressed that

Janpot:
Is this a breaking change?

bharatkashyap:
> Is this a breaking change?

Yes, since it removes `menuItems` on `<Account />`

docs/data/toolpad/core/components/account/AccountSlotsWallet.js
apedroferreira: Got it, sounds good then!

docs/data/toolpad/core/components/account/account.md
```diff
@@ -30,16 +30,45 @@ When signed out, the component renders as an inline sign in button within the da
 
 ## Customization
 
-### Slot Props
+The `<Account />` component is composed of multiple sub-components:
+
+- `<SignInButton />`
+- `<AccountPreview />`
+- `<Popover />`
```
apedroferreira, line 37: This is the only one of these 4 components that's actually just a standard Material UI component, right?
Not sure if that can be confusing in any way.

packages/toolpad-core/src/Account/Account.tsx
```diff
@@ -96,142 +84,179 @@ function Account(props: AccountProps) {
...
+      ) : (
+        <AccountPreview
+          variant="condensed"
+          handleClick={handleClick}
+          open={open}
+          {...slotProps?.preview}
+        />
+      )}
+      {slots?.popover ? (
+        <slots.popover />
```
apedroferreira, line 104: Should it spread `slotProps.popover` in this line too?

apedroferreira (approved):
Looks great, thanks for addressing the previous comments!
Approving already not to block as I have nothing else that should be a major issue.

About the `popover` and `popoverContent` slots, not sure if we usually have slots where one wraps / can override the other like that, but I guess that X must have some cases like it too?
So it's probably fine, mentioning it just in case! I guess there's not a much better way to allow full customization there.

packages/toolpad-core/src/Account/Account.tsx
```diff
@@ -96,142 +84,179 @@ function Account(props: AccountProps) {
...
-              {localeText?.signOutLabel}
-            </Button>
-          </SignOutContainer>
-        )}
-      </Popover>
+            ...slotProps?.popover?.slotProps,
+          }}
+        >
+          {slots?.popoverContent ? (
+            <slots.popoverContent />
```
apedroferreira, line 140: Oh and here not sure if it should spread `slotProps?.popoverContent` too.

Janpot:
* The alignment of the email line looks off:

  Screenshot 2024-10-18 at 18 23 46
    
  vs.
    
  Screenshot 2024-10-18 at 18 24 31

* The projects submenu on the account switcher looks a bit too much IMO, I'd just keep the accounts and instead add a button at the bottom to link another account

docs/data/toolpad/core/components/account/account.md
bharatkashyap: Yes, not sure if I need to clarify that in any way? Perhaps okay to leave it as it is and see if that inspires any confusion

PR #4189, opened 2024-10-05
[PageContainer] Avoid React invalid props warning

Janpot:
Avoid spreading these props to the root element

Closes https://github.com/mui/toolpad/issues/4184


packages/toolpad-core/src/PageContainer/PageContainer.test.tsx
```diff
@@ -112,4 +112,44 @@ describe('PageContainer', () => {
...
+    );
+
+    const breadcrumbs = screen.getByRole('navigation', { name: 'breadcrumb' });
+
+    const helloLink = within(breadcrumbs).getByRole('link', { name: 'Hello' });
+    expect(helloLink.getAttribute('href')).toBe('/hello');
+    expect(within(breadcrumbs).getByText('World')).toBeTruthy();
+  });
+
+  // TODO: Remove in the next major version
```
bharatkashyap, line 133: Should we remove it in the next minor itself and not wait for a major? Perhaps breaking changes are okay to be made in a minor while the project is in beta

bharatkashyap (approved):

packages/toolpad-core/src/PageContainer/PageContainer.test.tsx
Janpot: sure, but it's not even in the current minor yet

apedroferreira (approved):

PR #4227, opened 2024-10-07
[account] [docs] Add custom user details example

### LINKED ISSUE #4226: Customize user details section in `<Account/>`
### Summary

- From feedback: https://discord.com/channels/1131323012554174485/1248515176626524200/threads/1290737976934010880

> If i'm not wrong, the menuItems are rendered under the top part and not inside (rectangle in my above picture).
I really would like to display some informations in the top part of the component (where email is normaly displayed).

> I think it could be a great feature !

### Examples

_No response_

### Motivation

_No response_

**Search keywords**: account, user, prop, slot

bharatkashyap:
- Close #4226 

### To Do

- [x] We will export a `useSession()` hook for user to be able to extend it
- [x] Add custom session account example
- [x] Add `useSession` docs pages 

Screenshot 2024-10-13 at 4 32 35 PM

https://deploy-preview-4227--mui-toolpad-docs.netlify.app/toolpad/core/react-use-session/


packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -679,11 +679,13 @@ DashboardLayout.propTypes /* remove-proptypes */ = {
         iconButton: PropTypes.object,
         signInButton: PropTypes.object,
         signOutButton: PropTypes.object,
+        userDetailsContainer: PropTypes.object,
       }),
       slots: PropTypes.shape({
-        menuItems: PropTypes.elementType,
+        content: PropTypes.elementType,
```
apedroferreira, line 685: Maybe the name `content` is too generic, but it's just my opinion, I don't mind it.

packages/toolpad-core/src/Account/Account.tsx
```diff
@@ -166,19 +159,39 @@ function Account(props: AccountProps) {
...
-        </AccountInfoContainer>
+        {slots?.userDetailsContainer ? (
+          <slots.userDetailsContainer session={session} />
+        ) : (
+          <Box
+            sx={{
+              display: 'flex',
+              flexDirection: 'row',
+              justifyContent: 'space-between',
+              padding: theme.spacing(2),
```
apedroferreira, line 170: Is using a simple `2` here not the same thing? (not sure though)

docs/data/toolpad/core/components/account/account.md
```diff
@@ -30,16 +30,57 @@ When signed out, the component renders as an inline sign in button within the da
...
+##### Crypto Wallet
+
+{{"demo": "AccountSlotsInfo.js", "bg": "outlined" }}
+
+#### User Details Container
+
+By passing a custom component to the `userDetailsContainer` slot, you can override the section which displays the signed-in user's details. The custom component receives a `session` prop which contains the current authentication session:
+
+```tsx
+// ...
```
apedroferreira, line 68: This example doesn't immediately make sense to me: where is the `CustomSession` being used?

apedroferreira (changes requested):
Looks great!
Other than what I added in https://github.com/mui/toolpad/pull/4181# just added some opinions on naming and docs feedback.

docs/data/toolpad/core/components/use-session/use-session.md
```diff
@@ -0,0 +1,54 @@
...
+When using authentication features inside Toolpad Core, a `SessionContext` is created to share session information among all Toolpad Core components. This accepts a default value from the `session` prop of the `AppProvider`.
+
+```js
+<AppProvider session={session}>{props.children}</AppProvider>
+```
+
+The `session` can be created using any authentication provider of your choice. You can access the current value of the `SessionContext` inside Toolpad Core components by invoking the hook:
+
+```js
+const session = useSession();
```
apedroferreira, line 25: Should we show where to import it from here too?

apedroferreira (approved):
Looks great, really easy to understand!
Added some small documentation suggestions.

aress31:
I believe that useSession should return both session and setSession. In many parts of an application, such as the SignInPage, a successful sign-in requires setting the session. Currently, this can only be done at the top level in the AppProvider, forcing the setter to be passed down multiple levels through the component hierarchy. Allowing this hook to both read and update the session directly would significantly expand its usability across various use cases.

bharatkashyap:
> I believe that `useSession` should return both session and `setSession`. In many parts of an application, such as the `SignInPage`, a successful sign-in requires setting the session. Currently, this can only be done at the top level in the `AppProvider`, forcing the setter to be passed down multiple levels through the component hierarchy. Allowing this hook to both read and update the session directly would significantly expand its usability across various use cases.

I'll update the original issue you created with this and we can track this as a feature request to be discussed and potentially added (or not) to the roadmap - let me know if that's okay

PR #4236, opened 2024-10-10
[DashboardLayout] Add `sidebarFooter` slot

apedroferreira:
Closes https://github.com/mui/toolpad/issues/4101
Partial item in https://github.com/mui/toolpad/issues/4147

- Includes `mini` slot prop so that users can render footer content differently in mini-drawer
- Add test for slot content
- Add sidebar footer slot example to slots demo (copyright text) (once we have the new `Account` variant we can add document it separately and it will also serve as an example for using this slot): https://deploy-preview-4236--mui-toolpad-docs.netlify.app/toolpad/core/react-dashboard-layout/#customization
- Remove resolved `TODO` comments in `DashboardLayout` (unrelated)


docs/data/toolpad/core/components/dashboard-layout/DashboardLayoutSlots.js
```diff
@@ -124,7 +139,9 @@ function DashboardLayoutSlots(props) {
       theme={demoTheme}
       window={demoWindow}
     >
-      <DashboardLayout slots={{ toolbarActions: Search }}>
+      <DashboardLayout
+        slots={{ toolbarActions: Search, sidebarFooter: SidebarFooter }}
```
Janpot, line 143: I wonder about naming. Should we call this something like "NavigationFooter"? 🤔  I feel like it would make sense to give our surfaces a semantic name, rather one that is based on where it's positioned. Not sure though

Janpot (approved):
Some musings about naming, but overall looks fine.

docs/data/toolpad/core/components/dashboard-layout/DashboardLayoutSlots.js
apedroferreira: So far I've used the word "sidebar" for this kind of interfaces and it made the most sense to me in the places it was used, where it seemed like they were specific features of the sidebar itself and not just navigation.
The `hideNavigation` prop in `DashboardLayout` is an exception but it also hides the menu icon, so it's a bit more general.
For this particular slot I think we really mean the footer of the sidebar and no other type of navigation so I feel like it makes sense to name it this way? And it might be easier for users to understand what it really means visually.

It's also just my opinion of course, but the logic I've been following for this kind of naming is intentional.

apedroferreira:
Merging but if we agree to change any naming we can still do it!

PR #4237, opened 2024-10-10
[cli] Fix `auth.ts` file generation

bharatkashyap:
- Remove `NavigateButton.tsx` (not needed since we directly load the first page)
- Fix indentation errors and typos in the generated `auth.ts`


apedroferreira (approved):
LGTM

PR #4268, opened 2024-10-16
[docs] Docs/demo cleanup

apedroferreira:
- Add rule not to import directly from `@toolpad/core` in docs similar to `@mui/material` rule in Material UI repo
- Fix all imports in docs accordingly
- Use new `useDemoRouter` hook in every demo where it's possible

To check if nothing is broken (everything seems fine from what I've checked):
https://deploy-preview-4268--mui-toolpad-docs.netlify.app/toolpad/core/react-dashboard-layout/


Janpot (approved):
Great!

Also noticed we have both `/internal` and `/internals`. Maybe we should unify them?

apedroferreira:
> Great!
> 
> Also noticed we have both `/internal` and `/internals`. Maybe we should unify them?

I guess this should do? https://github.com/mui/toolpad/pull/4268/commits/9896d0d6e1f308dad0c5fec15c92295e59377db9

PR #4276, opened 2024-10-18
[DashboardLayout] Show nested navigation in mini-drawer

apedroferreira:
Initially was just trying to refactor the sidebar navigation in `DashboardLayout` to use the same navigation cache and utilities as the `PageContainer` component.

But this required adding nested navigation to the sidebar for consistency with the warnings we show when there are duplicated navigation items, because the navigation links should match in all variants of the sidebar:

https://github.com/user-attachments/assets/fe3e2d0c-705f-4d18-9f00-4f94ffe8a870

Also improved selected navigation item tests to cover an extra case.

**Inspiration:**

- https://minimals.cc/dashboard


douglaszaltron:
Nice!

apedroferreira:
Status: right now with `useActivePage` there's no way to identify a particular item when its properties are the exact same as another. This would be necessary to identify an item instead of another and show that item only as selected in the sidebar.

If multiple items could be shown as selected simultaneously this would not be an issue, otherwise we need to find a solution. Maybe we could generate item ids every time the navigation is changed, but this could be a bit expensive as we'd have to go through the whole navigation tree.

packages/toolpad-core/src/shared/navigation.tsx
```diff
@@ -165,3 +126,28 @@ export function getItemPath(navigation: Navigation, item: NavigationPageItem): s
...
+    return item.children.some((nestedItem) => {
+      if (!isPageItem(nestedItem)) {
+        return false;
+      }
+
+      if (nestedItem.children) {
+        return hasSelectedNavigationChildren(navigation, nestedItem, activePagePath);
+      }
+
+      return activePagePath === getItemPath(navigation, nestedItem);
```
bharatkashyap, line 148: To be sure, does this function work in cases where the nested navigation items are supplied via regex patterns? I might be missing something, but seems like the earlier function was checking for a `navigationItem.pattern` case and testing the path against the regex pattern in case it existed but I can't seem to find an analogous case here

packages/toolpad-core/src/DashboardLayout/DashboardSidebarSubNavigation.tsx
```diff
@@ -236,17 +281,44 @@ function DashboardSidebarSubNavigation({
...
-                ? nestedNavigationCollapseIcon
-                : null}
+              {navigationItem.children ? <ExpandMoreIcon sx={nestedNavigationCollapseSx} /> : null}
             </NavigationListItemButton>
+            {navigationItem.children && isMini ? (
+              <Grow in={navigationItemId === hoveredMiniSidebarItemId}>
+                <Box
+                  sx={{
+                    position: 'fixed',
+                    left: 62,
```
bharatkashyap, line 291: Is there no avoidable way to hard code the left and `pl` values here? I could be wrong but this might create a future bug if we make the size of the mini sidebar configurable - perhaps we can link these values to that constant (if indeed they derive from it)

bharatkashyap (approved):
Looks really good! Great addition to the feature-set. Some small comments on code.

royhuntley:
Related a little to this item...

If you have a navigation structure that a PAGE item also has children, when the navigation drawer is open (Wide), you cant actually navigate to the page, because clicking on it expands to make the children visible.  However, when the navigation drawer is closed (Narrow), you get taken to the page when you click on the item.

{
    segment: 'test',
    title: 'Test',
    children: [  {
      segment: 'orders',
      title: 'Orders',
      icon: <ShoppingCartIcon />,
    },
    {
      kind: 'divider'
    },
    {
      segment: 'orders',
      title: 'Orders',
      icon: <ShoppingCartIcon />,
    },

packages/toolpad-core/src/shared/navigation.tsx
apedroferreira: I believe the `activePagePath` that goes here as a parameter already accounts for pattern matching with the logic in the `useActivePage` hook. I'll check again just to be sure.

So the behavior should be that pattern matching in children items works, but only under the `segment`s in the parent items. If the parent also has a `pattern` it will probably be ignored in these cases and I think that's probably the way it should work.

apedroferreira:
Didn't want to abandon this old PR so fixed what made sense from the review suggestions + improved mini bar design as the previous one here wasn't great.
Updated the video above with the new mini bar design.

royhuntley:
What needs to happen to get this change released - its good work - the menu bar is so much more functional when minimised!On 13 Mar 2025, at 18:41, Pedro Ferreira ***@***.***> wrote:﻿
  Didn't want to abandon this old PR so fixed what made sense from the review suggestions + improved mini bar design as the previous one here wasn't great.
Updated the video above with the new mini bar design.—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you commented.Message ID: ***@***.***>



apedroferreira left a comment (mui/toolpad#4276)
Didn't want to abandon this old PR so fixed what made sense from the review suggestions + improved mini bar design as the previous one here wasn't great.
Updated the video above with the new mini bar design.

apedroferreira:
I'll merge this one soon if it's good with everyone!

douglaszaltron (approved):

bharatkashyap (approved):
Looks good!

DaliborHomola:
> I'll merge this one soon if it's good with everyone!

Yes, great

PR #4296, opened 2024-10-21
[core] Allow changing the width of the drawer in dashboard layout com…

garryxiao:


packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -286,7 +291,7 @@ function DashboardLayout(props: DashboardLayoutProps) {
 
   const getDrawerSharedSx = React.useCallback(
```
apedroferreira, line 292: In this `getDrawerSharedSx` callback, `sidebarExpandedWidth` also needs to be added at the end to the dependencies array.
The `react-hooks/exhaustive-deps` eslint rule should automatically show you an error about this.

apedroferreira (changes requested):
Looks great, thank you!
Can you also please autogenerate the documentation for the new prop?

You should be able to do it by running 2 scripts from the root package after addressing the comments below.

1. `pnpm proptypes`
2. `pnpm docs:build:api`

And then just add and commit the autogenerated changes!

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
garryxiao: Hi @apedroferreira, changes are made accordingly, please double check.

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -89,6 +89,11 @@ export interface DashboardLayoutProps {
    * @default false
    */
   hideNavigation?: boolean;
+  /**
+   * Width of the sidebar when expanded, in pixels.
+   * @default 320
+   */
+  sidebarExpandedWidth?: number;
```
Janpot, line 96: Do we want (is it possible) to allow lengths in percent? or even a calculation such as `min(200px, 20%)`? we don't have to do this as part of this PR, it's just a question.

garryxiao: My personal feedback is no. A fix width for the bar is more reasonable for design.

apedroferreira: I'll test if it works, if it does we can allow strings too, i guess it would be a simple change.

apedroferreira (approved):
Thanks a lot @garryxiao !!

PR #4301, opened 2024-10-23
[docs] Separate example folders for Core and Studio

### LINKED ISSUE #4298: [examples] Separate the examples in two directories (core & studio)
### Summary

Create two directories in the /examples, one for core and one for studio, so it's easier for people to find examples for the library they use. It's not obvious otherwise which example is for which project.

### Examples

_No response_

### Motivation

_No response_

**Search keywords**: examples

bharatkashyap:
Closes #4298


apedroferreira (approved):
Looks good!

I gave it a quick check and maybe it's just missing updating the path on the example READMEs?

apedroferreira:
The studio example deployments in Railway seem to be failing in the CI in `master`: https://github.com/mui/toolpad/commits/master/
I guess we need to update the paths somewhere else?

bharatkashyap:
> The studio example deployments in Railway seem to be failing in the CI in `master`: https://github.com/mui/toolpad/commits/master/ I guess we need to update the paths somewhere else?

Yep, I noticed this as well (and flagged on Slack, I guess we can continue this there!) but couldn't investigate further without access to the Railway apps

PR #4340, opened 2024-10-30
Allow theme switcher override with slots

apedroferreira:
Closes https://github.com/mui/toolpad/issues/4151 and https://github.com/mui/toolpad/issues/4291

- Include `ThemeSwitcher` component in `toolbarActions` slot so it can be hidden
- Export existing `ThemeSwitcher` and `ToolbarActions` (default slot) components
- Add example of popover menu & custom theme switcher in `DashboardLayout` demo page in docs

### DIFF
--- a/docs/data/toolpad/core/components/dashboard-layout/DashboardLayoutCustomThemeSwitcher.js
+++ b/docs/data/toolpad/core/components/dashboard-layout/DashboardLayoutCustomThemeSwitcher.js
@@ -1,0 +1,168 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import Box from '@mui/material/Box';
+import FormControlLabel from '@mui/material/FormControlLabel';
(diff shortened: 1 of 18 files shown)


bharatkashyap (approved):
Looks good! The demo and explanation of the `slots` seems clear to me. Perhaps we can have a diagram eventually when we have more slots

apedroferreira:
@bharatkashyap I just noticed this if you sign out in the "Account in sidebar" example:

Screenshot 2024-11-07 at 17 37 37

Probably no need to fix it in this release, just ran into it! Also not sure how to best handle it in the mini variant of the sidebar...

bharatkashyap:
> Probably no need to fix it in this release, just ran into it! Also not sure how to best handle it in the mini variant of the sidebar...

Thanks for that catch! I'll open an issue to track this

PR #4346, opened 2024-11-01
[DashboardLayout] Left-align header title in mobile viewport

apedroferreira:
Closes https://github.com/mui/toolpad/issues/4111

Screenshot 2024-11-01 at 20 03 24

Previously, the `DashboardLayout` header title in mobile viewports was center-aligned with absolute positioning.
However, when using the `toolbarActions` slot to add items to the header, problems with overlapping such as the one reported in https://github.com/mui/toolpad/issues/4111 could occur.

Simply using a statically positioned, left-aligned header title, just like the tablet and desktop viewports do, should solve this problem and even make more space available for additional actions in the header.

https://deploy-preview-4346--mui-toolpad-docs.netlify.app/toolpad/core/react-dashboard-layout


packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -348,31 +348,22 @@ function DashboardLayout(props: DashboardLayoutProps) {
...
-                    ml: 0.5,
-                    whiteSpace: 'nowrap',
-                  }}
-                >
-                  {applicationTitle}
-                </Typography>
-              </Stack>
-            </Link>
-          </Box>
+          <Link href="/" style={{ color: 'inherit', textDecoration: 'none' }}>
```
apedroferreira, line 351: It's just removing the `Box`, which only had conditional styles for mobile viewports.

Janpot:
There's something wrong with the icon button in codesandbox:

Screenshot 2024-11-01 at 21 49 05

Trying out the overflow behavior:


https://github.com/user-attachments/assets/002e3dea-179c-49cf-9add-cdfd81925d00

Some observations:
* The theme toggle stays aligned at the top (yet it does shift somewhat vertically)
* The title/menu stays center aligned.
* 🤔 I thought the theme toggle would (should) be part of the toolbar actions.

bharatkashyap:
> * 🤔 I thought the theme toggle would (should) be part of the toolbar actions.

That will happen with https://github.com/mui/toolpad/pull/4340

Janpot:
My expectation of overflow would differ than this implementation. I think the actions should wrap, the same way it does in the PageContainer:


https://github.com/user-attachments/assets/31c2bf9b-5805-4f50-814f-860abb977491

apedroferreira:
> My expectation of overflow would differ than this implementation. I think the actions should wrap, the same way it does in the PageContainer:
> 
>  Screen.Recording.2024-11-02.at.13.59.19.mov 
> We could even consider accepting an array of actions so that we can wrap them separately.
> 
> _edit: just noticed this is currently broken. It misses a flex-wrap. I fixed it in the devtools for this video._

Good feedback! Hadn't checked that.
It made it so it wraps now, the `marginLeft` trick seems to do it!

https://github.com/user-attachments/assets/9c9b2c88-7e6c-46de-8c3d-94c2365a986b

> We could even consider accepting an array of actions so that we can wrap them separately.

I gave it a brief attempt but wrapping and aligning multiple items to the right doesn't seem as straightforward, and the types for a slot like that aren't as automatic, plus the account button is already a separate slot and not sure if we want it to wrap separately from the actions. So I guess we could keep that for later?

bharatkashyap (approved):

Janpot:
> Here the app name gets trimmed

This doesn't feel like a fair comparison. that app doesn't optimize for mobile.

prakhargupta1:
> This doesn't feel like a fair comparison. that app doesn't optimize for mobile.

Hmm agree that's not the mobile view.

PR #3383, opened 2024-04-11
[docs] Start Toolpad Core docs

bharatkashyap:
- Closes https://github.com/mui/mui-toolpad/issues/3420

- The example referred to in the docs added by this PR is in #3617 

- Create an outline for the Introduction page: https://deploy-preview-3383--mui-toolpad-docs.netlify.app/toolpad/core/introduction/

PR #4350, opened 2024-11-01
[docs] Fix layout shift example page

oliviertassinari:
Open https://mui.com/toolpad/core/introduction/examples/, see the paint flash / layout shift of the footer. An issue from #3383.

I noticed this on https://github.com/mui/toolpad/pull/4344.

After: https://deploy-preview-4350--mui-toolpad-docs.netlify.app/toolpad/core/introduction/examples/


docs/src/modules/components/ExamplesGrid/ExamplesGrid.tsx
```diff
@@ -44,15 +44,11 @@ function CodeSandboxIcon() {
...
-    const importExamples = async () => {
-      const exampleContent = await import(`./${examplesFile}`);
-      setExamples(exampleContent.default);
-    };
-    importExamples();
-  }, [examplesFile]);
+  // @ts-ignore
+  const req = require.context('./', false);
+  // TODO: When migrating away from webpack, or when we have time, to create dedicate files for each
+  // demo. We shouldn't do this dynamic imports in the first place ⬇️
```
oliviertassinari, line 50:
```suggestion
  // demo. We shouldn't do those dynamic imports in the first place ⬇️
```

bharatkashyap:
I can see this still on the PR preview link:

https://github.com/user-attachments/assets/ff43e245-703d-46be-beb1-433b0e756673

oliviertassinari:
> layoutShift.mov

@bharatkashyap This is another issue, it's a problem with how docs-infra loads the font.

Here the PR fixes the setState: 

https://github.com/user-attachments/assets/cb98c8e4-11a6-4e67-a631-aa2263922483

docs/src/modules/components/ExamplesGrid/ExamplesGrid.tsx
```diff
@@ -44,15 +44,11 @@ function CodeSandboxIcon() {
...
-
-  React.useEffect(() => {
-    const importExamples = async () => {
-      const exampleContent = await import(`./${examplesFile}`);
-      setExamples(exampleContent.default);
-    };
-    importExamples();
-  }, [examplesFile]);
+  // @ts-ignore
+  const req = require.context('./', false);
```
Janpot, line 48: I'd just import them statically instead of doing proprietary webpack stuff.

```tsx
import coreExamples from './core-examples'
import studioExamples from './studio-examples'

const EXAMPLES_BY_PRODUCT = {
  'core-examples': coreExamples,
  'studio-examples': studioExamples
}

// ...

const examples = EXAMPLES_BY_PRODUCT[examplesFile]
```

Or better yet, just pass the array as a prop and import thecorrect file on the callsite.

oliviertassinari: Yeah, true, this indirection is strange. I have created one file per demo, it should have been done this way in the first place IMHO.

oliviertassinari:
I have noticed problems with the repository spending a bit of time on this:

- Docs files are types checked in the CI
- ESlint rule filenames/match-exported doesn't lint

Janpot (approved):
looks good. would maybe suggest to extract `${process.env.SOURCE_CODE_REPO}/tree/master` into its own variable and reuse between the conditional and the replace but not super important.

oliviertassinari:
> would maybe suggest to extract ${process.env.SOURCE_CODE_REPO}/tree/master into its own variable

We have something equivalent in https://github.com/mui/material-ui/blob/060c55cb219bfc33633f7e61484884ea6d37bfdf/packages/markdown/parseMarkdown.js#L403 for the markdown. We might need to merge those two logics.

bharatkashyap (approved):
Looks good!

PR #4370, opened 2024-11-05
[SignInPage] Improve default UI and customisation ability

### LINKED ISSUE #4066: [core] `DashboardLayout` theme overrides issues
1. Many styles from List and ListItem not being applied to Toolpad Core layout

> Styling seems to be working differently from the current Material UI components.In the custom theme, there are many styles for List and ListItem , but these aren’t being applied to the layout.

| [Toolpad Core Theme app](https://github.com/zanivan/toolpad-core-theme) | [Material UI Dashboard Template App](https://github.com/mui/material-ui/tree/master/docs/data/material/getting-started/templates/dashboard) |
|--------|--------|
|  |  |

2. Custom theme breaking for `focus` and `active` states
> The custom theme seems to be applied in most of the UI, but in some states such as focus and active are acting weird:

image
image

- Link to [full feedback](https://www.notion.so/mui-org/Toolpad-Core-review-58ef743943104bfbb1d57794c74fc0c0?pvs=4#48a0e82aa366477a9891db550d3d85bd)

**Search keywords**:

bharatkashyap:
- Closes https://github.com/mui/toolpad/issues/4067
- Closes https://github.com/mui/toolpad/issues/4337
- Closes #4066 

- Removes the default `LockIcon` - users can display an icon using `branding.logo` 
- Adds `title` and `subtitle` slots to `SignInPage` to override the defaults
- Changes the color of the `ListItemText` inside the `DashboardLayout` Navigation to `text.primary` when selected
- Changes to the default styling on the buttons and input fields
- Adds an `sx` prop to the outermost `<Box />` on the `SignInPage`
- Adds an `sx` prop to the `AccountPreview` component applied to the wrapping `<Stack />` in both `expanded` and `condensed` variants

<table>
  <tr>
    <th>Default</th>
    <th>Themed</th>
  </tr>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>


packages/toolpad-core/src/SignInPage/SignInPage.tsx
```diff
@@ -239,388 +250,421 @@ function SignInPage(props: SignInPageProps) {
...
+          sx={{
+            marginTop: 8,
+            display: 'flex',
+            flexDirection: 'column',
+            alignItems: 'center',
+            bgcolor: 'background.paper',
+            borderRadius: 1,
+            p: 4,
+            border: '1px solid',
+            borderColor: (theme) => alpha(theme.palette.grey[400], 0.4),
```
apedroferreira, line 274: If you're using functions in properties more than once I guess the whole `sx` can be a function as the first argument should be the `theme`?

packages/toolpad-core/src/SignInPage/SignInPage.tsx
```diff
@@ -156,6 +158,11 @@ export interface SignInPageSlots {
    * @default Link
    */
   signUpLink?: React.JSXElementConstructor<LinkProps>;
+  /**
+   * Custom information added to the top header beneath the title
+   * @default Typography
+   */
+  headerInfo?: React.ElementType;
```
apedroferreira, line 165: By header you mean the title, right? But we've been calling "header" to something more like the AppBar on top of pages. So maybe instead of "header" it could be "heading" or "title"?

packages/toolpad-core/src/SignInPage/SignInPage.tsx
```diff
@@ -239,388 +250,421 @@ function SignInPage(props: SignInPageProps) {
...
+                  ) : (
+                    <Box sx={{ marginBottom: (theme) => theme.spacing(1) }}>
+                      <InputLabel shrink htmlFor="email-passkey" sx={{ marginBottom: 0 }}>
+                        Email
+                      </InputLabel>
+                      <TextField
+                        required
+                        slotProps={{
+                          htmlInput: {
+                            sx: (theme) => ({
```
apedroferreira, line 383: Also if we're using the theme so much, we could just call it from the `useTheme` hook?

apedroferreira (changes requested):
Both versions look really cool from the images!
In the alert in the themed examples, you could highlight the test email and password somehow (bold, italic, a color, whichever looks better, probably bold).
Are you still going to add a playground? Not sure if I should be reviewing yet.

Just for `headerInfo` I think we can come up with a better name, other than that all looks good.

packages/toolpad-core/src/SignInPage/SignInPage.tsx
bharatkashyap: It's the section that could contain a sub-heading, given the heading of the Sign In Page is "Sign In", but I get your concern about the conflict with the "header" being the app bar. We could call it `secondaryTitle`?

apedroferreira: That sounds ok, on second thought maybe "headerInfo" isn't bad either but it could get confusing if we end up adding a header with the theme switcher I guess...
Or even something like "subHeading" could work probably. Feel free to go with what you think is best in the end, just thought I'd mention this.

bharatkashyap: Yes, on second thought I could probably add two slots called `title` and `subtitle` to be able to override that entire section. What do you think?

bharatkashyap:
> Are you still going to add a playground? Not sure if I should be reviewing yet.
Added a playground for you to test. Run `pnpm --filter playground-nextjs-themed dev`. I can move this into an example once we merge this PR, what do you think?

packages/toolpad-core/src/SignInPage/SignInPage.tsx
prakhargupta1: `title` and `subtitle` sound super clear to me.

apedroferreira:
> > Are you still going to add a playground? Not sure if I should be reviewing yet.
> 
> Added a playground for you to test. Run `pnpm --filter playground-nextjs-themed dev`. I can move this into an example once we merge this PR, what do you think?

Wow it looks really good!
Yeah, i guess this makes more sense as an example than a playground so we can just move it afterwards.

Some feedback from what I tested:

- Weird spacing above "main items" heading (too small):

Screenshot 2024-11-07 at 17 57 23

- I forgot to set the auth secret env variable at first and it just took me to this page:
 
Screenshot 2024-11-07 at 17 59 38

Not sure if it should show the error in the app itself instead?

- A border probably doesn't need to be there in the "mini" version of the sidebar account button:

Screenshot 2024-11-07 at 18 02 08

These don't need to be blocking as they're probably minor issues.
Also I'm noticing the theme can flash from light to dark when the page loads, I'll look into it as it shouldn't be happening.

apedroferreira (approved):

bharatkashyap:
> * Weird spacing above "main items" heading (too small):

> * A border probably doesn't need to be there in the "mini" version of the sidebar account button:

Good catch. Fixing these! 

> Not sure if it should show the error in the app itself instead?

It's planned to add an `AuthErrorPage` component that can be used to handle these errors (potentially after the `SignUpPage`) for now this is intentional to allow users to handle these errors with their own UI.

PR #4371, opened 2024-11-05
[examples] Fix Dependabot warning in Next.js example with passkey

apedroferreira:
Should fix https://github.com/mui/toolpad/security/dependabot/90


pnpm-lock.yaml
Janpot, line 1: 🤔 This shouldn't affect the lockfile, it's not a workspace, so we're not installing these dependencies as part of `npm install`

apedroferreira: It was unrelated it seems, these also update right now if I run `pnpm install && pnpm dedupe` on master.

bharatkashyap (approved):

PR #4373, opened 2024-11-06
[docs] Specify that dialogs payload is not restricted to string type

prakhargupta1:


docs/data/toolpad/core/components/use-dialogs/CustomDialogWithPayload.tsx
```diff
@@ -8,7 +8,7 @@ import DialogActions from '@mui/material/DialogActions';
 import TextField from '@mui/material/TextField';
 import Stack from '@mui/material/Stack';
 
-function MyCustomDialog({ payload, open, onClose }: DialogProps<string>) {
+function MyCustomDialog({ payload, open, onClose }: DialogProps<any>) {
```
Janpot, line 11: no, we need to advertise strong typing

```suggestion
function MyCustomDialog({ payload, open, onClose }: DialogProps<string>) {
```

prakhargupta1:
> What do you think about that?

The advanced example demonstrates this very well.👍

PR #4375, opened 2024-11-06
[docs] Add component as `payload` example to `useDialogs` docs

### LINKED ISSUE #4336: useDialogs Make Payload a Node Rather than a String
### Summary

The [dialog payload](https://mui.com/toolpad/core/react-use-dialogs/#with-dialog-payload) specifies that payload is expected to be of type `string`, which feels restrictive. For example, my use case involves passing a complex form as the payload. Expanding support to allow any type of payload would be highly beneficial.

### Examples

_No response_

### Motivation

Extending the usefulness of `useDialogs`.

**Search keywords**: payload useDialogs

bharatkashyap:
- Closes #4336

Screenshot 2024-11-06 at 4 30 09 PM

https://deploy-preview-4375--mui-toolpad-docs.netlify.app/toolpad/core/react-use-dialogs/#component-as-dialog-payload

### DIFF
--- a/docs/data/toolpad/core/components/use-dialogs/CustomDialogWithPayloadAdvanced.js
+++ b/docs/data/toolpad/core/components/use-dialogs/CustomDialogWithPayloadAdvanced.js
@@ -10,18 +10,7 @@ import DialogActions from '@mui/material/DialogActions';
 import TextField from '@mui/material/TextField';
 import Stack from '@mui/material/Stack';
 import MenuItem from '@mui/material/MenuItem';
-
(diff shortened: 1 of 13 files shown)


Janpot:
I think the ideal example for this use case uses a plain object, with multiple properties inside, one of which can be a react element

bharatkashyap:
> I think the ideal example for this use case uses a plain object, with multiple properties inside, one of which can be a react element

Agree, updated the example

apedroferreira (approved):

PR #4411, opened 2024-11-13
[docs] Add a new section for Integration 

prakhargupta1:
Preview:


prakhargupta1:
@bharatkashyap Not sure why Netlify preview fails. Can you have a look?

docs/data/toolpad/core/integrations/react-router.md
```diff
@@ -0,0 +1,399 @@
...
+title: React router - Integration
+---
+
+# React Router
+
+<p class="description">To integrate Toolpad Core into a single-page app (with Vite, for example) using React Router, follow these steps.</p>
+
+## Wrap all your pages in an `AppProvider`
+
+In your router configuration (e.g.: `src/main.tsx`), use a shared component or element (e.g.: `src/App.tsx`) as a root **layout route** that will wrap the whole application with the `AppProvider` from `@toolpad/core/react-router-dom`.
```
apedroferreira, line 11:
```suggestion
In your router configuration (e.g.: `src/main.tsx`), use a shared component or element (e.g.: `src/App.tsx`) as a root **layout route** that wraps the whole application with the `AppProvider` from `@toolpad/core/react-router-dom`.
```

docs/data/toolpad/core/integrations/react-router.md
```diff
@@ -0,0 +1,399 @@
...
+```tsx title="src/pages/orders.tsx"
+import * as React from 'react';
+import Typography from '@mui/material/Typography';
+
+export default function OrdersPage() {
+  return <Typography>Welcome to the Toolpad orders!</Typography>;
+}
+```
+
+You can then add these page components as routes to your React Router configuration (e.g.: `src/main.tsx`). By adding them as children of the layout route created in step 2, they will automatically be wrapped with that dashboard layout:
```
apedroferreira, line 138:
```suggestion
You can then add these page components as routes to your React Router configuration (e.g.: `src/main.tsx`). By adding them as children of the layout route created in step 2, they are automatically wrapped with that dashboard layout:
```

apedroferreira (changes requested):
Some Vale changes in my own copy :D

Also why did we remove the numbers from the headings?
In the React Router guide I sometimes mention "step 1" or "step 3" but this way it doesn't make sense anymore, we'd have to change it.

prakhargupta1:
> Also why did we remove the numbers from the headings?
> In the React Router guide I sometimes mention "step 1" or "step 3" but this way it doesn't make sense anymore, we'd have to change it.

Based on the core docs Integration section, I removed the numbers. There were two mentions of step 1, step 2 which I have now tackled.

apedroferreira (approved):
Thanks Prakhar!

bharatkashyap:
> Instead of a root level section, can we make it a sub-section of the Introduction? I think that makes more sense in the hierarchy of the navigation, what do you think?

We should have also added _redirects given that we went ahead with a new root level section, since the old URLs are now returning 404s: https://mui.com/toolpad/core/introduction/integration/ 

@prakhargupta1 Can you take this up? I'll redeploy the docs once this is merged

prakhargupta1:
Oh sorry, completely missed it. Thanks for pointing it out! On it now.

PR #4437, opened 2024-11-18
[DashboardLayout] Fix app bar items alignment

bharatkashyap:
| Before | After |
|--------|-------|
| Screenshot 2024-11-18 at 6 53 12 PM | Screenshot 2024-11-18 at 6 43 41 PM |


apedroferreira (changes requested):
Requesting changes just for adding `alignItems` in the stack that wraps the toolbar slots too.
Thanks for looking into this! I had missed it as I hasn't tested it with the account button in the top bar.

apedroferreira (approved):

PR #4442, opened 2024-11-19
[DashboardLayout] Add branding prop as override

apedroferreira:
Small change just to try to address this old discussion: https://github.com/mui/toolpad/discussions/4118

Add a `branding` prop to `DashboardLayout` that can be used to override the `AppProvider` branding.

Also slightly increased spacing between logo and title in the layout header bar, as I noticed it was currently too small for most logos.

### DIFF
--- a/docs/pages/toolpad/core/api/dashboard-layout.json
+++ b/docs/pages/toolpad/core/api/dashboard-layout.json
@@ -1,6 +1,10 @@
 {
   "props": {
     "children": { "type": { "name": "node" }, "required": true },
+    "branding": {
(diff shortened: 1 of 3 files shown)


bharatkashyap (approved):

bharatkashyap:
@apedroferreira Missed mentioning that we should document this override prop in this section: https://mui.com/toolpad/core/react-dashboard-layout/#branding

apedroferreira:
> @apedroferreira Missed mentioning that we should document this override prop in this section: https://mui.com/toolpad/core/react-dashboard-layout/#branding

Good point, we should do it, will add later!

PR #4443, opened 2024-11-20
[core] Remove affected deps for CVE-1333

bharatkashyap:
- From https://github.com/mui/toolpad/security/dependabot/95
- Cannot be automatically fixed
- I removed our dependencies that bring in the affected dependency (`cross-spawn <= 7`) and see no changes to the docs or the tests for `@mui/utils`


Janpot (changes requested):
this is not  going to be a sufficient replacement for the `title` dependency. I'd prefer to keep it.

PR #4446, opened 2024-11-21
CHORE: replace "Toolpad" default title with empty string

Petar-Dimitrov-AXA:
As discussed in issue https://github.com/mui/toolpad/issues/4445, I replaced the default "Toolpad" title with an empty string.


Petar-Dimitrov-AXA:
If that's the case, please consider adding a disableHeaderTitle prop. Setting the title to '' from client side feels kind a pointless. It would be ideal if there's a disableHeaderTitle boolean prop so that the title node will not render.

Janpot:
To add more options, perhaps a bit more versatile is to have a prop on the DashboardLayout to override the title. Set it to any value you want to override the branding context, or empty string to disable altogether.

PR #4480, opened 2024-11-26
[PageContainer] Allow full-size content inside

apedroferreira:
Closes https://github.com/mui/toolpad/issues/4324
Needed it already in the CRUD also for the list table to take up the full space in the layout.

As per the discussion in https://github.com/mui/toolpad/issues/4324:
- Allow full-size content inside `PageContainer` with `flex: 1` or `height: 100%`.
- Add `sx` prop to `PageContainer`.
- Separate `PageHeader` from `PageContainer` to optionally be used as a standalone component
- Documented both the cases above


docs/pages/toolpad/core/api/sign-out-button.json
```diff
@@ -1,5 +1,5 @@
 {
-  "props": {},
+  "props": { "children": { "type": { "name": "node" } } },
```
apedroferreira, line 2: Proptypes are now being generated for the components which don't match the name of the folder (secondary ones), I think that should be a good thing?

scripts/generateProptypes.ts
```diff
@@ -204,7 +204,7 @@ async function run(argv: HandlerArgv) {
         folderName = folderName.slice(9);
       }
 
-      return fileName === folderName;
+      return !fileName.endsWith('.test');
```
apedroferreira, line 207: Generated proptypes for all non-test files inside the folder of each component.

apedroferreira:
Looks like the CI is having out of memory issues again, we probably either need to manage dependencies to stay under or increase instance sizes?

bharatkashyap:
> Looks like the CI is having out of memory issues again, we probably either need to manage dependencies to stay under or increase instance sizes?

Yes, although I'm wondering what exactly triggered the CI out of memory issues

scripts/generateProptypes.ts
bharatkashyap: This is also generating proptypes for components marked as `@ignore` in the JSDoc comments - something to fix?

docs/data/toolpad/core/components/page-container/page-container.md
```diff
@@ -102,10 +102,22 @@ export default function Example() {
...
 
 {{"demo": "ActionsPageContainer.js", "height": 300}}
 
 ## Responsiveness
 
 The Page Container component inherits the properties of the Material&nbsp;UI [Container](https://mui.com/material-ui/react-container/) component. You can use its [`maxWidth`](https://mui.com/material-ui/api/container/#container-prop-maxWidth) and [`fixed`](https://mui.com/material-ui/api/container/#container-prop-fixed) properties to control the bounds of the page. Set `maxWidth` to `false` to disable the container altogether and have the content bleed right up to the edges of the page.
+
+## Standalone page header
+
+The `PageHeader` component included in `PageContainer` can be imported and used by itself if you wish to so, for more freedom of customization.
```
bharatkashyap, line 115:
```suggestion
The `PageHeader` component included in `PageContainer` can be imported and used by itself if you wish to do so, for more freedom of customization.
```

bharatkashyap (changes requested):
Looks good, just some suggestions on the docs demo and one flag on the proptypes script.

Also, I'm guessing this counts as a breaking change since we are renaming `PageContainerToolbar` to `PageHeaderToolbar`

docs/data/toolpad/core/components/page-container/PageContainerHeader.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const router = useDemoRouter('/orders');
+
+  const theme = useTheme();
+
+  return (
+    <AppProvider navigation={NAVIGATION} router={router} theme={theme}>
+      <Paper sx={{ width: '100%' }}>
+        {/* preview-start */}
+        <Container sx={{ my: 2 }}>
+          <PageHeader>Page content</PageHeader>
```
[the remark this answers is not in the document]
apedroferreira: Whoops, thanks for noticing that!

scripts/generateProptypes.ts
apedroferreira: I guess we can try to exclude them if it's easy to do.

apedroferreira: Also I just noticed every component is not listed in the left sidebar... I'll find a better way.

bharatkashyap (approved):
Looks good to me!

bharatkashyap:
Quick comment on second thought: The `PageHeader` demo: https://deploy-preview-4480--mui-toolpad-docs.netlify.app/toolpad/core/react-page-container/#standalone-page-header doesn't include how it would be used with the `DashboardLayout`. 

I'm wondering if we could show how we could use `PageHeader` inside the `DashboardLayout` versus the `PageContainer`, so that it doesn't break too much with the other demos. Wdyt?

bharatkashyap (changes requested):
On second thought: 

Apart from standalone usage, should we make the `PageHeader` composable inside the `PageContainer` as well? Just as we made the `ThemeSwitcher` composable inside `ToolbarActions`? 

Something like 
```tsx
<PageContainer>
  <PageHeader slots={{toolbar: CustomToolbar}} />
  {children}
</PageContainer>
```

This way, we could also move the `toolbar` slot to `PageHeader`, which better reflects its position anyway.

Wdyt? I know it's a tangential part of this PR, so feel free to extract the rest if this feels too blocking

apedroferreira:
> On second thought:
> 
> Apart from standalone usage, should we make the `PageHeader` composable inside the `PageContainer` as well? Just as we made the `ThemeSwitcher` composable inside `ToolbarActions`?
> 
> Something like
> 
> ```tsx
> <PageContainer>
>   <PageHeader slots={{toolbar: CustomToolbar}} />
>   {children}
> </PageContainer>
> ```
> 
> This way, we could also move the `toolbar` slot to `PageHeader`, which better reflects its position anyway.
> 
> Wdyt? I know it's a tangential part of this PR, so feel free to extract the rest if this feels too blocking

Sounds like a probably good idea, we can discuss it just to make sure.
The only drawback from that suggestion would be that the `PageContainer` wouldn't as well work out of the box, but we could probably fix that by giving it a slot for the header content with `PageHeader` as the default?

bharatkashyap:
> Sounds like a probably good idea, we can discuss it just to make sure. The only drawback from that suggestion would be that the `PageContainer` wouldn't work as well out of the box, but we could probably fix that by giving it a slot for the header content with `PageHeader` as the default?

I feel we can enforce a migration now to a composable API and remove the `header` and `toolbar` slots entirely given that it is likely to be the more popular way of customization. The breaking change would have to be adding a:

```diff
<PageContainer>
+  <PageHeader>
+     <PageHeaderBreadcrumbs /> // would contain the default breadcrumbs
+    <PageHeaderToolbar> // (optional, this would replace the `toolbar` slot)
+  </PageHeader>
   {children}
 </PageContainer>
```
to all uses of `PageContainer` to get the breadcrumbs out of the box.

We can discuss this tomorrow or in next week's grooming 👍🏻

Janpot:
What would be the difference between PageContainer and 

> On second thought:
> 
> Apart from standalone usage, should we make the `PageHeader` composable inside the `PageContainer` as well? Just as we made the `ThemeSwitcher` composable inside `ToolbarActions`?
> 
> Something like
> 
> ```tsx
> <PageContainer>
>   <PageHeader slots={{toolbar: CustomToolbar}} />
>   {children}
> </PageContainer>
> ```
> 
> This way, we could also move the `toolbar` slot to `PageHeader`, which better reflects its position anyway.
> 
> Wdyt? I know it's a tangential part of this PR, so feel free to extract the rest if this feels too blocking

what would be the difference between the PageContainer and a @mui/material Container? Wouldn't it be enough to just build a PageHeader component?

apedroferreira:
> > Sounds like a probably good idea, we can discuss it just to make sure. The only drawback from that suggestion would be that the `PageContainer` wouldn't work as well out of the box, but we could probably fix that by giving it a slot for the header content with `PageHeader` as the default?
> 
> I feel we can enforce a migration now to a composable API and remove the `header` and `toolbar` slots entirely given that it is likely to be the more popular way of customization. The breaking change would have to be adding a:
> 
> ```diff
> <PageContainer>
> +  <PageHeader>
> +     <PageHeaderBreadcrumbs /> // would contain the default breadcrumbs
> +    <PageHeaderToolbar> ... </PageHeaderToolbar>// (optional, this would replace the `toolbar` slot)
> +  </PageHeader>
>    {children}
>  </PageContainer>
> ```
> 
> to all uses of `PageContainer` to get the breadcrumbs out of the box.
> 
> We can discuss this tomorrow or in next week's grooming 👍🏻

One of the main advantages of using Toolpad is that it does standard things for you. I feel like if we separate the container into too many sub-components that always needs to used it kind of complicates things.

I prefer the approach where we provide a few components and more detailed customization can be made with slots.

> what would be the difference between the PageContainer and a @mui/material Container? Wouldn't it be enough to just build a PageHeader component?

Not a lot right now, but at the moment the `PageContainer` has styles that make it so that the content inside it can take up the full height inside the layout, and some default spacing around its content.

It seems like these wouldn't be difficult to do with a standard `Container`, especially if we provide an example on the docs on how to do full-size content in a page... So we could export the `PageHeader` only and include an example like that in its documentation. I could try that in this PR.

docs/data/toolpad/core/components/page-container/ActionsPageContainer.js
```diff
@@ -35,21 +39,34 @@ function PageToolbar({ status }) {
...
 // preview-end
 
-PageToolbar.propTypes = {
+CustomPageHeader.propTypes = {
   status: PropTypes.string.isRequired,
 };
 
 export default function ActionsPageContainer() {
   const router = useDemoRouter();
   const status = 'supesh';
```
bharatkashyap, line 66:
```suggestion
  const status = 'Active';
```

docs/data/toolpad/core/components/page-container/ActionsPageContainer.tsx
```diff
@@ -34,17 +38,26 @@ function PageToolbar({ status }: { status: string }) {
...
+    [status],
+  );
+
+  return <PageHeader slots={{ toolbar: CustomPageToolbarComponent }} />;
+}
 // preview-end
 
 export default function ActionsPageContainer() {
   const router = useDemoRouter();
   const status = 'supesh';
```
bharatkashyap, line 57:
```suggestion
  const status = 'Active';
```

bharatkashyap (approved):
Looks good! We should make sure to mark this as a breaking change in the release notes

PR #4486, opened 2024-11-27
CRUD component

apedroferreira:
Implement CRUD feature as specified in https://github.com/mui/toolpad/issues/4146

https://deploy-preview-4486--mui-toolpad-docs.netlify.app/toolpad/core/react-crud/

Tests and examples will be added in separate PRs as this PR has already grown too large.

To help review, the contents of this PR are mostly:
- Relevant changes for the new feature (new components) in `packages/toolpad-core`.
- Documentation added in `docs`, many examples but with a lot of repetition between them, feel free to just mostly take a look at the new documentation page for the `Crud` component.
- Added orders CRUD to the `playground`s for Vite, Next.js App Router and Next.js Pages Router.
- Other than that there are very few changes.

It took a while to reach this first version but still feel free to provide any major or minor feedback!

### DIFF
--- a/packages/toolpad-core/src/Crud/List.tsx
+++ b/packages/toolpad-core/src/Crud/List.tsx
@@ -429,9 +429,13 @@ function List<D extends DataModel>(props: ListProps<D>) {
             [`& .${gridClasses.columnHeader}:focus-within, & .${gridClasses.cell}:focus-within`]: {
               outline: 'none',
             },
-            [`& .${gridClasses.row}:hover`]: {
-              cursor: 'pointer',
-            },
+            ...(onRowClick
+              ? {
+                  [`& .${gridClasses.row}:hover`]: {
+                    cursor: 'pointer',
+                  },
+                }
+              : {}),
             ...slotProps?.dataGrid?.sx,
           }}
         />


docs/package.json
```diff
@@ -25,16 +25,17 @@
...
-    "@mui/material": "6.3.1",
+    "@mui/material": "6.4.1",
     "@mui/material-nextjs": "6.3.1",
-    "@mui/styles": "6.3.1",
-    "@mui/system": "6.3.1",
-    "@mui/utils": "6.3.1",
+    "@mui/styles": "6.4.1",
+    "@mui/system": "6.4.1",
+    "@mui/utils": "6.4.1",
+    "@mui/x-data-grid-pro": "7.26.0",
```
apedroferreira, line 38: Adding Data Grid pro for one of the examples in the docs. 
Should we skip this? A license might be needed not to show the watermark too.

packages/toolpad-core/src/Crud/List.tsx
```diff
@@ -0,0 +1,450 @@
...
+    return [
+      ...(fields.map((field) =>
+        field.type === 'longString' ? { ...field, type: 'string' } : field,
+      ) as GridColDef[]),
+      {
+        field: 'actions',
+        type: 'actions',
+        flex: 1,
+        align: 'right',
+        getActions: ({ id }) => [
```
apedroferreira, line 318: A pinned column would be very useful for these actions but is not a feature of the basic data grid.
Open for discussion, the current behavior might not be ideal.

packages/toolpad-core/src/Crud/Show.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+  invariant(dataSource, 'No data source found.');
+
+  const { fields, ...methods } = dataSource;
+  const { getOne, deleteOne } = methods;
+
+  const dialogs = useDialogs();
+  const notifications = useNotifications();
+
+  const [data, setData] = React.useState<D | null>(null);
```
apedroferreira, line 64: Not using any data fetching library (such as `react-query`) for now, so no caching of requests, for example.
This simple implementation could be a placeholder for using the MUI X data providers when they're ready.

packages/toolpad-core/src/Crud/FormPage.tsx
```diff
@@ -0,0 +1,376 @@
...
+    onSubmit,
+    onSubmitSuccess,
+    resetOnSubmit = true,
+    localeText,
+  } = props;
+  const { fields, validate } = dataSource;
+
+  const notifications = useNotifications();
+
+  const [formState, setFormState] = React.useState<{
```
apedroferreira, line 58: Not using any external form libraries, just keeping form values in state.
It made the implementation more difficult and very specific + adds an extra choice for the user, this implementation seems to work well so far.

apedroferreira (commented):
I've added some comments with possible things that I could've brought up earlier.

packages/toolpad-core/src/Crud/FormPage.tsx
apedroferreira: Update: I've made the `CrudForm` component being used by the `Create` and `Edit` components implementation-agnostic and exported, so it should be possible to use it with any form library in advanced customization use cases.

https://github.com/mui/toolpad/pull/4486/files#diff-774ee88d140c8d65cc0b84644c1a6ae920994e79a17c3d751235e53f04bd8cdd

packages/toolpad-core/src/Crud/Show.tsx
apedroferreira: Update: I've added caching with the `dataSourceCache` prop based on the approach from https://mui.com/x/react-data-grid/server-side-data/?srsltid=AfmBOooZZzlgdHeSqhZ6ZM7J9SKLqXZeo6W07kidqbp80tqql3xIeSHc#data-caching.

It's documented in the `Data caching` section of the component docs.

docs/package.json
bharatkashyap: I think it's a good example to demonstrate how you could combine the Crud component with paid X Datagrid offerings, I'd vote to keep that example there

packages/toolpad-core/src/Crud/Show.tsx
bharatkashyap: Makes sense, if someone wants to use a custom cache (as documented in the link above) they could pass in a custom Datagrid in the List component's `slot`, right?

apedroferreira: I didn't document using a custom cache but it would be same as the approach in that page. I'm recommending that users pass an instance like `new DataSourceCache()` to the `dataSourceCache` prop but they could use a different class with the same methods if they want to.

In the end we would just use the same exact cache feature as MUI X once they finish the server-side data and make it free/public, so this is kind of a placeholder.

packages/toolpad-core/src/Crud/Create.tsx
```diff
@@ -0,0 +1,220 @@
...
+    const manualCache = dataSourceCache ?? crudContext.dataSourceCache;
+    return typeof manualCache !== 'undefined' ? manualCache : new DataSourceCache();
+  }, [crudContext.dataSourceCache, dataSourceCache]);
+  const cachedDataSource = useCachedDataSource<D>(dataSource, cache) as NonNullable<
+    typeof props.dataSource
+  >;
+
+  const { fields, createOne, validate } = cachedDataSource;
+
+  const [formState, setFormState] = React.useState<{
```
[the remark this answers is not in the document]
apedroferreira: But that's only the initial state value, it doesn't run every render.
I think it should be ok, `.reduce` might be more efficient but not as readable.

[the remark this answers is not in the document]
apedroferreira: hm, i just looked it up and looks like it might run every render...
I guess that just passing a function with the same result still using `.map` and `.filter` would only run once though.

docs/data/toolpad/core/components/crud/crud.md
```diff
@@ -0,0 +1,356 @@
...
+:::
+
+## Data sources
+
+A **data source** is an object with a particular shape that must be used to configure the `Crud` component or its subcomponents.
+It includes:
+
+- A `fields` property, which extends the [MUI X Data Grid column definition](https://mui.com/x/react-data-grid/column-definition/) where different fields, header names and field types, among others, can be specified for your data.
+
+  An `id` field (string or number) is mandatory so that individual items can be identified.
```
prakhargupta1, line 43:
```suggestion
 - An `id` field (string or number) is mandatory so that individual items can be identified.
```

docs/data/toolpad/core/components/crud/crud.md
```diff
@@ -0,0 +1,356 @@
...
+
+  An additional type `longString` is provided for multiline form fields.
```
prakhargupta1, line 45:
```suggestion
 - An additional type `longString` is provided for multiline form fields.
```

docs/data/toolpad/core/components/crud/crud.md
```diff
@@ -0,0 +1,356 @@
+---
+productId: toolpad-core
+title: Crud
+components: Crud, CrudProvider, List, Show, Create, Edit, CrudForm
+---
+
+# Crud
```
prakhargupta1, line 7: Should it be CRUD or Crud?

docs/data/toolpad/core/components/crud/crud.md
```diff
@@ -0,0 +1,356 @@
...
+
+<p class="description">The Crud component provides UIs to interact with data from any data source.</p>
+
+With the `Crud` component and its subcomponents you can easily generate pages where items from your data source can be listed in a table, shown individually in detail, or created and edited with forms. All with minimal configuration from a single data source definition.
+
+## Demo
+
+The `Crud` component automatically generates CRUD pages for your data source, automatically integrated with your routing solution. All that is required is a `dataSource` definition and a `rootPath` for the CRUD pages.
+
+{{"demo": "CrudBasic.js", "height": 600, "iframe": true}}
```
prakhargupta1, line 17: Code highlight doesn't seem to be applied for the preview code in case of TS.

bharatkashyap (commented):
Looks great @apedroferreira! A huge addition to the component set. 

I think the only missing piece is integration with the new `LocalizationProvider` (https://github.com/mui/toolpad/pull/4620) so that we're not using any hard-coded strings and all text can be overridden.

docs/data/toolpad/core/components/crud/crud.md
```diff
@@ -0,0 +1,356 @@
...
+
+    return errors;
+  },
+  // ...
+}
+```
+
+This function will run automatically against the current values when the user tries to submit a form inside the `Crud` component, or changes any of its fields.
+
+#### Integration with external schema libraries
```
Janpot, line 251: Have you seen https://standardschema.dev/?

Janpot (approved):
I believe we can start with this initial feature set 👍

docs/data/toolpad/core/components/crud/crud.md
apedroferreira: These lines are kind of "under" the first point, that's why I structured them like this. I'll see if there's a better way to structure it.

docs/data/toolpad/core/components/crud/crud.md
apedroferreira: According to what we discussed about MUI component casing the component should be called `Crud`.

docs/data/toolpad/core/components/crud/crud.md
apedroferreira: I know, it doesn't seem like it's ready for components like `<Crud<Note>` but that's valid Typescript, I guess it's an issue with the highlighting...

docs/data/toolpad/core/components/crud/crud.md
apedroferreira: Hadn't seen it, we can probably make the `validate` function be compatible with this standard.

apedroferreira:
> Looks great @apedroferreira! A huge addition to the component set.
> 
> I think the only missing piece is integration with the new `LocalizationProvider` (#4620) so that we're not using any hard-coded strings and all text can be overridden.

Here are the localization changes if you want to take a look, anyway it seems to work fine just followed the work you did in the other components. https://github.com/mui/toolpad/pull/4486/commits/00342c9038617d5087538c0de017cd3c94d6a522

Tagging you for review just in case but I'll merge soon anyway probably.

apedroferreira:
Oh, and interpolation in these would be useful because the delete confirmation messages used to have the id of the item to be deleted.
But I guess we can wait until we add interpolation and then improve the delete messages.

apedroferreira:
> Oh, and interpolation in these would be useful because the delete confirmation messages used to have the id of the item to be deleted.
> But I guess we can wait until we add interpolation and then improve the delete messages.

Oh, and interpolation in these would be useful because the delete confirmation messages used to have the id of the item to be deleted.
But I guess we can wait until we add interpolation and then improve the delete messages.

docs/data/toolpad/core/components/crud/crud.md
apedroferreira: I've changed the format of the `validate` function to follow the Standard Schema, this should make it really straight forward to integrate with schema libraries that adhere to it: https://deploy-preview-4486--mui-toolpad-docs.netlify.app/toolpad/core/react-crud/#form-validation

apedroferreira:
Review feedback has been addressed, will merge and release tomorrow.
Tests and example can be merged right after the release.

PR #4489, opened 2024-11-28
[SignInPage] Remove `docs` context from component logic

bharatkashyap:
- From https://github.com/mui/toolpad/pull/4418#discussion_r1847722687
- Removed two instances of checking if the component is being used inside the docs
- Autofocus issue solved via adding a `slotProp` of `autoFocus: false` to each docs demo and removed handling it inside the component
- Conditionally setting the loading state of the Sign In Button on deeper analysis does not appear to be required at all really, simply removed


packages/toolpad-core/src/SignInPage/SignInPage.tsx
```diff
@@ -366,7 +364,7 @@ function SignInPage(props: SignInPageProps) {
                         const oauthResponse = await signIn?.(provider, undefined, callbackUrl);
                         setFormStatus((prev) => ({
                           ...prev,
-                          loading: oauthResponse?.error || docs ? false : prev.loading,
+                          loading: false,
```
mnajdova, line 367: Shouldn't we keep the `oauthResponse?.error` logic here? If not should we just remove the loading and depend on whether it is defined in the `prev`?

mnajdova (approved):
I left one clarification. Alsi I feel we are over-engineering the singleProvider const by using memo there, but we can handle it separately :)

packages/toolpad-core/src/SignInPage/SignInPage.tsx
bharatkashyap: The idea in this code was to use `oauthResponse?.error` being defined as a way to check that the operation is complete and set the loading state to false (but create an escape hatch. for the docs since no oauthResponse is created inside the docs). 

`prev.loading` will be true in all cases, and letting `loading` stay that way was done so that even if a response is received, the button would remain in a "loading" state till the OAuth redirect occurs, to visually indicate to the user that something is still "pending". This behaviour is present only in the oauth case, since none of the other auth methods involve an intermediate redirect. 

On second thought, I feel it's a nice detail to let the user know that the redirect is pending, so what I feel we can do is to simply remove the additional `docs` check here, and change the oauth docs demos to return a mock error with "this is a fake error message", so that the button goes back into the default state.

<table>
  <thead>
    <tr>
      <th>Current</th>
      <th>After this PR</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>
        <video src="https://github.com/user-attachments/assets/36b0358b-a479-473c-b951-5386bbdc0aea" controls></video>
      </td>
      <td>
        <video src="https://github.com/user-attachments/assets/04efa460-b3c3-47e7-8ae3-fd6b504a84b0" controls></video>
      </td>
    </tr>
  </tbody>
</table>

mnajdova: Yeah, that makes sense 👌

mnajdova (approved):

PR #4490, opened 2024-11-28
[DashboardLayout] Add `roles` to `navigation`

bharatkashyap:
- Closes https://github.com/mui/toolpad/issues/4120
- Add a `roles: string[]` to `NavigationPageItem`
- Add `groups?: string[]` and `roles?: string[]` to `Session`
- Check if `user.groups` or `user.roles` has a matching role in `hasMatchingRole`
- Remove pages from navigation and render a 404 on navigation to pages for which there is no access
- Remains agnostic to the method of providing the `roles` or `groups` on the session object


bharatkashyap:
> I'm hesitant building anything RBAC before we have the CRUD features and have set up a pro version.

We discussed this week that given https://github.com/mui/toolpad/issues/4120 is the most upvoted (and fastest growing in upvotes) issue, does it make sense to keep a minimal version of the `roles` interface blocked till the entirety of `CRUD` has landed? 

Once we have CRUD and the pro version, we can integrate the `roles` we define here with the ability to restrict some CRUD actions on some resources.

Wdyt?

Janpot:
My thinking was that there are more ways to do authorization than RBAC. Some users may only need a very simple setup , such as checking against an email address. We could serve more kinds of users with a less opinionated API. Then see which usage parents emerge and provide convenience APIs for the most used ones. I feel like now we're locking every use case into RBAC.

JuR-0:
As @Janpot mentioned, adding a simple callback driving the display of navigation items could solve any RBAC needs as well as many other.
Then RBAC would just be:
- A custom session
- A callback leveraging roles/groups as needed
The availability of content can be tackled at the page level by the user as a first approach
Is it the goal we are pursuing ? if yes I could work on a draft

janeisklar:
I gave the changes in your pull request a go, but realized that it rendered me unable to access any pages that I don't want to show up in the navigation. For example, my startpage is a simple redirect to another subpage which did not work anymore (as it showed the 404 instead) unless I added it to the navigation (which I don't want).

While I think it is useful to have a central place where I configure which permissions should be necessary for each and every page, I would need a mechanism to then have them not show up in the navigation at all.

Then again, the configuration of the navigation may not be the best place to configure permissions as it feels somewhat awkward to add paths that should not show up only so I can set up their permissions.

Looking forward to having this implemented at some point. Hope this feedback helps in some way:)

PR #4500, opened 2024-12-03
[examples] [docs] Add Vite example with Firebase auth

### LINKED ISSUE #4420: [docs] Add a Vite + functional auth example

bharatkashyap:
- Closes #4420
- Also fixes CodeSandbox and StackBlitz links not being visible: codesandbox -> codeSandbox, stackblitz -> stackBlitz


Janpot:
What's special about the themed example that we need it in more than one runtime? sure it doesn't do anything runtime specific, right?

Also feels to me like a vite+firebase example should stand on its own. Adding other concepts in there would make it a bit confusing I think.

bharatkashyap:
> What's special about the themed example that we need it in more than one runtime? sure it doesn't do anything runtime specific, right?
> 
> Also feels to me like a vite+firebase example should stand on its own. Adding other concepts in there would make it a bit confusing I think.

Fair enough, I can remove the theme and custom components from this example and make it focus on the Vite + Firebase integration

examples/core/firebase-vite/src/theme.ts
```diff
@@ -0,0 +1,13 @@
+'use client';
+import { createTheme } from '@mui/material/styles';
+import getMPTheme from './theme/getMPTheme';
+
+const lightTheme = createTheme(getMPTheme('light'));
+const darkTheme = createTheme(getMPTheme('dark'));
+
+const theme = {
```
apedroferreira, line 8: Can't we use a CSS variables theme instead? Also I can't find this `getMPTheme`, is this outdated?

apedroferreira (approved):
Nice, just some small things!

apedroferreira:
Oh i guess the screenshots are outdated also if we're not using that theme after all.

examples/core/firebase-vite/src/theme.ts
bharatkashyap: Yes, missed deleting this file (isn't used anywhere anymore)

PR #4513, opened 2024-12-05
Micro-improvements concerning types

gicrisf:
1. There is no `Navigation` type in `@toolpad/core/react-router-dom`. However, I found it in `@toolpad/core/AppProvider`;
2. Added explicit type `React.FC` for the `App` component (otherwise importing `* as React from 'react'` would be unnecessary).


gicrisf:
Sorry, I just found out that you're already working on this:
https://github.com/mui/toolpad/issues/4270

It was in the closed issues, so I missed it before creating the PR

Issue #4406, opened 2024-11-12
Add props to control sidebar's collapsed state in DashboardLayout component


Mohammad-210:
### Summary

This feature request is to enhance the DashboardLayout component by adding two new props:

collapsed: A boolean prop that controls the initial state of the sidebar (collapsed or expanded).
onToggleSidebar: A function prop that allows toggling the sidebar’s collapsed state from outside the DashboardLayout component.
This setup will make the sidebar’s state externally controllable, enabling more dynamic and flexible UI behaviors in various responsive scenarios.

### Examples

import React, { useState } from 'react';
import DashboardLayout from './DashboardLayout';

const AppProvider = () => {
  const [isSidebarCollapsed, setIsSidebarCollapsed] = useState(false);

  const handleToggleSidebar = (collapsed) => {
    setIsSidebarCollapsed(collapsed);
  };

  return (
    <DashboardLayout
      collapsed={isSidebarCollapsed}
      onToggleSidebar={handleToggleSidebar}
    >
      {/* Application Content */}
    </DashboardLayout>
  );
};

export default AppProvider;

### Motivation

This feature will improve DashboardLayout’s flexibility and control by allowing parent components to manage and toggle the sidebar’s collapsed state. With responsive design requirements, many apps need to control the sidebar's visibility dynamically based on user interactions or screen size. This enhancement will make DashboardLayout more adaptable for complex UI requirements and enhance user experience across devices. The issue i face is i need a state to ensure the sidebar is collapsed or not to manage the responsiveness of design of rendered children comp  

**Search keywords**: sidebar toggle, sidebar collapsed state, DashboardLayout, Material-UI, Toolpad sidebar, responsive sidebar, toggle sidebar from parent, sidebar props

bharatkashyap:
Agree with providing a way to access the state of the sidebar, and also imperatively setting it. An alternative way of doing this is through a `useSidebar` hook:

```tsx
const { collapsed, setCollapsed } = useSidebar();

const onClick = () => { if(!collapsed) { setCollapsed(); } 

```

This will enable components across the app to be able to access (and imperatively set) the sidebar's state instead of having to share props.

vikasgurjar:
@Mohammad-210 are you planning to work on this create a PR? If not, I can help with the PR

Janpot:
a controlled prop makes sense to me. a hook will be constraint to descendants only, which is probably too limiting.

Mohammad-210:
[...]

PR #4516, opened 2024-12-06
add prop sidebarCollapsed and onToggleSidebar

vikasgurjar:
Closes [#4406](https://github.com/mui/toolpad/issues/4406)
- [ ] I've read and followed the [contributing guide](https://github.com/mui/toolpad/blob/master/CONTRIBUTING.md#sending-a-pull-request) on how to create great pull requests.


vikasgurjar:
@Janpot @bharatkashyap  from previous convo [here](https://github.com/mui/toolpad/pull/4514)

> edit: actually, it looks like the data grid has the convention of "xyzOpen", "onXyzOpen" and "onXyzClose". Perhaps that's the closest in spirit?

> @Janpot Are you suggesting to have two different callbacks for open and close events? onSidebarOpen and onSidebarClosed?


> When changing any of the demos and prop types, make sure to run pnpm docs:typescript:formatted to format TS demos and pnpm:proptypes to autogenerate prop type changes
I ran that command before raising the PR. What I have noticed is [this lines](https://github.com/mui/toolpad/blob/29cdab0d1d15158932ed5b700836c6b99b8df8a9/docs/data/toolpad/core/components/dashboard-layout/DashboardLayoutSidebarCollapsedProp.js#L66) props are wrongly generated. Correct proptype should be Proptype.func.required I think.

Janpot:
Yes, I'm suggesting to make this similar to how we do this in other products of the MUI ecosystem. for example [`menuOpen`](https://mui.com/x/api/data-grid/data-grid-premium/) and friends in the datagrid. 

So something like `navigationMenuOpen`, `onNavigationMenuOpen`, `onNavigationMenuClose`. I have a slight preference or naming it semantically (navigation menu) over naming it according to where it's laid out (side bar). This leaves the option open to introduce other variants some day. Not sure how the team feels about this @bharatkashyap @apedroferreira did we already name it this way in other places?

apedroferreira:
> Yes, I'm suggesting to make this similar to how we do this in other products of the MUI ecosystem. for example [`menuOpen`](https://mui.com/x/api/data-grid/data-grid-premium/) and friends in the datagrid.
> 
> So something like `navigationMenuOpen`, `onNavigationMenuOpen`, `onNavigationMenuClose`. I have a slight preference of naming it semantically (navigation menu) over naming it according to where it's laid out (side bar). This leaves the option open to introduce other variants some day. Not sure how the team feels about this @bharatkashyap @apedroferreira did we already name it this way in other places? If so maybe it makes more sense to stay consistent?
> 
> _please wait a moment with implementing this to allow us get to a consensus._

I've been calling it `sidebar` when it's purely referred to inside the `DashboardLayout` component, but "navigationMenu" would also be fine and is more semantic in terms of the global app - we could make an effort to gradually rename things in this direction.

About referring to the property as "open", it does make more sense than referring to it as "collapsed" now that we're making it a controlled prop... So I agree with your suggestions.

vikasgurjar:
@Janpot @apedroferreira I have made changes as per the above suggestions, please review them when possible

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -174,6 +189,17 @@ function DashboardLayout(props: DashboardLayoutProps) {
   const [isNavigationFullyExpanded, setIsNavigationFullyExpanded] =
     React.useState(isNavigationExpanded);
 
+  React.useEffect(() => {
+    if (typeof navigationMenuOpen === 'boolean') {
+      setIsNavigationExpanded(navigationMenuOpen);
+      if (navigationMenuOpen) {
```
apedroferreira, line 195: When `navigationMenuOpen` is a boolean, we cannot use the current `isDesktopNavigationExpanded` or `isMobileNavigationExpanded` states the way they're currently being used in this component.

So besides this effect, we need to make it so that all checks that involve `isDesktopNavigationExpanded` or `isMobileNavigationExpanded` check `navigationMenuOpen` in those cases instead.

And every current call to `setIsDesktopNavigationExpanded` or `setIsMobileNavigationExpanded` can be ignored when `navigationMenuOpen` is a boolean, but the appropriate `onNavigationMenuOpen` or `onNavigationMenuClose` should be called instead.

`onNavigationMenuOpen` and `onNavigationMenuClose` should also still be called when `navigationMenuOpen` is `undefined`, however, so that they can be used for other cases.

Your current approach requires an extra render for the effect to run, which should not be necessary, and the `navigationMenuOpen` value can become out of sync with the UI when the internal state is changed.

Hopefully this makes sense and points you in the right direction!

packages/toolpad-core/src/DashboardLayout/DashboardLayout.test.tsx
```diff
@@ -410,4 +410,62 @@ describe('DashboardLayout', () => {
     // Ensure that main content is still rendered
     expect(screen.getByText('Hello world')).toBeTruthy();
   });
+
+  test('renders the sidebar in collapsed state when navigationMenuOpen is false', () => {
```
apedroferreira, line 414: Thanks a lot for adding tests, they look good!

docs/data/toolpad/core/components/dashboard-layout/dashboard-layout.md
```diff
@@ -126,6 +126,12 @@ The layout sidebar can be hidden if needed with the `hideNavigation` prop.
 
 {{"demo": "DashboardLayoutSidebarHidden.js", "height": 400, "iframe": true}}
 
+### Toggle sidebar
```
apedroferreira, line 129: This section needs to be a bit more complete as in referring to the whole concept of using "controlled state" for the navigation menu, as well as what kind of use case they could serve.

We also need to document the new callbacks, and how they could be used together with `navigationMenuOpen`.

Anyway, I can take care of the documentation part when the functionality is done unless you really want to tackle it!

docs/data/toolpad/core/components/dashboard-layout/DashboardLayoutSidebarCollapsedProp.tsx
```diff
@@ -0,0 +1,112 @@
...
+    <AppProvider
+      navigation={NAVIGATION}
+      router={router}
+      theme={demoTheme}
+      window={demoWindow}
+    >
+      <DashboardLayout navigationMenuOpen={navigationMenuOpen}>
+        <DemoPageContent
+          pathname={pathname}
+          toggleSidebar={() => toggleSidebar(!navigationMenuOpen)}
```
apedroferreira, line 107: We should also show the typical usage for the new callbacks in this example, which would change the `navigationMenuOpen` state accordingly.

docs/data/toolpad/core/components/dashboard-layout/DashboardLayoutSidebarCollapsedProp.tsx
```diff
@@ -0,0 +1,112 @@
...
+   * Remove this when copying and pasting into your project.
+   */
+  window?: () => Window;
+}
+
+export default function DashboardLayoutSidebarCollapsedProp(props: DemoProps) {
+  const { window } = props;
+
+  const [pathname, setPathname] = React.useState('/dashboard');
+  const [navigationMenuOpen, toggleSidebar] = React.useState<boolean>(true);
```
apedroferreira, line 85:
```suggestion
  const [navigationMenuOpen, setNavigationMenuOpen] = React.useState<boolean>(true);
```

apedroferreira (changes requested):
Thanks for the effort so far, it's looking good but I have many important suggestions!

apedroferreira:
> Hi team,
> 
> Just wondering if there's any update on this PR. It seems like a straightforward and useful addition, and I noticed it's been open since December 2024 with no major changes requested.
> 
> Is there anything blocking this from moving forward? We're interested in this feature and would love to see it merged soon.
> 
> Thanks for your work on the project!

Toolpad Core won't be actively maintained in the short term so we're not prioritizing this feature right now unfortunately...

We will however publish a template soon in the Material UI docs with more customizable equivalent versions of these components, including separate components for the sidebar and header with the navigation menu expanded state being controlled from the outside and passed to them as props.

pongells:
> Toolpad Core won't be actively maintained in the short term so we're not prioritizing this feature right now unfortunately...

this is sad news :( 
toolpad core seemed really promising, just missing few things like this PR and a tiny bit more customisation.

> We will however publish a template soon in the Material UI docs with more customizable equivalent versions of these components, including separate components for the sidebar and header with the navigation menu expanded state being controlled from the outside and passed to them as props.

what do you mean by template? an example using shared components replacing the toolpad core features?

prakhargupta1:
> what do you mean by template? an example using shared components replacing the toolpad core features?

Yes, a dashboard example built using Material UI core components, so that there is no dependency on Toolpad. Here is a PR for this: https://github.com/mui/material-ui/pull/46376

PR #4523, opened 2024-12-09
[DashboardLayout] Add navigation prop as override

apedroferreira:
Closes https://github.com/mui/toolpad/issues/4327
Also addresses comment in https://github.com/mui/toolpad/pull/4442#issuecomment-2500063838 about mentioning these override props in the documentation.

Add a `navigation` prop to `DashboardLayout` that can be used to override the `AppProvider` navigation.

https://deploy-preview-4523--mui-toolpad-docs.netlify.app/toolpad/core/react-dashboard-layout


packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -139,6 +145,9 @@ function DashboardLayout(props: DashboardLayoutProps) {
   const navigationContext = React.useContext(NavigationContext);
   const appWindowContext = React.useContext(WindowContext);
 
+  const branding = { ...brandingContext, ...brandingProp };
+  const navigation = navigationProp ?? navigationContext;
```
bharatkashyap, line 149: I think it's important to document that the `navigation` override in the `DashboardLayout` completely replaces the one passed in the parent `AppProvider`, whereas the `branding` prop is merged with the parent `branding`

docs/data/toolpad/core/components/dashboard-layout/dashboard-layout.md
```diff
@@ -46,6 +48,8 @@ The `navigation` prop in the [AppProvider](https://mui.com/toolpad/core/react-ap
 
 The flexibility in composing and ordering these different elements allows for a great variety of navigation structures to fit your use case.
 
+Optionally, a specific `navigation` can also be passed as a prop to the `DashboardLayout` component itself.
```
bharatkashyap, line 51: I think this might be better placed as an `:::info` callout so that it isn't missed; and perhaps the demo can explain the usage of this prop

bharatkashyap (commented):
Looks good! Just recommending some modifications to the docs

docs/data/toolpad/core/components/dashboard-layout/dashboard-layout.md
apedroferreira: I feel like there's no need to demo these besides the callout as it's not the recommended usage so don't want to cause confusion, and the usage is the exact same for in the `AppProvider` so there's nothing new to really exemplify, plus no good place for an extra demo unless we made a whole section about it...

We could make a whole section about it but also thinking it might be a bit overkill, as the sections where we have this information are already the places where readers would look for them.

bharatkashyap: Fair enough, your point about this not being recommended usage and only an additional option makes sense 👍🏻

docs/data/toolpad/core/components/dashboard-layout/dashboard-layout.md
```diff
@@ -46,6 +48,10 @@ The `navigation` prop in the [AppProvider](https://mui.com/toolpad/core/react-ap
 
 The flexibility in composing and ordering these different elements allows for a great variety of navigation structures to fit your use case.
 
+:::info
+Optionally, a specific `navigation` can also be passed as a prop to the `DashboardLayout` component itself, which takes complete precedence over any `navigation` set in the `AppProvider`.
```
bharatkashyap, line 52:
```suggestion
Optionally, a specific `navigation` object can also be passed as a prop to the `DashboardLayout` component itself, which takes complete precedence over any `navigation` prop passed in the `AppProvider`.
```

bharatkashyap (approved):
Looks good!

docs/data/toolpad/core/components/dashboard-layout/dashboard-layout.md
```diff
@@ -37,6 +37,8 @@ This can be done via the `branding` prop in the [AppProvider](https://mui.com/to
 {{"demo": "DashboardLayoutBranding.js", "height": 400, "iframe": true}}
 
 :::info
+Optionally, a specific `branding` can also be passed as a prop to the `DashboardLayout` component itself, which properties take precedence over any properties of the `branding` set in the `AppProvider`.
```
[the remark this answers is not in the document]
apedroferreira: I think `whose` is for people so `which` should be correct here? @samuelsycamore

docs/data/toolpad/core/components/dashboard-layout/dashboard-layout.md
apedroferreira: It's not an object, it's an array :D I think we don't need to be too specific here?

docs/data/toolpad/core/components/dashboard-layout/dashboard-layout.md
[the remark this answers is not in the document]
apedroferreira: Got it, will adjust it, thanks!
Also I just looked it up and I guess that "whose" is actually gramatically correct in this case?... didn't know that.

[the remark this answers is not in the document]
samuelsycamore: "Whose" can work, but as a rule of thumb, if you find yourself writing a long sentence that includes a "whose" clause with multiple subjects and objects, it's often more readable and less ambiguous if you just start a new sentence and restate the subject directly. (Does "whose" refer to "the branding object" or "the DashboardLayout component"? It's not entirely clear.)

PR #4555, opened 2024-12-24
[Account] Prevent account info overflowing on kebab menu

bharatkashyap:
- Closes https://github.com/mui/toolpad/issues/4552
- A simple `overflow="hidden"` does the trick 
- The `AccountPreview` will truncate the account info if placed inside a div with a set width, otherwise will expand

Screenshot 2025-01-08 at 7 14 25 PM

- https://deploy-preview-4555--mui-toolpad-docs.netlify.app/toolpad/core/react-dashboard-layout/#examples


packages/toolpad-core/src/Account/AccountPreview.tsx
```diff
@@ -99,13 +99,19 @@ function AccountPreview(props: AccountPreviewProps) {
...
-        <Stack direction="row" justifyContent="flex-start" spacing={2}>
+        <Stack direction="row" justifyContent="flex-start" spacing={2} overflow="hidden">
           {avatarContent}
           <Stack direction="column" justifyContent="space-evenly">
-            <Typography variant="body2" fontWeight="bolder" noWrap>
+            <Typography
+              variant="body2"
+              fontWeight="bolder"
+              // Only truncate when a longer text causes the text to overflow over the MoreIconButton
+              maxWidth={handleClick ? 180 : 'unset'}
```
apedroferreira, line 109: Looks good, but is there no way to make it overflow with the available width without us having to set a manual value?
I feel like it should be possible but haven't tried it.

Also what is the behavior if there is no options button? Does it truncate with an ellipsis after the sidebar width? I guess that's what it should do in every case...

apedroferreira (changes requested):
Added just one comment, I guess at least having an ellipsis even when there's no options button should be doable for consistency?

packages/toolpad-core/src/Account/AccountPreview.tsx
bharatkashyap: The idea to only truncate when there is an options button is to allow the popover to have the complete information, so something like:

https://github.com/user-attachments/assets/9848b2ee-becf-401e-b3a6-fb35a32a923d

We could truncate in all cases, but then do you suggest including a Tooltip when the text is inside the popover; something like: 

https://github.com/user-attachments/assets/37583309-75c9-45b2-9529-28a8f9715f36

apedroferreira: I meant that if you don't pass a `handleClick` and so there's no options button to show a popover, this is what it looks right now:

Screenshot 2025-01-07 at 10 44 02

So maybe all these overflows should ellipsize automatically? Probably not a huge deal though, it was just something I noticed.

bharatkashyap: Ah got it! Yes the ellipsis should happen in this case as well, good catch

apedroferreira (approved):
Nice, thanks for the changes, looks like there was a better solution after all! I wasn't 100% sure.

PR #4571, opened 2024-12-29
[PageHeader] Allow multiple breadcrumbs that are not links

null93:

Simply allows multiple breadcrumbs to exist that are not links. An application of this would be having a top level breadcrumb that you would not want to be a link but instead be static text.

For example, given this definition for breadcrumbs:

```js
[
    { title: "UOW" },
    { title: "Teams", path: "/teams" },
    { title: "New" },
]
```

Before, the last breadcrumb was always static text and all breadcrumbs before it were links.

Before

After, the breadcrumb is a link if it has a truthy value for `path` and it is static text if that value of `path` is falsy.

After


packages/toolpad-core/src/PageContainer/PageHeader.tsx
```diff
@@ -94,7 +94,7 @@ function PageHeader(props: PageHeaderProps) {
                   {getItemTitle(item)}
                 </Link>
               ) : (
-                <Typography key={item.path} color="text.primary">
+                <Typography key={item.title} color="text.primary">
```
Janpot, line 97: Let's use the `map` index here instead. We don't necessarily expect `title` to be uniqye

null93: Great thinking! Done!

Janpot (approved):

packages/toolpad-core/src/PageContainer/PageHeader.tsx
```diff
@@ -83,7 +83,7 @@ function PageHeader(props: PageHeaderProps) {
       <Breadcrumbs aria-label="breadcrumb">
         {resolvedBreadcrumbs
           ? resolvedBreadcrumbs.map((item, index) => {
-              return index < resolvedBreadcrumbs.length - 1 ? (
+              return item.path ? (
                 <Link
                   key={item.path}
```
Janpot, line 88:
```suggestion
                  key={index}
```

Janpot (changes requested):
Sorry, approved too fast. Let's use the same key everywhere. Also, can the `path` property be made optional now in the `BreadCrumb` [type](https://github.com/mui/toolpad/blob/b9962d1ec3a3407c283267da07a5afd41d454cbe/packages/toolpad-core/src/PageContainer/PageContainer.tsx#L18)?

Janpot (approved):
Looks great, thank you!

PR #4573, opened 2024-12-30
[internal] Improve release

oliviertassinari:
Same as https://github.com/mui/base-ui/pull/1245

### DIFF
--- a/packages/toolpad-core/package.json
+++ b/packages/toolpad-core/package.json
@@ -38,7 +38,7 @@
   },
   "homepage": "https://github.com/mui/toolpad#readme",
   "scripts": {
-    "clean": "rimraf build",
+    "clean": "rimraf --glob build \"*.tsbuildinfo\"",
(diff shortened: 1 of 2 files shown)


bharatkashyap (approved):

packages/toolpad-core/package.json
```diff
@@ -38,7 +38,7 @@
   },
   "homepage": "https://github.com/mui/toolpad#readme",
   "scripts": {
-    "clean": "rimraf build",
+    "clean": "rimraf --glob build \"*.tsbuildinfo\"",
```
Janpot, line 41: The two changes in this PR feel like they're at conflict with each other. Either we generate tsbuildinfo files in the build folder and then the old clean command would suffice. Or we generate them outside of the build folder and then we don't remove them during the copy script. But maybe there are reasons to keep this? I know there are differences in build between studio packages and core packages.

Janpot (approved):
Works for me

PR #4574, opened 2024-12-30
[SignInPage] Remove default `RememberMe` 

bharatkashyap:
### Breaking Change

- Remove the default`rememberMe` component

- Closes https://github.com/mui/toolpad/issues/4572

To Do:
- [x] we keep the `rememeberMe`
- [x] default: `null`
- [x] demo for users that are relying on the default ATM
- [ ] mark as breaking change in release notes


docs/data/toolpad/core/components/sign-in-page/SlotPropsSignIn.js
```diff
@@ -16,17 +16,19 @@ export default function SlotPropsSignIn() {
...
           )
         }
+        slots={{ checkbox: RememberCheckbox }}
         slotProps={{
+          form: { noValidate: true },
           emailField: { variant: 'standard', autoFocus: false },
           passwordField: { variant: 'standard' },
           submitButton: { variant: 'outlined' },
-          rememberMe: {
+          checkbox: {
```
apedroferreira, line 25: `checkbox` is more generic than `rememberMe` so I think it's an improvement, but could we make it more generic even as any type of content could go in this slot, such as a small disclaimer for example, or multiple checkboxes?
Not sure if something like `formFooter` would be a good name...
Anyway these are just some thoughts!

docs/data/toolpad/core/components/sign-in-page/SlotPropsSignIn.js
```diff
@@ -16,17 +16,19 @@ export default function SlotPropsSignIn() {
             `Signing in with "${provider.name}" and credentials: ${formData.get('email')}, ${formData.get('password')} and checkbox value: ${formData.get('tandc')}`,
           )
         }
+        slots={{ checkbox: RememberCheckbox }}
```
apedroferreira, line 19: What about `RememberMeCheckbox`, would it be just a bit more descriptive?

docs/data/toolpad/core/components/sign-in-page/sign-in-page.md
```diff
@@ -239,7 +239,7 @@ To enable deep customization beyond what is possible with custom props, the `Sig
 
 {{"demo": "SlotsSignIn.js", "iframe": true, "height": 540 }}
 
-You can use the `slotProps` prop to pass props to the underlying components of each slot:
+You can use the `slotProps` prop to pass props to the underlying components of each slot, and also to `form` element:
```
apedroferreira, line 242:
```suggestion
You can use the `slotProps` prop to pass props to the underlying components of each slot, and also to the `form` element:
```

docs/translations/api-docs/sign-in-page/sign-in-page.json
```diff
@@ -18,10 +18,10 @@
   },
   "classDescriptions": {},
   "slotDescriptions": {
+    "checkbox": "A custom checkbox placed in the credentials form",
```
apedroferreira, line 21: If we change the name as I proposed we'd have to update the description to something more generic too.

packages/toolpad-core/src/SignInPage/RememberCheckbox.tsx
```diff
@@ -0,0 +1,57 @@
...
+        />
+      }
+      label="Remember me"
+      {...props}
+      slotProps={{
+        typography: {
+          color: 'textSecondary',
+          fontSize: theme.typography.pxToRem(14),
+        },
+        ...props?.slotProps?.typography,
```
apedroferreira, line 40: Shouldn't we just spread this inside `typography` instead?

packages/toolpad-core/src/SignInPage/RememberCheckbox.tsx
```diff
@@ -0,0 +1,57 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import { FormControlLabel, Checkbox } from '@mui/material';
+import { useTheme } from '@mui/material/styles';
+
+interface RememberCheckboxProps {
```
apedroferreira, line 6: Can't we just allow all `slotProps` from `FormControlLabel` and make sure they're passed to the component?

docs/data/toolpad/core/components/sign-in-page/SlotPropsSignIn.js
apedroferreira: Also I see we have a `forgotPasswordLink` slot too, so if when implementing these you saw that there's a benefit to these slots being more specific then it should be fine.

apedroferreira (commented):
Took a look and left some comments!
Overall the idea of the "remember me" slot being more generic and optional with its own component seems good to me.

docs/data/toolpad/core/components/sign-in-page/SlotPropsSignIn.js
apedroferreira: Moving those to "plug-in" components seems good to me if they're complex enough!

packages/toolpad-core/src/SignInPage/SignInPage.tsx
```diff
@@ -596,49 +604,30 @@ function SignInPage(props: SignInPageProps) {
                           type: 'password',
                           label: 'Password',
                           id: 'password',
-                          placeholder: '*****',
+                          placeholder: '***** asdasd',
```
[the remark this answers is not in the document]
bharatkashyap: 😅 Thanks for the catch!

apedroferreira (approved):

PR #4576, opened 2025-01-04
[ci] Fix CI failing due to Recharts console error

bharatkashyap:
- "alpha" causes a `setState` call related warning to show up in the console and fail the test

### DIFF
--- a/test/integration/codeComponents/index.spec.ts
+++ b/test/integration/codeComponents/index.spec.ts
@@ -15,6 +15,13 @@ test.use({
   localAppConfig: {
     cmd: 'dev',
   },
+  ignoreConsoleErrors: [
+    // React setState warning inside "Recharts"
+    [/Cannot update a component \(`.*`\) while rendering a different component \(`.*`\)/],
+    {
+      scope: 'test',
+    },
+  ],
 });
 
 test('custom components can use external libraries', async ({ page }) => {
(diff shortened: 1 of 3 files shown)


test/integration/codeComponents/index.spec.ts
```diff
@@ -15,6 +15,13 @@ test.use({
   localAppConfig: {
     cmd: 'dev',
   },
+  ignoreConsoleErrors: [
+    // React setState warning inside "Recharts"
```
Janpot, line 19: Is there a bug in recharts? Or are we holding it wrong on toolpad side?

bharatkashyap: I couldn't find a Recharts bug report, but also couldn't find anything we're doing with Recharts in the test that's failing

Janpot: Is it reproducible in isolation?

bharatkashyap: Don't see any warnings in the console: https://codesandbox.io/p/sandbox/mutable-snow-x2kx55?file=%2Fsrc%2FApp.tsx

Janpot: As per DM, we should set `recharts` version from `alpha` to `^2.15.0`.

Janpot (approved):

PR #4597, opened 2025-01-12
[cli] Refactor setup

### LINKED ISSUE #4596: [cli] Fails to install example when passing a directory
- When passing in a non-default option in the path argument or option and passing in an example through `--example`, the CLI fails to create the required folder
- Introduced in https://github.com/mui/toolpad/pull/4434

bharatkashyap:
- Separate the CLI `index` code into separate files
- Fix #4596
- Adds better post-install instructions when installing examples using `--example`


packages/create-toolpad-app/src/core.ts
```diff
@@ -0,0 +1,58 @@
...
+  // eslint-disable-next-line no-console
+  console.log();
+  // eslint-disable-next-line no-console
+  console.log(
+    `${chalk.cyan('info')} - Creating Toolpad Core project in ${chalk.cyan(options.absolutePath)}`,
+  );
+  // eslint-disable-next-line no-console
+  console.log();
+
+  const pkg = await findCtaPackageJson();
```
Janpot, line 18: Better to remove this call and pass the version as an option.

packages/create-toolpad-app/src/package.ts
```diff
@@ -0,0 +1,27 @@
...
+      return 'pnpm';
+    }
+    if (userAgent.startsWith('npm')) {
+      return 'npm';
+    }
+  }
+  return 'pnpm';
+}
+
+export async function findCtaPackageJson() {
```
Janpot, line 22: There exist package json types in this workspace already. Should add a return type here.

Janpot (approved):
Few small suggestions. Otherwise seems to be good. 👍

packages/create-toolpad-app/src/core.ts
bharatkashyap: If we remove the call, how would we set the default value of the version in case nothing is passed in the version option?

Janpot: If there is no sensible default, then you make the option mandatory

PR #4621, opened 2025-01-20
[SIgnInPage] Add default `title`

bharatkashyap:
- Fixes https://github.com/mui/toolpad/issues/4618
- Fixes https://github.com/mui/toolpad/issues/4602
- Placing the `title` inside `SignInPage` should be okay with React 19: https://vercel.com/blog/whats-new-in-react-19#document-metadata-support


Janpot:
> should be okay with React 19

Don't we also support React 18 still?

bharatkashyap:
> > should be okay with React 19
> 
> Don't we also support React 18 still?

An option could be to set in a ` useEffect` with `document.title = "Sign In"

PR #4626, opened 2025-01-22
[SignInPage] Add Non-Null Assertion to Prevent Compiler Errors

JimZhouZZY:
Sometimes, the TypeScript compiler raises an error about ‘slotProps’ possibly being undefined, even though it is guaranteed to be defined under the given condition.

```
Failed to compile.

TS18048: 'slotProps' is possibly 'undefined'.
    34 | const mergeSlotSx = (defaultSx: SxProps<Theme>, slotProps?: { sx?: SxProps<Theme> }) => {
    35 |   if (Array.isArray(slotProps?.sx)) {
  > 36 |     return [defaultSx, ...slotProps.sx];
       |                           ^^^^^^^^^
    37 |   }
    38 |
    39 |   if (slotProps?.sx) {

```

To resolve this, I added a non-null assertion to ensure the compiler no longer raises this unnecessary error.

### DIFF
--- a/packages/toolpad-core/src/SignInPage/SignInPage.tsx
+++ b/packages/toolpad-core/src/SignInPage/SignInPage.tsx
@@ -40,7 +40,7 @@ import { BrandingContext, RouterContext } from '../shared/context';
 
 const mergeSlotSx = (defaultSx: SxProps<Theme>, slotProps?: { sx?: SxProps<Theme> }) => {
   if (Array.isArray(slotProps?.sx)) {
-    return [defaultSx, ...slotProps.sx];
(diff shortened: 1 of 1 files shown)


Janpot:
I can't reproduce this error, can you provide a reproduction? In any case we don't use the non-null assertion operator, there are better ways.

JimZhouZZY:
Thanks for the response!

After further investigation, I realized that the issue was due to differences in the build tools. The original project uses Babel, but I used the TypeScript compiler directly in my project after overriding `SignInPage.tsx`. I’ll close this PR now. Thanks again!

PR #4642, opened 2025-01-29
[docs] Fix redirection to GitHub example view

oliviertassinari:
#4350 was a bit too rushed. This creates a ton of 301 in the ahrefs crawls: https://app.ahrefs.com/site-audit/3833384/data-explorer?columns=pageRating%2Curl%2Ctraffic%2ChttpCode%2CcontentType%2Cdepth%2Credirect%2CincomingAllLinks&current=29-01-2025T022904&filterId=b3b75b6257a370fcf9f1f73befb5deb5&issueId=c64d8847-d0f4-11e7-8ed1-001e67ed4656&sorting=-pageRating

### DIFF
--- a/docs/src/modules/components/examples/ExamplesFeatured.tsx
+++ b/docs/src/modules/components/examples/ExamplesFeatured.tsx
@@ -51,7 +51,7 @@ export default function ExamplesFeatured(props: FeaturedExamplesProps) {
               }}
             >
               <Link
-                href={versionGitHubLink(example.href)}
+                href={example.href}
                 target="_blank"
                 sx={{
                   position: 'relative',
@@ -162,7 +162,7 @@ export default function ExamplesFeatured(props: FeaturedExamplesProps) {
                 </Box>
                 <Button
                   component="a"
-                  href={versionGitHubLink(example.href)}
+                  href={example.href}
                   size="small"
                   variant="outlined"
                   color="primary"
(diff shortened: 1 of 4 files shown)


docs/src/modules/components/examples/coreExamples.ts
```diff
@@ -109,7 +108,7 @@ const coreExamples = [
     src: '/static/toolpad/docs/core/tutorial-1.png',
     srcDark: '/static/toolpad/docs/core/tutorial-1.png', // TODO Fix
     href: 'https://mui.com/toolpad/core/introduction/tutorial/',
-    source: 'https://github.com/mui/toolpad/tree/master/examples/core/tutorial/',
+    source: 'https://github.com/mui/toolpad/tree/master/examples/core/tutorial',
```
oliviertassinari, line 111: GitHub supports both, they have no care for forcing one over the other, it's so cheap, lol, not even a canonical link. 
In any case, here, we can use what their links default to.

x same as for the other diffs.

docs/src/modules/components/examples/ExamplesFeatured.tsx
```diff
@@ -162,7 +162,7 @@ export default function ExamplesFeatured(props: FeaturedExamplesProps) {
                 </Box>
                 <Button
                   component="a"
-                  href={versionGitHubLink(example.href)}
+                  href={example.href}
```
oliviertassinari, line 165: `href` are live demos, no need to handle them. It's only relevant for `source`.

bharatkashyap (approved):

PR #4657, opened 2025-02-04
Fix Next.js Pages Router pathname in AppProvider

apedroferreira:
Use `pathname` in Next.js Pages Router adapter.

Closes https://github.com/mui/toolpad/issues/4645#issuecomment-2631343237

### DIFF
--- a/packages/toolpad-core/src/nextjs/NextAppProviderPages.tsx
+++ b/packages/toolpad-core/src/nextjs/NextAppProviderPages.tsx
@@ -8,7 +8,7 @@ import type { AppProviderProps, Navigate, Router } from '../AppProvider';
  * @ignore - internal component.
  */
 export function NextAppProviderPages(props: AppProviderProps) {
-  const { push, replace, asPath, query } = useRouter();
+  const { push, replace, pathname, query } = useRouter();
 
   const search = React.useMemo(() => {
     const params = new URLSearchParams();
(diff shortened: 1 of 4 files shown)


packages/toolpad-core/src/nextjs/NextAppProviderPages.tsx
```diff
@@ -8,7 +8,7 @@ import type { AppProviderProps, Navigate, Router } from '../AppProvider';
  * @ignore - internal component.
  */
 export function NextAppProviderPages(props: AppProviderProps) {
-  const { push, replace, asPath, query } = useRouter();
+  const { push, replace, pathname, query } = useRouter();
```
apedroferreira, line 11: @Janpot is this ok or is there a reason why you were using `asPath`?

Janpot: > is there a reason why you were using `asPath`?

only short-sightedness 🙂

Janpot (approved):

packages/toolpad-core/src/nextjs/NextAppProviderPages.tsx
marcomele: I believe the reason for using `asPath` is that `pathname` has a different form when running a nextjs app with `skipMiddlewareUrlNormalize: true` option for dynamic routes — `pathname` would look like `/route/[param]` for route `/route/[param].ts`. I believe the older solution of `asPath.split('?')[0]` worked better.

apedroferreira: Hi, yes, that's actually something I reverted recently in the CRUD component PR as I noticed some issues with using `pathname`. So in the next release it will be `asPath.split('?')[0]` (we didn't have the split with "?" before)

PR #4661, opened 2025-02-05
[core] Use router specific Link components

bharatkashyap:
- Fixes https://github.com/mui/toolpad/issues/4556
- Fixes https://github.com/mui/toolpad/issues/4556

### DIFF
--- a/packages/toolpad-core/src/AppProvider/AppProvider.tsx
+++ b/packages/toolpad-core/src/AppProvider/AppProvider.tsx
@@ -2,6 +2,8 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
 import { createTheme as createMuiTheme, Theme } from '@mui/material/styles';
+import { LinkProps as ReactRouterLinkProps } from 'react-router';
+import { LinkProps as NextLinkProps } from 'next/link';
 import { NotificationsProvider } from '../useNotifications';
 import { DialogsProvider } from '../useDialogs';
 import {
@@ -27,6 +29,7 @@ export interface Router {
   pathname: string;
   searchParams: URLSearchParams;
   navigate: Navigate;
+  Link?: React.JSXElementConstructor<NextLinkProps & ReactRouterLinkProps>;
 }
 
 export interface Branding {
--- a/packages/toolpad-core/src/nextjs/NextAppProviderPages.tsx
+++ b/packages/toolpad-core/src/nextjs/NextAppProviderPages.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import Link from 'next/link';
 import { asArray } from '@toolpad/utils/collections';
 import { useRouter } from 'next/router';
 import { AppProvider } from '../AppProvider';
@@ -41,6 +42,7 @@ export function NextAppProviderPages(props: AppProviderProps) {
       pathname,
       searchParams,
       navigate,
+      Link,
     }),
     [navigate, pathname, searchParams],
   );
--- a/packages/toolpad-core/src/react-router/ReactRouterAppProvider.tsx
+++ b/packages/toolpad-core/src/react-router/ReactRouterAppProvider.tsx
@@ -1,6 +1,6 @@
 'use client';
 import * as React from 'react';
-import { useSearchParams, useLocation, useNavigate } from 'react-router';
+import { useSearchParams, useLocation, useNavigate, Link } from 'react-router';
 import { AppProvider, type AppProviderProps, Navigate, Router } from '../AppProvider/AppProvider';
 
 function ReactRouterAppProvider(props: AppProviderProps) {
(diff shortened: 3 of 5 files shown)


packages/toolpad-core/src/AppProvider/AppProvider.tsx
```diff
@@ -2,6 +2,8 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
 import { createTheme as createMuiTheme, Theme } from '@mui/material/styles';
+import { LinkProps as ReactRouterLinkProps } from 'react-router';
+import { LinkProps as NextLinkProps } from 'next/link';
```
Janpot, line 6: This seems incorrect because:
* the package should never import from third party runtimes
* I don't think the implementations are 100% compatible

We need to define an abstract LinkProps type ourselves, and then make sure in the routeradapter to map the concrete implementation to the abstract properties.

packages/toolpad-core/src/shared/Link.tsx
```diff
@@ -5,15 +5,17 @@ import { RouterContext } from './context';
  * @ignore - internal component.
  */
 
-export interface LinkProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
+export interface LinkProps extends Omit<React.AnchorHTMLAttributes<HTMLAnchorElement>, 'href'> {
```
Janpot, line 8: I believe this should be a small interface. With only the properties we need for usage inside of Toolpad core.

packages/toolpad-core/src/shared/Link.tsx
```diff
@@ -5,15 +5,17 @@ import { RouterContext } from './context';
...
+  href: string;
+  to: string;
 }
 
 export const Link = React.forwardRef(function Link(
   props: LinkProps,
   ref: React.ForwardedRef<HTMLAnchorElement>,
 ) {
-  const { children, href, onClick, history, ...rest } = props;
+  const { children, href, onClick, history, to, ...rest } = props;
```
Janpot, line 18: We're just passing `href` internally. Remap correctly to `to` in the `react-router` adapter,

packages/toolpad-core/src/shared/Link.tsx
```diff
@@ -28,7 +29,11 @@ export const Link = React.forwardRef(function Link(
     };
   }, [routerContext, onClick, history]);
 
-  return (
+  return routerContext?.Link && href ? (
```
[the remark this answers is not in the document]
Janpot: also, the Link component that is exported from this module should have the same signature as the on in the `routerContext`. Ideally the current implementation is renamed to `DefaultLink` and we export a new version that wraps it:

```tsx
export const Link = React.forwardRef(function Link(
  props: LinkProps,
  ref: React.ForwardedRef<HTMLAnchorElement>,
) {
  const routerContext = React.useContext(RouterContext);
  const LinkComponent = routerContext?.Link ?? DefaultLink
  return <LinkComponent ref={ref} {...props} />
})
```

this to avoid unnecessary callback creation and other render logic running in the default link.

packages/toolpad-core/src/shared/Link.tsx
```diff
@@ -5,12 +5,13 @@ import { RouterContext } from './context';
  * @ignore - internal component.
  */
 
-export interface LinkProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
+export interface DefaultLinkProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
```
Janpot, line 8: Can't this just be the `LinkProps` that we use both for `DefaultLink` and `Link`?
```suggestion
export interface LinkProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
```

packages/toolpad-core/src/react-router/ReactRouterAppProvider.tsx
```diff
@@ -1,7 +1,13 @@
 'use client';
 import * as React from 'react';
-import { useSearchParams, useLocation, useNavigate } from 'react-router';
+import { useSearchParams, useLocation, useNavigate, Link as ReactRouterLink } from 'react-router';
 import { AppProvider, type AppProviderProps, Navigate, Router } from '../AppProvider/AppProvider';
+import { LinkProps } from '../shared/Link';
+
+const Link = React.forwardRef<HTMLAnchorElement, LinkProps>((props, ref) => {
+  const { href, ...rest } = props;
```
Janpot, line 8: I believe this should also handle the `history` prop?

packages/toolpad-core/src/nextjs/NextAppProviderPages.tsx
```diff
@@ -41,6 +42,7 @@ export function NextAppProviderPages(props: AppProviderProps) {
       pathname,
       searchParams,
       navigate,
+      Link,
```
Janpot, line 45: I believe this needs to handle the `history` prop?

Janpot (approved):
Looks great now. Maybe some tests?

bharatkashyap:
> Looks great now. Maybe some tests?

I think we can add effective tests for this only if we set up and run an actual app (avoid mocking), so I think we can take in a follow-up to this? I'll merge this and create an item for myself to do that

PR #4685, opened 2025-02-18
Add v7 alpha peer deps

Janpot:
Add peer dependency of the alpha packages


Janpot:
part of https://github.com/mui/toolpad/pull/4686

PR #4713, opened 2025-02-28
[CRUD] Add examples

apedroferreira:
New examples with mock employees CRUD for Vite, Next.js App Router and Next.js Pages Router.


examples/core/auth-nextjs-pages/src/data/orders.ts
```diff
@@ -0,0 +1,200 @@
+'use client';
```
Janpot, line 1: Would it make sense to keep this example about authentication and next.js, and create a new example for data sources?

apedroferreira: It would if we're going for many small examples instead of just a few with multiple things.
As we already have quite a few examples this seems to be the approach we're following, so I guess 3 new examples would be okay, I'll change it.

Janpot (approved):
Seems fine 👍

apedroferreira:
> Seems fine 👍

Thanks, just missing some images for examples I think.

Janpot:
> Thanks, just missing some images for examples I think.

Yep, sry, I had just read over the "Files changed" tab, didn't notice CI failed

apedroferreira:
> > Thanks, just missing some images for examples I think.
> 
> Yep, sry, I had just read over the "Files changed" tab, didn't notice CI failed

Should be ready now.

docs/src/modules/components/examples/coreExamples.ts
```diff
@@ -101,6 +101,33 @@ const coreExamples = [
     codeSandbox: true,
     stackBlitz: true,
   },
+  {
```
bharatkashyap, line 104: I think we should show them at the top of the list, for which you'd need to add them at the beginning just after the functional dashboard (since the component reverses this list when displaying the cards)

apedroferreira: Ok, just adjusted the order!

PR #4745, opened 2025-03-10
[code-infra] Revert ESM update for toolpad-core

Janpot:
Wait for both X and core to be in beta with their updated package layout before making this change to `@toolpad/core`. We can keep it for `@toolpad/utils`.


apedroferreira (approved):
Nice! I had actually ran into this too but wasn't sure if this was the fix.

apedroferreira:
Oh wait, I guess we need to update the versions for development, the playground apps and the examples accordingly?
But I can take care of that.

Janpot:
> Oh wait, I guess we need to update the versions

Yes, and I haven't tested what this does for Studio

apedroferreira:
> > Oh wait, I guess we need to update the versions
> 
> Yes, and I haven't tested what this does for Studio

Hopefully nothing too breaking, I'll try in a separate PR.

Janpot:
@apedroferreira I'm reverting this until X publishes their beta. Core v7 is not compatible with X v7.

Janpot:
> I guess we will need some kind of alternative there, didn't look to deep into it yet.

I'll fix this in core

PR #4746, opened 2025-03-10
Update MUI package versions to v7 in Toolpad Core

apedroferreira:
On top of https://github.com/mui/toolpad/pull/4745

These updates should be needed? I can't commit directly to https://github.com/mui/toolpad/pull/4745


apedroferreira:
A few issues I've found due to these recent changes but couldn't find the best fix for:
- `dev` command fails when it uses `build-types --watch` as the updated `builtTypes` script does not seem to be able to take those flags anymore
- `toolpad-studio-components` seems to have issues importing from `@toolpad/utils`
- I was trying to debug Toolpad Studio to make sure nothing broke there, but it throws errors when importing icons, might also be related to these changes

Janpot:
I'm reverting https://github.com/mui/toolpad/pull/4745 until X pushes their beta, so that we can release

PR #4762, opened 2025-03-14
fix: render divider as li in menu

rkristelijn:
- [x] I've read and followed the [contributing guide](https://github.com/mui/toolpad/blob/master/CONTRIBUTING.md#sending-a-pull-request) on how to create great pull requests.
- [x] I've updated the relevant documentation for any new or updated feature.
- [x] I've linked relevant GitHub issue with "Closes #4759".
- [x] I've added a visual demonstration in the form of a screenshot or video.

PR #4763, opened 2025-03-14
fix: render divider and header as li in menu

rkristelijn:
- [x] I've linked relevant GitHub issue with "Closes #4762 ".

Closes https://github.com/mui/toolpad/issues/4759


packages/toolpad-core/src/DashboardLayout/DashboardSidebarSubNavigation.tsx
```diff
@@ -141,6 +141,7 @@ function DashboardSidebarSubNavigation({
 
           return (
             <Divider
+              component="li"
```
apedroferreira, line 144: Can we just move this to after the `key` prop?

apedroferreira (changes requested):
Hi, you're right, thanks a lot for finding these issues and working on the fixes!
All good, just a small change in the order of props to be consistent.

packages/toolpad-core/src/DashboardLayout/DashboardSidebarSubNavigation.tsx
rkristelijn: sure, _leave the campground cleaner than you found it_ right? ;) 

Wouldn't a linter help with proper sorting e.g. https://github.com/rkristelijn/eslint-plugin-mui

apedroferreira: It would help, I guess we don't want to strictly enforce that but at least I like the prop order to be as consistent as possible :)

apedroferreira (approved):

rkristelijn:
I've ran `npx pnpm prettier:all` 

```
❯ npm ls pnpm
toolpad-project@0.0.1 /path/to/toolpad
└── (empty)

apedroferreira:
> I would recommend to either:
> 
> 1. add to README.md to use `corepack enable && corepack prepare pnpm --activate`
> 2. add in `.npmrc`: `packageManager=pnpm@8.15.4` or other version
> 3. (my preference) add the right version of `pnpm` as a dev dependency that can be used via `npx` (first uses local installs, then globals)

Yes, looks like we could improve the documentation about suggestion 1 at least. Feel free to even submit a PR with that or I can do it soon!

Other than that we do have `"packageManager": "pnpm@9.12.3"` in the root `package.json`, and it doesn't seem like we're following number 3 organization-wide.

PR #4773, opened 2025-03-17
[core] [i18n] Add interpolation to translation keys

bharatkashyap:
- Fixes https://github.com/mui/toolpad/issues/4727
- Allows interpolation for localization keys 
- Adds a `providerSignInTitle` of a `(provider: string) => string` type
- Sets `signInTitle` to a `string | ((brandingTitle?: string) => string)` type

**Breaking Change**

- Removes `oauthSignInTitle`, `passkeySignInTitle` and `magicLinkSignInTitle`


packages/toolpad-core/src/AppProvider/LocalizationProvider.tsx
```diff
@@ -105,7 +103,47 @@ LocalizationProvider.propTypes /* remove-proptypes */ = {
   /**
    * Locale for components texts
    */
-  localeText: PropTypes.object,
```
Janpot, line 108: I believe this was deliberate

Janpot (approved):
I believe the proptype of `localeText` was deliberate

packages/toolpad-core/src/AppProvider/LocalizationProvider.tsx
bharatkashyap: Yes, but I felt that having the explicit props is helpful at least on the `LocalizationProvider` docs, it is typed as `object` for every other component. We can revert it if you think it doesn't make sense

Janpot: > We can revert it if you think it doesn't make sense

It makes sense, but I think it makes more sense to be consistent with X

PR #4786, opened 2025-03-19
[CRUD] Expose subcomponent slots to top-level CRUD component

apedroferreira:
So we can use a single `Crud` component in https://github.com/mui/toolpad/pull/4785


bharatkashyap (approved):
Makes sense, but to be more future-looking, does it make more sense to shape the `Crud` slotss as `list: ListSlots, create: CreateSlots ...` and so on, given that we might need to do this in the future? But not to block, this is okay for me in its current form

PR #4788, opened 2025-03-19
feat: lock pnpm version via npx

### LINKED ISSUE #4787: request: use pnpm in dev dependency
### Summary

see discussion in https://github.com/mui/toolpad/pull/4763#issuecomment-2737417404

pnpm is not a dev dependency so you need to install it globally. This way the version is not locked and you might have weird situations and issues logged based on this.

recomandation: add pnpm to dev dependency and use npx to run. npx first uses the repo version (that is now installed, and version locked)

_No response_

### Examples

_No response_

### Motivation

_No response_

**Search keywords**: pnpm locked version

rkristelijn:
- [x] I've linked relevant GitHub issue with Closes #4787

```bash
# get the version from package.json
❯ npm pkg get devDependencies.pnpm
"9.12.3"
# install a different version globally
❯ npm i pnpm@8 -g
# print the version of pnpm from within the repo root
❯ pnpm --version
# observe it can be different and therefore can cause issues
8.15.9
# whereas if you use npm pnpm, it always picks from the dev dependencies
❯ npx pnpm --version
9.12.3
```


Janpot (commented):
You can run pnpm through npx if you want but we won't recommend it. I think this PR can be closed.

rkristelijn:
Can you elaborate a bit? I think my argument is valid, would like to understand your viewpoint

Janpot:
`pnpm` is our package manager of choice, if you want to contribute we recommend installing it globally. If you don't want to do that, you can use something like [`corepack`](https://nodejs.org/api/corepack.html). Both of these methods will respect the [`packageManager`](https://nodejs.org/api/packages.html#packagemanager) field in our `package.json` and automatically use the configured version of `pnpm`. All maintainers of the mui ecosystem have a global `pnpm`, we don't need `npx pnpm`.

rkristelijn:
@Janpot Thanks for the clarification, I would like to add this in the CONTRIBUTING.md, is that ok?

PR #4805, opened 2025-03-26
[docs] Attempt to solve blank space issue in Material UI Drawer docs

apedroferreira:


Janpot (approved):

Janpot:
Interesting that no eslint rule exists to catch that unused ref

apedroferreira:
> Interesting that no eslint rule exists to catch that unused ref

Well it was being used for the container, but once I removed it there was no eslint warning even though the ref wasn't doing anything anymore, I guess there might be a React rule for that.

PR #4809, opened 2025-03-27
[DashboardLayout] Allow custom navigation page items

apedroferreira:
Closes https://github.com/mui/toolpad/issues/4413.
Closes https://github.com/mui/toolpad/issues/4607 both for the hidden and disabled cases.

Also helps with https://github.com/mui/toolpad/issues/4733 but additional state management would be needed from the user, so maybe there is a better solution for that in the future.

Also makes it possible to easily mark items as `expanded`, `selected` or `disabled`, or use external links with a custom `href`.

Check the new documentation here:
- https://deploy-preview-4809--mui-toolpad-docs.netlify.app/toolpad/core/react-dashboard-layout/#custom-page-items
- https://deploy-preview-4809--mui-toolpad-docs.netlify.app/toolpad/core/api/dashboard-sidebar-page-item/


davorian:
I'm awaiting this!!!

docs/pages/toolpad/core/api/dashboard-sidebar-page-item.json
```diff
@@ -0,0 +1,31 @@
+{
+  "props": {
```
Janpot, line 2: 🤔  I would have expect this to just have a `item` prop. Does it need all these other props? This is just to render the default item right? I'm not sure we should expose more API on it.

apedroferreira: If you look at the expanded demo [here](https://deploy-preview-4809--mui-toolpad-docs.netlify.app/toolpad/core/react-dashboard-layout/#custom-page-items) I was just trying to make this type of changes as easy as possible as well as reusing the existing component.

Would it help if these props were additional options maybe? And the first parameter was the item?

apedroferreira:
I've changed the `renderPageItem` typing to `renderPageItem?: (item: NavigationPageItem, params: { mini: boolean }) => React.ReactNode;` and updated the docs accordingly. The `mini` parameter still seems useful to render things differently in the minified sidebar.

It's using context under the hood to pass default values to the page item component without exposing all that API to the users, who can only set a few simple optional props now if they wish to override certain behaviors.

Examples in https://deploy-preview-4809--mui-toolpad-docs.netlify.app/toolpad/core/react-dashboard-layout/#custom-page-items

Janpot:
👍 

> The `mini` parameter still seems useful to render things differently in the minified sidebar.

I suppose the alternative would be `renderMiniPageItem`?

Janpot:
> I'll change it.

for the record, that's not the propose of my comment. I'm fine with what there is

apedroferreira:
> > I'll change it.
> 
> for the record, that's not the propose of my comment. I'm fine with what there is

Ok, I didn't have a huge preference between the 2 solutions but I guess that as a user I would prefer a single function prop, not to have to use 2 separate ones. So if it's ok I guess we can go with this.

docs/data/toolpad/core/components/dashboard-layout/DashboardLayoutCustomPageItems.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+export default function DashboardLayoutCustomPageItems(props: DemoProps) {
+  const { window } = props;
+
+  const router = useDemoRouter('/dashboard');
+
+  // Remove this const when copying and pasting into your project.
+  const demoWindow = window !== undefined ? window() : undefined;
+
+  const renderPageItem = React.useCallback(
```
bharatkashyap, line 174: I think it makes sense to add this renderPageItem definition to the demo preview as well

```suggestion
// preview-start
  const renderPageItem = React.useCallback(
```

docs/data/toolpad/core/components/dashboard-layout/DashboardLayoutCustomPageItems.tsx
```diff
@@ -0,0 +1,214 @@
...
+      }
+      if (item.title === 'Expanded Folder') {
+        return <DashboardSidebarPageItem item={item} expanded />;
+      }
+      if (item.title === 'Custom Item') {
+        return <CustomPageItem item={item} mini={mini} />;
+      }
+
+      return <DashboardSidebarPageItem item={item} />;
+    },
```
bharatkashyap, line 198:
```suggestion
// preview-end
    },
```

bharatkashyap (approved):
Big improvement! Looks good in the demo/docs 👍🏻

PR #4843, opened 2025-04-14
[docs] [examples] Don't use `checkboxSelection` in CRUD examples

bharatkashyap:
- Fix https://github.com/mui/toolpad/issues/4836
- Remove `checkboxSelection` as the prop we use in the demo since it doesn't do much
- Also remove it from the themed example


apedroferreira (approved):
Thanks for this!

Maybe a more useful column to pin would be the actions column (edit & delete), i would love to always have it pinned but it's a paid feature... (and probably feel free to give it the header name / name `actions` or something if that would be needed for that). But I can also change this myself later!

PR #4844, opened 2025-04-14
[DashboardLayout] Highlight and text alignment for mini

bharatkashyap:
- Fix https://github.com/mui/toolpad/issues/4835

- use the `MINI_DRAWER_WIDTH`  for the `ul` and passing `scrollbar-gutter: stable` to the `nav` when `isMini` is set, everything being the same in all other modes.

| **Before** | **After** |
|---|---|
|  | <video width="500" alt="Screenshot 2025-04-14 at 3 10 11 PM" src="https://github.com/user-attachments/assets/8d4dc7e7-9cc7-47df-a076-93634c96906a" /> |

### DIFF
--- a/packages/toolpad-core/src/DashboardLayout/DashboardSidebarSubNavigation.tsx
+++ b/packages/toolpad-core/src/DashboardLayout/DashboardSidebarSubNavigation.tsx
@@ -253,7 +253,7 @@ function DashboardSidebarSubNavigation({
                   sx={{
                     position: 'relative',
                     top: isMini ? -6 : 0,
-                    left: isMini ? 5 : 0,
+                    left: 0,
                   }}
                 >
                   <ListItemIcon
@@ -287,9 +287,9 @@ function DashboardSidebarSubNavigation({
                       sx={{
                         position: 'absolute',
                         bottom: -18,
-                        left: '50%',
+                        left: '45%',
                         transform: 'translateX(-50%)',
-                        fontSize: 10,
+                        fontSize: 9,
                         fontWeight: 500,
                         textAlign: 'center',
                         whiteSpace: 'nowrap',
--- a/packages/toolpad-core/src/DashboardLayout/DashboardSidebarSubNavigation.tsx
+++ b/packages/toolpad-core/src/DashboardLayout/DashboardSidebarSubNavigation.tsx
@@ -127,6 +127,7 @@ function DashboardSidebarSubNavigation({
         mb: depth === 0 && !isPopover ? 4 : 0.5,
         pl: (isPopover ? 1 : 2) * (isPopover ? depth - 1 : depth),
         minWidth: isPopover && depth === 1 ? 240 : 'auto',
+        width: isMini ? MINI_DRAWER_WIDTH : 'auto',
       }}
     >
       {subNavigation.map((navigationItem, navigationItemIndex) => {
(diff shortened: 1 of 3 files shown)


packages/toolpad-core/src/DashboardLayout/DashboardSidebarSubNavigation.tsx
```diff
@@ -287,9 +287,9 @@ function DashboardSidebarSubNavigation({
                       sx={{
                         position: 'absolute',
                         bottom: -18,
-                        left: '50%',
+                        left: '45%',
```
apedroferreira, line 290: I wouldn't change this, this is what is centering the text and it goes together with `transform: 'translateX(-50%)'` below.
And also the font size shouldn't be what is causing the issue and it shouldn't be changed unless necessary.

Can we not just set a minimum width somewhere so that the scrollbar doesn't make things narrower?
The current changes are only trying to disguise the problem but the content is not centered.
I can take a deeper look if it helps.

apedroferreira (changes requested):

apedroferreira (approved):
This seems a lot better, thanks a lot for the changes!

apedroferreira:
Looks like there's a horizontal scrollbar when the sidebar is in mini mode now: https://deploy-preview-4809--mui-toolpad-docs.netlify.app/toolpad/core/react-dashboard-layout/#custom-page-items

Maybe it'll be fixed if we remove a few pixels from the width?

bharatkashyap:
> Looks like there's a horizontal scrollbar when the sidebar is in mini mode now: https://deploy-preview-4809--mui-toolpad-docs.netlify.app/toolpad/core/react-dashboard-layout
> 
> Maybe it'll be fixed if we remove a few pixels from the width?

We could do that or hide scrollbars on the container nav of the list, which approach do you think makes more sense? Either make sense to me, although the `overflow-x: hidden` on the container nav seems to be more robust since we never want horizontal scrollbars to appear in the sidebar

apedroferreira:
> > Looks like there's a horizontal scrollbar when the sidebar is in mini mode now: https://deploy-preview-4809--mui-toolpad-docs.netlify.app/toolpad/core/react-dashboard-layout
> > Maybe it'll be fixed if we remove a few pixels from the width?
> 
> We could do that or hide scrollbars on the container nav of the list, which approach do you think makes more sense? Either make sense to me, although the `overflow-x: hidden` on the container nav seems to be more robust since we never want horizontal scrollbars to appear in the sidebar

I would try to figure out why there is an overflow now - if it's related with having set a width in the items, I would try to fix something related to that. I don't mind taking a look soon.

bharatkashyap:
> > > Looks like there's a horizontal scrollbar when the sidebar is in mini mode now: https://deploy-preview-4809--mui-toolpad-docs.netlify.app/toolpad/core/react-dashboard-layout
> > > Maybe it'll be fixed if we remove a few pixels from the width?
> > 
> > 
> > We could do that or hide scrollbars on the container nav of the list, which approach do you think makes more sense? Either make sense to me, although the `overflow-x: hidden` on the container nav seems to be more robust since we never want horizontal scrollbars to appear in the sidebar
> 
> I would try to figure out why there is an overflow now - if it's related with having set a width in the items, I would try to fix something related to that. I don't mind taking a look soon.

The scrollbar is due to providing an explicit width to the list now - reducing the width makes the scrollbar go away but I haven't tested this fix rigorously over all device sizes. I can open a PR with that fix and you can take it forward/review it

PR #4854, opened 2025-04-16
Remove MUI X libraries as peer dependencies

apedroferreira:
Remove MUI X libraries as peer dependencies, as it's not necessary for them to be.

Addresses https://github.com/orgs/mui/projects/9/views/1?pane=issue&itemId=106384645

Closes https://github.com/mui/toolpad/issues/4840

### DIFF
--- a/packages/toolpad-core/package.json
+++ b/packages/toolpad-core/package.json
@@ -42,6 +42,8 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.27.0",
+    "@mui/x-data-grid": "^8.0.0 || ^8.0.0-beta",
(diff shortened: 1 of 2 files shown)


Janpot (approved):

bharatkashyap:
Could this cause issues with people who have v7 X Packages installed in their project separately? Should we keep v7 as our dependency version till v8 is stable?

Janpot:
> Could this cause issues with people who have v7 X Packages installed in their project separately?

core v7 and x v7 are incompatible. so they shouldn't use x v7 with toolpad in the first place. if they don't want to use a beta version of x they shouldn't use toolpad. it's more unstable than a beta version of x.

besides all of that. I don't see why two versions of the grid couldn't be together in one project

PR #4861, opened 2025-04-18
[DashboardLayout] Prevent horizontal scrollbars by adding tolerance

bharatkashyap:
- Resolve https://github.com/mui/toolpad/pull/4844#issuecomment-2814024050


apedroferreira (approved):
Looks like this works, thanks!

apedroferreira (changes requested):
Wait I tried it in the playground and it still might not be working?
Looking into this, it works in the demos...

Screenshot 2025-04-18 at 12 10 46

apedroferreira (approved):

Issue #4841, opened 2025-04-14
[CRUD] Incorrect spacing of edit/list view


bharatkashyap:
### Steps to reproduce

- Start up a new create-toolpad-app 
- Navigate to the employees page
- Click on an employee

### Current behavior

<img width="1058" alt="Image" src="https://github.com/user-attachments/assets/00d8cc91-9973-4671-9aeb-bef6153ad1d5" />

### Expected behavior

- The cards should be spaced more evenly in the page

**Search keywords**: CRUD

PR #4863, opened 2025-04-18
Update dependencies in `create-toolpad-app` and examples

apedroferreira:
Closes https://github.com/mui/toolpad/issues/4841

Also the latest version of `create-toolpad-app` isn't working for me because the dependencies there were never updated to match the latest release versions... I thought that something like this would have been caught during the release. 
But this should fix it.


apedroferreira:
Looks like it fixes https://github.com/mui/toolpad/issues/4841 if there was an issue (I never managed to reproduce it)

I've tried the vite started without authentication and all looks good:
Screenshot 2025-04-18 at 18 02 30

bharatkashyap:
> Closes #4841
> 
> Also the latest version of `create-toolpad-app` isn't working for me because the dependencies there were never updated to match the latest release versions... I thought that something like this would have been caught during the release. But this should fix it.

I think a way to future-proof the release to check for this is to add an `--example` test to the smoke tests to test not just the bootstrapping of new apps but also if the examples have been updated correctly: https://github.com/mui/mui-toolpad/blob/master/RELEASE.md

bharatkashyap (approved):
Explicit versions work better than `latest` since `latest` can mean different things in dependency managers' caches

apedroferreira:
> > Closes #4841
> > Also the latest version of `create-toolpad-app` isn't working for me because the dependencies there were never updated to match the latest release versions... I thought that something like this would have been caught during the release. But this should fix it.
> 
> I think a way to future-proof the release to check for this is to add an `--example` test to the smoke tests to test not just the bootstrapping of new apps but also if the examples have been updated correctly: https://github.com/mui/mui-toolpad/blob/master/RELEASE.md

I think the examples are fine to update later, but in the current version even running `npx create-toolpad-app` seems to fail for me.

PR #4864, opened 2025-04-18
[docs] Fix theme switcher in docs demos

apedroferreira:
Since the update to MUI v7, the theme switcher in all demos seems to be broken: https://mui.com/toolpad/core/react-dashboard-layout/

I think I've identified the problem as we need to set a `container` based on the `iframe` window in the Emotion cache being used.
Added a new `DemoProvider` component thats needs to wrap each demo that uses a full application with theme switching. A more hidden alternative could be to again use some specific docs context for this. Not sure which is better...

https://deploy-preview-4864--mui-toolpad-docs.netlify.app/toolpad/core/react-dashboard-layout/#demo


docs/data/toolpad/core/components/dashboard-layout/DashboardLayoutBasic.js
```diff
@@ -99,22 +101,32 @@ function DashboardLayoutBasic(props) {
 
   const router = useDemoRouter('/dashboard');
 
-  // Remove this const when copying and pasting into your project.
+  // Remove these variables when copying and pasting into your project.
   const demoWindow = window !== undefined ? window() : undefined;
+  const demoEmotionCache = React.useMemo(
```
apedroferreira, line 106: Adding this in each demo that uses a whole `AppProvider` would solve this issue.

bharatkashyap:
Does it make sense to show a warning/error in the console/screen when the `DemoProvider` is used outside by devs in development/production? We use the origin and `NODE_ENV` to detect this? The value would be to ensure that even if a dev forgets to remove these when copying and pasting, there is a guardrail that prevents from this code making it to production

apedroferreira:
> Does it make sense to show a warning/error in the console/screen when the `DemoProvider` is used outside by devs in development/production? We use the origin and `NODE_ENV` to detect this? The value would be to ensure that even if a dev forgets to remove these when copying and pasting, there is a guardrail that prevents from this code making it to production

Sounds good, just added it.

apedroferreira:
As discussed: we can merge this but a future alternative to get rid of this clutter in the demos would be to add a new attribute to documentation demos in MUI infra to set this up for us without showing it to the user.

bharatkashyap (approved):

PR #4876, opened 2025-04-21
Fix showing single-select values, add tests

apedroferreira:
Closes https://github.com/mui/toolpad/issues/4806

Show proper `singleSelect` field value in CRUD "Show" UI.
Adds tests to cover these scenarios.


bharatkashyap:
Should we add a demo in the CRUD docs to illustrate this behaviour?

apedroferreira:
> Should we add a demo in the CRUD docs to illustrate this behaviour?

I don't think so, it's just a different way to configure multi-select fields but it's the same as for the MUI X Data Grid. What we have in the demos is enough in my opinion.

bharatkashyap (approved):

PR #4881, opened 2025-04-23
[DashboardLayout] Remove `DashboardLayout` navigation prop

apedroferreira:
Deprecate `navigation` prop in `DashboardLayout` for the next 2 versions.
The `navigation` should always be set in the `AppProvider` instead.

Closes https://github.com/mui/toolpad/issues/4856


packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -75,9 +75,11 @@ export interface DashboardLayoutProps {
    * @default null
    */
   branding?: Branding | null;
+  // @TODO: Remove deprecated prop from v0.16.0 onwards
   /**
    * Navigation definition for the dashboard. [Find out more](https://mui.com/toolpad/core/react-dashboard-layout/#navigation).
    * @default []
+   * @deprecated Set the navigation in the [AppProvider](https://mui.com/toolpad/core/react-app-provider) instead.
    */
   navigation?: Navigation;
```
Janpot, line 84: Probably also want to show a warning at runtime.

apedroferreira: added in https://github.com/mui/toolpad/pull/4881/commits/9551fd1d36a117bb5e03dbe5ea016452775068c3

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -149,6 +151,14 @@ function DashboardLayout(props: DashboardLayoutProps) {
     sx,
   } = props;
 
+  React.useEffect(() => {
```
Janpot, line 154: This should only run in development, and we also have this [utility](https://github.com/mui/toolpad/blob/077f338f38acef9557a40b648a48cc26af0a92ba/packages/toolpad-utils/src/warnOnce.ts#L7). Doesn't seem to be used anywhere though

apedroferreira: didn't know about it, used it here, I guess it should be used on render like this? https://github.com/mui/toolpad/pull/4881/commits/de4d2c7c30452261885d6fc594b685b8219b89f3

packages/toolpad-core/src/DashboardLayout/DashboardLayout.tsx
```diff
@@ -149,6 +152,12 @@ function DashboardLayout(props: DashboardLayoutProps) {
     sx,
   } = props;
 
+  if (navigationProp && process.env.NODE_ENV === 'development') {
```
Janpot, line 155: We almost exclusively check against `'production'` in our codebases

```suggestion
  if (navigationProp && process.env.NODE_ENV !== 'production') {
```

Janpot (approved):
We always use `process.env.NODE_ENV !== 'production'`

bharatkashyap (approved):

PR #4883, opened 2025-04-23
Remove .js extensions from Next.js AppProvider imports

apedroferreira:
Closes https://github.com/mui/toolpad/issues/4862 as it seems to fix it? Tried it in the repro: https://github.com/a88zach/issue-4862

Not sure exactly why it's happening though @Janpot


Janpot (changes requested):
Can't do that because https://github.com/mui/toolpad/pull/4686#issuecomment-2708631483

I can't install https://github.com/a88zach/issue-4862 without errors, and it uses outdated versions of all our libraries. This reproduction should be scoped down, and the conflicting peer dependencies removed.

Wr could try the [trick](https://github.com/mui/material-ui/blob/e2bf4ca06a1c699213d06616f295e90011026623/packages/mui-material-nextjs/src/v13-pagesRouter/pagesRouterV13Document.tsx#L6) that we do in core

a88zach:
> I can't install https://github.com/a88zach/issue-4862 without errors, and it uses outdated versions of all our libraries. This reproduction should be scoped down, and the conflicting peer dependencies removed.


You should be able to install using the `--force` flag as noted in the issue description and the readme.

This repo came directly from your examples using `npx create-toolpad-app@latest --example auth-nextjs-themed`

apedroferreira:
As I've commented in https://github.com/mui/toolpad/issues/4862#issuecomment-2830859345, the standard example with no changes seems not to have this issue, so it would be better to identify what exactly is causing the reloads on navigation before trying to find a solution.

I've closed this PR as we probably need more info on the problem to find a suitable solution.

PR #4888, opened 2025-04-25
Remove auth-nextjs-themed

Janpot:
Proposing to remove the example. it seems to be more of a template than an example. An example should be terse, to the point and demonstrate a single aspect of the product.


JAugustusSmith:
I'd prefer it be left in. When I'm reviewing a product to see if it's useful, I load up the template. This is the template I used to decide I should use toolpad

apedroferreira:
I agree that this example is different to the others, and adds a lot of maintenance overhead due to its complexity, which is even worse if people use it as a starter.

Other than that the only current issue it seems to have is that the `@react-spring/web` seems incompatible with the current dependencies somehow, worth fixing separately. I can take a look at that.

PR #4890, opened 2025-04-25
[examples] Remove React Spring from themed example

apedroferreira:
As mentioned in https://github.com/mui/toolpad/pull/4888, the React Spring dependency is preventing the `auth-nextjs-themed` example from properly installing as it's not compatible with React 19 and not being actively maintained enough.

This replacement with the MUI `Collapse` component seems to be similar enough and makes it so we don't need React Spring.

**Before:**

https://github.com/user-attachments/assets/25914ec9-9e5e-4c7d-b335-f981711083e6

**After:**

https://github.com/user-attachments/assets/32a86d18-7124-484c-85f1-5bc3c9b80310


bharatkashyap (approved):

bharatkashyap:
Sounds good! Better solution imo than removing the example outright which seems to suggest that the maintenance overhead with our own version upgrades is so great that we can't take it on.

apedroferreira:
> Sounds good! Better solution imo than removing the example outright which seems to suggest that the maintenance overhead with our own version upgrades is so great that we can't take it on.

It's ok, we can just discuss it! Thanks for approving, just gonna need to merge first https://github.com/mui/toolpad/pull/4891 for the build to pass, it seems.

PR #4945, opened 2025-05-17
Use named imports for locales

yuito-it:
## Summary

This PR adds an `index.js` file under the `locales/` directory to explicitly export available locale modules (currently `hiIN`). This enables imports like:

```ts
import { hiIN } from "@toolpad/core/locales";
```

## Motivation

Previously, consumers attempting to `import @toolpad/core/locales` encountered errors because:

- The module was implemented as a flat file (hiIN.js)
- There was no index.js in the locales/ folder
- The package.json exports field did not expose this path explicitly

but this method matches how Material UI locales are imported.

As a result, valid-looking import paths like:

```ts
import { hiIN } from "@toolpad/core/locales";
```

**would fail at runtime**

This PR fixes that by introducing an explicit index that re-exports the locales, 

## Changes
- Added `locales/index.js` that re-exports `hiIN`
- No locale logic was changed—just improved how they're consumed

## Benefits
- Fixes a broken import path (bugfix)
- Improves DX by enabling clean, named imports
- Easy to scale: new locales can be added via the index file

## Notes
This fix avoids having to restructure each locale into its own folder or clutter the `exports` field with specific entries. It keeps the implementation simple while solving a real usability issue.


yuito-it:
Also I'm going to send anothor PR to translate Japanese, if this PR was merged.

bharatkashyap:
Tested the [localization demo](https://deploy-preview-4945--mui-toolpad-docs.netlify.app/toolpad/core/react-sign-in-page/#localization) with the named import - works

yuito-it:
@bharatkashyap I have a question for you.
When are you going to merge this pull request?
I'm affected by this bug, so I'd like you to merge it quickly.

PR #4968, opened 2025-05-21
[CRUD] Allow custom form fields

apedroferreira:
Closes https://github.com/mui/toolpad/issues/4837

Adds new `renderFormField` property to data source fields to be able to render anything as a form field. Receive arguments for `value`, `onChange` and `error` to connect with the form logic.

Also added slots for every field type in the form, as well as some overall improvements around all this.

### DIFF
--- a/playground/vite/src/data/orders.tsx
+++ b/playground/vite/src/data/orders.tsx
@@ -1,7 +1,16 @@
+import * as React from 'react';
 import { DataModel, DataSource, DataSourceCache } from '@toolpad/core/Crud';
 import { z } from 'zod';
+import Box from '@mui/material/Box';
+import Chip from '@mui/material/Chip';
+import FormControl from '@mui/material/FormControl';
+import FormHelperText from '@mui/material/FormHelperText';
+import InputLabel from '@mui/material/InputLabel';
+import MenuItem from '@mui/material/MenuItem';
+import Select, { SelectChangeEvent } from '@mui/material/Select';
 
-type OrderStatus = 'Pending' | 'Sent';
+type OrderStatus = 'pending' | 'sent';
+type OrderTag = 'gift' | 'fragile' | 'wholesale';
 
 export interface Order extends DataModel {
   id: number;
@@ -10,6 +19,7 @@ export interface Order extends DataModel {
   status: OrderStatus;
   itemCount: number;
   fastDelivery: boolean;
+  tags: OrderTag[];
   createdAt: string;
   deliveryTime?: string;
 }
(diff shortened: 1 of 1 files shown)


bharatkashyap (approved):
Great addition! Also the improvement with the `slots` is important, do we need to additional document it/add a demo for it?

apedroferreira:
> Great addition! Also the improvement with the `slots` is important, do we need to additional document it/add a demo for it?

Thanks! I guess I can add a small example, at least with `slotProps` or so, just not demo every slot exhaustively.

apedroferreira:
Added a demo just with `slotProps` for the forms (I guess that replacing a whole `slot` would be a lot more niche) and restructured the CRUD docs page a little with the "Customization" section:

https://deploy-preview-4968--mui-toolpad-docs.netlify.app/toolpad/core/react-crud/#custom-form-fields
https://deploy-preview-4968--mui-toolpad-docs.netlify.app/toolpad/core/react-crud/#slots

bharatkashyap:
> Added a demo just with `slotProps` for the forms (I guess that replacing a whole `slot` would be a lot more niche) and restructured the CRUD docs page a little with the "Customization" section:
> 
> https://deploy-preview-4968--mui-toolpad-docs.netlify.app/toolpad/core/react-crud/#custom-form-fields https://deploy-preview-4968--mui-toolpad-docs.netlify.app/toolpad/core/react-crud/#slots

Looks good! For the `slot`, what props would the component receive, if any? Even though it feels advanced, I think there's really great value in this feature and by not having a demo we would hamper its discoverability - so we probably need to educate users into being able to pass in their own customised form field components. Wdyt? It can be a follow up to this PR

apedroferreira:
> > Added a demo just with `slotProps` for the forms (I guess that replacing a whole `slot` would be a lot more niche) and restructured the CRUD docs page a little with the "Customization" section:
> > https://deploy-preview-4968--mui-toolpad-docs.netlify.app/toolpad/core/react-crud/#custom-form-fields https://deploy-preview-4968--mui-toolpad-docs.netlify.app/toolpad/core/react-crud/#slots
> 
> Looks good! For the `slot`, what props would the component receive, if any? Even though it feels advanced, I think there's really great value in this feature and by not having a demo we would hamper its discoverability - so we probably need to educate users into being able to pass in their own customised form field components. Wdyt? It can be a follow up to this PR

The `slots` and `slotProps` there have the same properties, and there kind of is a better way to customize individual fields with `renderFormField`. Using a slot would be helpful to override all text fields in general to your own custom version of it, for example...
I can keep it noted to maybe document this even more extensively in a demo or so.

PR #5135, opened 2026-08-06
fix(DashboardLayout): render NavigationItem.action as ListItem secondfix(DashboardLayout): render NavigationItem.action as ListItem's secondaryAction

### LINKED ISSUE #4747: [DashboardLayout] Render `NavigationItem.action` in as `ListItem`'s `secondaryAction`
### Search keywords

dashboard layout navigation item action secondary action

### Latest version

- [x] I have tested the latest version

### Summary

When a `NavigationItem` is not a parent item (i.e. it's a button in expanded sidebar view), and an `action` is provided, pointer event on the `action` component also propagate to the `ListItem` component. Following MUI practices, the `action` component would be better rendered as the `ListItem`'s `secondaryAction` (or wrapped in the `SecondaryAction` child component).

### Examples

_No response_

### Motivation

_No response_

forurmah:
Fixes #4747

When a `NavigationItem` has an `action` (e.g. a button rendered in the expanded sidebar), it was rendered inside the clickable `NavigationListItemButton`, so pointer events on the action also triggered the parent nav item's click handler.

This moves `item.action` out of the button and passes it via `ListItem`'s built-in `secondaryAction` prop instead, following MUI's standard pattern for this composition — the action now renders as a sibling of the button rather than a nested clickable element, so clicks no longer propagate.


forurmah:
Looks like the CI failures are unrelated to this change — the CircleCI logs show:
"Orb codecov/codecov@5.4.3 not loaded. To use this orb, your organization must enable the 'Allow uncertified public orbs' feature in Org Settings > Security."
This seems to be an org-level CircleCI config issue rather than something caused by this PR.

